PR #65, opened 2021-03-25
Subscription price

nicolasburtey:
Start adding a WebSocket connection for price updates. This will be used in the mobile app.

@krtk6160 we should use the same way to dynamically confirm payment on lnpage (and also to have usd equivalent price).


nicolasburtey:
@samerbuna should this PR being close?

samerbuna:
> @samerbuna should this PR being close?

Yep. Closing...

PR #148, opened 2021-04-17
Active payments limiting

krtk6160:


nicolasburtey:
this would be great to rebase this PR @krtk6160 so we can also merge it. it's an important PR security wise.

PR #193, opened 2021-05-12
2fa

nicolasburtey:


default.yaml
```diff
@@ -90,10 +90,18 @@ limits:
...
+
+# blacklistedIPs: ['1.2.3.4', '5.6.7.8']
+
 fees:
   withdraw: 2000 # 2000 sats flat per tx
   deposit: 0.003 # 0.3 percent
 
+twoFactor:
+  enabled: true
+  threshold: 300000
```
nicolasburtey, line 103: not clear what the unit is

src/Lightning.ts
```diff
@@ -263,12 +263,21 @@ export const LightningMixin = (superclass) => class extends superclass {
...
 
+
     const { tokens, mtokens, username: input_username, destination, pushPayment, id, routeHint, messages, memoInvoice, payment, cltv_delta, features, max_fee } = await this.validate(params, lightningLogger)
-    const { memo: memoPayer } = params
+    const { memo: memoPayer, twoFactorToken: token } = params
 
     // not including message because it contains the preimage and we don't want to log this
     lightningLogger = lightningLogger.child({ decoded: { tokens, destination, pushPayment, id, routeHint, memoInvoice, memoPayer, payment, cltv_delta, features }, params })
 
+    if(yamlConfig.twoFactor?.enabled && this.user.twoFactor.secret && tokens > this.user.twoFactor.threshold) {
```
nicolasburtey, line 273: maybe we should have a default in `schema.ts` for twoFactor instead of having to use a `?` in `yamlConfig.twoFactor?.enabled`

src/userWallet.ts
```diff
@@ -219,4 +222,61 @@ export abstract class UserWallet {
...
+
+    try {
+      await this.user.save()
+      return true
+    } catch (err) {
+      throw new DbError("Unable to save 2fa secret", {forwardToClient: true, logger: this.logger, level: 'error', err})
+    }
+  }
+
+  static validate2fa = ({ secret, token, logger }): boolean => {
```
nicolasburtey, line 257: it's odds to have this one being a static and not the other one. why is that?

src/__tests__/helper.ts
```diff
@@ -143,3 +144,9 @@ export const mineBlockAndSync = async ({ lnds, blockHeight }: { lnds: Array<any>
   }
   await Promise.all(promiseArray)
 }
+
+export const set2FA = async ({ wallet, secret }) => {
+  const token = generateToken(secret)!.token
```
nicolasburtey, line 149: is there a way to type this in a way we don't need the `!`?

src/userWallet.ts
krtk6160: I want to call this from inside `Lightning.ts`

src/__tests__/helper.ts
```diff
@@ -143,3 +144,9 @@ export const mineBlockAndSync = async ({ lnds, blockHeight }: { lnds: Array<any>
   }
   await Promise.all(promiseArray)
 }
+
+export const set2FA = async ({ wallet, secret }) => {
```
nicolasburtey, line 148: this seems somewhat redundant with `save2fa`

nicolasburtey (commented):
couple of comments

src/userWallet.ts
nicolasburtey: oh gotcha. but this is not a great design. some sure what the right solution here right now but something to think about.

src/Lightning.ts
krtk6160: We can have a default in the dbMetadata schema but I don't think it's the right place for this to be. I was thinking about the `twoFactor.enabled` in a similar way to how we are doing `proxyChecking.enabled` and thus yamlConfig seemed the right place for it. We can of course remove this check altogether in the future and just rely on whether the user's object has a secret or not (in which case there won't be a system-wide twofactor enable/disable flag)

src/__tests__/helper.ts
krtk6160: Agreed, I'll refactor it.

src/userWallet.ts
krtk6160: Added a comment to it. Idk what the right way to do this is. Maybe Samer can help here, but I am unable to tag him.

default.yaml
krtk6160: Fixed

src/__tests__/helper.ts
krtk6160: I tried to not have the `!` but it seems because of the way `generateToken` is implemented in the `node-2fa` library, the value returned can indeed be null.

krtk6160:
Closing as duplicate of #271

PR #199, opened 2021-05-17
improve resiliency by handling several lnd

nicolasburtey:


src/schema.ts
```diff
@@ -41,9 +41,21 @@ const invoiceUserSchema = new Schema({
...
+
+  paid: {
+    type: Boolean,
+    default: false
+  }
+
 })
 
-invoiceUserSchema.index({ "uid": 1 })
+invoiceUserSchema.index({ "uid": 1, paid: 1 })
```
krtk6160, line 58: If we don't need quotes then we can unquote `uid` as well.

src/lndHealth.ts
```diff
@@ -0,0 +1,47 @@
...
+    // if we get disconnected, we need to recreate the lnd object
+    const paramIndex = _.findIndex(params, {socket})
+    params[paramIndex] = {
+      ...params[paramIndex],
+      lnd: authenticatedLndGrpc(params[paramIndex]).lnd
+    }
+
+    active = false
+  }
+  _.find(params, {socket})!.active = active
```
krtk6160, line 42: What is this syntax (`!.active`)?

src/__tests__/01-connection.ts
```diff
@@ -20,10 +21,26 @@ it('I can connect to bitcoind', async () => {
...
 })
 
-it('I can connect to bank lnd', async () => {
-	const { public_key } = await getWalletInfo({ lnd: lndMain })
-	expect(public_key.length).toBe(64 + 2)
-})
+const lnds = [lnd1, lnd2, lndonchain]
+for (let item in lnds) {
+  it(`I can connect to lnd index ${item}`, async () => {
+    const { public_key } = await getWalletInfo({ lnd: lnds[item] })
```
krtk6160, line 27: Are `item` and `lnds[item]` not same?

src/__tests__/02b-onchain-receive.ts
```diff
@@ -43,7 +45,8 @@ beforeEach(async () => {
   initBlockCount = await bitcoindDefaultClient.getBlockCount()
   initialBalanceUser0 = (await walletUser0.getBalances()).BTC
 
-  amount_BTC = +(1 + Math.random()).toPrecision(9)
+  // amount_BTC = +(1 + Math.random()).toPrecision(9)
```
krtk6160, line 48: Why are we switching to constant amount_BTC instead of random?

src/ledger/balanceSheet.ts
```diff
@@ -23,7 +20,7 @@ export const updateUsersPendingPayment = async () => {
   }
 }
 
-export const getBalanceSheet = async () => {    
+export const getLedgerAccounts = async () => {
```
krtk6160, line 23: I think `getBalanceSheet` was a more apt name for this function than `getLedgerAccounts`

src/lndHealth.ts
```diff
@@ -0,0 +1,47 @@
+import { authenticatedLndGrpc, getWalletInfo } from 'lightning';
+import _ from "lodash";
+import { baseLogger } from "./logger";
+import { ILndParamsAuthed } from "./lndAuth"
+
+const refresh_time = 5000 // ms
+
+// TODO replace with
+// const asyncForever = require('async/forever');
+const loop = async ({socket, params}: {socket: string, params: ILndParamsAuthed}) => {
```
krtk6160, line 10: Can we not just call isUp inside the setTimeout to remove the need for recursion? Maybe I'm missing why there's a necessity for recursion here

src/lndUtils.ts
```diff
@@ -31,7 +44,7 @@ export const lndBalances = async () => {
   )
 
   const total = chain_balance + channel_balance + pending_chain_balance + opening_channel_balance + closing_channel_balance
-  return { total, onChain: chain_balance + pending_chain_balance, offChain: channel_balance, opening_channel_balance, closing_channel_balance }
+  return { total: total, onChain: chain_balance + pending_chain_balance, offChain: channel_balance, opening_channel_balance, closing_channel_balance }
```
krtk6160, line 47: This change seems to be redundant

src/__tests__/09f-routing.ts
```diff
@@ -37,9 +34,9 @@ it('records routing fee correctly', async () => {
 
   await updateRoutingFees()
 
-  const {balance} = await MainBook.balance({
-    accounts: revenueFeePath
-  })
+  // const {balance} = await MainBook.balance({
```
krtk6160, line 37: Why is this test removed?

src/__tests__/03-create-channels.ts
```diff
@@ -115,16 +119,30 @@ it('opens private channel from lndOutside1 to lndOutside2', async () => {
...
+  const socket = `lnd2:9735`
+  await openChannelTesting({ lnd: lnd1, other_lnd: lnd2, socket })
+
+  const { channels } = await getChannels({ lnd: lnd1 })
+  expect(channels.length).toEqual(channelLengthMain + 1)
+  
+  const channel = _.find(channels, {partner_public_key: getLnds()[1].pubkey})
+  const input = {fee_rate: 0, base_fee_tokens: 0, transaction_id: channel!.transaction_id, transaction_vout: channel!.transaction_vout}
+
+  await updateRoutingFees({lnd: lnd1, ...input});
```
krtk6160, line 142: Why is this needed?

src/Lightning.ts
```diff
@@ -638,8 +680,29 @@ export const LightningMixin = (superclass) => class extends superclass {
...
-    let invoice
+  // return whether the invoice has been paid of not
+  async updatePendingInvoice({ hash, lock, pubkey }: {hash: string, lock: any, pubkey?: string}): Promise<boolean> {
+    let invoice, pubkey_
+    
+    // if a pubkey has been provided, it means the invoice has not been set as paid in mongodb
+    // so not need for a round back trip to mongodb
+    if (!pubkey) {
+      let paid
+      ({ pubkey: pubkey_, paid } = await InvoiceUser.findOne({ _id: hash }));
```
krtk6160, line 691: Why not fetch everything here and remove the call on #L733?

krtk6160 (commented):
First round of comments

src/lndHealth.ts
nicolasburtey: I'm saying to typescript that I know active property actually exist and is not null. it's because the type associated with it could be null. but actually I've refactored it and this is not needed any more

src/__tests__/01-connection.ts
nicolasburtey: item is an index. ie: `item = 0`, then `item = 1`, etc

I'm using `let item in lnds` instead of `let item of lnds` because I want to show the index if we can't connect to some lnd

src/__tests__/02b-onchain-receive.ts
nicolasburtey: had set a fix amount for debugging stuff. eventually we should set the random seed to always return the same result.

src/ledger/balanceSheet.ts
nicolasburtey: from an accounting point of view, balance sheet is really about liability + assets. here this function is returning some other stuff, that I think is best describe as ledger: https://www.freshbooks.com/hub/accounting/what-is-a-ledger

src/__tests__/09f-routing.ts
nicolasburtey: not sure any more. added it back.

src/__tests__/03-create-channels.ts
nicolasburtey: we want to be able to receive on-us transaction even if one of the 2 nodes we have is offline.

ie: 
1/ user A show a QR code from node 1 and send the invoice to someone else.
2/ node 1 goes offline for whatever reason
3/ user B pays the invoice. because the invoice is part of the `isMyNode()` set, basically it settle offline, with 0 fee, even if one node generated the invoice and another node would have paid for it through a common channel.

only caveat here is that graphql would not delete the invoice from node A

and so because it settle in layer 3 "off lightning", there is no fee currently. so I wanted to have the same no-fee connection between 2 nodes of the same group. 

now that I think about it, this is not really necessary, and we could still have fees between those.

src/Lightning.ts
nicolasburtey: good question.

the idea is that the first check here is to see whether the invoice has been paid. we only do it if the pubkey is not passed as an argument, which is when it's called from graphql directly (from updatePendingInvoice or noauthUpdatePendingInvoice).

if it is, we can just return the function here. the benefits of doing that is that we only lock the account if the invoice has been paid but is not yet updated in medici. 

if we wanted to have a single call to InvoiceUser, we should basically have to lock the account on every request, which is not ideal. if we don't use a lock, we could have a race condition and payment could be validated multiple time. 

does that make sense?

src/lndHealth.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    active = false
+  }
+  _.find(params, {socket}).active = active
+  baseLogger.info({socket, active}, "lnd pulse")
+
+  setTimeout(async function () {
+    // TODO check if this could lead to a stack overflow
+    isUpLoop({socket, params})
```
krtk6160, line 35: I still have the same question regarding this. From above -
"Can we not just call isUpLoop inside the setTimeout to remove the need for recursion? Maybe I'm missing why there's a necessity for recursion here"

Why can't we just do this -> `setTimeout(() => isUpLoop(), refresh_time)` and have the setTimeout be outside the isUpLoop function?

src/Lightning.ts
krtk6160: Yep

src/__tests__/03-create-channels.ts
krtk6160: > only caveat here is that graphql would not delete the invoice from node A

Is that a problem? (Apart from just slightly bloating the db)

> now that I think about it, this is not really necessary, and we could still have fees between those.

Yeah, there will be 0 fees irrespective of the fees set for the channel because they're directly connected.

krtk6160 (commented):
One question, but otherwise LGTM

src/lndHealth.ts
nicolasburtey: oh right ok.
so what we want to do here is to update the array and set `{active: true}` or `{active: false}` over time depending on whether lnd responds. so we want a regular call.

it's like the liveness probe for kubernetes if you want. 

if lnd1 if offline, then graphql will switch automatically to lnd2. 

does that answer your question?

krtk6160: I mixed up setInterval and setTimeout, my bad.

Why not use `setInterval` instead of `setTimeout`? That will remove the need for recursion.

krtk6160 (commented):
Last question on the isUpLoop

src/entrypoint/cron.ts
```diff
@@ -1,15 +1,44 @@
...
+const deleteExpiredInvoices = async () => {
+  // this should be longer than the invoice validity time
+  const delta = 2 // days
+  
+  const date = new Date();
+  date.setDate(date.getDate() - delta);
+  InvoiceUser.deleteMany({timestamp: {lt: date}})
+}
+
+const deleteFailedPaymentsAllLnds = async () => {
```
krtk6160, line 20: This and `deleteExpiredInvoices` should be moved to lndUtils imo

src/types.ts
```diff
@@ -56,14 +56,21 @@ export interface IOnChainPayment {
...
   description: string
-  created_at: Date
-  hash?: string
-  destination?: string
   type: TransactionType
+  hash?: string
+  fee: number,
+  feeUsd: number,
+  username: string,
+  // destination?: string
```
krtk6160, line 69: Why is this being removed?

src/lndUtils.ts
```diff
@@ -141,3 +166,146 @@ export const updateRoutingFees = async () => {
...
+
+export const offchainLnds = getLnds({type: "offchain"}) 
+
+// only returning the first one for now
+export const getActiveLnd = () => {
+  const lnds = getLnds({active: true, type: "offchain"})
+  if (lnds.length === 0) {
+    throw Error("no active lnd to send/receive a payment offchain")
+  }
+  // this favor lnd1
```
krtk6160, line 288: Have you given more thought to making this function act like a loadbalancer (alternating between lnd1 and lnd2)?

src/lndUtils.ts
```diff
@@ -141,3 +166,146 @@ export const updateRoutingFees = async () => {
...
+  return lnds[0]
```
krtk6160, line 289: Have you given more thought to making this function act like a loadbalancer (alternating between lnd1 and lnd2)?

src/OnChain.ts
```diff
@@ -251,64 +256,75 @@ export const OnChainMixin = (superclass) => class extends superclass {
...
-    //   description: undefined,
-    //   fee: undefined,
-    //   id: '60dfde7a0c5209c1a8438a5c47bb5e56249eae6d0894d140996ec0dcbbbb5f83',
-    //   is_confirmed: false,
-    //   is_outgoing: false,
-    //   output_addresses: [Array],
-    //   tokens: 100000000,
-    //   transaction: '02000000000...' }
+    const pubkeys: string[] = this.user.onchain_pubkey
+    let user_matched_txs: GetChainTransactionsResult["transactions"] = []
```
krtk6160, line 260: Not sure if it's a good practice to use typings from other packages in this manner. It's hard to find the type without digging into the `Lightning` package's code. While it probably isn't a problem in an editor like VS Code (with it's Ctrl+click functionality), I think it makes the dev experience slightly worse if you're using a lighter editor or reviewing on web.

src/OnChain.ts
```diff
@@ -251,64 +256,75 @@ export const OnChainMixin = (superclass) => class extends superclass {
...
+      user_matched_txs = [...user_matched_txs, ...lnd_incoming_filtered.filter(tx => _.intersection(tx.output_addresses, addresses).length > 0)]
+  
     }
 
-    const user_matched_txs = lnd_incoming_filtered.filter(tx => _.intersection(tx.output_addresses, this.user.onchain_addresses).length > 0)
-
     return user_matched_txs
   }
 
-  async getTransactions(): Promise<Array<ITransaction>> {
```
krtk6160, line 310: Why is this being removed? Is it because the type is now being inferred automatically?

src/Lightning.ts
```diff
@@ -72,6 +65,15 @@ export const LightningMixin = (superclass) => class extends superclass {
 
     const expires_at = this.getExpiration(moment()).toDate()
     
+    let lnd: AuthenticatedLnd, pubkey: string
+
+    try {
+      ({ lnd, pubkey } = getActiveLnd())
+    } catch (err) {
+      this.logger.error("no active lnd to create an invoice")
```
krtk6160, line 73: Probably better to import and throw new CustomError (if you don't want to create a new error class just for this error)

src/Lightning.ts
```diff
@@ -179,15 +185,15 @@ export const LightningMixin = (superclass) => class extends superclass {
...
+    const value = JSON.stringify({...route, pubkey})
     await redis.set(key, value, 'EX', 60 * 5); // expires after 5 minutes
 
     lightningLogger.info({ redis: { key, value }, probingSuccess: true, success: true }, "successfully found a route")
     return route.fee
   }
 
   // FIXME this should be static
-  async validate(params: IFeeRequest, lightningLogger) {
+  async validate({params, logger}: {params: IFeeRequest, logger: Logger}) {
```
krtk6160, line 196: Shifting this function to `lndUtils.ts` might make sense

src/Lightning.ts
```diff
@@ -338,14 +353,23 @@ export const LightningMixin = (superclass) => class extends superclass {
           })
         })
         
-        await transactionNotification({ amount: sats, user: payeeUser, hash: id, logger: this.logger, type: "paid-invoice" })
+        transactionNotification({ amount: sats, user: payeeUser, hash: id, logger: this.logger, type: "paid-invoice" })
```
krtk6160, line 356: Has the effect of removing `await` been tested here?

krtk6160 (commented):
More comments

src/entrypoint/cron.ts
nicolasburtey: yup

src/types.ts
nicolasburtey: not being used currently. destination is meant for push payment to other wallet but this is just an idea I have been working  on months back that would need a lot more thoughts.

src/lndUtils.ts
nicolasburtey: yup could be an option. specifically if we get higher and higher load. a random() would be good for that.

but I think for now the main issue is not as much as the load on a single lnd as the need for a failover. 

we can revisit when we have more input with higher load.

src/OnChain.ts
nicolasburtey: do you have a recommendation here?
expect duplicating the type to our repo I'm not sure what other there would be. 

on the code editor side, I think that when working with Typescript, it doesn't make a lot of sense to use an editor that doesn't provide type hints.

src/OnChain.ts
nicolasburtey: yup

src/Lightning.ts
nicolasburtey: yep good point

src/Lightning.ts
nicolasburtey: yup

src/Lightning.ts
nicolasburtey: not really but is it necessary?

`transactionNotification` return `Promise<void>` and basically we don't care about the result of this function for what the below this function. 

so awaiting for this is unnecessary. it will increase the response time to the request to the user. the notification can be sent in parallel.

src/lndUtils.ts
krtk6160: It might make sense to mock this function as a loadbalancer just for the tests (for now) to make sure the tests work irrespective of which lnd is returned. This might help catch some bugs earlier.

src/OnChain.ts
krtk6160: > do you have a recommendation here?
> expect duplicating the type to our repo I'm not sure what other options there would be.

Nope, not apart from duplicating it.

> on the code editor side, I think that when working with Typescript, it doesn't make a lot of sense to use an editor that doesn't provide type hints.

Yeah kinda makes sense. It will still be hard for someone who's just looking through the code on github as opposed to cloning and viewing. We'll have a lot more people in the former category than the latter. Something to consider.

src/Lightning.ts
krtk6160: Just wondering if the transactionNotification method will still be executed completely if the parent function returns while it's in the middle of completing

src/OnChain.ts
krtk6160: Maybe a comment that has the structure of the type (an example) makes sense. This is something we've already been using and it definitely helps imo

krtk6160 (approved):

src/Lightning.ts
bodymindarts: [Stackoverflow](https://stackoverflow.com/questions/45594596/async-function-without-await-in-javascript) thinks it will complete.
> As you assumed, if no await is present the execution is not paused and your code will then be executed in a non-blocking manner.

bodymindarts:
The GitHub action that runs the integration test wasn't failing because the `jest` exit code was being eaten. I added a commit to fix this and now the `integration test` GitHub action (correctly) fails.

src/OnChain.ts
bodymindarts: I think in general we should start refactoring to not rely on the third party dependencies on so many places in the code.

Have 1 file that talks to the dependency and acts as a wrapper exposing our own types internally. That way if things change upstream you only have to touch the wrapper and not the rest of the code.

But this obviously requires quite a bit of cleanup and is best done step-by-step.

nicolasburtey:
> The GitHub action that runs the integration test wasn't failing because the jest exit code was being eaten. I added a commit to fix this and now the integration test GitHub action (correctly) fails.

yep it was on todo list before merging the PR to make sure those tests also run correctly. thanks! will push the change on docker-compose to fix it.

PR #203, opened 2021-05-24
Add lndmon scraping

krtk6160:


charts/monitoring/values.yaml
```diff
@@ -198,6 +198,37 @@ prometheus:
         - /etc/config/alerts
 
       scrape_configs:
+        - job_name: 'lnd'
```
nicolasburtey, line 201: can we add those configs to the lndmon chart instead?

it would remove a hard dependency to the monitoring chart in case someone doens't want to have lndmon

krtk6160: Nope, it's not possible to add a prometheus config without explicitly specifying it in the prometheus values.yaml. The closest we can get is: Specifying a configmap in `extraConfigmapMounts` in this file, and then having the template for that configmap in the lndmon chart. However this I don't think this solution makes sense because it doesn't remove the hard dependency, it just shifts the scrape job definition onto a different file.

nicolasburtey: I'm not sure to really follow. pod like graphql have this annotation:

```
  podAnnotations:
    prometheus.io/path: /swagger/metrics
    prometheus.io/port: "4000"
    prometheus.io/scrape: "true"
```

and make prometheus fetch the data from this pod (with `scrape_interval` and `scrape_timeout` being the default). why can't we do the same here?

krtk6160: We are actually already doing this, but the difference lies in the relabelling. So in our lnd scrape config, we are doing:
```
- source_labels: [__meta_kubernetes_pod_label_app_kubernetes_io_instance
  action: replace
  target_label: pod
```

What this does is it sets the `pod` label to the value of the `app.kubernetes.io/instance` label of the pod being scraped. We are doing the same with the namespace as well. This is required because the lndmon grafana dashboards expect the labels to be named pod and namespace.

Now our original generic scrape config actually does relabelling in a different manner, which is why the formats are incompatible.

PR #204, opened 2021-05-24
use helpers.tpl for jwt secret handling

krtk6160:


charts/galoy/templates/jwt-secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: jwt-secret
+  labels:
+    app.kubernetes.io/managed-by: Helm
```
nicolasburtey, line 6: is there some label or annotation to add so that this secret is not deleted even if the chart is uninstall?

nicolasburtey (commented):
one question

charts/galoy/templates/jwt-secret.yaml
krtk6160: Yes. I have added it.

nicolasburtey (approved):

PR #209, opened 2021-05-30
Enable redis sentinel

krtk6160:


initLocalTest.sh
```diff
@@ -165,6 +165,8 @@ helmUpgrade galoy \
   $INFRADIR/galoy/
 
 kubectlWait app.kubernetes.io/instance=galoy
+kubectlWait statefulset.kubernetes.io/pod-name=galoy-redis-node-1
+kubectlWait statefulset.kubernetes.io/pod-name=galoy-redis-node-2
```
nicolasburtey, line 169: we should wait for 3 nodes and just 2?

src/__tests__/01-connection.ts
```diff
@@ -44,13 +44,13 @@ it('I can connect to mongodb', async () => {
 it('I can connect to redis', async () => {
 	const value = "value"
 
-	redisClient.on("error", function(error) {
-		throw new Error("redis issue");
+	redis.on("error", function(error) {
+    console.log({error})
```
nicolasburtey, line 48: we should probably use baseLogger here and add an error message for enhanced "queriness" if something goes wrong

src/redis.ts
```diff
@@ -1,10 +1,29 @@
...
-export const getAsyncRedisClient = () => {
-  clientAsync = clientAsync ?? asyncRedis.decorate(redisClient);
-  return clientAsync
+if(process.env.LOCAL === 'true') {
+  const REDIS_0_INTERNAL_IP = `${process.env.REDIS_0_INTERNAL_IP}:6379` 
+  const REDIS_1_INTERNAL_IP = `${process.env.REDIS_1_INTERNAL_IP}:6379`
+  const REDIS_2_INTERNAL_IP = `${process.env.REDIS_2_INTERNAL_IP}:6379`
+  
+  natMap = {
+    [REDIS_0_INTERNAL_IP]: { host: process.env.REDIS_0_DNS, port: process.env.REDIS_0_PORT },
```
nicolasburtey, line 11: we should pass the default port if nothing is passed in the `REDIS_0_PORT` as a env variable

charts/galoy/templates/deployment.yaml
```diff
@@ -103,10 +103,20 @@ spec:
...
+            - name: REDIS_0_SENTINEL_PORT
+              value: "26379"
+            - name: REDIS_0_DNS
+              value: "galoy-redis-node-0.galoy-redis-headless"
+            - name: REDIS_1_SENTINEL_PORT
+              value: "26379"
+            - name: REDIS_1_DNS
+              value: "galoy-redis-node-1.galoy-redis-headless"
+            - name: REDIS_2_SENTINEL_PORT
+              value: "26379"
```
nicolasburtey, line 117: if we use the default value we probably don't need to pass anything for the port

nicolasburtey:
few comments

initLocalTest.sh
krtk6160: Actually these two lines aren't needed because of the line above them which is waiting for instance=galoy. I've removed it.

src/redis.ts
krtk6160: This branch is for local development. The default port will probably never work with a minikube loadbalancer because the port range starts at `30000` and the default port is `26379`

charts/galoy/templates/deployment.yaml
krtk6160: Yep, removed.

src/__tests__/01-connection.ts
krtk6160: Is that needed for the test? It will automatically print it in the right place if the test fails, so we don't really need to search for anything here.

nicolasburtey: oh yeah nevermind. I though the `redis.on("error", fn)` was on a non test file.

nicolasburtey: that said, we probably still want to have the same thing in src/redis.ts to log redis error if they arise.

nicolasburtey (commented):
one last question on error logging

src/redis.ts
```diff
@@ -26,4 +23,7 @@ connectionObj = {
 }
 
 export const redis = new Redis(connectionObj);
+redis.on('error', (err) => baseLogger.fatal("Redis error", err))
```
nicolasburtey, line 26: the way the pino logger works is that the message should be the last argument.
ie: https://github.com/pinojs/pino/blob/master/docs/api.md#error

I don't think we need to have a fatal for this. error is probably good.

src/redis.ts
```diff
@@ -1,16 +1,13 @@
...
-  const REDIS_0_INTERNAL_IP = `${process.env.REDIS_0_INTERNAL_IP}:6379` 
-  const REDIS_1_INTERNAL_IP = `${process.env.REDIS_1_INTERNAL_IP}:6379`
-  const REDIS_2_INTERNAL_IP = `${process.env.REDIS_2_INTERNAL_IP}:6379`
+  const REDIS_0_INTERNAL_IP = `${process.env.REDIS_0_INTERNAL_IP}:6379`
   
   natMap = {
-    [REDIS_0_INTERNAL_IP]: { host: process.env.REDIS_0_DNS, port: process.env.REDIS_0_PORT },
-    [REDIS_1_INTERNAL_IP]: { host: process.env.REDIS_1_DNS, port: process.env.REDIS_1_PORT },
-    [REDIS_2_INTERNAL_IP]: { host: process.env.REDIS_2_DNS, port: process.env.REDIS_2_PORT }
+    [REDIS_0_INTERNAL_IP]: { host: process.env.REDIS_0_DNS, port: process.env.REDIS_0_PORT }
```
nicolasburtey, line 10: could you explain this syntax here?
`{[ip]: {}}` 
I'm not familiar with the use of square bracket for the key of an object.

krtk6160: Check [this](https://stackoverflow.com/a/11508490/1124303) out

nicolasburtey (approved):
this looks good to me. some extensive testing on testnet before deploying to mainnet to ensure there are no regression. also see how the pod react when 1 or 2 redis are unavailable

PR #212, opened 2021-06-04
Fix routing revenue job

krtk6160:


src/lndUtils.ts
```diff
@@ -84,7 +84,10 @@ export const getRoutingFees = async ({ lnd, before, after }): Promise<Record<str
   const dateGroupedForwards = _.groupBy(forwards, e => new Date(e.created_at).toDateString())
 
   // returns revenue for each date by reducing all forwards for each date
-  return (_.mapValues(dateGroupedForwards, e => e.reduce((sum, {fee_mtokens}) => sum + +fee_mtokens, 0) / 1000))
+  const feePerDate = _.mapValues(dateGroupedForwards, e => e.reduce((sum, {fee_mtokens}) => sum + +fee_mtokens, 0) / 1000)
+
+  // returns an array of objects where each object has key = date and value = fees
+  return (_.map(feePerDate, (v, k) => ({[k]: v})))
```
nicolasburtey, line 90: parenthesis is not needed here?

nicolasburtey (approved):
have not checked the algo but this PR sounds it fixed the prior bug we had of launching too much promise simultaneously. I'll let you check the fees calculated map what we have with lncli, bos or rtl.

src/lndUtils.ts
krtk6160: Fixed.

PR #213, opened 2021-06-05
reduce generic error class usage

krtk6160:


src/Lightning.ts
```diff
@@ -174,23 +174,20 @@ export const LightningMixin = (superclass) => class extends superclass {
         total_mtokens: payment ? mtokens : undefined,
       }));
     } catch (err) {
-      const error = "error getting route / probing for route"
-      lightningLogger.error({ err, max_fee, probingSuccess: false, success: false }, error)
-      throw new LoggedError(error)
+      const error = "Error probing for route"
```
nicolasburtey, line 177: I think there is no need to create an error message here, if this will be always the same across all the `RouteFindingError`. you can hard code the message directly within `RouteFindingError` so that you don't have to pass an error message independantly. (this apply to a lot of other error as well that are using the same error message across all their instances)

nicolasburtey: similarly for the error just above this one, `forwardToClient: false` is probably always the same across all the instance of an error

src/Lightning.ts
```diff
@@ -324,7 +320,7 @@ export const LightningMixin = (superclass) => class extends superclass {
         }
 
         if (String(payeeUser._id) === String(this.user._id)) {
-          const error = 'User tried to pay himself'
+          const error = 'User tried to pay themselves'
```
nicolasburtey, line 323: User is singular here. are you sure it should be themselves?

src/Lightning.ts
```diff
@@ -712,8 +706,7 @@ export const LightningMixin = (superclass) => class extends superclass {
 
       } catch (err) {
         const error = `issue updating invoice`
-        this.logger.error({ err, invoice }, error)
-        throw new LoggedError(error)
+        throw new DbError(error, {forwardToClient: true, logger: this.logger, level: 'error', err, invoice})
```
nicolasburtey, line 709: I think we should have 2 errors here. one for the first mongodb fetch regarding `InvoiceUser` and the second for `addTransactionLndReceipt`. so that if an error occurs we know more precisely at what level it occured.

nicolasburtey: actually the way it's structured, I also think this error could cover the case if we can't acquire the lock? not sure this try catch is actually helpful. something to give a try probably, or least add a TODO / FIXME to revisit later.

src/entrypoint/graphql.ts
```diff
@@ -357,7 +357,7 @@ export async function startApolloServer() {
       // i.e. catch-all errors will not be forwarded
       if(log = err.extensions?.exception?.log) {
         const errObj = { message: err.message, code: err.extensions.code }
-        log(errObj)
+        log({...errObj, ...err.extensions.metadata})
```
nicolasburtey, line 360: could you expend on what this is doing? (maybe also add some comment in the code to make it a bit more clear?)

src/Lightning.ts
krtk6160: Themself can be used as well (it's usage is more recent), however for formal purposes, themselves is generally preferred. See [this](https://www.lexico.com/grammar/themselves-or-themself).

nicolasburtey (commented):
a couple of comments

nicolasburtey:
also need some rebasing

src/Lightning.ts
krtk6160: I have moved the try catches inside the redlock function. Do we want to enclose the entire redlock function in another try catch?

nicolasburtey (approved):

PR #214, opened 2021-06-05
optimize script for local dev

krtk6160:
We are already updating the repos at the beginning of `initLocalTest.sh` so using a skip flag for helm dep build makes sense. On my machine, it reduced times by 50% and 66% for galoy and monitoring charts respectively. Eventually we will be removing this script, so I think the added 'ugliness' is okay for now because it provides a much faster local dev experience.

Edit: I am keeping `--skip-refresh` only for local dev because it's unavailable on CircleCI for some reason, not sure why. The very flag is missing from the `helm dep build` command despite the helm version being recent.


nicolasburtey:
pushed to master (with a slight refactoring)

PR #216, opened 2021-06-09
Fix error severity

krtk6160:


src/entrypoint/graphql.ts
```diff
@@ -244,7 +244,7 @@ const resolvers = {
 const isAuthenticated = rule({ cache: 'contextual' })(
   async (parent, args, ctx, info) => {
     if(ctx.uid === null) {
-      return new Error(`${util.inspect({ message: 'Not authorised!', request: ctx.request.body }, false, Infinity)}`)
+      throw new AuthorizationError(undefined, {logger: graphqlLogger, request: ctx.request.body})
```
nicolasburtey, line 247: oh we were returning an error which doesn't make sense.
can you verify the throw error is handled correctly? (ie: it doesn't make graphql restart?)

krtk6160: Yep, already did that. Works as expected.

src/Lightning.ts
```diff
@@ -315,8 +315,7 @@ export const LightningMixin = (superclass) => class extends superclass {
         }
 
         if (String(payeeUser._id) === String(this.user._id)) {
-          const error = 'User tried to pay themselves'
-          throw new SelfPaymentError(error, {logger: lightningLoggerOnUs})
+          throw new SelfPaymentError(undefined, {logger: lightningLoggerOnUs})
```
RobZuazua, line 318: if you never override the message field, why have a parameter for it ? 

Seems like it will be cleaner to remove this parameter until you need it

src/LightningUserWallet.ts
```diff
@@ -20,7 +21,7 @@ export class LightningUserWallet extends OnChainMixin(LightningMixin(UserWallet)
   async addEarn(ids) {
 
     if (this.user?.twilio?.carrier?.type === "voip") {
-      throw new Error("reward can only be given on non voip-based phone")
+      throw new CustomError("reward can only be given on non voip-based phone", "VOIP_REWARD_RESTRICTED", {forwardToClient: true, logger: this.logger, level: 'warn', metadata: undefined})
```
RobZuazua, line 24: I would recommend creating an enum with all your custom error messages so that you have a singular place to organize these error messages and update/reuse them.

`throw new CustomError(Errors.VOIP_BASED_PHONE,...) `

```
enum Errors {
  VOIP_BASED_PHONE = "reward can only be given on non voip-based phone",
  ...
}
```

src/OnChain.ts
```diff
@@ -100,8 +100,7 @@ export const OnChainMixin = (superclass) => class extends superclass {
         }
 
         if (String(payeeUser._id) === String(this.user._id)) {
-          const error = 'User tried to pay himself'
-          throw new SelfPaymentError(error, {logger: onchainLoggerOnUs})
+          throw new SelfPaymentError(undefined, {logger: onchainLoggerOnUs})
```
RobZuazua, line 103: same comment as above. 

Not a fan of passing undefined unless absolutely necessary

src/text.ts
```diff
@@ -161,7 +161,7 @@ export const login = async ({ phone, code, logger, ip }: ILogin): Promise<string
         user.twilio = result
         await user.save()
       } catch (err) {
-        subLogger.error({err}, "impossible to fetch carrier")
+        subLogger.warn({err}, "impossible to fetch carrier")
```
RobZuazua, line 164: what's the effect of the code failing here? 

does this have any downstream effects that an unknowing person like me should know? 

Just thinking it would be a good practice to comment or log the effects on not being able to fetch the carrier.

src/LightningUserWallet.ts
krtk6160: The mapping between error codes and error messages isn't 1 <-> 1. For the same error code, we sometimes have a different error message.

src/Lightning.ts
krtk6160: We don't override it for this particular error type, but we do for other error types. We're trying to keep the error classes consistent, which is why this parameter hasn't been removed here (I did consider it).

src/LightningUserWallet.ts
RobZuazua: I am suggesting that you have an enum to house the different string constants related to errors. 

This will allow the error objects to have many different types of error messages.

If you need a new text message, it can be a new item in the enum

```
throw new CustomError(Errors.VOIP_BASED_PHONE,...)
throw new CustomError(Errors.OTHER_ERROR,...)
```

```
enum Errors {
  VOIP_BASED_PHONE = "reward can only be given on non voip-based phone",
  ...
}
```

src/Lightning.ts
RobZuazua: Oh, I see where your head is.

I would ask: what is the value of keeping error classes consistent? 

Each Error Class is fundamentally a different object. 

Trying to add this implicit connection between your classes leads to confusion, because people like me have no context that you are doing something like that.

If you want to connect these classes, would recommend using composition.

src/text.ts
krtk6160: It's a non-critical failure. I've added a comment to make it clear.

src/LightningUserWallet.ts
krtk6160: I can see what you're suggesting here, however consolidating all the error messages in one place doesn't seem straightforward here. At the moment, some of our error messages are defined as default parameters in custom error classes, or they're scattered across the functions that throw them (this includes variable error messages).

I don't think it'll be a worthwhile improvement to move the errors from error.ts into an enum, while leaving the others as they are.

src/Lightning.ts
krtk6160: > Oh, I see where your head is.
> 
> I would ask: what is the value of keeping error classes consistent?
> 
> Each Error Class is fundamentally a different object.
> 
> Trying to add this implicit connection between your classes leads to confusion, because people like me have no context that you are doing something like that.

I disagree. The reason we're trying to keep these classes consistent in their naming as well as parameters is because it removes a lot of guessing work as to how an error class in our codebase works. All of them accept the exact same arguments and behave more or less the same way.

I agree there's scope for improvement here (especially with the degree of repetition across the error classes). 

You're welcome to take a stab at improving the error handling if you'd like. We appreciate and welcome all contributions :)

PR #221, opened 2021-06-12
fetch compute zone from env var

krtk6160:


nicolasburtey:
move thos commit to debug-circleci

PR #224, opened 2021-06-12
Improve installation

krtk6160:


.gitignore
```diff
@@ -18,7 +18,18 @@ src/tool/users_balance.csv
 src/tool/record_transactions_*.csv
 
 *.DS_Store
-configs
+charts/configs/*
+!charts/configs/lnd
+charts/configs/lnd/*
+!charts/configs/lnd/regtest.yaml
```
nicolasburtey, line 24: those files are not commited in this PR?

krtk6160: They are

.gitignore
```diff
@@ -18,7 +18,18 @@ src/tool/users_balance.csv
...
+!charts/configs/lnd/regtest-outside.yaml
+!charts/configs/lnd/localdev.yaml
+!charts/configs/lnd/localdev-outside.yaml
+
+!charts/configs/bitcoind
+charts/configs/bitcoind/*
+!charts/configs/bitcoind/regtest.yaml
+!charts/configs/bitcoind/localdev.yaml
```
nicolasburtey, line 32: I feel those config file could be in a different folder but maybe this is for another PR

charts/configs/bitcoind/localdev.yaml
```diff
@@ -0,0 +1,2 @@
+service:
```
nicolasburtey, line 1: maybe add a comment why we need that (necessary for minikube to be able to access from outside the cluster?)

nicolasburtey (approved):
just one comment but good to go for me

.gitignore
nicolasburtey: those files are also duplicated in the config repo. probably should be deleted there to avoid confusion.

krtk6160: I was planning on keeping a set in configs as well to allow us to overwrite these files without having to commit them to the public galoy repo. On second thought maybe it's something we wouldn't require, so I'll remove it.

PR #230, opened 2021-06-14
Added caveat when running locally on mac, link

keblek:
added link for minikube tutorial as well

I also used `nvm` to manage `node`, should that be added as well?

### DIFF
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -20,7 +20,7 @@ There are multiple ways to install the galoy servers. We recommend first trying 
 This will setup all the required pods. Note that this will setup the environment in **regtest** mode. Read more about regtest mode [here](https://developer.bitcoin.org/examples/testing.html#regtest-mode).
 
 #### Caveats
-1. If installing on macOS, this script will clash with the default grep, [here is a workaround](https://stackoverflow.com/questions/16658333/grep-p-no-longer-works-how-can-i-rewrite-my-searches)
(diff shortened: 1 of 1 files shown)


INSTALLATION.md
```diff
@@ -10,15 +10,18 @@ There are multiple ways to install the galoy servers. We recommend first trying
...
-1. Make sure minikube is working correctly by deploying the hello-world starter application
+1. Make sure minikube is working correctly by deploying [hello-world starter application](https://minikube.sigs.k8s.io/docs/start/)
 2. Run `git clone https://github.com/galoymoney/galoy && cd galoy`
 3. Run `yarn install`
 4. Run `yarn add-charts`
 
 This will setup all the required pods. Note that this will setup the environment in **regtest** mode. Read more about regtest mode [here](https://developer.bitcoin.org/examples/testing.html#regtest-mode).
+
+#### Caveats
+1. If installing on macOS, this script will clash with the default grep, [here is a workaround](https://stackoverflow.com/questions/16658333/grep-p-no-longer-works-how-can-i-rewrite-my-searches)
```
nicolasburtey, line 23: you're right. I forgot I had installed it:

```
$ brew list | grep grep
grep
```

maybe we can suggest installing grep with `brew install grep` for now? eventually we should get rid of this dependancy

nicolasburtey (approved):

INSTALLATION.md
keblek: Having a section for ubuntu and mac would be the way to go forward, but yeah more info is better.

PR #231, opened 2021-06-15
expanding a little on the data store being used/stored

nicolasburtey:


ARCHITECTURE.md
```diff
@@ -26,8 +30,41 @@ The graphql server can scale horizontally with additional pods.
...
+There is typically a very low load transaction wise in regard to the wallet. Typically 1 transaction per week when a rebalancing is needed.
+
+### lnd 
+
+Lnd handles all of the transactions the users make.
+Lnd stores the data in a bbolt database internally.
+
+Every successful payment within lnd is being recorded on mongodb.
+
+Offchain-wise, nothing is needed within the lnd side to present a transaction to a user after it has been executed.
```
krtk6160, line 50: This and the line below this seems a little confusing to me

krtk6160 (commented):
Just one comment otherwise LGTM

ARCHITECTURE.md
```diff
@@ -26,8 +30,41 @@ The graphql server can scale horizontally with additional pods.
...
+
+offchain-wise, nothing is needed within the lnd side to present a transaction to a user after it has been executed.
+onchain-wise, it's possible that a user shared an onchain address. this is currently tied to lnd. (we intend to not use lnd for onchain transaction to remove this dependancy and simplify lnd instance recyling)
+
+### Mongodb
+
+Mongodb is the database storing all users transactions. It's the source of truth for the customers' accounts.
+
+To achieve strong consistency, it's currently setup with:
+- Write access need to be valided by 2-out-of-3 pods before being considered succesful
```
bodymindarts: I have been reading a bit in the MongoDB docs to see what consistency guarantees there are. It appears with Majority write concern (as is being used in the deploy scripts) we are indeed guaranteed that the write only succeeds in a replica set once a majority of nodes have persisted the data. (https://docs.mongodb.com/manual/reference/write-concern/)

The problem as far as I can tell is that reads have no such guarantee. By default the read concern is 'local' which will only query the primary or 'available' if querying a secondary. Only with read concern 'linearisable' is there guarantee of consistent reads where all data that is being read has the guarantee of being written to a majority of nodes. https://docs.mongodb.com/manual/reference/read-concern-linearizable/ .

I cannot currently see any place in the code base where this is configured so my assumption is we aren't using it. I also do not have enough awareness of the exact order of processing in the code to asses what inconsistencies we risk via this behaviour. But it doesn't make me feel confident about things working correctly at scale.

ARCHITECTURE.md
```diff
@@ -26,8 +30,41 @@ The graphql server can scale horizontally with additional pods.
...
+There is no transaction used yet (the main inconsistency if this were to happen would be between the journal entries and the transaction entries). A [contribution on this](https://github.com/flash-oss/medici/issues/23) is welcome. It has not been a focus yet because there hasn't been any issue on this yet. 
+
+### Redis
 
-Mongodb is the main database which is storing the transactions. It's the source of truth for the customers' accounts.
+Redis is used a distributed lock, such that the number of replicas for the node/graphql can be scaled as needed.
+Redis is also used for query rate limiting , and to store ephemeral data such as result of route probing.
 
-It's currently setup by default with auto replication on a 2-out-of-3 setup, and writes are only accepted as successful when a majority (2) of the nodes have 
+Redis can be though as a cache. The data can be deleted as needed (it would reset the lock, so some attention needed to be paid on this if some live queries are running)
```
bodymindarts: Redis is explicit about not providing any consistency guarantees:

>> Redis Cluster is not able to guarantee strong consistency. In practical terms this means that under certain conditions it is possible that Redis Cluster will lose writes that were acknowledged by the system to the client.
https://redis.io/topics/cluster-tutorial

Using it as a distributed lock doesn't seem a viable solution when running at scale. 2 concurrent attempts to acquire a lock may both complete successfully with unknown consequences.

bodymindarts:
I think the change of the wording in this document is good and I would merge it as it has clarified some things. In regards to the persistence solutions (Redis / Mongo) we should be clear that we currently don't currently guarantee consistency.

We need to discuss what strategies we have to move toward a solution that is guaranteed to remain consistent even when running at scale.

ARCHITECTURE.md
bodymindarts: Actually the write side also has problems. We are using `majority` concern but not specifying the `j` option to only acknowledge a write when the operation has persisted to disk https://docs.mongodb.com/manual/reference/write-concern/ is set to false.

Hence we don't have consistent writes either... just in memory.

nicolasburtey:
@bodymindarts I'm merging it. would you mind creating an issue by coping your different inputs highlighted here so we can discuss those over there?

ARCHITECTURE.md
nicolasburtey: this is the page to look at for the lock we are using: https://redis.io/topics/distlock

bodymindarts:
Done https://github.com/GaloyMoney/galoy/issues/232

PR #234, opened 2021-06-15
Remove all trailing whitespaces and add missing newlines at EOF

samerbuna:
NO CODE CHANGE


nicolasburtey:
may be relevant to add a doc a code policy with the trailing whitespaces and missing newlines altogether with this PR. and maybe other stuff like variable name convention.

I don't have any strong opinion on this. but there is currently not much consistency in this repo.

bodymindarts:
I think removing all needles whitespace and other noise from the repo is a good idea but we should add a linter config in the process so that it will be auto-enforced in the future.

A project-wide enforced code style is a good thing to reduces the amount of non functionality-critical discussions. 
A 1 time cleanup however doesn't solve this.

Perhaps @samerbuna could add a https://palantir.github.io/tslint/ or something similar.

bodymindarts:
Looks like there already is a [task](https://github.com/GaloyMoney/galoy/blob/main/package.json#L4) for this. Its just not being enforced or not configured.

bodymindarts:
@samerbuna can you join our [public slack](https://join.slack.com/t/galoymoney-workspace/shared_invite/zt-rvnhsdb5-72AZCD_jzw6_Q05aCs0SgA) to discuss.

samerbuna:
Closing this PR as it'll be part of a new PR that adopts typescript-eslint

PR #239, opened 2021-06-15
Add eslint configurations (based on old tslint) and fix all straightforward issues

samerbuna:
Added the following new rules:
 - no-unused-vars
 - comma-dangle (now required)
 - semi (not allowed any more)
 - no-trailing-spaces
 - eol-last

Temporarily disabled 3 existing rules to get this big commit merged ASAP
 - no-shadow
 - no-unnecessary-type-assertion
 - no-floating-promises


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "galoy",
   "scripts": {
-    "lint": "tslint --project tsconfig.json",
+    "lint": "eslint src/ --ext .js,.jsx,.ts,.tsx",
```
nicolasburtey, line 4: don't think we have any code in .js

package.json
```diff
@@ -35,8 +35,10 @@
     "@grpc/grpc-js": "^1.2.11",
     "@grpc/proto-loader": "^0.5.6",
     "@profusion/apollo-validation-directives": "^2.1.3",
+    "apollo-server-errors": "^2.5.0",
     "apollo-server-express": "^2.21.2",
     "axios": "^0.21.1",
+    "bip66": "^1.1.5",
```
nicolasburtey, line 41: this was a missing dependency provided by another package?

src/Lightning.ts
```diff
@@ -581,7 +577,7 @@ export const LightningMixin = (superclass) => class extends superclass {
     // we're doing the the Transaction.find after the lock to make sure there is no race condition
     // note: there might be another design that doesn't requiere a lock at the uid level but only at the hash level,
     // but will need to dig more into the cursor aspect of mongodb to see if there is a concurrency-safe way to do it.
-    await redlock({ path: this.user._id, logger: lightningLogger, lock }, async (lock) => {
+    await redlock({ path: this.user._id, logger: lightningLogger, lock }, async () => {
```
nicolasburtey, line 580: lock was not used I guess?

src/__mocks__/exchange.ts
```diff
@@ -1,19 +1,21 @@
...
-            exchange: expect.any(Number),
-        }),
-        BTC: expect.objectContaining({
-            total: expect.any(Number), // how to set toBeGreaterThanOrEqual(0) ?
-            exchange: expect.any(Number),
-            onchain: expect.any(Number),
-            offchain: expect.any(Number),
-        }),
-    })
+// import { getBalances } from "../exchange"
```
nicolasburtey, line 3: I'm wondering why this test even passed. I think it can be deleted I'm not sure what it was doing but clearly outdated and no longer used.

src/__tests__/01a-price.ts
```diff
@@ -107,17 +102,17 @@ beforeAll(async () => {
...
 it('test updating price', async () => {
   await price.update()
   // test it doesn't throw an error
 })
 
-it('test fetching last 24 hours', async () => {
-  const priceHistory = await price.lastCached()
-  // expect(priceHistory.length).toBe(25)
-})
+// it('test fetching last 24 hours', async () => {
```
nicolasburtey, line 112: I guess this test was not very useful as is but why commenting it?

src/__tests__/08-specter.ts
```diff
@@ -66,46 +66,47 @@ it('createWallet', async () => {
...
-  
-  // console.log({initBitcoindBalance, bitcoindBalance, lndBalance, initLndBalance})
-  // expect(bitcoindBalance).toBe(initBitcoindBalance - sats)
-  // expect(lndBalance).toBe(initLndBalance + sats)
-
-  
-  // const balance = await SpecterWallet.listWallets()
-  // expect(balance).toBe(0)
-})
+// TODO: Fix or remove. Expectations were commented out
```
nicolasburtey, line 91: if there is no `expect()` the test is failing with the new setup?

src/__tests__/09d-text-twilio.ts
```diff
@@ -27,41 +27,41 @@ afterAll(async () => {
...
 
 
 it('test fetching carrier and adding this info to User', async () => {
-  const getCarrier = (_) => new Promise(function (resolve, reject) { resolve(resp) });
+  const getCarrier = () => new Promise(function (resolve) { resolve(resp) })
 
   try {
     const phone = "+1650555000"
-    const result = await getCarrier(phone)
+    const result = await getCarrier()
```
nicolasburtey, line 50: phone should be provided I think

nicolasburtey: oh it's a test, maybe that doesn't matter here.

src/__tests__/09d-text-twilio.ts
```diff
@@ -27,41 +27,41 @@ afterAll(async () => {
 	await mongoose.connection.close()
 })
 
-const phone = "add phone number here with extension (ie: +1...)"
+// const phone = "add phone number here with extension (ie: +1...)"
```
nicolasburtey, line 30: ok I see why phone is not provided

src/csvAccountExport.ts
```diff
@@ -35,47 +35,48 @@ export class CSVAccountExport {
...
-    }))
+    const ledger = await MainBook.ledger({
+      account,
+    })
 
-    transactions.forEach(tx => tx = tx.toObject())
-    transactions.forEach(tx => tx.meta = JSON.stringify(tx.meta))
-    transactions.forEach(tx => tx.unix = moment(tx.datetime).unix())
-    transactions.forEach(tx => tx.date = moment(tx.datetime).format('L'))
+    const transactions = ledger.results.map(tx => {
```
nicolasburtey, line 70: 👍

nicolasburtey:
couple of feedback. will finish parsing tomorrow

.eslintrc.json
```diff
@@ -0,0 +1,64 @@
...
+    "no-void": "error",
+    "prefer-const": "error",
+
+    "no-unused-vars": "error",
+    "comma-dangle": ["error", "always-multiline"],
+    "semi": ["error", "never"],
+    "no-trailing-spaces": "error",
+    "eol-last": "error"
+
+    // TODO: Fix or remove
```
bodymindarts, line 53: What is the purpose of keeping this comment here?

bodymindarts (commented):
On merging this PR it would be great to start enforcing this so that all contributors use the same code style and we never need to discuss style issues.

It would be great if the prettier command `yarn prettier -w .` would be successful so that integration with IDEs / language servers will automatically keep everything nice.

If its too much to handle in 1 PR we can also merge this and then continue.

tsconfig.json
```diff
@@ -18,8 +18,10 @@
   "compileOnSave": true,
   "include": [
     "src/entrypoint/*.ts",
-    "src/__tests__/*.ts",
+    "src/postman/*.ts",
+    "src/**/__tests__/*.ts",
+    "src/__mocks__/*.ts",
     "src/debug/*.ts",
-    "src/**/__tests__/.ts"
+    "src/*.d.ts",
```
nicolasburtey, line 25: is there a reason to include mock and .d.ts in the include here? I would only the entry points might be needed

src/utils.ts
```diff
@@ -143,8 +138,8 @@ export const caseInsensitiveRegex = (input) => {
 export const inputXOR = (arg1, arg2) => {
   const [[key1, value1]] = Object.entries(arg1)
   const [[key2, value2]] = Object.entries(arg2)
-  if(!(!value1 != !value2)) {
-    throw new ValidationError(`Either ${key1} or ${key2} is required, but not both`, {logger: baseLogger});
+  if(!(!value1 !== !value2)) {
```
nicolasburtey, line 141: @krtk6160 could there be a reason for the `==` here?

src/text.ts
```diff
@@ -151,11 +151,11 @@ export const login = async ({ phone, code, logger, ip }: ILogin): Promise<string
 
     // TODO
     // if (yamlConfig.carrierRegexFilter)  {
-    // 
+    //
     // }
     //
     // only fetch info once
-    if (user.twilio.countryCode == undefined) {
+    if (user.twilio.countryCode === undefined) {
```
nicolasburtey, line 158: may worth check this is not null returned

src/interface.ts
```diff
@@ -7,18 +7,18 @@ export interface Balances {
 }
 
 export interface IWallet {
-  getBalances(lock?): Promise<Balances>
+  getBalances(): Promise<Balances>
```
nicolasburtey, line 10: there might be some issue with the type here. maybe IWallet as a type is not being used, but getBalances may require a lock when it's called within a thread that have already acquired the lock.

here is why (from a design perspective at least):
- if a user call `getBalances()` directly, then no lock is pass to the function, and getBalances will acquire the lock by itself.
- a function like `pay()` will need to know the balance of the user during the payment process. the lock would already been acquired prior to `getBalance()`, so it's passing it `lock` downstream to `getBalance()`

nicolasburtey (commented):
thanks for taking the initiative @samerbuna this is appreciated and much need.
I've multiple questions, many arise from the fact I'm not a typescript expert (but want to move to a fully typed codebase, so I'm learning)

nicolasburtey:
> If its too much to handle in 1 PR we can also merge this and then continue.

I would think enforcing the rules can be in another step.

we also want to have some guideline into how configure our IDE to match the spec we have here. me and @krtk6160 are running on vscode. not sure for @bodymindarts ?

some context to @bodymindarts, we had a discussion with @samerbuna in dm, I shared with him that I'm not the best fan of semicolons. I think it's reduce readability. but semi colons might be default with a linter, which may be why there is some custom rules.

bodymindarts:
I'm a vim user but it shouldn't matter. I'm pretty sure `prettier` can integrate with any editor.

Regarding semi-colons yes/no. I would also prefer not to have any but as long as `prettier` adds them automatically (on save for example) I really don't care. Just want the minimum amount of maintenance / discussion around style as possible moving forward. So whatever ts community thinks are widely adopted defaults we should use IMO with perhaps only a handful of exceptions.
Aligning with wider ts community is to lower cognitive overhead when onboarding.

samerbuna:
I prefer doing the mass prettier formatting in a separate PR. That will be a no-code change commit.

Regarding semi-colons, there is no harm in leaving them off as long as linting and type-checking are actively enforced. Prettier can be configured to remove semi-colons on save.

package.json
samerbuna: I'll change that to just .ts

package.json
samerbuna: Yep

src/Lightning.ts
samerbuna: No. and it was not used in so many other places.

src/__mocks__/exchange.ts
samerbuna: I think it was not run at all.

src/utils.ts
krtk6160: Don't think there'll be an issue with this change. At the moment this is only used by the admin-panel (afaik) so there shouldn't be any impact on the users in case things go wrong.
Need to test this out, but there's an issue with the local dev setup scripts right now so unable to do so. Working on fixing that.

src/__tests__/01a-price.ts
samerbuna: Hoping someone will revive it one day :)

src/__tests__/08-specter.ts
samerbuna: Tests pass with no expectations.

.eslintrc.json
```diff
@@ -0,0 +1,64 @@
...
+    "es6": true,
+    "node": true
+  },
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "project": "tsconfig.json",
+    "sourceType": "module"
+  },
+  "plugins": ["eslint-plugin-import", "@typescript-eslint"],
+  "rules": {
```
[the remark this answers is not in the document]
samerbuna: I prefer the recommended rules but started with the existing ones. Let me see how much of trouble switching to recommended will be in the codebase.

.eslintrc.json
samerbuna: These were existing TSlint rules with existing code violations. I was not sure if we want to keep them or drop them.

bodymindarts: Drop them

tsconfig.json
samerbuna: Usually, `src/**/*` is included for typescript checking. `typescript-eslint` was not picking up on some files with the existing includes.

src/utils.ts
samerbuna: The expression here is a triple-negative one! I'd love to flip it to just one negative.

src/text.ts
samerbuna: We can add a null check here to be safe.

bodymindarts:
I have no more comments. `yarn lint` exits `0` for me. If @nicolasburtey is happy we should merge and follow up with the `prettier` PR.

.eslintrc.json
nicolasburtey: 🙄

tsconfig.json
nicolasburtey: here in my mind I had placed the entrypoints of the application. so that when we do the transpilation, only the necessary files are being "copied" over. 

maybe there should be no reason for files to exist if not imported eventually.

src/interface.ts
nicolasburtey: `getBalances` in `UserWallet` might not be type correctly? (`IWallet`?)
those types has not been well maintained and would need more attention!

nicolasburtey:
> I have no more comments. `yarn lint` exits `0` for me. If @nicolasburtey is happy we should merge and follow up with the `prettier` PR.

just want to make the PR is not breaking the test, but for some reason circleci doesn't run the PR. probably a config to fix on the github side.

bodymindarts:
@samerbuna can you rebase the branch off of https://github.com/GaloyMoney/galoy/tree/docker-compose and run the integration tests?

samerbuna:
This PR now passes on the `docker-compose` branch (but it does not include it). I've also merged in the latest in `main`.

bodymindarts:
Thats awesome thank you. Really appreciate the work your doing with this cleanup.

nicolasburtey:
merging it

PR #242, opened 2021-06-16
Fix local dev scripts

krtk6160:


nicolasburtey:
what was the issue?

bodymindarts:
Trying to run the tests Redis wasn't coming up properly on Minikube.
I'm just creating a `docker-compose` file now to simplify local dev. That way devs don't need to understand all the complexity of k8s / Minikube + helm etc just to run the tests.

PR #248, opened 2021-06-18
do recursive merging for the config file. 

nicolasburtey:
so that if the custom yaml override some property but not all property of a nested object, the missing property would be fetch from the default.yaml config file

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import yaml from 'js-yaml';
+import _ from "lodash";
 import { baseLogger } from "./logger";
 
 let defaultContent = fs.readFileSync('./default.yaml', 'utf8');
@@ -16,4 +17,4 @@ try {
   }
 }
 
-export const yamlConfig = { ... defaultConfig, ...customConfig }
+export const yamlConfig = _.merge(defaultConfig, customConfig)
--- a/src/text.ts
+++ b/src/text.ts
@@ -71,7 +71,7 @@ export const requestPhoneCode = async ({ phone, logger, ip }: {phone: string, lo
   try {
     const ipDetails = await fetchIP({ip})
     if(ipDetails?.type == "VPN") {
-      logger.warn({ip}, "RequestPhoneCode: called by a VPN IP")
+      logger.warn({ip, type: ipDetails?.type}, "RequestPhoneCode: IP type not supported")
       return false
     }
   } catch(err) {


src/config.ts
```diff
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import yaml from 'js-yaml';
+import _ from "lodash";
```
bodymindarts, line 3: naming this import `_` is hard to understand later in the code. How about `import { merge } from "loadsh";`?

bodymindarts (commented):
Just a small naming nitpick

src/config.ts
bodymindarts: Oh I see using `_` seems to be common for lodash https://www.blazemeter.com/blog/the-correct-way-to-import-lodash-libraries-a-benchmark. So then its fine

krtk6160 (approved):

krtk6160:
A merging system where,
A:
```
x:
- y: abc
```
and B:
```
x:
- y: pqr
```

when merged give:
```
x:
- y: abc
- y: pqr
```

Might be useful if we want to merge `test_accounts`

src/config.ts
nicolasburtey: yup it's the standard

nicolasburtey:
> A merging system where,
> A:
> 
> ```
> x:
> - y: abc
> ```
> 
> and B:
> 
> ```
> x:
> - y: pqr
> ```
> 
> when merged give:
> 
> ```
> x:
> - y: abc
> - y: pqr
> ```
> 
> Might be useful if we want to merge `test_accounts`

is this a question?

what you describe is what should happen.

but if you have :
```
x:
 y: a
```
merged with 

```
x:
 y: b
```
then you would have
`{x: y: {b}}`

krtk6160:
> > A merging system where,
> > A:
> > ```
> > x:
> > - y: abc
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and B:
> > ```
> > x:
> > - y: pqr
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > when merged give:
> > ```
> > x:
> > - y: abc
> > - y: pqr
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Might be useful if we want to merge `test_accounts`
> 
> is this a question?

A suggestion

> what you describe is what should happen.

Nope, that does not happen with this PR. I just tested it locally. We get this:
```
// merged
x:
  - y: pqr
```


But we don't want that for the `test_accounts` data structure at least, because each entry there would be a separate account.

PR #250, opened 2021-06-19
Refactor ip filtering

krtk6160:


src/utils.ts
```diff
@@ -180,3 +180,7 @@ export const fetchIPDetails = async ({ip, user, logger}): Promise<void> => {
     }
   }
 }
+
+export const isIPTypeAllowed = ({type}) => !yamlConfig.blacklistedIPTypes?.includes(type)
```
nicolasburtey, line 184: you should stick with either allowed or blacklisted but not mix both

nicolasburtey (approved):

Issue #18, opened 2020-06-04
use public/private key pair instead of symmetrical key for JWT_Secret 


nicolasburtey:
we should be using a private/public key pair, so that the client can check the signature of the JWT token being generated. This may prevent an attack in the middle attack where a client could otherwise be connected to a forged backend (for some reason like a DNS attack).

nicolasburtey:
irrelevant as switching to Ory will solve the issue

Issue #19, opened 2020-07-31
being able to regenerate auth token periodically


nicolasburtey:

nicolasburtey:
moving to Ory will solve the issue

PR #252, opened 2021-06-21
Use RSA to sign jwt token (Fixes GaloyMoney/galoy#18)

dolcalmi:
changed:
- jwt signing from HS256 to RS256
- jwt env variable from randAlpha to genPrivateKey

added:
- endpoint /.well-known/jwks.json (based on auth0 implementation)
- node-jose library to generate JSON Web Key Sets (jwks)


src/entrypoint/graphql.ts
```diff
@@ -376,11 +377,17 @@ export async function startApolloServer() {
...
     expressJwt({
       secret: process.env.JWT_SECRET,
-      algorithms: ["HS256"],
+      algorithms: ["RS256"],
       credentialsRequired: false,
       requestProperty: 'token',
     }))
 
+  // JWKS endpoint to allow clients to verify jwt tokens
+  app.get("/.well-known/jwks.json", async function(req, res) {
```
nicolasburtey, line 386: is the `.well-known` the standard way to do it? all of our API interaction currently rely on graphql, so I think that would make sense to provide the public key through the same API.

nicolasburtey:
Great job for this first PR.

I've run the test locally and it passes all the tests, although I don't think we have any test that are focus on the JWT

One of the main things that we need to think of, as part of this PR, is the key upgrade from the existing users. A few things that are coming in mind:

- existing mobile wallet will have the HS256 key in their keystore. so when they will call the graphql API, this will currently fail. 
- there should be a way to be both compatible with HS256 and RS256 while the mobile upgrade is ongoing
- mobile upgrade should consist of several items:
if the mobile app is on the current version, then it should continue to work with the HS256 (until we decide to force an update to remove the deprecated HS256 API)
if the mobile has download a new version compatible with RS256, and detect that it still has the "legacy" JWT, then there should be an API call to the backend to ask for an new JWT with the RS256 key.

I think this "refresh" token strategy could probably be thought in the context of this issue: https://github.com/GaloyMoney/galoy/issues/19. currently token doesn't have an expiry, but making sure token are only valid for a limited period of time until a refresh is needed would be a good security practice.

src/entrypoint/graphql.ts
bodymindarts: There are many systems I know that rely on a rest api to query the public key. This would probably enable compatibility with 3rd party auth libraries. The path isn't standard but its what `auth0` currently exposes.

dolcalmi: let me know which one you want...

dolcalmi:
ok.. in theory we can add both algorithms to the middleware but there is a warning in the library
```
Do not mix symmetric and asymmetric (ie HS256/RS256) algorithms: Mixing algorithms without further validation can potentially result in downgrade vulnerabilities.
```
I just checked the code (express-jwt and jsonwebtoken)  and basically it does the same that we must do (validate alg field in jwt token header) so I think there is no problem.

let me know if it is ok to proceed

nicolasburtey:
> Mixing algorithms without further validation can potentially result in downgrade vulnerabilities.

I'm not sure what this is supposed to mean. that the security we get is as best as least secured one? this would make sense/be obvious. but maybe they're saying something else?

I would just try to see how this works as a starting point.

src/entrypoint/graphql.ts
nicolasburtey: on the mobile side we use apollo-client. maybe the easiest would be see if a particular option would have a smoother integration.

dolcalmi:
> Great job for this first PR.
> 
> I've run the test locally and it passes all the tests, although I don't think we have any test that are focus on the JWT
> 
> One of the main things that we need to think of, as part of this PR, is the key upgrade from the existing users. A few things that are coming in mind:
> 
> * existing mobile wallet will have the HS256 key in their keystore. so when they will call the graphql API, this will currently fail.
> * there should be a way to be both compatible with HS256 and RS256 while the mobile upgrade is ongoing
> * mobile upgrade should consist of several items:
>   if the mobile app is on the current version, then it should continue to work with the HS256 (until we decide to force an update to remove the deprecated HS256 API)
>   if the mobile has download a new version compatible with RS256, and detect that it still has the "legacy" JWT, then there should be an API call to the backend to ask for an new JWT with the RS256 key.
> 
> I think this "refresh" token strategy could probably be thought in the context of this issue: #19. currently token doesn't have an expiry, but making sure token are only valid for a limited period of time until a refresh is needed would be a good security practice.

Just finish the update:

- After run helm upgrade: 
  - if there is no secret then a new RSA private key is created (rsaSecret field is not created)
  - if there is a secret and is not RSA private key then rsaSecret field is added with a RSA private secret
  - if there is a secret and is RSA private key then no changes are made to secret (if rsaSecret exists, it is removed)
- JWT:
  - the middleware uses "secret callback" to check "alg" in jwt header. This add support for both algorithms (HS256 and RS26)
  - all new tokens are generated with RS256
- JWKS: there is no particular implementation in apollo-client. `.well-known` is the "most standard" way.
- Mobile app: with this implementation no changes are required

bodymindarts:
Stale

PR #255, opened 2021-06-21
Rename lnd deployment

krtk6160:


initLocalTest.sh
```diff
@@ -131,7 +132,7 @@ fi
 set +e
```
nicolasburtey, line 132: redundant?

nicolasburtey (approved):
LGTM

initLocalTest.sh
krtk6160: Nope, the script will exit if the templates folder is empty (which will be for people running it that don't have access to our configs repo)

PR #259, opened 2021-06-21
Make some ESLint/tsc cleanups

samerbuna:
- Remove husky
- Rename eslint task to eslint-check and add a tsc-check task (there will also be a prettier-check)
- Remove unnecessary TS configs and eslint ignore file


bodymindarts:
How come `.eslintignore` was redundant? And what does `tsc-check` do? Is it a dry-run of the type-script compiler to check for syntax errors?

samerbuna:
Missed adding the `ignorePatterns` property. I just did to ignore top-level .js and lib/coverage. node_modules is ignored by default.

samerbuna:
The `tsc-check` task is a dry-run of all type checks.

PR #260, opened 2021-06-22
Configure Prettier and add a check task for it

samerbuna:
Also remove the ESLint formatting rules as they'll now be enforced through Prettier

(This commit should fail the GitHub Action check as Prettier is not run yet)


.gitignore
```diff
@@ -1,10 +1,3 @@
-## Compiled JavaScript files
-**/*.js
```
krtk6160, line 2: Why is this being removed?

krtk6160 (commented):
One question on the `.gitignore` changes

.gitignore
samerbuna: Too generic. It was causing top-level .js files to be ignored. The lib folder is already ignored

.prettierignore
```diff
@@ -1,2 +1,5 @@
-build
 coverage
+lib
+tmp
+
+*.yaml
```
bodymindarts, line 5: Need to add the whole `charts` directory here.

samerbuna: Added.

bodymindarts:
Can we remove `*.md` and `*.yml`. Or simply _only_ check `*.{js,ts,json}` ie. javascript related files? Also can we have a task to execute the prettier command and a section in `DEV.md` regarding the usage of prettier (can also be a separate PR).

bodymindarts:
If I run
```
$ yarn prettier . -w
$ git diff samerbuna/prettier
```
On the commit before the one where you push all the changes I have this diff:
```diff --git a/.eslintrc.json b/.eslintrc.json
index 8f2a6906..86896c5e 100644
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -33,7 +33,6 @@
     "no-throw-literal": "error",
     "no-void": "error",
     "@typescript-eslint/no-unused-vars": "off", // Duplicate
-    "@typescript-eslint/no-extra-semi": "off", // Prettier work
 
     // Temporarily disabled recommended rules
     "@typescript-eslint/explicit-module-boundary-types": "off",
diff --git a/.gitignore b/.gitignore
index ce8f43e2..9674501f 100644
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,6 @@ src/tool/records.csv
 src/tool/users_balance.csv
 src/tool/record_transactions_*.csv
 
-.vscode
 *.DS_Store
 charts/configs/*
 !charts/configs/lnd
diff --git a/.prettierignore b/.prettierignore
index 610f0593..bd9856a3 100644
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,7 +1,5 @@
 coverage
 lib
 tmp
-charts
-.vscode
 
 *.yaml
diff --git a/src/__tests__/09b-integration.ts b/src/__tests__/09b-integration.ts
index 81497ba7..4e08da4c 100644
--- a/src/__tests__/09b-integration.ts
+++ b/src/__tests__/09b-integration.ts
@@ -45,8 +45,8 @@ it("rate limit limiterRequestPhoneCode", async () => {
   }
 
   try {
+    // @ts-expect-error: TODO
     const {
-      // @ts-expect-error: TODO
       errors: [{ code }],
     } = await mutate({ mutation, variables: { phone } })
     expect(code).toBe("TOO_MANY_REQUEST")
diff --git a/src/schema.graphql b/src/schema.graphql
index a032dce2..3cc22060 100644
--- a/src/schema.graphql
+++ b/src/schema.graphql
@@ -169,8 +169,9 @@ type Query {
     # @deprecated
     buildVersion: String
     # @deprecated
-    os: String # @deprecated
-  ): BuildParameter
+    os: String
+  ): # @deprecated
+  BuildParameter
   nodeStats: NodeStats
 
   getLastOnChainAddress: LastOnChainAddress
```

Shouldn't we arrive at the same exact formatting?

samerbuna:
I rebased the branch for clarity and added a docs section. The first 2 commits are for configurations/docs. The last commit is the no-change Prettier run, except:

- The ts-expect-error needs to be moved post formatting.
- For some reason, schema.graphql changes on the second Prettier run. Seems like a Prettier bug.

bodymindarts:
Okay even after running
```
yarn prettier -w .
```
twice I'm still getting this diff:
```
diff --git a/src/__tests__/09b-integration.ts b/src/__tests__/09b-integration.ts
index 81497ba7..4e08da4c 100644
--- a/src/__tests__/09b-integration.ts
+++ b/src/__tests__/09b-integration.ts
@@ -45,8 +45,8 @@ it("rate limit limiterRequestPhoneCode", async () => {
   }
 
   try {
+    // @ts-expect-error: TODO
     const {
-      // @ts-expect-error: TODO
       errors: [{ code }],
     } = await mutate({ mutation, variables: { phone } })
     expect(code).toBe("TOO_MANY_REQUEST")

```

bodymindarts:
Okay now I understand your comment:
> The ts-expect-error needs to be moved post formatting.
Why doesn't it move each time once you have moved it manually?

samerbuna:
> Why doesn't it move each time once you have moved it manually?

It affects the line right after it (which Prettier changes)

PR #261, opened 2021-06-22
Fix no-non-null-assertion and most of no-explicit-any

samerbuna:


src/utils.ts
```diff
@@ -120,7 +120,10 @@ export function timeout(delay, msg) {
   })
 }
 
-export async function measureTime(operation: Promise<any>): Promise<[any, number]> {
+// FIXME: This function is not used anywhere. Delete?
```
krtk6160, line 123: This can be deleted

src/__tests__/02b-onchain-receive.ts
```diff
@@ -152,7 +152,10 @@ it("identifies unconfirmed incoming on chain txn", async () => {
   expect(sendNotification.mock.calls[0][0].data.type).toBe("onchain_receipt_pending")
 
   const satsPrice = await getCurrentPrice()
-  const usd = (btc2sat(amount_BTC) * satsPrice!).toFixed(2)
+  if (!satsPrice) {
```
bodymindarts, line 155: Is this explicit enough? Does it only check for `null`/`undefined`?

src/ln-service.d.ts
```diff
@@ -1 +1,3 @@
 declare module "ln-service"
+
+type primitive = string | boolean | number
```
krtk6160, line 3: Any reason why this should be here and not in `types.ts`?

src/__tests__/02c-onchain-send.ts
```diff
@@ -170,8 +170,12 @@ it("Sends onchain payment _with memo", async () => {
   const { address } = await createChainAddress({ format: "p2wpkh", lnd: lndOutside1 })
   const paymentResult = await userWallet0.onChainPay({ address, amount, memo })
   expect(paymentResult).toBe(true)
-  const txs = await userWallet0.getTransactions()
-  expect((first(txs) as any).description).toBe(memo)
+  const txs: Record<string, string>[] = await userWallet0.getTransactions()
+  const firstTxs = first(txs)
+  if (!firstTxs) {
+    throw Error("No transactions found")
```
bodymindarts, line 176: Do we really need to throw an Error here? Isn't there a more elegant way to fail the test at this point? Or can move the check out of the test code into the production code and be explicit / safer with what type we are returning? (Same questions apply to the change higher up).

src/__tests__/02c-onchain-send.ts
```diff
@@ -184,12 +188,22 @@ it("makes onchain on-us transaction with memo", async () => {
...
+  }
+  expect(firstTxs.description).toBe(memo)
 
   // receiver should not know memo from sender
   const txsUser3 = await userWallet3.getTransactions()
-  expect((first(txsUser3) as any).description).not.toBe(memo)
+  firstTxs = first(txsUser3)
+  if (!firstTxs) {
+    throw Error("No transactions found")
```
bodymindarts, line 204: Same issue as above. Is this the idiomatic way to fail a test? Can't we guarantee the return type to be not nil?

src/notifications/payment.ts
```diff
@@ -29,7 +29,7 @@ export const transactionNotification = async ({
   let title
 
   if (satsPrice) {
-    const usd = (amount * satsPrice!).toFixed(2)
+    const usd = (amount * satsPrice).toFixed(2)
```
bodymindarts, line 32: What is the implication of removing the `!` here?

src/utils.ts
bodymindarts: Isn't there a linting option to flag dead code? Might be good to enforce that we don't have any unused code on `main` branch.

bodymindarts (changes requested):

src/notifications/payment.ts
nicolasburtey: `!` after a variable is a typescript directive to avoid type checking

src/__tests__/02c-onchain-send.ts
nicolasburtey: my guess:

`first` can return an object or `null` if `firstTxs` array is empty. so because we exist the loop if `firstTxs` is null, then the now after this `if` that `firstTxs` is not null. 

my (bad) workaround, as highlighted below, was casting the result of the array as `any` (either with `any` or the use of `!`)

src/utils.ts
samerbuna: I see an eslint plugin but it has not been updated in 3 years. There are other tools that can do that though. I'll research this some more.

src/ln-service.d.ts
samerbuna: I actually thought this was a global types file. My bad. Will fix.

I think we need to move all type definitions to a global .d.ts file though. I'll do that in a separate PR.

src/__tests__/02b-onchain-receive.ts
samerbuna: Well. JS is messed up, so this also check for 0, false, and NAN. The exception in this test actually used to ignore a falsy value in `satsPrice`. I think the test should fail in that case. No?

src/__tests__/02c-onchain-send.ts
samerbuna: I wish TypeScript would respect tests expectations, but it does not. I assumed the test HAS to have at least one transaction here, otherwise it should fail.

src/__tests__/02c-onchain-send.ts
samerbuna: Is it never null? That would be cleaner.

src/notifications/payment.ts
samerbuna: In this particular case, it's not significant since there is a `satsPrice` check.

src/__tests__/02c-onchain-send.ts
bodymindarts: Thats what I mean can't we encode this expectation in form of a test assertion `expect(firxtTxs).notTo(be(null))` or something like that.

samerbuna: Yes, but unfortunately, TypeScript would still complain that's the variable might be undefined.

src/types.ts
```diff
@@ -1,13 +1,19 @@
...
 export type ISuccess = boolean
 
+export type Logger = PinoLogger
+
+export type Primitive = string | boolean | number
+
 export interface ILightningWalletUser {
-  user: any
-  logger: any
+  // FIXME: Add a type for user here.
```
vindard, line 14: Adding a potential approach here for typing `User`: https://tomanagle.medium.com/strongly-typed-models-with-mongoose-and-typescript-7bc2f7197722

krtk6160: Thanks for sharing that article @vindard. I remember we referred to it when thinking about typing mongoose a few months ago, but only now I noticed that the first section makes a reference to `typegoose`. They don't consider it because it 'adds too much abstraction'.

Can @bodymindarts and @samerbuna also look at that library and share their opinion on it, and whether we can consider using it?

bodymindarts: From reading the article my opinion is:
- using interfaces to add typing to the objects returned from the database 👍 
- using objects returned from the database as a place for domain functions 👎 

There is a big misunderstanding when talking about "the model". What model? Persistence model? Data model? Domain model? These are all distinct things and when things get complex need to be treated explicitly.

The current approach to add domain logic to the database entities is very common and okay-ish in the short term for small projects.... but long term it has a lot of negative consequences.

Anyway that's a topic to tackle in the future. For now adding any typing (for example via the interface approach mentioned in the article) would be a win!

PR #267, opened 2021-06-25
Reward users with a username via pushPayment

krtk6160:
At the moment, when a user completes a quiz question, a new invoice is generated from their wallet and immediately paid by the funder wallet. This is inefficient because it puts unnecessary load on LND. We can instead make pushPayments to users that have a username set, whereas for users without a username we can continue with the older invoice approach.

### DIFF
--- a/src/LightningUserWallet.ts
+++ b/src/LightningUserWallet.ts
@@ -44,8 +44,18 @@ export class LightningUserWallet extends OnChainMixin(LightningMixin(UserWallet)
         )
 
         if (userPastState.earn.findIndex((item) => item === id) === -1) {
-          const invoice = await this.addInvoice({ memo: id, value: amount })
-          await lightningFundingWallet.pay({ invoice, isReward: true })
+          if (this.user.username) {
+            await lightningFundingWallet.pay({
+              destination: this.getNodePubkey(),
+              username: this.user.username,
+              amount,
+              isReward: true,
+              memo: id,
+            })
+          } else {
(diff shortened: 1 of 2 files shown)


nicolasburtey:
arg yes. it's a good optimization but likely that many users would not have set their username yet when they will do the quiz. haven't thought about this one.

src/LightningUserWallet.ts
```diff
@@ -44,8 +44,18 @@ export class LightningUserWallet extends OnChainMixin(LightningMixin(UserWallet)
         )
 
         if (userPastState.earn.findIndex((item) => item === id) === -1) {
-          const invoice = await this.addInvoice({ memo: id, value: amount })
-          await lightningFundingWallet.pay({ invoice, isReward: true })
+          if (this.user.username) {
```
nicolasburtey, line 47: so this optimization would benefits only if the user already have set his username. which my guess (we could look at the data, don't have yet a practical way to do it) is that this would generally not the case

instead of a username, the pay function could accept a uid. this would solve this particular issue. this would be a bigger refactor.

but now the bigger question is whether this invoice creating is a main contribution to the lock issue, or really this is something else.

nicolasburtey: the issue for instance could be that redis lock takes too long to acquire for a high throughput, and the issue arise where there are too many users doing the quiz at the same time.

bodymindarts:
Stale

PR #269, opened 2021-06-26
Use dev dependencies to run yarn build for Docker

samerbuna:
While this solves the build issue, I think the better solution would be to not run `yarn build` in Docker at all and instead run it locally prior to docker build, and then COPY `lib` instead of `src`. Let me know if you want me to update the PR for that (note that'll require changing where `docker build` is run so that `yarn build` is run before it)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,14 +6,15 @@ RUN apk update && apk add git
 
 COPY ./*.json ./yarn.lock ./
 
-RUN yarn install --frozen-lockfile  --production
+RUN yarn install --frozen-lockfile
 
 COPY ./src ./src
 
-RUN yarn global add typescript
 RUN find ./src -type d -depth -name '__tests__' -exec rm -r {} \; -prune
 
 RUN yarn build
+
+RUN yarn install --frozen-lockfile --production
 
 FROM gcr.io/distroless/nodejs:14
 COPY --from=BUILD_IMAGE /app/lib /app/lib


nicolasburtey:
should I revert my commit to main I just did while merging this pr? I did that as a quick fix.

nicolasburtey:
>  run it locally prior to docker build

do you mean copying the lib content from using a multi-stage Docker build?

samerbuna:
> should I revert my commit to main I just did while merging this pr? I did that as a quick fix.

No.

samerbuna:
> do you mean copying the lib content from using a multi-stage Docker build?

Maybe I am not understanding this, but doesn't COPY use the local Docker client files? Why not have it use `lib` instead of `src`?

nicolasburtey:
if I understand your question, `COPY` will copy from the host. `COPY --from=BUILD_IMAGE` will copy from a previous build.
Also worth noting we have 2 Dockerfile. one for the test (Dockerfile-debug) and one for production (Dockerfile). the one in production is distroless and doesn't have any debug tool.

nicolasburtey:
> While this solves the build issue, I think the better solution would be to not run yarn build in Docker at all and instead run it locally prior to docker build, and then COPY lib instead of src. Let me know if you want me to update the PR for that (note that'll require changing where docker build is run so that yarn build is run before it)

if I understand correctly this message, it make me think that you are thinking about a multi stage build. This is actually what we are doing, but using a Docker multi stage process, which is better practice than building locally on the host

bodymindarts:
Definitely shouldn't run `yarn build` locally. Need to to have a deterministic way to create a production build and a test-build (ie in docker). It would be good if both were testable locally so that we can verify the build execution succeeds locally and in CI.

nicolasburtey:
`docker build .` can be helpful to check that the docker will build correctly. also `docker build -f Dockerfile-debug .` for the debug docker.

samerbuna:
> Definitely shouldn't run `yarn build` locally. Need to to have a deterministic way to create a production build and a test-build (ie in docker). It would be good if both were testable locally so that we can verify the build execution succeeds locally and in CI.

Ok then I think we can go for this PR as is.

Issue #130, opened 2021-04-15
have a config that would enable/disable the ip log information


nicolasburtey:
haven't look deep into it but there are some of the issues:
- lots of logger.error (moved to in logger.info)
- seems ip is not stored the array

PR #270, opened 2021-06-27
Config enable/disable the ip and proxy log information (Fixes GaloyMoney/galoy#130)

dolcalmi:
(Fixes GaloyMoney/galoy#130)
> on the ip stuff, would be good to have a way to 1/ deactive ip logging 2/ deactive proxy checking (from the yaml)
- Added new config (ipChecking)
- Nest proxyChecking in the new configuration
- Added new validations in context and fetchIPDetails


src/entrypoint/graphql.ts
```diff
@@ -357,9 +357,12 @@ export async function startApolloServer() {
           { lastConnection: new Date() },
           { new: true },
         )
-        if (yamlConfig.proxyChecking.enabled) {
-          fetchIPDetails({ ip, user, logger })
+
+        if (yamlConfig.ipChecking.enabled) {
+          const checkProxy = yamlConfig.ipChecking.proxyChecking.enabled
+          fetchIPDetails({ ip, user, logger, checkProxy })
```
nicolasburtey, line 363: I think we could just initiate `checkProxy` within `fetchIPDetails` directly if the only reason `checkProxy` const is being created here is to pass it to the `fetchIPDetails` function.

default.yaml
```diff
@@ -87,8 +87,10 @@ limits:
       1: 5000000
       2: 100000000
 
-proxyChecking:
+ipChecking:
```
nicolasburtey, line 90: `ipRecording` might be more accurate?

nicolasburtey (commented):
two minors comments, overall the PR looks good

dolcalmi:
> two minors comments, overall the PR looks good

fixed

src/utils.ts
```diff
@@ -160,15 +160,14 @@ export const fetchIPDetails = async ({ ip, user, logger }): Promise<void> => {
...
-    // const {data} = await axios.get(`http://proxycheck.io/v2/${ip}?key=${PROXY_CHECK_APIKEY}&vpn=1&asn=1`)
-    // ipinfo = data[ip]
-
-    ipinfo = await fetchIP({ ip })
+    if (yamlConfig.ipRecording.proxyChecking.enabled) {
+      ipinfo = await fetchIP({ ip })
+    }
   } catch (error) {
     logger.info({ error }, "Failed to fetch ip details")
   } finally {
```
nicolasburtey, line 173: I guess there is no need to go into this branch if proxyCheck is not enabled?

```
    if (!res.nModified) {
      await User.findOneAndUpdate(
        { "_id": user._id, "lastIPs.ip": { $ne: ip } },
        { $push: { lastIPs: { ip, ...ipinfo, Type: ipinfo?.type } } },
      )
    }
```

nicolasburtey (commented):
another comment after checking where `ipinfo` was being used.

maybe the flow could be refactor, we could just return after 
```
   const res = await User.updateOne(
      { "_id": user._id, "lastIPs.ip": ip },
      { $set: { "lastIPs.$.lastConnection": Date.now() } },
    )
```

if proxy is not enabled.

dolcalmi:
do you want to merge PR https://github.com/GaloyMoney/galoy/pull/250/ before the refactor?

krtk6160:
> do you want to merge PR #250 before the refactor?

We've merged that PR. You should rebase and then we can merge this as well.

PR #271, opened 2021-06-27
Add 2FA

krtk6160:


src/schema.graphql
```diff
@@ -201,6 +202,12 @@ type Query {
   getLevels: [Int]
   getLimits: Limits
   getWalletFees: WalletFees
+  generate2fa: TwoFactorSecret
```
bodymindarts, line 205: Shouldn't this be under `Mutation`? `Generate` implies a side effect. `Queries` should be side effect free.

src/schema.graphql
```diff
@@ -286,4 +293,10 @@ type Mutation {
   setAccountStatus(uid: ID!, status: AccountStatus!): User
 
   setLevel(uid: ID!, level: Int!): User
+
+  save2fa(secret: String!, token: String!): Boolean
+
+  delete2fa(token: String!): Boolean
+
+  validate2fa(token: String!): Boolean
```
bodymindarts, line 301: Does this have a side effect? If not it could be under `Query`.

src/userWallet.ts
```diff
@@ -240,6 +242,79 @@ export abstract class UserWallet {
     return usdValue
   }
 
+  generate2fa = () => {
+    const { secret, uri, qr } = generateSecret({
+      name: yamlConfig.name,
+      account: this.user.phone,
+    })
+    /*
+    { secret: 'XDQXYCP5AC6FA32FQXDGJSPBIDYNKK5W',
```
bodymindarts, line 251: Why is the comment being committed?

bodymindarts (commented):
Its looks pretty good to me. Just a few issues with the GraphQL schema and perhaps you could fix the warnings for the `check-code` task.

src/userWallet.ts
krtk6160: We have a few comments like this across the codebase where we have an example of the return value if it's not immediately clear.

src/schema.graphql
krtk6160: There's no side effect. Just returns a new `secret` but does not interact with data in any way.

src/schema.graphql
krtk6160: Shifted

bodymindarts:
Tod should be fine via a type or a unit test. That would communicate the same but without the risk of getting out of sync

src/__tests__/01b-create-users.ts
```diff
@@ -134,9 +137,32 @@ describe("username tests", () => {
...
-  // })
+  it("sets account status correctly", async () => {
+    await AdminOps.setAccountStatus({ uid: userWallet2.user._id, status: "locked" })
+    userWallet2 = await getUserWallet(2)
+    await expect(userWallet2.user.status).toBe("locked")
+  })
+
+  it("set 2fa for user0", async () => {
+    const { secret } = userWallet0.generate2fa()
+    const token = generateToken(secret)!.token
```
nicolasburtey, line 148: can you fix the typescript warning?

src/__tests__/01b-create-users.ts
```diff
@@ -134,9 +137,32 @@ describe("username tests", () => {
...
-  //   await AdminOps.setAccountStatus({uid: userWallet2._id, status: 'locked'})
-  //   await expect(userWallet2.status).toBe('locked')
-  // })
+  it("sets account status correctly", async () => {
+    await AdminOps.setAccountStatus({ uid: userWallet2.user._id, status: "locked" })
+    userWallet2 = await getUserWallet(2)
+    await expect(userWallet2.user.status).toBe("locked")
+  })
+
+  it("set 2fa for user0", async () => {
```
nicolasburtey, line 146: maybe `generate2fa` instead of `set` here would be better for consistency

src/userWallet.ts
```diff
@@ -240,6 +242,79 @@ export abstract class UserWallet {
     return usdValue
   }
 
+  generate2fa = () => {
+    const { secret, uri, qr } = generateSecret({
```
nicolasburtey, line 246: some other typescript error to fix

nicolasburtey:
also need to be rebased

src/__tests__/01b-create-users.ts
krtk6160: I'm not sure how to do that. This is a part of the node-2fa .d.ts file - 
```
export declare function generateToken(secret: string): {
    token: string;
} | null;
```

As you can see the return value for generateToken could be null. We can probably override this by declaring our own file but that doesn't seem to be an improvement over the non-null assertion operator we're using currently.

src/userWallet.ts
krtk6160: I don't see what the error is here? Is it that `qr` is an unused var?

src/__tests__/01b-create-users.ts
krtk6160: `generate2fa` function in our codebase just generates a secret, whereas `save2fa` actually sets it. I'll change it to `save2fa`.

src/schema.graphql
bodymindarts: What is the source of the new secret? Is anything persisted? Anything with the word `generate` implies a creation (which in GQL is handled via the `Mutation` path). If we are not creating something perhaps the word is wrong.

krtk6160: I would argue that when it comes to generating crypto-related keys/secrets, there's typically no side effect with functions that have `generate` in their name. For example: the `generateKeyPair` function from node's inbuilt crypto module comes to mind. 
Having said that I do agree that something like `generateNew2FASecret` would be a better name

nicolasburtey: we don't persist anything yet at this time of the call. we create/pass a secret that the client will send back to the server once the client verified within the mobile that he has saved it by providing a 6 digits associated to the secret.

I don't think not saving the temporary secret can be a source of attack but maybe something to consider

nicolasburtey:
@krtk6160 you may have taken one of my commit from main that I mistakenly pushed. you may want to rebase (to remove the `"noImplicitAny": true,`)

src/__tests__/01b-create-users.ts
krtk6160: > @krtk6160 you may have taken one of my commit from main that I mistakenly pushed. you may want to rebase (to remove the `"noImplicitAny": true,`)

Yep, thanks.

krtk6160:
> @krtk6160 you may have taken one of my commit from main that I mistakenly pushed. you may want to rebase (to remove the `"noImplicitAny": true,`)

Yep, thanks.

krtk6160:
> @krtk6160 you may have taken one of my commit from main that I mistakenly pushed. you may want to rebase (to remove the `"noImplicitAny": true,`)

Yep, thanks.

src/schema.graphql
bodymindarts: > for example: the generateKeyPair function from node's inbuilt crypto module comes to mind.

This may be generally true in the context of crypto-functions but GraphQL has its own conventions and here we are designing a public GQL interface. Not a crypto-library API.

I'm not sure what the idiomatic way to handle this in GQL would be and the GQL layer needs reworking anyway so as long as nothing is persisted we can probably leave it like this for the time being.

Another thing to consider is that `Queries` can typically be combined with various caching layers which would break here because each call returns a different response.

so TLDR - I believe its probably a `Mutation` but not a big deal to leave it as is for now. Needs revisiting when we re-design the whole GQL layer.

src/userWallet.ts
bodymindarts: My comment
> Then this should be done via a type or a unit test. That would communicate the same but without the risk of getting out of sync

Was referring to this. We shouldn't have issues that refer to Adding Tests separately. Test code must be written together with (potentially even before) the production code and merged together.

krtk6160:
> Tod should be fine via a type or a unit test. That would communicate the same but without the risk of getting out of sync

Actually there's already a type for that function:
```
(alias) generateSecret(options?: Options | undefined): {
    secret: string;
    uri: string;
    qr: string;
}
```
This gives a limited amount of info, but by including an example we can give a lot more info. It might not be helpful for the `secret` value, but for `uri` and `auth` I think it helps, because it shows the type of scheme being used to generate them. 

Is that the right way to think about this? Or should we eliminate the risk of being out-of-sync when the return value changes (which I believe is very low) and let the user dig a bit deeper to find this info?

@nicolasburtey wdyt?

src/userWallet.ts
krtk6160: Ah okay, gotcha

src/Lightning.ts
```diff
@@ -290,6 +291,23 @@ export const LightningMixin = (superclass) =>
         params,
       })
 
+      if (
+        yamlConfig.twoFactor?.enabled &&
```
nicolasburtey, line 295: what is the goal of this settings?

src/Lightning.ts
```diff
@@ -290,6 +291,23 @@ export const LightningMixin = (superclass) =>
...
+        this.user.twoFactor.secret &&
+        tokens > this.user.twoFactor.threshold
```
nicolasburtey, line 297: it should be related to velocity here, not just the current amoung being paid?

src/userWallet.ts
```diff
@@ -238,6 +240,79 @@ export abstract class UserWallet {
...
+  static validate2fa = ({ secret, token, logger }): boolean => {
+    if (!secret) {
+      throw new TwoFactorError("no 2fa has been set", { logger })
+    }
+
+    if (verifyToken(secret, token)) {
+      return true
+    }
+
+    throw new TwoFactorError(undefined, { logger })
```
nicolasburtey, line 292: should we mention that code is wrong here? or is the default message?
the `undefined` is not obvious.

src/Lightning.ts
krtk6160: Deploy PR with this disabled, make sure nothing breaks, then enable this

src/Lightning.ts
```diff
@@ -290,6 +291,23 @@ export const LightningMixin = (superclass) =>
...
+      ) {
+        if (!token) {
+          throw new TwoFactorError("Need a 2FA code to proceed with the payment", {
+            logger: lightningLogger,
+          })
+        }
+
+        UserWallet.validate2fa({
```
nicolasburtey, line 305: I guess the assumption here is that we throw if the token is not correct. but I don't think it's clear by reading the code. 

specially at the function return a boolean.

I think: either we don't return a boolean and have a `throwIfNotValid()` or we use the boolean and throw the error locally.

src/__tests__/01b-create-users.ts
```diff
@@ -33,6 +37,16 @@ it("add user0/funder/Dealer", async () => {
 
 const username = "user0"
 
+const generateTokenHelper = ({ secret }) => {
+  const generateTokenResult = generateToken(secret)
+  expect(generateTokenResult).toBeTruthy()
+  if (generateTokenResult) {
+    return generateTokenResult.token
+  } else {
+    fail("generateToken returned null")
```
nicolasburtey, line 46: `fail` is part of the `jest` toolbox?

krtk6160: `fail` is from jasmine which jest uses internally afaik

src/__tests__/04b-lightning-payment.ts
```diff
@@ -303,6 +305,36 @@ functionToTests.forEach(({ fn, name, initialFee }) => {
     const { BTC: finalBalance } = await userWallet1.getBalances()
     expect(finalBalance).toBe(initBalance1)
   }, 60000)
+
+  it(`fails to pay above 2fa threshold without 2fa token`, async () => {
+    // this is needed because this test runs twice
+    if (!userWallet0.user.twoFactor.secret) {
+      const { secret } = userWallet0.generate2fa()
```
nicolasburtey, line 312: let's use a `// eslint-disable-next-line  @typescript-eslint/no-non-null-assertion` to remove warning from elint? (also present on github which is nice)

src/userWallet.ts
krtk6160: We already do that in `error.ts` in the definition of the `TwoFactorError`

src/__tests__/04b-lightning-payment.ts
```diff
@@ -303,6 +305,36 @@ functionToTests.forEach(({ fn, name, initialFee }) => {
     const { BTC: finalBalance } = await userWallet1.getBalances()
     expect(finalBalance).toBe(initBalance1)
   }, 60000)
+
+  it(`fails to pay above 2fa threshold without 2fa token`, async () => {
+    // this is needed because this test runs twice
+    if (!userWallet0.user.twoFactor.secret) {
```
nicolasburtey, line 311: why do you check this as part of the test?

krtk6160: Because this code block is only needed for the first run of this test, but not the second one (it will throw if we try to set 2fa secret if it's already present)

src/entrypoint/graphql.ts
```diff
@@ -72,14 +72,15 @@ const resolvers = {
   Query: {
     me: async (_, __, { uid, user }) => {
       const { phone, username, contacts, language, level } = user
```
nicolasburtey, line 74: we can't do it that way? `const { phone, username, contacts, language, level, twoFactorEnabled } = user`? we should I think

krtk6160: I refactored this a bit to change how twoFactorEnabled is being determined. 

But did you mean we should add another field (`twoFactorEnabled`) directly to the schema?

src/Lightning.ts
krtk6160: I'm going with the former option

src/Lightning.ts
nicolasburtey: > Deploy PR with this disabled, make sure nothing breaks, then enable this.

By default this is not enabled at the user level? So deploying this will be a no-op for a user perspective until he decides to activate it (once we have the integration in the mobile app obviously). 

> Could also be used as a switch to disable any 2FA related stuff across the system (Maybe someone doesn't want to deploy galoy with 2FA?)

I think this is unlikely once we have a good 2FA system, but this is a good point. I think eventually there should be settings to configure the auth part of galoy.

src/Lightning.ts
```diff
@@ -290,6 +291,26 @@ export const LightningMixin = (superclass) =>
...
+        this.user.twoFactor.secret &&
+        twoFactorLimitHit
+      ) {
+        if (!token) {
+          throw new TwoFactorError("Need a 2FA code to proceed with the payment", {
+            logger: lightningLogger,
+          })
+        }
+
+        UserWallet.throwIf2faNotValid({
```
nicolasburtey, line 307: @jrbuhl93 would you see that as an anti-pattern following our last discussion?

src/OnChain.ts
```diff
@@ -134,7 +134,7 @@ export const OnChainMixin = (superclass) =>
           if (payeeUser) {
             const onchainLoggerOnUs = onchainLogger.child({ onUs: true })
 
-            if (await this.user.limitHit({ on_us: true, amount })) {
+            if ((await this.user.remainingOnUsLimit()) < amount) {
```
nicolasburtey, line 137: here `this.user.remainingOnUsLimit()` is a function, above it's a property? probably one of them is not right

src/Lightning.ts
```diff
@@ -305,7 +326,7 @@ export const LightningMixin = (superclass) =>
           if (isMyNode({ pubkey: destination }) || destination === "") {
             const lightningLoggerOnUs = lightningLogger.child({ onUs: true, fee: 0 })
 
-            if (await this.user.limitHit({ on_us: true, amount: tokens })) {
+            if ((await this.user.remainingOnUsLimit) < tokens) {
```
nicolasburtey, line 329: I would generally avoid doing an await + comparison + if in the same line. you can extract   `remainingOnUsLimit` and set it as a const in the line above for more clarify?

src/__tests__/01b-create-users.ts
```diff
@@ -132,9 +134,33 @@ describe("username tests", () => {
     await expect(userWallet2.setUsername({ username })).rejects.toThrow()
   })
 
-  // FIXME: failing for some reason
-  // it('sets account status correctly', async () => {
-  //   await setAccountStatus({uid: userWallet2._id, status: 'locked'})
-  //   await expect(userWallet2.status).toBe('locked')
-  // })
+  it("sets account status correctly", async () => {
```
nicolasburtey, line 137: name of the test seems outdated

src/__tests__/04b-lightning-payment.ts
nicolasburtey: maybe refactor the test to not run this here in that case?
it could be done in the test related to the user settings

src/schema.graphql
nicolasburtey: > Another thing to consider is that Queries can typically be combined with various caching layers which would break here because each call returns a different response.

this is a very valid point. we're not caching anything at the GQL layer yet but we should at some point

nicolasburtey: the counter argument is that even if we were to return a cached secret / user, this would not be issue I guess (as long as the secret being generated is obviously different for each user)

src/__tests__/04b-lightning-payment.ts
nicolasburtey: you could create an active virtual property from `userWallet0.user.twoFactor`. such that if `secret` is set, `active` return true

src/schema.graphql
```diff
@@ -118,6 +118,8 @@ type User {
   created_at: String
   coordinate: Coordinate
   status: AccountStatus
+  twoFactorEnabled: Boolean
+  remainingTwoFactorLimit: Int
```
nicolasburtey, line 122: still used?

src/entrypoint/graphql.ts
```diff
@@ -184,6 +193,7 @@ const resolvers = {
       return uid
     },
     getLevels: () => Levels,
+    getUserLimits: (_, __, { wallet }) => wallet.getUserLimits(),
```
nicolasburtey, line 196: could we refactor and not have a `getLimits` and `getUserLimits`? not clear what would be the difference when just looking at the function name

nicolasburtey (commented):
couple more comments. my main feedback is regarding the `Limits` and `UserLimits` query. I think we should refactor this and start to adopt best practice for graphql to have an easy to understand graphql API just by looking at schema.graphql

src/__tests__/01b-create-users.ts
krtk6160: Nope. The test checks if the account status is set correctly by the `setAccountStatus` function.

src/Lightning.ts
krtk6160: > > Deploy PR with this disabled, make sure nothing breaks, then enable this.
> 
> By default this is not enabled at the user level? So deploying this will be a no-op for a user perspective until he decides to activate it (once we have the integration in the mobile app obviously).

Yes, that's correct. My first point is moot.

> > Could also be used as a switch to disable any 2FA related stuff across the system (Maybe someone doesn't want to deploy galoy with 2FA?)
> 
> I think this is unlikely once we have a good 2FA system, but this is a good point. I think eventually there should be settings to configure the auth part of galoy.

Yes, which is why I didn't remove this. Do you want this to be removed for now and added back later?

If we do want to do this in the future then I think we should leave this here and if we end up adopting a different kind of switching approach (something other than the yaml based approach we're using currently for switches) then we can refactor this.

src/OnChain.ts
krtk6160: Nope, it is a method defined on the UserSchema. Are you referring to it being a field in the `userWallet.ts` file?

src/Lightning.ts
krtk6160: Done

src/schema.graphql
krtk6160: Removed

src/__tests__/04b-lightning-payment.ts
krtk6160: Yep, I've created a `twoFactorEnabled` virtual on the schema, however that doesn't solve the issue here.
This test is primarily checks whether the payment is denied if a payment go over the 2FA limit without providing a valid 2fa token. For that we need the 2fa to be enabled on that particular user.

I could leave 2fa enabled from the user creation test, but that adds an ordering dependency, which is something we definitely do not want.

I can create a helper function that does the 2FA enabling part, but again, that won't allow us to remove this part from this test, it will just move the declaration to a different file and call that.

src/Lightning.ts
```diff
@@ -271,7 +272,7 @@ export const LightningMixin = (superclass) =>
         features,
         max_fee,
       } = await validate({ params, logger: lightningLogger })
-      const { memo: memoPayer } = params
+      const { memo: memoPayer, twoFactorToken: token } = params
```
nicolasburtey, line 275: why the renaming? we could keep `twoFactorToken`?

src/schema.ts
```diff
@@ -261,31 +272,65 @@ UserSchema.virtual("oldEnoughForWithdrawal").get(function (this: typeof UserSche
...
-    : [{ type: { $ne: "on_us" } }]
+const getTimestampYesterday = () => Date.now() - MS_PER_DAY
 
-  const limit = yamlConfig.limits[on_us ? "onUs" : "withdrawal"].level[this.level]
+// FIXME: Improve function name
+UserSchema.methods.remainingTwoFactorLimit = async function () {
+  const threshold = this.twoFactor.threshold
+
+  const txnType = [
+    { type: "on_us" },
```
nicolasburtey, line 286: @krtk6160 we need to differentiate on_us to on_us_receipt and on on_us_sent. otherwise the velocity check would also apply for money received.

we could open a ticket and do this in another PR 

cc @samerbuna for the graphql type

src/Lightning.ts
nicolasburtey: we have `token` and `tokens`

src/schema.ts
krtk6160: https://github.com/GaloyMoney/galoy/issues/320

src/Lightning.ts
krtk6160: The `node-2fa` library uses the `token` variable name, so to keep naming consistent in functions that use functions from the library, I am renaming `twoFactorToken` to `token`.
But I do see your point that having both `token` and `tokens` in the same function can be confusing. I'll refactor the name

src/schema.ts
```diff
@@ -261,31 +272,65 @@ UserSchema.virtual("oldEnoughForWithdrawal").get(function (this: typeof UserSche
...
 
   const outgoingSats =
     (
       await User.getVolume({
-        after: timestampYesterday,
+        after: getTimestampYesterday(),
         txnType,
         accounts: this.accountPath,
       })
     )?.outgoingSats ?? 0
```
nicolasburtey, line 299: I think the getVolume function could be improved. the fact that we need to go `?.outgoingSats ?? 0` to parse the value of the function is not great. this should probably be integrated within the function itself.

src/Lightning.ts
```diff
@@ -290,6 +291,26 @@ export const LightningMixin = (superclass) =>
         params,
       })
 
+      const remainingTwoFactorLimit = await this.user.remainingTwoFactorLimit()
+
+      if (
```
nicolasburtey, line 296: this logic of looking at if twoFactor is enabled it not present for onchain. is that intended?

src/Lightning.ts
nicolasburtey: I think a more elegant way to do it could be to have `remainingTwoFactorLimit` returning infinite (or MaxInt) when it's not enabled. this would avoid this additional if here. (the pay function is already big we should try to not increase it's size too much)

src/schema.ts
```diff
@@ -261,31 +272,65 @@ UserSchema.virtual("oldEnoughForWithdrawal").get(function (this: typeof UserSche
...
+  const withdrawalLimit = yamlConfig.limits["withdrawal"].level[this.level]
+
+  const outgoingSats =
+    (
+      await User.getVolume({
+        after: getTimestampYesterday(),
+        txnType: [{ type: { $ne: "on_us" } }],
+        accounts: this.accountPath,
+      })
+    )?.outgoingSats ?? 0
```
nicolasburtey, line 316: same remarks here on the repetition of `?.outgoingSats ?? 0`

nicolasburtey (commented):
Review the PR and have a couple more feedbacks

src/schema.graphql
samerbuna: IMO, a non-idempotent operation should be a mutation.

src/schema.ts
```diff
@@ -233,6 +233,17 @@ const UserSchema = new Schema({
     enum: ["active", "locked"],
     default: "active",
   },
+
+  twoFactor: {
+    secret: {
+      type: String,
+      default: undefined,
```
samerbuna, line 240: Why a default is needed here?

src/schema.graphql
```diff
@@ -123,6 +123,7 @@ type User {
   created_at: String
   coordinate: Coordinate
   status: AccountStatus
+  twoFactorEnabled: Boolean
```
samerbuna, line 126: Can we make this not nullable?

src/schema.graphql
```diff
@@ -205,18 +206,25 @@ type Query {
...
+  uri: String
 }
 
-type Limits {
-  oldEnoughForWithdrawal: Int
-  withdrawal: Int
-  onUs: Int
+type WalletFees {
+  deposit: Float
 }
```
samerbuna, line 229: Can we make the types fields not nullable?

src/schema.ts
```diff
@@ -261,31 +272,65 @@ UserSchema.virtual("oldEnoughForWithdrawal").get(function (this: typeof UserSche
...
-UserSchema.methods.limitHit = async function ({
-  on_us,
-  amount,
-}: {
-  on_us: boolean
-  amount: number
-}) {
-  const timestampYesterday = Date.now() - MS_PER_DAY
+UserSchema.virtual("twoFactorEnabled").get(function (this: typeof UserSchema) {
+  return !!this.twoFactor.secret
```
samerbuna, line 276: Not a fan of `!!`. If we don't have a helper for this logic, maybe introduce something like `hasValue(this.twoFactor.secret)`

test/integration/01b-create-users.spec.ts
```diff
@@ -132,9 +134,34 @@ describe("username tests", () => {
...
+  })
+
+  it("delete 2fa for user0", async () => {
+    const token = generateTokenHelper({ secret: userWallet0.user.twoFactor.secret })
+    const result = await userWallet0.delete2fa({ token })
+    expect(result).toBeTruthy()
+    userWallet0 = await getUserWallet(0)
+    expect(userWallet0.user.twoFactorEnabled).toBeFalsy()
+  })
 })
```
samerbuna, line 167: Is there a reason the `userWallet0` variable is re-used across many scopes?  Wouldn't this introduce test order dependency?

krtk6160: Yep, that's how we're doing our tests at the moment. They're all ordered and you can't run most of them independently.
Just to be clear, this is not introducing any new dependency that's not already there. I think removing the ordering dependency will be part of a much bigger refactor.

src/schema.ts
krtk6160: I've refactored getVolume to have the check inside the function itself

src/Lightning.ts
krtk6160: Nope, not at all. I missed that. I've added it, and also a test for it.

src/schema.ts
krtk6160: Fixed

src/schema.ts
krtk6160: Good question. I just removed it and it doesn't seem to cause any issues. I don't remember the exact error I solved by adding this, but it seems it's not there anymore.

src/schema.graphql
krtk6160: Done

src/schema.graphql
krtk6160: Done

src/schema.ts
krtk6160: Checking for `x != null` seems to be a common way in JS/TS, so I've refactored this check that way. Wdyt?

src/schema.ts
```diff
@@ -261,31 +271,56 @@ UserSchema.virtual("oldEnoughForWithdrawal").get(function (this: typeof UserSche
...
+  return this.twoFactor.secret != null
+})
+
+const getTimestampYesterday = () => Date.now() - MS_PER_DAY
+
+// FIXME: Improve function name
+UserSchema.methods.remainingTwoFactorLimit = async function () {
+  const threshold = this.twoFactor.threshold
+
+  const txnType = [
```
nicolasburtey, line 284: @krtk6160 warning to be fixed here

src/schema.ts
```diff
@@ -340,7 +379,7 @@ UserSchema.virtual("userIsActive").get(async function (this: typeof UserSchema)
     accounts: this.accountPath,
   })
 
-  return volume?.outgoingSats > 1000 || volume?.incomingSats > 1000
+  return volume.outgoingSats > 1000 || volume.incomingSats > 1000
```
nicolasburtey, line 382: those magic strings should be at minimum set as a `const`. but even better could be set in our config yaml file

src/schema.ts
```diff
@@ -261,31 +271,56 @@ UserSchema.virtual("oldEnoughForWithdrawal").get(function (this: typeof UserSche
...
-}) {
-  const timestampYesterday = Date.now() - MS_PER_DAY
+UserSchema.virtual("twoFactorEnabled").get(function (this: typeof UserSchema) {
+  return this.twoFactor.secret != null
+})
+
+const getTimestampYesterday = () => Date.now() - MS_PER_DAY
+
+// FIXME: Improve function name
+UserSchema.methods.remainingTwoFactorLimit = async function () {
```
nicolasburtey, line 281: during the last code review (@krtk6160 you were not here but a recording should be available) @bodymindarts made a great point that those business logic function should not be part of the schema definition. we can move everything related to volume/velocity to UserWallet.ts, this would be a better place for this type of code.

src/Lightning.ts
nicolasburtey: @krtk6160 I think that should be considered/addressed

nicolasburtey (commented):
Some more feedback on the review, mainly related to the fact that we want to have a better separation of concern between db logic and business logic.

src/schema.ts
krtk6160: Yep, @bodymindarts and I had a discussion on the same about a different PR of mine in one of our pair programming sessions. We decided to not change schema.ts much because
1. The plan was to get my pending backend PRs merged quickly so I could switch to the release engineering stuff completely
2. Let the schema related refactor be part of the bigger upcoming refactor

I can refactor this in this PR itself, but it won't be a small refactor. I think it makes sense to do it in a separate PR. wdyt?

I can refactor this, or let it be like this depending on what makes sense.

src/schema.ts
krtk6160: Good catch, fixed.

src/schema.ts
krtk6160: Fixed

src/schema.ts
nicolasburtey: ok fair enough. let's not do this as part of this PR.

src/Lightning.ts
nicolasburtey: @krtk6160 this is my last comment on this PR

krtk6160: I've completely removed the system level limit

nicolasburtey (dismissed):

default.yaml
```diff
@@ -121,3 +121,5 @@ onChainWallet:
 
 twoFactor:
   threshold: 300000 # sats
+
+userActivenessMonthlyVolumeThreshold: 1000 # sats
```
bodymindarts, line 125: Missing EOF. Looks like you haven't configured your editor

bodymindarts (approved):

nicolasburtey:
this PR needs to be rebased

test/helpers/index.ts
```diff
@@ -46,3 +47,13 @@ export const resetDatabase = async (mongoose) => {
       await db.dropCollection(collectionName)
     })
 }
+
+export const generateTokenHelper = ({ secret }) => {
+  const generateTokenResult = generateToken(secret)
+  expect(generateTokenResult).toBeTruthy()
```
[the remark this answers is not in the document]
krtk6160: > please don't test helpers (except),

Okay, I've removed the expect.

> just use `return generateTokenResult.token` also else is not necessary

Actually I have to use `generateTokenResult?.token` otherwise there's a TS warning due to the object possibly being null. That's the reason why I created the helper, so that the null checking would be done here and won't be needed in multiple places.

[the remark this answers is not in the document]
dolcalmi: if you wan to call the fail is ok (although is not necessary) but else is not necessary:

```js
export const generateTokenHelper = ({ secret }) => {
  const generateTokenResult = generateToken(secret)

  if (generateTokenResult) {
    return generateTokenResult.token
  }

  fail("generateToken returned null")
}
```
or just 

```js
export const generateTokenHelper = ({ secret }) => {
  const { token } = generateToken(secret)
  return token
}
```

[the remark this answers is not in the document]
krtk6160: Yes you're right, I somehow missed that I had a return in the if block.

I've refactored it now.

However your second suggestion will not work due to the same reason I mentioned above, and I get this error -> `Property 'token' does not exist on type '{ token: string; } | null'.ts(2339)`
The reason being generateTokenResult returns either an object with a token field, or null.

src/schema.ts
```diff
@@ -233,6 +233,16 @@ const UserSchema = new Schema({
...
     default: "active",
   },
+
+  twoFactor: {
+    secret: {
+      type: String,
+    },
+    threshold: {
+      type: Number,
+      default: yamlConfig.twoFactor.threshold,
```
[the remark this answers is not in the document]
krtk6160: Schema level config refactoring will be done in another PR, as this has been open for a very long time.

test/integration/02-user-wallet/02-send-onchain.spec.ts
```diff
@@ -356,3 +358,26 @@ describe("UserWallet - onChainPay", () => {
...
+it("fails to pay above 2fa limit without 2fa token", async () => {
+  const remainingLimit = await userWallet0.user.remainingTwoFactorLimit()
+  const address = await bitcoindDefaultClient.getNewAddress()
+  expect(
+    userWallet0.onChainPay({ address, amount: remainingLimit + 1 }),
+  ).rejects.toThrowError(TwoFactorError)
+})
+
+it(`Makes large payment with a 2fa code`, async () => {
+  const address = await bitcoindDefaultClient.getNewAddress()
```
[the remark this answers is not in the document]
krtk6160: I tried both the ways - sending to `RANDOM_ADDRESS` and also sending to an address controlled by `lndOutside1`, but the specterwallet test continues to fail despite that.

bodymindarts (approved):

samerbuna (approved):

dolcalmi (approved):

PR #274, opened 2021-06-28
Onchain pay user send all

jotapea:


src/OnChain.ts
```diff
@@ -100,14 +100,15 @@ export const OnChainMixin = (superclass) =>
     }
 
     // amount in sats
-    async onChainPay({ address, amount, memo }: IOnChainPayment): Promise<ISuccess> {
+    async onChainPay({ address, amount, memo, sendAll = false }: IOnChainPayment): Promise<ISuccess> {
```
jotapea, line 103: Should a check comparing the balance.total_in_BTC vs amount be added when sendAll? Or amount stops being a relevant parameter in this case?

src/OnChain.ts
```diff
@@ -200,33 +211,58 @@ export const OnChainMixin = (superclass) =>
...
+          if (!sendAll) {
+            amountToSend = amount
+
+            // case where there is not enough money available within lnd on-chain wallet
+            if (onChainBalance < amount + estimatedFee) {
+              // TODO: add a page to initiate the rebalancing quickly
+              throw new RebalanceNeededError(undefined, {
+                logger: onchainLogger,
+                onChainBalance,
+                amountToSend,
```
jotapea, line 223: Is the rename from 'amount' to 'amountToSend' ok for the error metadata?

src/__tests__/02c-onchain-send.ts
```diff
@@ -133,6 +135,82 @@ it("Sends onchain payment successfully", async () => {
...
+
+  await sleep(1000)
+  console.log(JSON.stringify(sendNotification.mock.calls))
+
+  // expect(sendNotification.mock.calls.length).toBe(2)  // FIXME: should be 1
+
+  // TODO Still showing amount, find where this happens...
+  // expect(sendNotification.mock.calls[0][0].title).toBe(
+  //   getTitle["onchain_payment"]({ amount: initialBalanceUser1 }),
+  // )
```
jotapea, line 193: This has not been solved yet...

src/OnChain.ts
jotapea: Added comments where 'amount' is still used, which should be affected if is sendAll and amount is not specified or is not relevant.

nicolasburtey:
thanks for the work on this PR @jotapea . could you rebase it? 

we have had a big PR merged yesterday that made us go from 1 lnd node to multiple lnd node.

src/OnChain.ts
krtk6160: Any reason for the rename here?

jotapea: Hello Kartik! That was originally implemented that way, but then I decided to add the named parameter so that no change would be needed. https://github.com/GaloyMoney/galoy/pull/274/commits/eb8e7ac257be22672ba52632e39b477044c7eaff

jotapea:
@nicolasburtey yes I can do the rebase, but I also think there is still work to be done. In the onChainPay method the 'amount' parameter is still used in checks and my understanding is that if 'sendAll' applies, then amount stops being relevant. Is this a correct assumption? Like how would the app use the api? If sendAll is chosen then amount would be sent as 0? https://github.com/GaloyMoney/galoy/pull/274/commits/5c6d713e2e796ad451b174e6c5c9651b08ca7ba4

PR #275, opened 2021-06-29
Onchain pay user send all

jotapea:


nicolasburtey (commented):
I have skimmed through the code (not the test yet) and this seems to be a very good PR. thanks for making it! we'll review more in depth when the tests are completed

jotapea:
You're welcome! I removed the TODO comments from the tests, these were kept by mistake from the rebase. The tests are passing for me.

I'll be waiting for next steps.

jotapea:
@nicolasburtey PR updated.

src/OnChain.ts
```diff
@@ -98,23 +98,31 @@ export const OnChainMixin = (superclass) =>
...
+        sendAll,
       })
 
-      if (amount <= 0) {
-        const error = "Amount can't be negative"
-        throw new ValidationError(error, { logger: onchainLogger })
-      }
+      if (!sendAll) {
+        if (amount <= 0) {
+          const error = "Amount can't be negative"
```
nicolasburtey, line 114: "Amount can't be negative, and can only be zero if sendAll = true" may be more appropriate here.

jotapea: But that is the branch if sendAll = false, which I just kept the original flow intact.

nicolasburtey: yup I know it's part of the original flow. but now that we have another branch where `amount = 0` as part of the flow, I think it's worth being a bit more specific on the error (which was already a bit too vague to get started, as the error message mention negative result but the check is also actually negative or zero).

jotapea: Made the update.

src/__tests__/02e-onchain-send.ts
```diff
@@ -165,6 +243,40 @@ it("makesOnchainOnUsTransaction", async () => {
   }
 })
 
+it("makesOnchainOnUsSendAllTransaction", async () => {
+  try {
```
nicolasburtey, line 247: do we need a try catch here?

nicolasburtey: the test might enter in the catch block without failing. I would either remove the try catch and let the test fails, or add a `fail()` comment in the catch block if this is needed for some reasons.

other the test could have an issue and would be reported as failed by jest

nicolasburtey (commented):
one last comment on the test, once this has been addresses this should be good to be merged.

src/__tests__/02e-onchain-send.ts
jotapea: Makes sense. That test was based on the 'makesOnchainOnUsTransaction' test which also has the try. Will remove it from both.

jotapea:
Tests updated.

nicolasburtey:
ok, launching github action CI tests. will merge if it does pass

jotapea:
Ok the try catch seems to have been hiding some issues, having the same failed test on my side now. Will work on this today.

nicolasburtey:
ok awesome. you'll also need to run `prettier` for the test pass.

jotapea:
Cool, will do.

jotapea:
@nicolasburtey tests fixed and updated by now using new test accounts. Also ran prettier and all tests are passing for me now.

src/schema.graphql
```diff
@@ -91,6 +91,7 @@ type OnChain {
     address: String! @pattern(regexp: "^[A-Za-z0-9]+$", policy: THROW)
     amount: Int! @range(min: 0, policy: THROW)
     memo: String @stringLength(max: 1024, policy: THROW) # TODO: add limit on the mobile app
+    sendAll: Boolean
```
nicolasburtey, line 94: I think we could have a cleaner API here by splitting the pay function in 2:
```
 pay(
    address: String! @pattern(regexp: "^[A-Za-z0-9]+$", policy: THROW)
    amount: Int! @range(min: 0, policy: THROW)
    memo: String @stringLength(max: 1024, policy: THROW) # TODO: add limit on the mobile app
): Success

payAll(
    address: String! @pattern(regexp: "^[A-Za-z0-9]+$", policy: THROW)
    memo: String @stringLength(max: 1024, policy: THROW) # TODO: add limit on the mobile app
): Success
```

The reason is that it would be harder to misuse the graphql API. either you pay with an amount, or payAll and no amount is needed.

We would need to add a resolver payAll that would pass the `{ amount: 0, sendAll: true}` to the pay function in the backend, but otherwise nothing would change.

What do you think?

nicolasburtey (commented):
just one comment, otherwise the PR looks good and ready to be merged.
what was the issue with the tests previously that those last 4 commits had fixed?

jotapea:
It was two things, related to walletUser2 already being used in a test. The makesOnchainOnUsSendAllTransaction test was failing because of this and also its funding should have been using the onUs level 1 limit. This funding was the new test added.

src/schema.graphql
jotapea: I agree, I'll work on it.

jotapea:
Split into pay and payAll done.

PR #276, opened 2021-06-29
Add option to send sms through SMSala in yaml

jrbuhl93:
We will need to add SMSALA_API_ID, SMSALA_API_PASSWORD, and SMSALA_SENDER_ID. Message me and I can provide all of these.


src/__tests__/09d-text-twilio.ts
```diff
@@ -28,14 +28,21 @@ afterAll(async () => {
 })
```
krtk6160, line 29: File should be renamed

krtk6160 (commented):
One minor comment but otherwise LGTM

src/text.ts
```diff
@@ -99,7 +122,16 @@ export const requestPhoneCode = async ({
     }
 
     await PhoneCode.create({ phone, code })
-    await sendText({ body, to: phone, logger })
+
+    
+    if (yamlConfig.sms_provider === "Twilio") {
```
nicolasburtey, line 127: I think the capitalization can be issue here. one will have to remember that this is "Twilio" and not "twilio" and "SMSala". maybe only lower case + `toLowerCase()` would avoid falling into this potential pitfall.

src/text.ts
```diff
@@ -24,19 +25,41 @@ const getTwilioClient = () => {
...
+
+export const sendSMSalaText = async({ body, to, logger }) => {
+  try {
+    const base_url = "http://api.smsala.com/api/SendSMS"
+    const api_id = process.env.SMSALA_API_ID
+    const api_password = process.env.SMSALA_API_PASSWORD
+    const sms_type = "T"
+    const encoding = "T"
+    const sender_id = process.env.SMSALA_SENDER_ID
+    const phoneNumber = to.replace(/\D/g,'')
```
nicolasburtey, line 51: what is the goal of this regex?

@bodymindarts here is where I think a comment would be useful. As a comment minimalist do you agree with that? :)

src/text.ts
```diff
@@ -24,19 +25,41 @@ const getTwilioClient = () => {
...
+
+    let url = `${base_url}?api_id=${api_id}&api_password=${api_password}`
+    url = url + `&sms_type=${sms_type}&encoding=${encoding}&sender_id=${sender_id}`
+    url = url + `&phonenumber=${phoneNumber}&textmessage=${body}`
+    await axios.get(url)
+  } catch (err) {
+    logger.fatal({ err }, "impossible to send text through SMSala")
```
nicolasburtey, line 58: I don't think this is what should be consider a `fatal` error. probably `logger.error()` is more appropriate. we need to document those error level but it is expect as some point that this function will throw, ie if their server is offline. this is not ideal but not big deal. 

fatal error is more for something that require immediate developer attention. like if a transaction brings the state of our database no longer have an assets = liabilities mismatch, which should not be possible under a double entry accounting standard.

src/text.ts
```diff
@@ -24,19 +25,41 @@ const getTwilioClient = () => {
...
       body,
     })
   } catch (err) {
-    logger.fatal({ err }, "impossible to send text")
+    logger.fatal({ err }, "impossible to send text through Twilio")
     return
   }
 
-  logger.info({ to }, "sent text successfully")
+  logger.info({ to }, "sent text through Twilio successfully")
```
nicolasburtey, line 40: @krtk6160 will this match our current query in grafana of plotting the number of successful text being send over time?

nicolasburtey (commented):
great first pass on this PR. couple of comments that should be addressed

src/text.ts
krtk6160: No I don't think it will. Good catch. We might need to add a field that represents a successful text msg (common across every send text log, independent of the provider)

nicolasburtey:
probably a missing piece is to add the carrier in our db when we doing a query. 
we can add carrier in this collection: https://github.com/GaloyMoney/galoy/blob/b47d96a2ea71da834318dc02e5530e5c45d55824/src/schema.ts#L395

the main reason is that currently the provider we use is a single point of failure for the account authentication. if the provider is evil, he could intercept the text and login with the information provider to the carrier. the starting point from here is to know which carrier text has been sent from to at least understand the issue. down the road there are a lot to do to mitigate this issue more broadly (this is basically the User Auth project: https://github.com/GaloyMoney/galoy/wiki/User-authentication)

src/text.ts
jrbuhl93: Do I need to change this logging message back to "sent text successfully" before we accept this PR?

krtk6160: I'd suggest changing the text back, and then also adding something like `provider: xyz` so that we can differentiate between them in logs if needed

nicolasburtey: yup better to have those variables in the first argument of the log function, and keep the message generic otherwise. this is how Pino should typically be used

src/text.ts
```diff
@@ -98,8 +123,17 @@ export const requestPhoneCode = async ({
...
 
-    await PhoneCode.create({ phone, code })
-    await sendText({ body, to: phone, logger })
+    await PhoneCode.create({ phone, code, sms_provider })
+
+    
+    if (sms_provider === "twilio") {
+      await sendTwilioText({ body, to: phone, logger })
+    } else if (sms_provider === "smsala") {
+      await sendSMSalaText({ body, to: phone, logger })
```
nicolasburtey, line 132: this could be refactored: `{ body, to: phone, logger }` and declared before the branch to avoid the repetition

src/text.ts
```diff
@@ -193,18 +227,20 @@ export const login = async ({
...
-        const result = await getCarrier(phone)
-        user.twilio = result
-        await user.save()
-      } catch (err) {
-        // Carrier fetching is a non-critical operation
-        // Primarily useful for analytics
-        // Hence failure should be handled with a warn instead of an error
-        subLogger.warn({ err }, "impossible to fetch carrier")
+
+    if (yamlConfig.sms_provider === "Twilio") {
```
nicolasburtey, line 231: case issue is still present here

src/text.ts
```diff
@@ -24,19 +25,42 @@ const getTwilioClient = () => {
...
+    let url = `${base_url}?api_id=${api_id}&api_password=${api_password}`
+    url = url + `&sms_type=${sms_type}&encoding=${encoding}&sender_id=${sender_id}`
+    url = url + `&phonenumber=${phoneNumber}&textmessage=${body}`
+    await axios.get(url)
+  } catch (err) {
+    logger.error({ err, provider: "smsala" }, "impossible to send text")
+    return
+  }
+
+  logger.info({ to, provider: "smsala" }, "sent text successfully")
```
nicolasburtey, line 63: `const provider = "smsala"` above the `try catch` would allow the repetition of `{ provider: "smsala" }`

nicolasburtey (commented):
a few more questions/inputs, otherwise looks good.

src/text.ts
bodymindarts: I [agree](https://xkcd.com/1171/)

nicolasburtey:
need to run prettier. otherwise looks good. I'll merge it tomorrow after the multi lnd has been deployed to mainnet.

nicolasburtey:
actually @jrbuhl93 one thing is still missing. we need to pass the env variable to the pod. you can do that here: https://github.com/GaloyMoney/galoy/blob/main/charts/galoy/templates/deployment.yaml

you can just take example from `twilio-secret`. I'll take care of creating the kubernetes secret on my end.

src/text.ts
```diff
@@ -193,18 +229,20 @@ export const login = async ({
...
-        const result = await getCarrier(phone)
-        user.twilio = result
-        await user.save()
-      } catch (err) {
-        // Carrier fetching is a non-critical operation
-        // Primarily useful for analytics
-        // Hence failure should be handled with a warn instead of an error
-        subLogger.warn({ err }, "impossible to fetch carrier")
+
+    if (yamlConfig.sms_provider.toLowerCase() === "twilio") {
```
nicolasburtey, line 233: @jrbuhl93 for now I actually think we still want to call the `getCarrier` API. you can remove this `if` here (it's giving metadata on the phone number, specifically we want to check whether it's a voip number, because we don't give sats for the quiz from a voip number)

src/schema.ts
```diff
@@ -407,6 +407,7 @@ const PhoneCodeSchema = new Schema({
     type: Number,
     required: true,
   },
+  sms_provider: String,
```
nicolasburtey, line 410: I think `sms_` might be obvious/implied given the collection. `provider` would probably be enough?

nicolasburtey (approved):
just a cosmetic comment but good to me otherwise

@krtk6160 maybe you can sync with @jrbuhl93 for the deployment of this PR and roll this out in testnet. you can ask Salim to give it a try on testnet afterwards?

PR #277, opened 2021-06-30
Handle onchain fee estimation error

krtk6160:


src/OnChain.ts
```diff
@@ -91,7 +92,13 @@ export const OnChainMixin = (superclass) =>
         const { lnd } = getActiveOnchainLnd()
 
         const sendTo = [{ address, tokens: amount ?? defaultAmount }]
-        ;({ fee } = await getChainFeeEstimate({ lnd, send_to: sendTo }))
+        try {
+          ;({ fee } = await getChainFeeEstimate({ lnd, send_to: sendTo }))
```
nicolasburtey, line 96: do you know in what scenario this would typically happen?

nicolasburtey:
bitfinex is down and it seems it is affecting our test for some reasons. didn't know we had this dependancy

src/OnChain.ts
krtk6160: Nope, not sure. But I just saw loads of uncaught errors in our logs which was causing the pager duty alerts.
We can either have this as an error in which case we'll continue to have this counted in our pager duty alerts.
Or if we think it's a non-issue then we can keep it at warn.

nicolasburtey: I think we want to try to understand why that could be the case. maybe amount is very low?

krtk6160: Yep, I think it's because of a low amount. I tried a bunch of low amount queries to getFee and it increased those errors in our logs

nicolasburtey: oh yeah if there were not the dust check, then this might be the reason

src/OnChain.ts
```diff
@@ -171,6 +178,10 @@ export const OnChainMixin = (superclass) =>
 
           // normal onchain payment path
 
+          if (amount < yamlConfig.onchainDustAmount) {
```
nicolasburtey, line 181: we don't want to put this in the lock. it's better to throw prior to the lock if we throw and acquiring the lock is not needed to know if this condition is true or not

krtk6160: I'm doing it inside the lock because that's where we determine whether it's on onchain_on_us or a regular onchain payment. The intention being that I want to allow a dust amount if it's an `onchain_on_us` but no if it's `onchain`.

nicolasburtey: yup you're right

nicolasburtey (dismissed):
this needs a rebase. otherwise looks good for me

src/OnChain.ts
```diff
@@ -86,12 +87,22 @@ export const OnChainMixin = (superclass) =>
       if (payeeUser) {
         fee = 0
       } else {
+        if (amount && amount < yamlConfig.onchainDustAmount) {
+          throw new DustAmountError(undefined, { logger: this.logger })
```
bodymindarts, line 91: This might be a good place to start introducing a new way to return errors via a union type like discussed in code review.

src/OnChain.ts
bodymindarts: Can the `onchainDustAmount` be passed in as constructor config instead of read directly from `yaml`? That would make the code more modular and easier to test.

bodymindarts (commented):
I know this PR is in line with the patterns established in the code base up until now. But sooner-or-later we want to start moving to more robust coding practices. This is a small focused PR that might be a good place to start adding best-practices in one of the critical areas of the code base.

wdyt?

PR #278, opened 2021-06-30
Add new make commands

krtk6160:


Makefile
```diff
@@ -2,13 +2,21 @@ start-deps:
 	docker-compose up -d
 	direnv reload
 
+start: start-deps
+	yarn build && node lib/entrypoint/graphql.js | yarn pino-pretty -c -l
```
bodymindarts, line 6: Don't you need to source `.envrc` here too?

Makefile
```diff
@@ -2,13 +2,21 @@ start-deps:
...
+
+watch:
+	yarn nodemon -V -e ts,graphql -w ./src -x make start
+
 clean-deps:
 	docker-compose rm -sfv
 
 reset-deps: clean-deps start-deps
 
+restart-integration: reset-deps integration
```
bodymindarts, line 16: Is there a need for both `restart-integration` and `integration`? Serious question in your workflow do you ever need both? If not I think we can just use `integration`:
```
integration: restart-deps
  . ./.envrc && yarn jest --forceExit --bail --runInBand --verbose $$TEST | yarn pino-pretty -c -l
```

Makefile
krtk6160: It works without it. Could you try it out on your machine too?

Makefile
krtk6160: Done

PR #283, opened 2021-07-01
Increase lnd scrape job timeout and interval

krtk6160:


charts/monitoring/values.yaml
```diff
@@ -200,8 +200,8 @@ prometheus:
       scrape_configs:
         - job_name: 'lnd'
 
-          scrape_interval: 20s
-          scrape_timeout: 10s
+          scrape_interval: 300s
+          scrape_timeout: 15s
```
nicolasburtey, line 204: let's use 60s here?

PR #286, opened 2021-07-01
Add restart-integration command

krtk6160:
@bodymindarts Actually I realized I do use the "non-clean slate" integration command a lot. For example sometimes I want to re-run an idempotent test file multiple times while building a feature and having the `make integration` command do `reset-deps` made my dev workflow much slower.


bodymindarts:
can we name it `reset-integration` instead of `restart`?

That way we are keeping the convention `restart` = restart process but don't wipe state. `reset` = wipe state and restart.
Like with `start-deps` and `reset-deps`

PR #287, opened 2021-07-02
Standardize label injection

krtk6160:


charts/galoy/charts/lnpay/templates/deployment.yaml
```diff
@@ -4,6 +4,9 @@ metadata:
   name: {{ include "lnpay.fullname" . }}
   labels:
     {{- include "lnpay.labels" . | nindent 4 }}
```
nicolasburtey, line 6: do we still need this `lnpay.labels`?

charts/galoy/charts/lnpay/templates/deployment.yaml
```diff
@@ -19,6 +22,9 @@ spec:
       {{- end }}
       labels:
         {{- include "lnpay.selectorLabels" . | nindent 8 }}
```
nicolasburtey, line 24: do we still need this `lnpay.selectorLabels`?

charts/galoy/charts/lnpay/templates/deployment.yaml
nicolasburtey: or maybe, aslk differently, what is the difference between the include here and the the include below?

nicolasburtey (commented):
one question but look good to me otherwise

nicolasburtey:
tests are failing thought

samerbuna:
@krtk6160 please rebase if this is still needed or close if not.

PR #291, opened 2021-07-03
Config enable/disable the ip and proxy log information (Fixes GaloyMoney/galoy#130)

dolcalmi:
(Fixes GaloyMoney/galoy#130)
> on the ip stuff, would be good to have a way to 1/ deactive ip logging 2/ deactive proxy checking (from the yaml)
- Added new config (ipRecording)
- Nest proxyChecking in the new configuration
- Added new validations in context and updateIPDetails
- Refactor ipinfo updates (remove double try/catch)

this PR replace https://github.com/GaloyMoney/galoy/pull/270 (the rebase created an overhead)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -157,34 +157,26 @@ export const updateIPDetails = async ({ ip, user, logger }): Promise<void> => {
     return
   }
 
-  let ipinfo
-
   try {
     // skip axios.get call if ip already exists in user object
     if (user.lastIPs.some((ipObject) => ipObject.ip === ip)) {
       return
     }
 
-    if (yamlConfig.ipRecording.proxyChecking.enabled) {
-      ipinfo = await fetchIP({ ip })
+    const res = await User.updateOne(
+      { "_id": user._id, "lastIPs.ip": ip },
+      { $set: { "lastIPs.$.lastConnection": Date.now() } },
+    )
+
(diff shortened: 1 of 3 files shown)


nicolasburtey (approved):
LGTM

src/utils.ts
```diff
@@ -157,32 +157,26 @@ export const updateIPDetails = async ({ ip, user, logger }): Promise<void> => {
...
-        { $set: { "lastIPs.$.lastConnection": Date.now() } },
+    const res = await User.updateOne(
+      { "_id": user._id, "lastIPs.ip": ip },
+      { $set: { "lastIPs.$.lastConnection": Date.now() } },
+    )
+
+    if (!res.nModified && yamlConfig.ipRecording.proxyChecking.enabled) {
+      const ipinfo = await fetchIP({ ip })
+      await User.findOneAndUpdate(
+        { "_id": user._id, "lastIPs.ip": { $ne: ip } },
```
krtk6160, line 174: Is the `$ne` check still necessary if we already check the `res.nModified`?

krtk6160 (commented):
One question otherwise LGTM

src/utils.ts
dolcalmi: well, really not sure about the real usage of the api or parallel reqs, etc... but since there is no a global lock is better to have it

krtk6160 (approved):

PR #295, opened 2021-07-05
fix current log issue related to cancelled invoice

nicolasburtey:
the previous comment was misleading. it was expected that there is a mismatch between lnd and mongodb for the invoice after activating the auto invoice deleting on lnd. this PR should fix the issue.

### DIFF
--- a/src/Lightning.ts
+++ b/src/Lightning.ts
@@ -816,10 +816,17 @@ export const LightningMixin = (superclass) =>
       // so we need a branch to return true in case the payment
       // has been managed off lnd.
       if (invoice.is_canceled) {
-        this.logger.warn(
-          { hash, user: this.user },
-          "cancelled invoice. nothing to do. this should not happen(?)",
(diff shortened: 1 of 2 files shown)


krtk6160 (approved):

src/Lightning.ts
```diff
@@ -902,29 +909,9 @@ export const LightningMixin = (superclass) =>
             })
           }
         })
-        // this should not happen
       } else {
+        this.logger.error({ invoice }, "there is no reason to be in this branch")
```
krtk6160, line 913: Why do we need this log message?

nicolasburtey: because there is no reason to be in the branch :) but if we do, we need to look into it (It's similar to the error I was reporting in this PR which is actually a state that I didn't had in mind/didn't think we could go through when writing the code). I will have to review another time the logic of this else to make sure we can't go there.

PR #304, opened 2021-07-06
Okex dealer bot

sebastienverreault:
Extracting ccxt functionality into generic exchange object to be used for any exchanges

### DIFF
--- a/src/dealer/DealerWallet.ts
+++ b/src/dealer/DealerWallet.ts
@@ -10,25 +10,20 @@ import { UserWallet } from "../userWallet"
 import ccxt from "ccxt"
 import assert from "assert"
 
-import { GenericExchange, ApiConfig } from "./GenericExchange";
+import { GenericExchange, SupportedExchanges, ApiConfig } from "./GenericExchange";
 
-enum Exchanges {
-    FTX = "ftx",
-    OKEX = "okex5",
-}
-
-const activeExchangeId = Exchanges.FTX;
+const activeExchangeId = SupportedExchanges.FTX;
 
 const activeApiConfig = new ApiConfig(undefined, undefined, undefined);
 
-switch (activeExchangeId as Exchanges) {
-    case Exchanges.FTX:
+switch (activeExchangeId as SupportedExchanges) {
+    case SupportedExchanges.FTX:
         activeApiConfig.apiKey = process.env.FTX_KEY
         activeApiConfig.secret = process.env.FTX_SECRET
         activeApiConfig.password = process.env.FTX_PASSWORD
         break;
 
-    case Exchanges.OKEX:
+    case SupportedExchanges.OKEX:
         activeApiConfig.apiKey = process.env.OKEX_KEY
         activeApiConfig.secret = process.env.OKEX_SECRET
         activeApiConfig.password = process.env.OKEX_PASSWORD
--- a/src/dealer/DealerWallet.ts
+++ b/src/dealer/DealerWallet.ts
@@ -79,7 +79,7 @@ export class DealerWallet extends OnChainMixin(UserWallet) {
     async createDepositAddress() {
         // create a new address
         // is not enabled on FTX
-        // return this.ftx.createDepositAddress("BTC")
+        // return this.exchange.createDepositAddress("BTC")
         return Error("not implemented")
     }
 
@@ -123,7 +123,7 @@ export class DealerWallet extends OnChainMixin(UserWallet) {
 
     async getExchangeBalance() {
         const balance = await this.exchange.fetchBalance()
-        this.logger.debug({ balance }, "this.ftx.fetchBalance result")
+        this.logger.debug({ balance }, "this.exchange.fetchBalance result")
         return {
             sats: btc2sat(balance.total.BTC ?? 0),
             usdPnl: balance.total.USD,
@@ -150,7 +150,7 @@ export class DealerWallet extends OnChainMixin(UserWallet) {
 
         // tmp for log
         const result = await this.exchange.privateGetAccount()
-        this.logger.debug({ result }, "full result of this.ftx.privateGetAccount")
+        this.logger.debug({ result }, "full result of this.exchange.privateGetAccount")
 
         const {
             result: {
@@ -163,7 +163,7 @@ export class DealerWallet extends OnChainMixin(UserWallet) {
         } = result
         this.logger.debug(
             { collateral, positions, chargeInterestOnNegativeUsd, marginFraction },
-            "value kept from this.ftx.privateGetAccount",
+            "value kept from this.exchange.privateGetAccount",
         )
 
         const positionBtcPerp = _.find(positions, { future: this.symbol })
@@ -335,7 +335,7 @@ export class DealerWallet extends OnChainMixin(UserWallet) {
             try {
                 withdrawalResult = await this.exchange.withdraw(currency, btcAmount, address)
             } catch (err) {
-                const error = "this.ftx.withdraw() error issue"
+                const error = "this.exchange.withdraw() error issue"
                 subLogger.error({ withdrawalResult }, error)
                 throw new Error(err)
             }
--- a/src/dealer/DealerWallet.ts
+++ b/src/dealer/DealerWallet.ts
@@ -14,26 +14,25 @@ import { GenericExchange, SupportedExchanges, ApiConfig } from "./GenericExchang
 
 const activeExchangeId = SupportedExchanges.FTX;
 
-const activeApiConfig = new ApiConfig(undefined, undefined, undefined);
+const activeApiConfig = new ApiConfig(activeExchangeId, undefined, undefined, undefined, undefined);
 
 switch (activeExchangeId as SupportedExchanges) {
     case SupportedExchanges.FTX:
-        activeApiConfig.apiKey = process.env.FTX_KEY
-        activeApiConfig.secret = process.env.FTX_SECRET
-        activeApiConfig.password = process.env.FTX_PASSWORD
+        activeApiConfig.apiKey = process.env.FTX_KEY;
+        activeApiConfig.secret = process.env.FTX_SECRET;
+        activeApiConfig.password = process.env.FTX_PASSWORD;
         break;
 
     case SupportedExchanges.OKEX:
-        activeApiConfig.apiKey = process.env.OKEX_KEY
-        activeApiConfig.secret = process.env.OKEX_SECRET
-        activeApiConfig.password = process.env.OKEX_PASSWORD
+        activeApiConfig.apiKey = process.env.OKEX_KEY;
+        activeApiConfig.secret = process.env.OKEX_SECRET;
+        activeApiConfig.password = process.env.OKEX_PASSWORD;
         break;
 
     default:
         break;
 }
 
-const defaultSymbol = "BTC-PERP"
 
 // TODO: move to the yaml config
 const simulateOnly = true
@@ -49,7 +48,7 @@ export class DealerWallet extends OnChainMixin(UserWallet) {
 
     constructor({ user, logger }: ILightningWalletUser) {
         super({ user, logger })
-        this.exchange = new GenericExchange(activeExchangeId, activeApiConfig, defaultSymbol);
+        this.exchange = new GenericExchange(activeApiConfig);
         this.symbol = this.exchange.GetSymbol();
         this.logger = logger.child({ topic: "dealer" })
     }
--- a/src/dealer/DealerWallet.ts
+++ b/src/dealer/DealerWallet.ts
@@ -9,29 +9,28 @@ import { baseLogger } from "../logger"
 import { UserWallet } from "../userWallet"
 import assert from "assert"
 
-import { GenericExchange, SupportedExchanges, ApiConfig } from "./GenericExchange";
+import { GenericExchange, SupportedExchanges, ApiConfig } from "./GenericExchange"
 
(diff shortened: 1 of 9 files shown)


src/dealer/DealerWallet.ts
```diff
@@ -0,0 +1,515 @@
...
+}
+
+const activeApiConfig = new ApiConfig(activeExchangeId, key, secret, password)
+
+// TODO: move to the yaml config
+const simulateOnly = true
+
+export type IBuyOrSell = "sell" | "buy" | null
+
+// FtxDealer is a user of the wallet, therefore we are inheriting from UserWallet
```
nicolasburtey, line 31: comment probably needs to be updated

src/dealer/GenericExchange.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    this.symbol = exchangeMapping[this.exchangeId]
+  }
+
+  public has() {
+    const has = this.exchange.has
+    return has
+  }
+
+  public has2() {
```
nicolasburtey, line 55: ?

src/dealer/GenericExchange.ts
```diff
@@ -0,0 +1,163 @@
...
+    const has = this.exchange.has
+    return has
+  }
+
+  public name() {
+    const name = this.exchange.name
+    return name
```
nicolasburtey, line 62: `return this.exchange.name` ? we want to have concise code. there is no need to define a const here.

src/dealer/GenericExchange.ts
```diff
@@ -0,0 +1,163 @@
...
+    return await this.exchange.loadMarkets(true)
+  }
+
+  public async getFilteredInstrument(base: string, type: string) {
+    const markets = await this.exchange.loadMarkets(true)
+    let data = ""
+
+    for (const symbol in markets) {
+      try {
+        if (!symbol.includes("MOVE")) {
```
nicolasburtey, line 109: MOVE? https://ftx.global/move-contracts?

src/dealer/DealerWallet.ts
```diff
@@ -0,0 +1,515 @@
...
+if (!key || !secret || !password) {
+  throw new Error(`Missing ${activeExchangeId} exchange environment variables`)
+}
+
+const activeApiConfig = new ApiConfig(activeExchangeId, key, secret, password)
+
+// TODO: move to the yaml config
+const simulateOnly = true
+
+export type IBuyOrSell = "sell" | "buy" | null
```
bodymindarts, line 29: I don't like this being potentially `null`.
Can we have the type be `IBuyOrSell` and then where its consumed use `arg: IBuyOrSell | null`? Its better if the types we define have uniform behaviour and the `null` is called out locally at point of usage.

src/dealer/DealerWallet.ts
```diff
@@ -0,0 +1,515 @@
...
+    return {
+      // dealer is the only one account with the a negative balance for USD
+      // FIXME: look if there is a cleaner design than just have a - sign here
+      usd: -USD,
+      satsLnd,
+      satsFtx,
+    }
+  }
+
+  async has() {
```
bodymindarts, line 63: Can we have a more descriptive name here? Has what exactly?

src/dealer/DealerWallet.ts
```diff
@@ -0,0 +1,515 @@
...
+  async createDepositAddress() {
+    // create a new address
+    // is not enabled on FTX
+    // return this.exchange.createDepositAddress("BTC")
+    return Error("not implemented")
+  }
+
+  async exchangeDepositAddress() {
+    // same address is returned each time
+    const { address } = await this.exchange.fetchDepositAddress("BTC")
```
bodymindarts, line 76: This can be returned directly

bodymindarts: Also `"BTC"` should be a constant here. Don't use magic values.

src/dealer/DealerWallet.ts
```diff
@@ -0,0 +1,515 @@
...
+
+    const {
+      netSize = 0,
+      estimatedLiquidationPrice,
+      collateralUsed,
+      maintenanceMarginRequirement,
+    } = positionBtcPerp ?? {}
+
+    // TODO: check this is the intended settings
+    assert(chargeInterestOnNegativeUsd === true)
```
bodymindarts, line 169: What is the consequence of this failing at runtime?

src/dealer/DealerWallet.ts
```diff
@@ -0,0 +1,515 @@
...
+
+    assert(netSize <= 0)
+
+    return {
+      // making netSize positive to simplify calculation
+      // we should always be short so netSize should initially
+      // always be negative
+      btc: -netSize,
```
bodymindarts, line 177: Instead of a comment can we codify this expectation via a function name or type?
This kind of implicit conversion can be dangerous and introduce subtle bugs if somebody isn't aware or forgets.
Some ideas how to improve it:
Instead of `btc` call the value: `btcShortPosition` or something else more descriptive.
Instead of `~` create an explicit function `normalizePositionSize`. Even if its just 1 line it'll help communicate what we are doing.
Perhaps introduce a wrapper type `NormalizedPositionSize(int)` that only gets returned from that function.

I think this approach will reduce the likelihood of introducing errors later on.

src/dealer/DealerWallet.ts
```diff
@@ -0,0 +1,515 @@
...
+  }
+
+  // we need to rebalance when price is increasing/decreasing.
+  // if price increase, then the short position risk being liquidated
+  // so we need to sned btc from the node to the exchange
+  // if price decrease, then there would be too much btc on the exchange
+  // the account won't be at the irsk of being liquidated in this case
+  // but then the custody risk of the exchange increases
+  static isRebalanceNeeded({ usdLiability, btcPrice, usdCollateral }) {
+    type IDepositOrWithdraw = "withdraw" | "deposit" | null
```
bodymindarts, line 207: Can we avoid using `| null` in our types. What exactly does `null` signify here? Does it mean `no action required`? Then make the type `type IDepositOrWithdraw = "withdraw" | "deposit" | "neither" or something like that.

src/dealer/DealerWallet.ts
```diff
@@ -0,0 +1,515 @@
...
+    return { btcAmount, depositOrWithdraw }
+  }
+
+  static isOrderNeeded({ usdLiability, usdExposure, btcPrice }) {
+    const { ratio } = this.getExposureRatio({ usdLiability, usdExposure })
+
+    let usdOrderAmount
+    let buyOrSell: IBuyOrSell = null
+
+    yamlConfig.hedging.LOW_SAFEBOUND_RATIO_SHORTING
```
bodymindarts, line 259: Can we inject this in the constructor and have it as a local field? Calling out to a config file from a method makes things harder to test.

src/dealer/DealerWallet.ts
```diff
@@ -0,0 +1,515 @@
...
+    }
+
+    return { btcAmount, depositOrWithdraw }
+  }
+
+  static isOrderNeeded({ usdLiability, usdExposure, btcPrice }) {
+    const { ratio } = this.getExposureRatio({ usdLiability, usdExposure })
+
+    let usdOrderAmount
+    let buyOrSell: IBuyOrSell = null
```
bodymindarts, line 257: Here again... please use a more descriptive value for `uninitialised` than `null`.

src/dealer/DealerWallet.ts
```diff
@@ -0,0 +1,515 @@
...
+
+    yamlConfig.hedging.LOW_SAFEBOUND_RATIO_SHORTING
+
+    try {
+      // long (exposed to change in price in BTC)
+      // will loose money if BTCUSD price drops
+      if (ratio < yamlConfig.hedging.LOW_BOUND_RATIO_SHORTING) {
```
bodymindarts, line 264: Here again the `yamlConfig` shouldn't be referenced directly. There should be some kind of `strategy` object that gets passed in in the constructer so we can test different configurations without needing to call out to `yaml`

src/dealer/DealerWallet.ts
```diff
@@ -0,0 +1,515 @@
...
+
+      let withdrawalResult
+
+      subLogger = subLogger.child({ memo, address })
+
+      try {
+        withdrawalResult = await this.exchange.withdraw(currency, btcAmount, address)
+      } catch (err) {
+        const error = "this.exchange.withdraw() error issue"
+        subLogger.error({ withdrawalResult }, error)
```
bodymindarts, line 328: Shouldn't log error here. If it is an expected error we should catch it specifically and translate to a return value otherwise rethrow and log at a higher level.

src/dealer/DealerWallet.ts
```diff
@@ -0,0 +1,515 @@
...
+        .commit()
+
+      subLogger.info({ memo, address }, "deposit rebalancing succesful")
+    }
+  }
+
+  async executeOrder({ buyOrSell, btcAmount }) {
+    let order, orderStatus
+
+    // let orderId = 6103637365
```
bodymindarts, line 378: What are these comments for?

bodymindarts (changes requested):
In general I think you need to look at:
- use more explicit naming in functions and fields eg `has`, `has2`. `btc` vs `netBtcShortPosition` etc. Longer names are always useful if they communicate better.
- You have a lot of comments with `TODO` and `Fixme` and others. I'm not sure which of these are really relevant at this point and which are left here by accident. Please remove all comments that contain code and review which of the `TODO`s and `Fixme`s are really requirements (then add a GitHub ticket) or nice to haves (then delete it). I'm not saying all the comments have to go, but I do sense that there are more here than necessary
- Review your use of `try/catch`
- Review you adding `| null` to your defined types

src/dealer/DealerWallet.ts
nicolasburtey: yup this is code from me. it could be something like `"sell"| "buy" | "noAction"`?

src/dealer/DealerWallet.ts
nicolasburtey: this is a wrapper around a ccxt function. `has()` give the list of function/property the exchange has enabled.

src/dealer/DealerWallet.ts
bodymindarts: Do we need the `buy`/`sell` to be a string? Are we parsing this type from the return of an API call? If its something we control ourselves then I think an enum would be more appropriate:
```
export enum TradeAction {
  Buy,
  Sell,
  NoAction,
}
```

src/dealer/DealerWallet.ts
bodymindarts: Well since we're wrapping the library its a great opportunity to add a more descriptive name so its clearer at the point of usage in our code. Perhaps `exchangeHasProperty` or something like that?

src/dealer/DealerWallet.ts
sebastienverreault: Done! If using LN with the exchange as we propose to do with OKEx, should the DealerWallet inherit from the LightningMixin?

src/dealer/GenericExchange.ts
sebastienverreault: Cleaned up.

src/dealer/GenericExchange.ts
sebastienverreault: Done and updated other similar code

src/dealer/GenericExchange.ts
sebastienverreault: They have a ton of different derivatives, MOVEs are some of them and they were noisy in the plethora of instruments this api method lists. 
I have removed if for now as it is not relevant to getting a first version of the okex dealer online

src/dealer/DealerWallet.ts
sebastienverreault: - I will review the names to communicate more on a second pass, I was trying to alter the code to a minimum, so left the comments and todos also as is.
- try/catch was on my list also, we have some that just "safe guard" our own asserts but mostly they might be better place around the 3rd party api library
- I agree with the typed strings, there's a few occurrences, I will review them
- done with extra vars pre return...

src/dealer/DealerWallet.ts
nicolasburtey: it means the exchange account was not set properly @bodymindarts. it's a one time settings. probably only applies to FTX.

nicolasburtey (commented):
why did you create whole new file instead of amending the existing one?

nicolasburtey:
this is not a good engineer practice: image

src/dealer/OkexPerpetualSwapStrategy.ts
```diff
@@ -58,31 +59,31 @@ export class OkexPerpetualSwapStrategy implements IHedgingStrategy {
...
       } else {
         if (hedgingOrder.output.IsSimulation) {
-          logger.debug(`Calculated a SIMULATED new hedging order: ${hedgingOrder}`)
+          logger.debug({
+            msg: "Calculated a SIMULATED new hedging order",
+            order: hedgingOrder,
+          })
         } else {
-          logger.debug(`Calculated a new hedging order: ${hedgingOrder}`)
+          logger.debug({ msg: "Calculated a new hedging order", order: hedgingOrder })
```
nicolasburtey, line 78: this is not how the logger should be used. logger documentation can be read here: https://github.com/pinojs/pino/

typically you have `log.info({ variables }, "your logging message without any variable")

nicolasburtey:
the build is failing also. need to fix the `Check code` build stage

PR #307, opened 2021-07-07
injecting apollo studio env

nicolasburtey:
@bodymindarts @krtk6160 I'll inject the secret manually. curious how to think about this in the context of an automated deployment

### DIFF
--- a/charts/galoy/templates/deployment.yaml
+++ b/charts/galoy/templates/deployment.yaml
@@ -140,22 +140,6 @@ spec:
                 secretKeyRef:
                   name: network
                   key: network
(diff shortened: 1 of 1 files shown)


samerbuna (approved):

bodymindarts:
> @bodymindarts @krtk6160 I'll inject the secret manually. curious how to think about this in the context of an automated deployment

Pre-provisioning the secret in Kubernetes that gets read via `secretKeyRef:` will be a step in the deployment process. Before the secret gets provisioned in Kubernetes it will be stored in the secret backend of the CI server.

bodymindarts (approved):

PR #314, opened 2021-07-08
Fix Dockerfile-debug issue with CI

dolcalmi:
Must fix Directory /app/test in the roots[0] option was not found.

I dont have access to run CircleCI so would be great if runs before merge PR


bodymindarts:
Have you verified that it works locally with this fix?

bodymindarts (approved):

samerbuna (approved):

PR #315, opened 2021-07-08
Add payer memo sharing

### LINKED ISSUE #301: share memo from the payer to the payee
Currently when a payment is being sent with payToUsername, the memo is not shared from the payer to the payee. Salim have given the feedback that this is a critically missing features, because it prevent sending a message to the payee like "$10 for the pizza" or "electricity bill payment account 1234"

This has not been done historically for privacy reason, but it seems this is very much needed. 

Maybe there could be a switch "share to the user" to enabled/disable memo sharing. or maybe everyone wants memo sharing and we should just change the behavior here.

krtk6160:
Fixes #301


src/types.ts
```diff
@@ -200,3 +200,14 @@ export const OnboardingEarn = {
   moneyLaundering: 500,
   tweet: 1000,
 }
+
+// TODO: Add types for payer, payee and metadata
+export type IAddTransactionOnUsPayment = {
+  description: string
+  sats: number
+  metadata: Record<string, unknown>
+  payerUser: any
```
nicolasburtey, line 209: use of `any` is no longer allowed with our eslint/typescript rules

default.yaml
```diff
@@ -117,3 +117,5 @@ lnds:
 onchainDustAmount: 5000
 # TODO
 # carrierRegexFilter: ""
+
+shareMemoWithPayee: true
```
nicolasburtey, line 121: I don't think we want a global level settings for that value. It doesn't make sense to switch this on/off really.

eventually we could have a way, within the query. to choose whether one user can select to share the memo or not. 

for now I would share the memo and see what feedback we have once this is merged.

nicolasburtey (commented):
some feedback on this PR

src/types.ts
krtk6160: `any` does give a warning (not an error), so I'm using our `typeof User` workaround for now. I don't know if there's a better way to do it (without adding types to our mongoose code).
@samerbuna Do you have a suggestion on this?

default.yaml
krtk6160: Fixed

src/types.ts
samerbuna: I don't think `typeof User` is a good idea but I see it used in the mongoose schema too, so maybe leave it for now and create a new issue to refactor all mongoose types to use interfaces extending mongoose documents.

src/Lightning.ts
```diff
@@ -374,6 +374,7 @@ export const LightningMixin = (superclass) =>
                 payerUser: this.user,
                 payeeUser,
                 memoPayer,
+                shareMemoWithPayee: pushPayment ? true : false,
```
samerbuna, line 377: If `pushPayment` can only have boolean values then we don't need the ternary here.

Maybe also consider renaming `pushPayment` to something that indicates it being a boolean.

krtk6160: I think `isPushPayment` should work

samerbuna: Get rid of the ternary?

  `shareMemoWithPayee: isPushPayment,`

krtk6160: > Get rid of the ternary?
> 
> `shareMemoWithPayee: isPushPayment,`

Yep that's what I did

krtk6160: Fixed now

samerbuna (dismissed):

src/ledger/transaction.ts
```diff
@@ -79,14 +80,16 @@ export const addTransactionOnUsPayment = async ({
   payerUser,
   payeeUser,
   memoPayer,
-}) => {
+  shareMemoWithPayee,
```
nicolasburtey, line 83: why is this needed?

isn't that if there is a memo then this could be shared, and if there is not then this will be null?

src/ledger/transaction.ts
```diff
@@ -79,14 +80,16 @@ export const addTransactionOnUsPayment = async ({
...
+}: IAddTransactionOnUsPayment) => {
   const dealerPath = await dealerMediciPath()
 
   const entry = MainBook.entry(description)
 
   entry
     .credit(customerPath(payeeUser._id), sats * payeeUser.ratioBtc, {
       ...metadata,
+      memoPayer: shareMemoWithPayee ? memoPayer : null,
```
nicolasburtey, line 92: ie: why is there a need to do that instead of just having a `memoPayer` that will just resolved as null if there is no memo?

nicolasburtey (commented):
one comment

src/ledger/transaction.ts
krtk6160: The function is called with `shareMemoWithPayee: true` only if it's a push payment.
From #301 ->
> Currently when a payment is being sent with payToUsername, the memo is not shared from the payer to the payee.

src/ledger/transaction.ts
krtk6160: Because we want to have the sender's memo be shared with the receiver only if it's a pushpayment. If it's an invoice based payment, and we directly pass memoPayer, then the receiver's memo (if any) will get overwritten and not be returned as part of the `Wallet` query

nicolasburtey: ok that makes sense. we'll need to document those edge cases when we'll be working on the API documentation

nicolasburtey:
I have been doing some client interview today. and it seems on strike (which I believe is sharing memo information) some people are trying to scam user by sending a 1 sat tx with the message "click here to pay 10 dollars from strike". 

I'm not sure how we should be thinking about that. currently there is no such issue because of 

maybe @bodymindarts you would have some idea about that? we could filtering the message, or making is more clear in the UI I'm not too sure how to think about it yet. we should also know if some users are sending a lot of 1 sat payment to other users, it could be a sign that something like this is happening.

krtk6160:
> I have been doing some client interview today. and it seems on strike (which I believe is sharing memo information) some people are trying to scam user by sending a 1 sat tx with the message "click here to pay 10 dollars from strike". 

I don't quite understand how this scam works. Do you mean to say that the receiver looks at the memo and believes it's a 10$ tx rather than looking at the actual amount received?

nicolasburtey:
> I don't quite understand how this scam works. Do you mean to say that the receiver looks at the memo and believes it's a 10$ tx rather than looking at the actual amount received?

what I have understood: the receiver would think he could received 10$ by tapping on this transaction and click on "pay". but instead of receiving $10 he would send $10.

nicolasburtey:
this PR needs a rebasing

@krtk6160 we have discussed an idea today about the spam/scam issue related to memo. this idea would be to share he memo only above a certain amount. but we can do it in a later PR.

nicolasburtey (approved):

bodymindarts (approved):

Issue #272, opened 2021-06-27
use a bitcoind hot wallet instead of lnd


nicolasburtey:
the current hot onchain wallet is based on lnd. we want to move the onchain hot wallet to bitcoind (or maybe [bdk](https://bitcoindevkit.org/) could be an alternative, haven't dive into it, but there is an initiative for [proof of reserve](https://github.com/ulrichard/bdk-reserves) which is something we want to get into) for the following reasons:

### disposable lnd
 the lnd nodes are currently being considered disposable. there is no state in lnd that are not being duplicated into mongodb in regards to balances/transactions. so deleting a lnd instance would not create any data loss for the uses perspective.

the reason we want to be able to do that is because nodes can have issues and eventually die (like an irreparable corrupt database).

if lnd dies, then it's possible to recover the funds using a recovery procedure and starts anoter another lnd. (our architecture will allow multi lnd also).

aftter lnd is offline, no more payment can be received on the disposed lnd: even if there were some invoices issued by the disposed lnd that another party tried to pay, the payment would not go through because of the liveness requirement to receive a payment.

all the above logic of payment doesn't apply for the onchain wallet. a user could have shared an onchain address, and received a payment later. if the fees are high (and moving higher) maybe a user will be waiting for its transaction to confirm on chain for weeks/months.

the current issue is that the onchain addresses of the users are tied to a particular lnd. so if this lnd is being disposed, the pending/future onchain payment send to this node would still be eventually received, but unless the node is being restarted, those payment would never be credited to the associated user.

so the best way to mitigate this is create a separate the hot onchain wallet that doesn't have as much state to maintain as a lnd node. a hot bitcoin wallet could be restored easily with an xpriv to another bitcoind node.

this is the key reason as to why moving from the current hot onchain wallet relying on lnd to something that require less state and would likely never need to be recycled.

### not focus on onchain
lightning labs doesn't focus on onchain wallet as much as lightning (which is understandable). some example of issues that have been pending: 
- https://github.com/GaloyMoney/galoy/blob/505c2011aeb4db8f27f18d51f95d3596f6b37a4e/src/entrypoint/trigger.ts#L58
- https://github.com/lightningnetwork/lnd/issues/5244

I would expect bitcoind to not have such issues

jotapea:
[...]

PR #317, opened 2021-07-09
Bitcoind for onchain

jotapea:
WIP: Starting PR for https://github.com/GaloyMoney/galoy/issues/272


bodymindarts:
@jotapea please pay attention to your commit messages we are trying to follow [this guide](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) for standard commit messages (first line 50 characters).

And the `check-code` task is failing on GitHub actions. Perhaps you could setup your editor to run `prettier` on save to avoid this.

jotapea:
@bodymindarts Will do, thanks for the feedback!

test/integration/02e-onchain-send.spec.ts
```diff
@@ -62,6 +62,7 @@ afterAll(async () => {
 
 const amount = 10040 // sats
 
+// Starts to fail from here...
```
dolcalmi, line 65: if it is a timeout fail is because waitUntilBlockHeight. Basically in lnd the "pruning channel graph" is taking more than the timeout in jest.
- you can check docker logs to verify that this is the case
- increase timeout in jest setup https://github.com/GaloyMoney/galoy/blob/main/test/jest.setup.js

src/OnChain.ts
```diff
@@ -59,6 +60,111 @@ export const getOnChainTransactions = async ({
...
+    }[],
+  ): Promise<number>
+
+  // return txid
+  abstract sendToAddress(address: string, amount: number): Promise<string>
+
+  abstract getTxnFee(id: string): Promise<number>
+}
+
+class LndOnChainClient extends PayOnChainClient {
```
dolcalmi, line 87: is a highly coupled class with getActiveOnchainLnd. Maybe is better pass the lnd client as parametter (the same for BitcoindClient)

test/integration/02e-onchain-send.spec.ts
jotapea: Thanks for this, modifying the timeout has allowed me to retry the tests when these show the following error:

```
FAIL test/integration/02a-fund-bitcoind.spec.ts (36.247 s)
  ✓ funds bitcoind wallet (3697 ms)
  ✕ funds outside lnd node (30268 ms)

  ● funds outside lnd node

    thrown: "Exceeded timeout of 30000 ms for a test.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."

      66 | })
      67 |
    > 68 | it("funds outside lnd node", async () => {
```

But this seems to be "unrelated?" to the the main failure that the tests are showing, which is that the balance sheet is no longer balanced.

src/OnChain.ts
jotapea: The main purpose of the classes at the moment is for ease of development purposes. After the bitcoind client works as a replacement to the lnd client, then we could reconsider the purpose and structure of these classes.

dolcalmi: but you have only one instance (line 68) and in the same file

src/OnChain.ts
```diff
@@ -217,16 +323,17 @@ export const OnChainMixin = (superclass) =>
             throw new TransactionRestrictedError(error, { logger: onchainLogger })
           }
 
-          const { lnd } = getActiveOnchainLnd()
+          // const { lnd } = getActiveOnchainLnd()
```
jotapea, line 326: @dolcalmi Notice that this is the same call that is now in line 90 (this.client = getActiveOnchainLnd().lnd), so it should work just like it used to.

test/integration/02a-fund-bitcoind.spec.ts
```diff
@@ -65,6 +63,22 @@ it("funds bitcoind wallet", async () => {
   expect(balance).toBe(initialBitcoinWalletBalance + blockReward * numOfBlock)
 })
 
+// This new test fails afterFinished: checkIsBalanced()
```
jotapea, line 66: All tests still pass after I renamed the default wallet from "" (unnamed) to "default". But adding the test to create a second wallet ("hot") fails just after creation. Made a comment of this failure in the balance sheet: https://github.com/GaloyMoney/galoy/pull/317/files#diff-67ab19f16c472b30989f27f5755650158189756bbf68874767a0ff7196c50a05

@nicolasburtey @krtk6160 how should I proceed? Why is the "default" wallet not being considered for the balance sheet tests yet?

dolcalmi: not sure if you can change the default wallet. The current production environment use "" as default wallet

test/integration/02e-onchain-send.spec.ts
bodymindarts: @jotapea FYI tests are NOT idempotent so you always have to run `make reset-deps` between runs.

test/integration/02a-fund-bitcoind.spec.ts
nicolasburtey: @jotapea will need to dig deeper to understand why it's failing but some feedback on top of my head that could be relevant:
- since bitcoind 0.20 (or 0.21?) the behavior of bitcoind changed in regards to defauilt wallet and now no default wallet is being created automatically. the code has been udpated to accomodate this but IIRC it was a bit hacky
- we currently only use the bitcoin wallet in production is for the multisig that is created by specter. the other use of bitcoind is currently only for the test (maybe we should have 2 bitcoind instance to make this more explicit... one that is for the wallet we use, and another bitcoind that represents the "outside world")
- the test instantiated a separate wallet just because we need to have some BTC coming from somewhere. lnd is not getting the BTC directly from the coinbase rewards because the funds are not available immediately and I had issues with that: lnd typically doesn't expect to be the recipient of a coinbase rewards that has timelock restriction. (need to wait X blocks before the coinbase rewards is spendable)

test/integration/02e-onchain-send.spec.ts
jotapea: Yes I'm running them with 'make integration' the first time and then 'make reset-integration' the rest of the times.

test/integration/02a-fund-bitcoind.spec.ts
jotapea: @nicolasburtey Renamed the second test from 02a-fund-bitcoind to 02a-fund-outside

test/integration/08-specter.spec.ts
```diff
@@ -62,38 +62,39 @@ it("createWallet", async () => {
   // expect(balance).toBe(0)
 })
 
-it("deposit to bitcoind", async () => {
```
jotapea, line 65: This test is being commented out (at least for now). Many of SpecterWallet methods are tightly coupled to the outside (default) wallet being bypassed in tests. In addition to also considering lnd as the onchain wallet.

jotapea:
FYI, the tests are passing in my machine, different from: https://github.com/GaloyMoney/galoy/pull/317/commits/a5d04335d24dd94ef3c40bc215b4b51542d37358

test/integration/02b-onchain-receive.spec.ts
```diff
@@ -183,32 +183,33 @@ it("createsBitcoindHotWallet", async () => {
...
 })
 
 it("fundingFunderWithOnchainTxFromBitcoind", async () => {
-  //   await bitcoind_onchain_funding({ walletDestination: funderWallet })
-  await lnd_onchain_funding({ walletDestination: funderWallet })
+  await bitcoind_onchain_funding({ walletDestination: funderWallet })
+  //   await lnd_onchain_funding({ walletDestination: funderWallet })
 })
 
+// Now tests start to fail here (same reason: checkIsBalanced)
```
jotapea: This is what I plan to continue working on... but @nicolasburtey let me know if this is the best path or if I should focus on something else.

jotapea:
To replicate the architecture of listening for lnd's chain_transaction and block events with bitcoind, [ZeroMQ](https://github.com/bitcoin/bitcoin/blob/master/doc/zmq.md) needs to be enabled https://github.com/GaloyMoney/galoy/issues/346.

jotapea:
Dividing this into multple tasks so that we can review and merge smaller branches, these being preparation for bitcoind fully replacing lnd for onchain transactions:

- First: [Replace default bitcoind wallet with multiple purpose specific wallets](https://github.com/GaloyMoney/galoy/issues/353)
- [Listen to bitcoind with ZeroMQ](https://github.com/GaloyMoney/galoy/issues/346)
- ...
- Finally this one: [Bitcoind for onchain](https://github.com/GaloyMoney/galoy/pull/317)

src/OnChain.ts
```diff
@@ -59,14 +66,205 @@ export const getOnChainTransactions = async ({
...
 }
 
+export const getOnChainTransactionsBitcoind = async ({
+  incoming, // TODO?
+}: {
+  incoming: boolean
+}) => {
+  try {
+    // TODO? how many transactions back?
+    const transactions: [] = await bitcoindHotWalletClient.listTransactions(null, 1000)
```
nicolasburtey, line 76: 1000 should be a separate const

src/OnChain.ts
nicolasburtey: is this still relevant @dolcalmi ?

src/OnChain.ts
```diff
@@ -59,14 +66,205 @@ export const getOnChainTransactions = async ({
...
+    return id
+  }
+
+  async getTxnFee(id: string): Promise<number> {
+    const outgoingOnchainTxns = await getOnChainTransactions({
+      lnd: this.client,
+      incoming: false,
+    })
+    // eslint-disable-next-line
+    let fee
```
nicolasburtey, line 211: why are `fee` and `fee_` necessary?

PR #318, opened 2021-07-09
Enable redis password auth

krtk6160:


charts/galoy/values.yaml
```diff
@@ -87,7 +87,7 @@ redis:
     persistence:
       enabled: true
   auth:
-    enabled: false
+    password: password
```
nicolasburtey, line 90: password is automatically generated?

krtk6160:
Nope. The redis chart differs from the mondo chart in this aspect. We need
to provide the password explicitly via the values.

charts/galoy/values.yaml
krtk6160: Nope. The redis chart differs from the mondo chart in this aspect. We need
to provide the password explicitly via the values.

nicolasburtey: how is the password defined in that case currently? You'll manually set it?

nicolasburtey (approved):

charts/galoy/values.yaml
krtk6160: Yes, via the values.yaml. For testnet and mainnet we have our regular flow of injecting from the `configs` repo and for regtest we have it in the default values.yaml file

PR #321, opened 2021-07-11
Pass config via constructor

krtk6160:


src/types.ts
```diff
@@ -14,6 +14,11 @@ export interface ILightningWalletUser {
   // FIXME: Add a type for user here.
   user: unknown
   logger: Logger
+  config: Config
+}
+
+export type Config = {
```
bodymindarts, line 20: Can this type have a more descriptive name? `WalletConfig` perhaps?

bodymindarts (changes requested):

src/types.ts
krtk6160: This is "the" config. Do we still need to make this more descriptive despite that fact?

src/types.ts
```diff
@@ -10,12 +10,26 @@ export type Logger = PinoLogger
...
   logger: Logger
 }
 
+export type UserWalletConstructorArgs = WalletConstructorArgs & {
+  config: UserWalletConfig
+}
+
+export type UserWalletConfig = {
+  dustThreshold: number
+  limits: {
```
bodymindarts, line 25: you can just use the `TransactionLimits` type here and call the functions in the code. No need to have _2_ layers of config wrapping

bodymindarts (approved):

src/Lightning.ts
```diff
@@ -59,8 +63,11 @@ export const delay = (currency) => {
 
 export const LightningMixin = (superclass) =>
   class extends superclass {
+    readonly config: UserWalletConfig
+
     constructor(...args) {
       super(...args)
+      this.config = args[0].config
```
samerbuna, line 70: Minor improvement. Maybe give `args[0]` a name here for better readability.

src/OnChain.ts
```diff
@@ -87,7 +89,7 @@ export const OnChainMixin = (superclass) =>
       if (payeeUser) {
         fee = 0
       } else {
-        if (amount && amount < yamlConfig.onchainDustAmount) {
+        if (amount && amount < this.config.dustThreshold) {
```
samerbuna, line 92: Shouldn't this be `onChainWallet.dustThreshold`?

src/config.ts
```diff
@@ -18,3 +20,19 @@ try {
 }
 
 export const yamlConfig = _.merge(defaultConfig, customConfig)
+
+export class TransactionLimits {
```
dolcalmi, line 24: This type of implementation must have tests (unit tests).

src/OnChain.ts
bodymindarts: I suggested to have the object specific config flattened. I like this because it reduces the amount of structural information that must be known at the point of usage. This also decouples the object specific config from the mechanism with by which the global config is provisioned. Achieving this decoupling was the primary goal of this refactoring.

src/Lightning.ts
vindard: Agreed, it looks like the eventual instantiations only ever have one object argument right now anyway ([here](https://github.com/GaloyMoney/galoy/blob/67e50d07b686f2d2a4c6a5e5ed01ab007986ce66/src/walletFactory.ts#L22), and [here](https://github.com/GaloyMoney/galoy/blob/67e50d07b686f2d2a4c6a5e5ed01ab007986ce66/src/walletFactory.ts#L36)).

Were you all expecting multiple arguments to those classes later on?

bodymindarts: I think going with 1 argument that is typed with a `XxxConstructorArgs` type and de-structuring via the `{fieldA, fieldB}` type syntax is a good pattern to go with. Using this `varargs` approach rarely makes sense.

vindard: Done in #424

PR #326, opened 2021-07-13
del InvoiceUser entry when unable to locate the invoice

nicolasburtey:
It seems there are different states of deletion for an invoice:
- first state is the invoice is written as "cancelled" but the reference is still present in the bbolt database.
- second state is that the invoice is being pruned and the reference is no longer present in the bbolt database.

we were currently not managing the second case correctly. this PR should fix it. I'm not sure how we can test this programmatically.

I'm not sure what make the invoice transition from 1 to 2. maybe a lnd restart with db compaction?
(is that correct @alexbosworth?)

### DIFF
--- a/src/Lightning.ts
+++ b/src/Lightning.ts
@@ -772,13 +772,13 @@ export const LightningMixin = (superclass) =>
       lock
       pubkey?: string
     }): Promise<boolean> {
-      let invoice, pubkey_
+      let invoice, pubkeyInvoiceUser
 
       // if a pubkey has been provided, it means the invoice has not been set as paid in mongodb
       // so not need for a round back trip to mongodb
       if (!pubkey) {
         let paid
-        ;({ pubkey: pubkey_, paid } = await InvoiceUser.findOne({ _id: hash }))
+        ;({ pubkey: pubkeyInvoiceUser, paid } = await InvoiceUser.findOne({ _id: hash }))
 
         if (paid) {
           return true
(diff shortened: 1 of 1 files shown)


src/Lightning.ts
```diff
@@ -807,8 +807,24 @@ export const LightningMixin = (superclass) =>
         // TODO: we should not log/keep secret in the logs
         this.logger.debug({ invoice, user: this.user }, "got invoice status")
       } catch (err) {
-        const error = `issue fetching invoice`
-        this.logger.warn({ err, invoice }, error)
+        const invoiceNotFound = "unable to locate invoice"
+        try {
```
bodymindarts, line 811: Why is this a `try/catch` instead of an `if/else`?

src/Lightning.ts
```diff
@@ -807,8 +807,24 @@ export const LightningMixin = (superclass) =>
...
+          this.logger.error(
+            { err, err2, invoice },
+            "issue fetching invoice. unknown error",
+          )
+          return false
+        }
+
+        this.logger.info({ err, invoice }, invoiceNotFound)
+        try {
+          await InvoiceUser.deleteOne({ _id: hash, uid: this.user._id })
```
bodymindarts, line 823: Should this always be tried here?
I think this code block 811 - 827 could be structured clearer via explicit if/else control flow.

krtk6160 (approved):

bodymindarts (commented):
I find it hard to follow the logic of the `try/catch` stuff here. I would advocate to re-write it with regular `if/else` control flow and only wrapping the external api calls.

That being said I understand this is some kind of emergency so perhaps we should merge and clean up later.

src/Lightning.ts
nicolasburtey: I agree this should be a if/else but the reason I did a try/catch instead if because the error being returned from the lightning library I believe is loosely defined (and currently not typed). so I'm not sure there is always an `err` property at `err[2]` and and that this `err` contains a `details` property. would need to parse the alexbosworth/lightning library to make this is the case. I have had previous issue with it in the past. (it's also an pseudo emergency PR)

PR #329, opened 2021-07-14
fix invoiceUser not existant issue

nicolasburtey:
this is related to https://github.com/GaloyMoney/lnpage/issues/57

### DIFF
--- a/src/Lightning.ts
+++ b/src/Lightning.ts
@@ -5,6 +5,7 @@ import {
   cancelHodlInvoice,
   createInvoice,
   getInvoice,
+  GetInvoiceResult,
   getPayment,
   payViaPaymentDetails,
   payViaRoutes,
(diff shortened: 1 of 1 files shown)


nicolasburtey:
@samerbuna I have to a `(invoice as GetInvoiceResult)` because I checked above that invoice should exist `if (!invoice)` so I'm not sure why this check is necessary. would love your insights on it.

samerbuna:
> @samerbuna I have to a `(invoice as GetInvoiceResult)` because I checked above that invoice should exist `if (!invoice)` so I'm not sure why this check is necessary. would love your insights on it.

It's a bit annoying but because the `invoice` variable is used in an async function after, TypeScript strictly cannot guarantee that the value of `invoice` has not changed before the async function gets executed. This is why using `const` instead of `let` is preferable since that makes that guarantee, basically. You can wrap the invoice getting logic in a function to make it possible to use `const` in this particular case.

nicolasburtey:
@samerbuna but I can't have a const because I'm wrapping up the call in a `try` `catch`?

samerbuna:
> @samerbuna but I can't have a const because I'm wrapping up the call in a `try` `catch`?

Right, that's why you'll need to wrap that whole block in a function that returns `invoice | undefined`

Issue #282, opened 2021-06-30
Improve testing structure


bodymindarts:
To start improving our test setup we want to bring the folder structure in line with common practices:
```
__tests__/{integration,helpers,...}
src
```
In other words `__tests__` should be at the root not inside `src`. The current long running test should go under `test/integration` as it depends on the runtime dependencies.

This opens up the possibility to create a new folder structure that mirrors the folders under `src` with 1 file per file in the production code for unit testing.

All tasks and build steps that depends on the current folder structure will need reviewing and adapting.

This was discussed as a first step to improve the testing with @dolcalmi.

nicolasburtey:
I don't think it's wise to put `__tests__` outside of `src/`. a lot of the config file that are within `src` (tsconfig, package.json, node_modules...) are necessary to run the tests. and they are within `src/`

bodymindarts:
All of the files you mention are in the root of the repo not within src. 🤔

bodymindarts:
To be clear I am proposing to move to this kind of structure:
```
$ tree repo
repo
├── __tests__
│   ├── fileTest.ts
│   ├── helpers
│   ├── integration
│   │   └── integrationTest.ts
│   └── subfolder
│       └── otherFileTest.ts
├── package.json
└── src
    ├── file.ts
    └── subfolder
        └── otherFile.ts
```


This is a [very](https://github.com/reduxjs/redux) [common](https://github.com/lodash/lodash) [structure](https://github.com/bitcoinjs/bitcoinjs-lib) in the [node js](https://github.com/nodejs/node) [ecosystem](https://github.com/npm/cli).

nicolasburtey:
oups yeah you're right! nevermind my comment above.

krtk6160:
Should probably use `tests` as the dir name instead of `__tests__`.

bodymindarts:
To apply the common pattern `test` is probably the right name.

krtk6160:
You're right, I misremembered the name.

nicolasburtey:
`__tests__` is the default folder for jest AFAIK. this can be overridden of course.

bodymindarts:
I don't really care what the folder name is `__tests__` or `test` or whatever... my point is regarding the structure.

nicolasburtey:
[...]

PR #310, opened 2021-07-08
Tests refactor - 1st stage

dolcalmi:
Move tests and mocks to test folder. Includes:
  - path changes
  - module import fixes
  - jest mocks and config fixes
  - neither the package/make/ci commands nor logic has been changed

Refs:
- https://github.com/GaloyMoney/galoy/issues/282
- https://galoymoney-workspace.slack.com/archives/C025EMR24KB/p1625698461365500

PR #340, opened 2021-07-17
Tests refactor - 3rd stage

dolcalmi:
This is a major refactor of the tests, please review carefully:

- **Structure change**:  same as `src` when is not an integration setup
- **New LNDs channel setup**: to avoid fee calc and syncing issues when close a channel. You can check the new setup in `test/integration/01-setup/01-lightning-channels.spec.ts`
- Helpers refactor
- Removed --forceExit need.
- New readme for test folder
- .skip was added to tests to be done or pending tests (previously commented)
- **General cleaning**:
  - delete files without tests
  - move "test files without tests" to debug folder (as @nicolasburtey requested)
  - move tests that have no dependencies to unit tests
  - Fix some commented tests
  - Clean unnecessary comments and all console.log from test folder
  - names updates. if you want to debug specific test or describe block please use .only (I added a comment in the new readme)
  - refactor mining and wait for sync methods

**src changes**

- PR [337](https://github.com/GaloyMoney/galoy/pull/337)

**Refs**:

- https://github.com/GaloyMoney/galoy/issues/282
- https://galoymoney-workspace.slack.com/archives/C025EMR24KB/p1625698461365500


test/README.md
```diff
@@ -0,0 +1,109 @@
+# Test
+
+## Prerequisites
+
+If you want to run it locally, you will need the following things properly installed on your computer.
+
+* [Direnv](https://direnv.net/)
+* [Node.js](https://nodejs.org/)
+* [Yarn](https://yarnpkg.com//)
```
krtk6160, line 9: Minor typo -> `https://yarnpkg.com//` should be `https://yarnpkg.com`

.circleci/config.yml
```diff
@@ -73,8 +73,8 @@ jobs:
       - run:
           name: Run jest test
           command: |
-            kubectl exec --namespace=$NAMESPACE $(kubectl get pods --namespace=$NAMESPACE --selector=app=testpod --output=jsonpath={.items..metadata.name}) -- /nodejs/bin/node node_modules/.bin/jest --ci --reporters=default --reporters=jest-junit --forceExit
-            # export JEST_JUNIT_OUTPUT_DIR=\"./artifacts\" && jest --ci --reporters=default --reporters=jest-junit --forceExit
+            kubectl exec --namespace=$NAMESPACE $(kubectl get pods --namespace=$NAMESPACE --selector=app=testpod --output=jsonpath={.items..metadata.name}) -- /nodejs/bin/node node_modules/.bin/jest --ci --reporters=default --reporters=jest-junit
+            # export JEST_JUNIT_OUTPUT_DIR=\"./artifacts\" && jest --ci --reporters=default --reporters=jest-junit
```
nicolasburtey, line 77: this is great

test/helpers/lightning.ts
```diff
@@ -0,0 +1,260 @@
...
+  openChannel,
+  sendToChainAddress,
+  subscribeToChannels,
+  subscribeToGraph,
+  updateRoutingFees,
+} from "lightning"
+
+import { bitcoindClient, RANDOM_ADDRESS } from "./bitcoinCore"
+
+export * from "lightning"
```
nicolasburtey, line 25: what is the goal of this?

test/integration/01-setup/01-connection.spec.ts
```diff
@@ -0,0 +1,70 @@
+import { setupMongoConnection } from "src/mongodb"
```
nicolasburtey, line 1: what is the reason to have the filename being `.spec.ts`?

test/integration/01-setup/01-metadata.spec.ts
```diff
@@ -0,0 +1,7 @@
+import { DbMetadata } from "src/schema"
```
nicolasburtey, line 1: this test is really to populating the database when doing manual request from the mobile app. the mobile app expect a `minBuildNumber` and `lastBuildNumber` to be returned from graphql.

maybe it shouldn't be there?

nicolasburtey: a db "`hydratation`"

test/integration/01-setup/01-lightning-channels.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+    const { channels } = await getChannels({ lnd: lnd1 })
+    expect(channels.length).toEqual(channelLengthMain + 1)
+
+    const { balance: finalFeeInLedger } = await MainBook.balance({
+      account: lndFeePath,
+      currency: "BTC",
+    })
+    expect(finalFeeInLedger - initFeeInLedger).toBe(channelFee * -1)
+
+    await setChannelFees({ lnd: lnd1, channel, base: 1, rate: 0 })
```
nicolasburtey, line 85: why do we set the rate at 0 here?

if we want to test how routing fees are affecting our balance sheet we should not have 0 rates.

test/integration/lnd-utils.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+  pay,
+} from "test/helpers"
+
+afterAll(async () => {
+  jest.restoreAllMocks()
+})
+
+describe("lndUtils", () => {
+  describe("updateRoutingFees", () => {
+    it("sets routing fee correctly", async () => {
```
nicolasburtey, line 20: what is the goal of this test?

test/integration/specter-wallet.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+      sats,
+      fee,
+      bitcoindBalance,
+      initBitcoindBalance,
+    })
+    expect(lndBalance).toBe(initLndBalance - sats - fee)
+  })
+
+  // TODO: Fix or remove. Expectations were commented out
+  it.skip("withdraw from bitcoind", async () => {
```
nicolasburtey, line 71: now that specter has it's API available, we should be able to do more integration/testing on that.

https://github.com/cryptoadvance/specter-desktop/pull/1232

nicolasburtey (commented):
Skimmed through it and it seems to be a very good PR. Haven't look into the details (and I think most of the diff is the result changing filename, so hard to review what changes on the code side versus what is highlighted on github here just because it's in a new directory). 
I have a couple of questions.

test/helpers/lightning.ts
dolcalmi: I tried to move moved all external references in tests to helpers, this is the first step to a next stage where we can mock classes, functions, etc to add more tests without runtime environment dependency

test/integration/01-setup/01-lightning-channels.spec.ts
dolcalmi: from lnd conf `total fee = basefee + (amount * feerate / 1000000)` so if we set feerate to 0 the total fee will be base fee (it is easier to calc fees in the tests)

ref: https://github.com/lightningnetwork/lnd/blob/master/sample-lnd.conf#L495

test/integration/lnd-utils.spec.ts
dolcalmi: basically test updateRoutingFees method (previous test https://github.com/GaloyMoney/galoy/blob/main/test/integration/09f-routing.spec.ts and related to https://github.com/GaloyMoney/galoy/blob/main/test/integration/04b-lightning-payment.spec.ts#L559). The method basically test how "routing fee" mainbook entry is set and the logic regarding routingFeeLastEntry in metadata

test/integration/01-setup/01-connection.spec.ts
dolcalmi: It is a common naming convention and It is easier to filter from helpers and configuration files. We discussed it in the 1st stage PR https://github.com/GaloyMoney/galoy/pull/310#pullrequestreview-701722944

test/integration/01-setup/01-metadata.spec.ts
dolcalmi: uhmm I thought it was required for other tests.. if is not testing anything and if it is not a setup then we can remove it. do you agree?

dolcalmi: although if the same environment is used to run e2e tests in mobile app then we should leave it as it is

nicolasburtey: there is no e2e test (yet) on the mobile side. 

I don't think we should just remove it but it doesn't need to be part of the integration test. it can be part of a script to be run before someone want to launch graphql to test the mobile app.

test/helpers/lightning.ts
nicolasburtey: ok

test/helpers/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  // arrives before wallet balances updates in lnd
+  await waitUntilChannelBalanceSyncAll()
+
+  const { assetsLiabilitiesDifference, bookingVersusRealWorldAssets } =
+    await balanceSheetIsBalanced()
+  expect(assetsLiabilitiesDifference).toBeFalsy() // should be 0
+
+  // FIXME: because safe_fees is doing rounding to the value up
+  // balance doesn't match any longer. need to go from sats to msats to properly account for every msats spent
+  expect(Math.abs(bookingVersusRealWorldAssets)).toBeLessThanOrEqual(5) // should be 0
```
nicolasburtey, line 28: maybe this could be 1 or 0 not that there is no variable routing fees?

the reason this was here is because lnd switch it's primary unit of account from `sats` to `msats` and because we only managed sats as of now in our books, we were doing rounding errors and a result the books were no longer accurate.

if we charge a routing fee of 1 sats/tx (is it 1 sats or 1 msats?) then we can reduce the error threshold here.

test/integration/01-setup/01-lightning-channels.spec.ts
nicolasburtey: ok make sense. have a new comment on rounding error in a helper file as a result of this.

test/integration/lnd-utils.spec.ts
nicolasburtey: how do you ensure that there is no direct channel between lndOutside1 and lndOutside2? shouldn't the order of test for this matters?

test/helpers/index.ts
dolcalmi: it is [1 sat](https://github.com/dolcalmi/galoy/blob/refactor_test_stg3a/test/helpers/lightning.ts#L191). I will run the tests changing the threshold

test/integration/lnd-utils.spec.ts
dolcalmi: :) https://github.com/dolcalmi/galoy/blob/refactor_test_stg3a/test/integration/01-setup/01-lightning-channels.spec.ts#L43 that is the reason to use a new channel configuration and only setup channels in the setup script

nicolasburtey: I'm not sure to full understand. 

https://github.com/dolcalmi/galoy/blob/a923ee88f2f084966ecac900521b2d95f2afe03a/test/integration/01-setup/01-lightning-channels.spec.ts

this file is run only once or for every tests?

dolcalmi: no, the channel setup only runs once for the integration test, what I did was to no setup the link between lndOutside1 and lndOutside2 (the channel closing was causing problems after finish the lightning sending tests) so all the payments between them go through lnd1

test/integration/utils.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+import { lndOutside1, lndOutside2, createInvoice, pay } from "test/helpers"
+
+describe("isInvoiceAlreadyPaidError", () => {
+  it("decodes error correctly", async () => {
+    const { request } = await createInvoice({ lnd: lndOutside2, tokens: 1000 })
+    await pay({ lnd: lndOutside1, request })
+    try {
+      await pay({ lnd: lndOutside1, request })
+    } catch (err) {
+      expect(isInvoiceAlreadyPaidError(err)).toBeTruthy()
```
bodymindarts, line 11: This is a bad way to expect an error.
- if this is just about decoding the error we should simply have a test that tests a util that does the decoding and not go through the `pay` function.
- if the `pay` function doesn't throw then the `catch` block is never executed and the test passes silently

If we really want to test that `pay` throws an exception we should use something like the [jest toThrow method](https://jestjs.io/docs/expect#tothrowerror). [Usage example in stack overflow](https://stackoverflow.com/questions/46042613/how-to-test-the-type-of-a-thrown-exception-in-jest).
But better would be to really only unit-test the decoding without the `pay` function in the call stack.

test/unit/config.spec.ts
```diff
@@ -2,7 +2,7 @@ import { defaultConfig } from "src/config"
 
 describe("config.ts", () => {
   describe("defaultConfig", () => {
-    it("should have hedging and name properties", () => {
+    it("has hedging and name properties", () => {
```
bodymindarts, line 5: This shouldn't be a test but a type.

test/unit/dealer/ftx-dealer-wallet.spec.ts
```diff
@@ -0,0 +1,146 @@
+import { redis } from "src/redis"
+import { FtxDealerWallet } from "src/dealer/FtxDealerWallet"
+
+beforeAll(async () => {
+  // avoids to use --forceExit and the need of a running redis
+  redis.disconnect()
```
bodymindarts, line 6: If we don't need redis for this file then we should rather have a way _not_ to start in except for the files were we need it instead of always starting it and then disabling. OR? Its better to explicitly require something (like a running redis) than implicitly require everything and then explicitly have to remove it when its not needed.

bodymindarts (commented):
Lots of files added I'm not sure what is your code and what was there before.

A few high-level / structural comments:
- lets please keep a root `DEV.md` instead of having 2 places where information regarding dev is written.
- Can we explicitly start the dependencies instead of having `redis.disconnect()` where its not needed?
- can we drop the `01` prefixes in the files (keep them in the folders)?

test/unit/dealer/ftx-dealer-wallet.spec.ts
dolcalmi: This is part of what we must try to gradually migrate. The `src/dealer/FtxDealerWallet` does not use redis explicitly but imports other files that use it and there is no way (with the current structure) to handle "disconnect" (similar to the PR related to express httpServer).

test/README.md
```diff
@@ -0,0 +1,109 @@
...
+```bash
+$ TEST=config yarn test:unit
+# or
+$ TEST=config make unit
+```
+where `config` is the name of the file `config.spec.ts`
+
+#### Integration
+
+Example to run `test/integration/01-setup/01-connection.spec.ts`
```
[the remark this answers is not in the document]
dolcalmi: yes, the sequencer use the filename without the folder :S

test/README.md
```diff
@@ -0,0 +1,109 @@
...
+$ make start-deps
+
+# or
+$ make reset-deps
+```
+Everytime the dependencies are re-started the environment must be reloaded via `direnv reload`. When using the [make command](../Makefile) this will happen automatically.
+
+## Known issues
+
+* **Test suite timeouts**: increase jest timeout value in [jest.setup.js](./jest.setup.js). Example:
```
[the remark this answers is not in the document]
dolcalmi: depends on the machine :( mine is slow, yours and our CI runners are faster so there is no need to increase it

test/README.md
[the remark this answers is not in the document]
bodymindarts: too bad

test/README.md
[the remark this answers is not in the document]
bodymindarts: then can we have it overridden by an env var? If someone wants a bigger timeout and needs to change code for that they might commit it by accident.

test/README.md
```diff
@@ -0,0 +1,109 @@
...
+  ```js
+  jest.setTimeout(120000) // 120 seconds
+  ```
+* **LNDs don't (re)start**: delete `*.macaroon` files from [dev folder](../dev/lnd) and re-create [runtime dependencies](#runtime-dependencies):
```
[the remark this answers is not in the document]
dolcalmi: really not sure exaclty why but sometimes after change the lnd conf (or maybe bitcoind.conf) and restart the runtime the containers start and stop again.. is not common but happened to me 2 times

test/integration/utils.spec.ts
dolcalmi: :) really I tried to [don't change](https://github.com/GaloyMoney/galoy/blame/633e205d3194dadca4360055d1a2256d7619adae/test/integration/09g-utils.spec.ts#L13) the internals of the tests but I will fix it

test/README.md
dolcalmi: will move test readme to dev.md (@bodymindarts suggestion), this section will be removed :)

test/integration/01-setup/01-metadata.spec.ts
bodymindarts: Yeah some kind of
```
make prep-mobile-tests
```
that will bring everything up ready to run the mobile tests or something like that.

dolcalmi: uhmmm I moved to debug folder (according to the conversation with Nicolas) https://github.com/GaloyMoney/galoy/pull/340/commits/3353ed95c63c0204258a34c7bd891032c3717074 
Also, not sure what involves prep environment for mobile testing but we can add it then

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -415,6 +415,7 @@ describe("UserWallet - Lightning Pay", () => {
         // https://github.com/alexbosworth/ln-service/issues/122
         await settleHodlInvoice({ lnd: lndOutside1, secret })
 
+        await sleep(5000)
```
bodymindarts, line 418: Is there a way we can poll when we can continue instead of sleep 5 seconds? This number seems arbitrary and is probably not optimal for everyone.

test/helpers/index.ts
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/340/commits/25087ca130b49e79a9a0cfe9ea94279b1164b68b now run with 0 but I had to add some sleeps to cancel hodl invoice

test/integration/02-user-wallet/02-send-lightning.spec.ts
bodymindarts: Can also leave this for a future pass.

dolcalmi: was in the original test, I tried to remove syncing nodes, waiting for pending balance and subscribing to update invoice event... seems to be the cancel event and the sync issues between lnds

this is the related issue https://github.com/GaloyMoney/galoy/pull/340/commits/1e36a02847c49efbad52ba0c8ba738e94ddef20b

test/unit/config.spec.ts
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/340/commits/ccdb806d4ae012f1c95fb0803b6120b1c3cc8509

test/README.md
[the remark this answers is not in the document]
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/340/commits/6405301fcf880eb2c62a326d60811a5db8a3f67e

DEV.md
```diff
@@ -45,36 +45,93 @@ $ make watch
...
+### Run specific test file
+
+To execute a specific test file:
+
+#### Unit
+
+Example to run `test/unit/config.spec.ts`
+
+```bash
+$ TEST=config yarn test:unit
```
bodymindarts, line 89: I think you've actually deleted that particular file right?

test/README.md
```diff
@@ -0,0 +1,109 @@
...
+  ```bash
+  $ make reset-deps
+  ```
+* **Integration tests running slow**: we use docker to run dependencies (redis, mongodb, bitcoind and 4 lnds) so the entire test suite is disk-intensive.
+  * Please make sure that you are running docker containers in a solid state drive (SSD)
+  * Add the next setting to [lnd.conf](../dev/lnd/lnd.conf) and re-create runtime dependencies
+    ```
+    # ./dev/lnd/lnd.conf
+    routing.assumechanvalid=true
```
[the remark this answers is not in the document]
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/340/commits/6405301fcf880eb2c62a326d60811a5db8a3f67e

DEV.md
dolcalmi: :man_facepalming:

test/integration/utils.spec.ts
dolcalmi: Fixed https://github.com/dolcalmi/galoy/commit/1e36a02847c49efbad52ba0c8ba738e94ddef20b and https://github.com/GaloyMoney/galoy/pull/340/commits/e8db0d20874b238ac2ce4173846763a605385ec3

DEV.md
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/340/commits/566b8414e428184b880baea8d47b622d71a857d7

dolcalmi:
@bodymindarts @nicolasburtey please remember that I am working in an external repo so CI is not running for this PR. 

before doing the merge could you run it?

nicolasburtey:
> before doing the merge could you run it?

probably because you pushed from your own repo, it's not possible test from the PR just from github interface. I guess it come to some settings on circleci related to security.

@bodymindarts do you have an easy solutions in mind?

test/integration/02-user-wallet/02-send-lightning.spec.ts
dolcalmi: let me try one last thing before giving up on this... (also maybe is not needed in your environment, only in mine)

dolcalmi: ok, a bit messy but I think is finally fixed https://github.com/GaloyMoney/galoy/pull/340/commits/7401863bcf37db1d293eeb5269d892afd353224a

dev/lnd/lnd.conf
```diff
@@ -26,3 +26,4 @@ noseedbackup=1
 bitcoind.rpchost=bitcoind:18443
 keysend-hold-time=2s
 tlsextradomain=lnd
+routing.assumechanvalid=true
```
nicolasburtey, line 29: maybe you can add a comment on why this is necessary?

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+beforeEach(async () => {
+  await waitUntilSyncAll()
+})
+
+describe("Bitcoind", () => {
+  it("should be funded mining 10 blocks", async () => {
+    const numOfBlock = 10
+    try {
+      const { name } = await bitcoindClient.createWallet(defaultWallet)
```
nicolasburtey, line 27: this could be its own test: "create default wallet"

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    } catch (error) {
+      baseLogger.warn({ error }, "bitcoind wallet already exists")
+    }
+
+    const bitcoindAddress = await bitcoindClient.getNewAddress()
+    await bitcoindClient.mineAndConfirm(numOfBlock, bitcoindAddress)
+    const balance = await bitcoindClient.getBalance()
+
+    // invalid only when there are problems deleting channels and
+    // they get confirmed while mineAndConfirm (101 blocks)
```
nicolasburtey, line 39: not sure to fully understand the tests/comments here

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    const { chain_balance: initialBalance } = await getChainBalance({ lnd: lnd1 })
+    const sats = initialBalance + btc2sat(amount)
+    await fundLnd(lnd1, amount)
+    const { chain_balance: balance } = await getChainBalance({ lnd: lnd1 })
+
+    expect(balance).toBe(sats)
+
+    const metadata = { type: "onchain_receipt", currency: "BTC", pending: "false" }
+
+    await MainBook.entry("funding tx")
```
nicolasburtey, line 63: this tweak is no longer necessary. we can just used getOnchainAddress from lnd and pay directly to that address

the reason I used this tweak before is because I experience where there could be 3 internal lnd, and the first 2 would only have lightning capability, not onchain capability. but this is no longer the case. not lnd1 has onchain capability so we don't need to explicitly add some funds in the books.

dev/lnd/lnd.conf
dolcalmi: uhmm is not necessary, that is why I had it in the readme (optional). This just improve performance in regtest (for slow environments).

https://github.com/GaloyMoney/galoy/commit/6405301fcf880eb2c62a326d60811a5db8a3f67e

test/integration/01-setup/01-lightning-channels.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+  channelLengthMain = (await getChannels({ lnd: lnd1 })).channels.length
+  channelLengthOutside1 = (await getChannels({ lnd: lndOutside1 })).channels.length
+})
+
+afterEach(async () => {
+  await checkIsBalanced()
+})
+
+// Setup the next network
+// lnd2 <- lnd1 <-> lndOutside1
```
nicolasburtey, line 44: is it intented the fact we have two arrow on this side: `lnd1 <-> lndOutside1`?
I would expect one node to open the channel. maybe it's because there are 2 channels (just trying to guess from the comment, not looking at the code yet)

test/integration/02-user-wallet/02-index.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+      expect(result).toBe(true)
+    })
+
+    it("return false if username does not exist", async () => {
+      const result = await usernameExists({ username: "user" })
+      expect(result).toBe(false)
+    })
+  })
+
+  describe("getOnchainFee", () => {
```
nicolasburtey, line 121: I would feel the getOnchainFee, getInvoice would benefits to be independant/in another file of the username stuff. one is purely mongodb related, the other involved bitcoind and lnd. what do you think?

nicolasburtey (commented):
still need to review more files but some more comments as I'm going through it

dev/lnd/lnd.conf
dolcalmi: ref: https://github.com/lightningnetwork/lnd/blob/6a97e6432d8d003752d4827a4a8bf05d16b279b6/rpcserver.go#L2523 let me know what to do... cc @bodymindarts

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
dolcalmi: uhmm if I remove the code checkIsBalanced fail... also, onchainTransactionEventHandler cant be used to balance the book because it needs an associated user

dev/lnd/lnd.conf
bodymindarts: I like adding a comment in the dev lnd.conf just to be clear that this is not a setting that should ever be used in prod incase someone copies the config. So a hint why it makes sense in dev but shouldn't be used anywhere else would be good.

bodymindarts:
> > before doing the merge could you run it?
> 
> probably because you pushed from your own repo, it's not possible test from the PR just from github interface. I guess it come to some settings on circleci related to security.
> 
> @bodymindarts do you have an easy solutions in mind?

I would:
- give commit access to regular contributors so they can push branches to main repo
- Make sure the essential functionality is tested in the context of GitHub actions so we can give safe feedback to everyone
- be ready to fix main branch quickly if a PR merge unexpectedly creates an issue
- if a merge did create an issue think about why GitHub actions didn't catch it

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
dolcalmi: didn't deleted after remove block reward removal, just removed https://github.com/GaloyMoney/galoy/pull/340/commits/7eba5e062a25234804eac0b372da682cff2a8ce2

test/integration/02-user-wallet/02-index.spec.ts
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/340/commits/e27cfbf08c2c0eb3d39671d82592efd6fec82653

test/integration/01-setup/01-lightning-channels.spec.ts
dolcalmi: I just added in that way because the setup open two channels.

[lnd -> lndOutside1](https://github.com/dolcalmi/galoy/blob/refactor_test_stg3a/test/integration/01-setup/01-lightning-channels.spec.ts#L63)
[lndOutside1 -> lnd](https://github.com/dolcalmi/galoy/blob/refactor_test_stg3a/test/integration/01-setup/01-lightning-channels.spec.ts#L89)

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/340/commits/7eba5e062a25234804eac0b372da682cff2a8ce2

test/helpers/index.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const checkIsBalanced = async () => {
+  await updateUsersPendingPayment()
+  // wait for balance updates because invoice event
+  // arrives before wallet balances updates in lnd
+  await waitUntilChannelBalanceSyncAll()
+
+  const { assetsLiabilitiesDifference, bookingVersusRealWorldAssets } =
+    await balanceSheetIsBalanced()
+  expect(assetsLiabilitiesDifference).toBeFalsy() // should be 0
```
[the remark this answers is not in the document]
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/340/commits/4fe99cc230de5849e6376d9d8c45b5112a3837f7

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
nicolasburtey: yes because here you're creating a new address directly by calling `getChainBalance` from lnd. if you were to create a new address using the UserWallet class, then the received funds would be associated with the User and there would be no more accounting gap

it's basically what we're doing on the receive onchain test:

```
describe("FunderWallet - On chain", () => {
  it("receives on-chain transaction", async () => {
    const funderWallet = await getFunderWallet({ logger: baseLogger })
    await sendToWallet({ walletDestination: funderWallet })
  })
})
```

dolcalmi: No, after lnd funding, the test fail in 01-funds-bitcoind-lnds.spec.ts

test/integration/01-setup/01-lightning-channels.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+// lnd2 <- lnd1 -> lndOutside2
+// this setup avoids close channels for routing fees tests
+describe("Lightning channels", () => {
```
nicolasburtey, line 47: don't you think this test could be rewritten?
we could just have an object defining the channel we want to have:

```
[{
initiator: lnd1,
partnre: lnd2,
basefee: 1,
...
},{
initiator: lnd1,
partnre: lnd2,
basefee: 1,
}]
```

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
dolcalmi: this was the previous test https://github.com/GaloyMoney/galoy/blob/1bcf8b11ff30030c786aa4a671bea69443de06f9/test/integration/02b-onchain-receive.spec.ts#L154 really I cant find the trigger that handle the payment for the first fund transaction to lnd (has not associated user)

nicolasburtey: oh ok the user are creating later now?

nicolasburtey: yup this is true that this remove the user dependancy. I have not thought about it. ok let's close this.

dolcalmi: is the first btc transaction to be able to create channels, really not sure how is balanced in production because I cant find where it handles the transactions not associated with users.

Please check https://github.com/GaloyMoney/galoy/blob/main/src/entrypoint/trigger.ts#L102 it just return when no user is associated

dolcalmi: if you like I can just receive onchain from funder user but... when you move onchain txs to bitcoind how will you fund/initialize the lnds?

nicolasburtey: > is the first btc transaction to be able to create channels, really not sure how is balanced in production because I cant find where it handles the transactions not associated with users.

we (typically) don't have transaction not associated with users. nor we should I believe. so this is more a test related stuff.

nicolasburtey: > if you like I can just receive onchain from funder user but... when you move onchain txs to bitcoind how will you fund/initialize the lnds?

not sure to get your question

dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/340/commits/86e070819b14741b3c5033c5156b2259b7358f23 but @jotapea please keep in mind

if on chain transactions are moved from lnds to bitcoind then there is no way to fund (and register in balance sheet) the funds sent to lnds (to open channels).

nicolasburtey: this is a good point... but we can funds directly a channel from a cold storage/bitcoind wallet. 

using the open function from [bos](https://github.com/alexbosworth/balanceofsatoshis ) as a command line tool now to do it but we could integrate this into our codebase. (bos#open is a wrapper above lnd api to make this possible)

bodymindarts:
Can we get this merged and address the rest of the issues post-merge? There are still lots of improvements to individual tests but this is already so much work that I think we should get it into main to prevent to much rebase pain.

Also I'm available this evening to see if it breaks circle ci.

bodymindarts (approved):

nicolasburtey (approved):

PR #344, opened 2021-07-19
(`SpecterWallet`) Inject config via constructor

vindard:
## Description

This PR refactors the `SpecterWallet` class to have the config defined and explicitly passed via its constructor instead of having config values called directly from global state (`yamlConfig`).

The global state is instead used to create the config that gets injected explicitly at `SpecterWallet` instantiation.

### DIFF
--- a/src/SpecterWallet.ts
+++ b/src/SpecterWallet.ts
@@ -18,20 +18,18 @@ export class SpecterWallet {
     this.config = config
   }
 
-  // below static method are {wallet} agnostics from bitcoin-core api perspective
-
-  static async listWallets() {
+  async listWallets() {
     return await bitcoindDefaultClient.listWallets()
   }
 
-  static async createWallet() {
+  async createWallet() {
     return await bitcoindDefaultClient.createWallet({
       wallet_name: "specter/coldstorage",
     })
   }
 
   async setBitcoindClient(): Promise<string> {
-    const wallets = await SpecterWallet.listWallets()
+    const wallets = await this.listWallets()
 
     const pattern = this.config.onchainWallet
     const specterWallets = _.filter(wallets, (item) => item.includes(pattern))
--- a/src/SpecterWallet.ts
+++ b/src/SpecterWallet.ts
@@ -1,6 +1,5 @@
 import { createChainAddress, sendToChainAddress } from "lightning"
 import _ from "lodash"
-import { yamlConfig } from "./config"
 import { bitcoindAccountingPath, lndAccountingPath, lndFeePath } from "./ledger/ledger"
 import { getActiveOnchainLnd, lndsBalances } from "./lndUtils"
 import { MainBook } from "./mongodb"
@@ -101,7 +100,7 @@ export class SpecterWallet {
     }
 
     const { total, onChain } = await lndsBalances()
-    const { action, sats, reason } = SpecterWallet.isRebalanceNeeded({
+    const { action, sats, reason } = this.isRebalanceNeeded({
       lndBalance: total,
       onChain,
     })
@@ -127,12 +126,12 @@ export class SpecterWallet {
     }
   }
 
-  static isRebalanceNeeded({ lndBalance, onChain }) {
+  isRebalanceNeeded({ lndBalance, onChain }) {
     // base number to calculate the different thresholds below
-    const lndHoldingBase = yamlConfig.rebalancing.lndHoldingBase
+    const lndHoldingBase = this.config.lndHoldingBase
 
-    const ratioTargetDeposit = yamlConfig.rebalancing.ratioTargetDeposit
-    const ratioTargetWithdraw = yamlConfig.rebalancing.ratioTargetWithdraw
+    const ratioTargetDeposit = this.config.ratioTargetDeposit
+    const ratioTargetWithdraw = this.config.ratioTargetWithdraw
 
     // threshold for when we need to move money from cold storage to the lnd wallet
     const thresholdLowBound = (lndHoldingBase * 70) / 100
@@ -153,7 +152,7 @@ export class SpecterWallet {
       let action: string | undefined = "deposit"
       let reason: string | undefined
 
-      const minOnchain = yamlConfig.rebalancing.minOnchain
+      const minOnchain = this.config.minOnchain
 
       if (onChain - sats < minOnchain) {
         action = undefined
--- a/src/SpecterWallet.ts
+++ b/src/SpecterWallet.ts
@@ -4,7 +4,7 @@ import { bitcoindAccountingPath, lndAccountingPath, lndFeePath } from "./ledger/
 import { getActiveOnchainLnd, lndsBalances } from "./lndUtils"
 import { MainBook } from "./mongodb"
 import { getOnChainTransactions } from "./OnChain"
-import { Logger, SpecterWalletConfig } from "./types"
+import { Logger, SpecterWalletConfig, SpecterWalletConstructorArgs } from "./types"
 import { UserWallet } from "./userWallet"
 import { BitcoindClient, bitcoindDefaultClient, btc2sat, sat2btc } from "./utils"
 
@@ -13,7 +13,7 @@ export class SpecterWallet {
   readonly logger: Logger
   readonly config: SpecterWalletConfig
 
-  constructor({ logger, config }) {
+  constructor({ logger, config }: SpecterWalletConstructorArgs) {
     this.logger = logger.child({ topic: "bitcoind" })
     this.config = config
   }
--- a/src/SpecterWallet.ts
+++ b/src/SpecterWallet.ts
@@ -187,7 +187,7 @@ export class SpecterWallet {
     let id
 
     try {
-      ; ({ id } = await sendToChainAddress({
+      ;({ id } = await sendToChainAddress({
         address,
         lnd,
         tokens: sats,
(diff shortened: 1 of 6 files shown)


src/SpecterWallet.ts
```diff
@@ -1,19 +1,21 @@
...
 import { BitcoindClient, bitcoindDefaultClient, btc2sat, sat2btc } from "./utils"
 
 export class SpecterWallet {
   bitcoindClient
-  logger
+  readonly logger: Logger
+  readonly config: SpecterWalletConfig
 
-  constructor({ logger }) {
+  constructor({ logger, config }) {
```
vindard, line 16: I wasn't sure if there was any way to include an 'ConstructorArgs' type here like you all did [at this line](https://github.com/GaloyMoney/galoy/pull/321/files#diff-0c83656ad06f75fb5b1c410eda5125a6fa79a5669ae0a211e174f76209c79813R14) before

test/integration/08-specter.spec.ts
```diff
@@ -3,17 +3,27 @@
...
 import { MainBook, setupMongoConnection } from "src/mongodb"
+import { SpecterWalletConfig } from "src/types"
 import { SpecterWallet } from "src/SpecterWallet"
 import { UserWallet } from "src/userWallet"
 import { bitcoindDefaultClient, sleep } from "src/utils"
 import { checkIsBalanced, mockGetExchangeBalance, RANDOM_ADDRESS } from "./helper"
 
 let specterWallet
 
+const specterWalletConfig: SpecterWalletConfig = {
```
vindard, line 19: I explicitly made a new config here instead of importing the one made in the `src/entrypoint/cron.ts` file

.gitignore
```diff
@@ -35,3 +35,5 @@ charts/configs/bitcoind/*
 coverage
 lib
 artifacts
+
+.node-version
```
vindard, line 39: I hope this is ok. I use the `nodenv` package locally to manage my node virtual environments and this is an artefact from that.

vindard (commented):
Some small comments

src/SpecterWallet.ts
```diff
@@ -124,10 +126,10 @@ export class SpecterWallet {
 
   static isRebalanceNeeded({ lndBalance, onChain }) {
     // base number to calculate the different thresholds below
-    const lndHoldingBase = yamlConfig.rebalancing.lndHoldingBase
+    const lndHoldingBase = this.config.lndHoldingBase
```
vindard, line 129: I just noticed that this wouldn't work with constructor injection. Converted the PR back to draft until this is addressed.

src/SpecterWallet.ts
```diff
@@ -29,7 +31,7 @@ export class SpecterWallet {
   async setBitcoindClient(): Promise<string> {
     const wallets = await SpecterWallet.listWallets()
 
-    const pattern = yamlConfig.rebalancing.onchainWallet ?? "specter"
+    const pattern = this.config.onchainWallet ?? "specter"
```
bodymindarts, line 34: Default should be in config instantiation.

src/SpecterWallet.ts
vindard: From call with `bodymindarts`, we will ignore this for now since there is other work that might address this. For this **static** method we will stick with global configs

src/SpecterWallet.ts
vindard: Based on call with `bodymindarts` I decided to skip this for now since the args here are finite and already typed.

bodymindarts (approved):

src/config.ts
```diff
@@ -36,3 +37,11 @@ export class TransactionLimits {
 
   oldEnoughForWithdrawalLimit = () => this.config.oldEnoughForWithdrawal / MS_IN_HOUR
 }
+
+export const getSpecterWalletConfig = (config): SpecterWalletConfig => ({
```
dolcalmi, line 41: @bodymindarts really not sure if we must pass here "config" (yaml), if we have this function in the same file really not sure about needing 2 references. We can remove this method and do a cast (if this is just a parsing method) or just use this method (much better) without the param or at least as optional (if have no value we can load the default configuration file)

bodymindarts: You are right we don’t need the function argument. We can treat yamlConfig as in scope within this file unless there is a clear need somewhere. 

I don’t like the idea of a cast as we want to have a consistent way of loading defaults or flattening structure or handling other config related logic so we’ve decided a function is the way to go.

vindard: Good catch, I had noticed that `yamlConfig` was local as well but I was thinking to maintain the pattern we setup just above in `TransactionLimits` to pass the config explicitly.

I think you're right though that if we're moving all these things into the `config.ts` module then we can remove that redundancy. For now, since nothing needs a different config I think I'll remove the parameter entirely, and if we need it for anything later on we can also add it back as an optional param which shouldn't affect any existing code.

dolcalmi:
@vindard in regards to isRebalanceNeeded static method, I think we have two options, pass the config or make it non-static (I can't see other place where it is neeed)

vindard:
> @vindard in regards to isRebalanceNeeded static method, I think we have two options, pass the config or make it non-static (I can't see other place where it is neeed)

@bodymindarts and I had chatted about that. He said that those static methods may be in flux in other branches so we should leave one for now until the other branches land.

bodymindarts:
> @vindard in regards to isRebalanceNeeded static method, I think we have two options, pass the config or make it non-static (I can't see other place where it is neeed)

@dolcalmi is right. Its only needed in 1 place and not where I thought I saw it before. Probably correct to make it an instance method in this case. When I searched the code base on our zoom call I confused it with the `FtxDealerWallet.isRebalanceNeeded` method which has the same name.

vindard:
> @dolcalmi is right. Its only needed in 1 place and not where I thought I saw it before. Probably correct to make it an instance method in this case. When I searched the code base on our zoom call I confused it with the `FtxDealerWallet.isRebalanceNeeded` method which has the same name.

Ok yea I just checked through again and I think it could be that we were looking at an older version of the codebase before I refactored this branch to the latest tip.

Based on a read-through of that `SpecterWallet` class I agree that making it an instance method makes sense, and there are two other static methods in there that we should probably tackle as well since we're in here already.

@nicolasburtey left [this comment](https://github.com/GaloyMoney/galoy/commit/c827b1b45ae739c146a87574d87ccb68ab4b17be#diff-692142bf2211c206c719066858e18ecf677ec75238587e7da7e101713bb88aa2R21) about those static methods in there when they were first made. I wouldn't mind if he weighed in to make sure there aren't any unforeseen gotchas from converting these methods to instance methods. Specifically we're looking at:
- `listWallets()`
- `createWallet()`
- `isRebalanceNeeded(...)`

src/SpecterWallet.ts
krtk6160: We can just do this -> `constructor({ logger, config }: { logger: Logger; config: SpecterWalletConfig })`
Or am I missing something?

vindard: Ah yes, this is what I was missing thanks! I'm still finding my way with Typescript.

I can include it in another PR where I'm thinking to tighten up the typing for some constructor functions as well.

dolcalmi (approved):

src/SpecterWallet.ts
```diff
@@ -29,7 +32,7 @@ export class SpecterWallet {
   async setBitcoindClient(): Promise<string> {
     const wallets = await SpecterWallet.listWallets()
 
-    const pattern = yamlConfig.rebalancing.onchainWallet ?? "specter"
+    const pattern = this.config.onchainWallet
```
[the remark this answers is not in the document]
vindard: Interesting point.

I know @bodymindarts wanted to still keep some of these concerns separate though. Could another approach be to pass global configs directly into the `SpecterWallet` constructor and then automatically call `getSpecterWalletConfig` inside the constructor?

.gitignore
krtk6160: I don't think we want to allow this. I remember @bodymindarts saying stuff like this should be part of the [global](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) gitignore.

vindard: Yup, it's since been removed

src/SpecterWallet.ts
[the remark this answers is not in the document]
vindard: @dolcalmi would this sort of change work? https://github.com/GaloyMoney/galoy/pull/344/commits/7187ea636f271014d7c87ce6e02e2009f84ee57d

I also changed the `??` check to a tighter `||` check since empty strings are no longer valid.

vindard:
Adding [this reference](https://github.com/GaloyMoney/galoy/pull/354/files#diff-692142bf2211c206c719066858e18ecf677ec75238587e7da7e101713bb88aa2R12) to where we make non-empty strings no longer valid `onChainWallet` values for the https://github.com/GaloyMoney/galoy/pull/344/commits/7187ea636f271014d7c87ce6e02e2009f84ee57d changes here.

dolcalmi (approved):

vindard:
[Rebased](https://github.com/GaloyMoney/galoy/compare/7187ea636f271014d7c87ce6e02e2009f84ee57d..849e54cd6494e780a41ec297c606cf3d0a4c152b) to new tip of main to resolve merge conflict.

Re-linking [this comment](https://github.com/GaloyMoney/galoy/commit/7187ea636f271014d7c87ce6e02e2009f84ee57d#r53802421) dropped from the PR.

dolcalmi (approved):

bodymindarts (approved):

PR #345, opened 2021-07-19
Start an admin-server serving a separate GraphQL schema

samerbuna:
- 3 top-level queries in the new schema
- New core schema tests (not mounted yet)


charts/galoy/values.yaml
```diff
@@ -16,6 +16,15 @@ image:
   repository: gcr.io/galoyapp/galoy
 
 deployment:
+- name: graphql-admin
+  args: ["lib/entrypoint/graphql-admin-server.js"]
+  healthz: true
+  targetPort: 4001
+  replicaCount: 1
+  podAnnotations:
+    prometheus.io/path: /swagger/metrics
```
nicolasburtey, line 25: I think we want to get rig of swagger. we already have a lot of data from apollo studio.

nicolasburtey: and apollo studio data is a lot more relevant to us than what swagger is providing

src/entrypoint/graphql-admin-server.ts
```diff
@@ -0,0 +1,187 @@
...
+const pino_http = PinoHttp({
+  logger: graphqlLogger,
+  wrapSerializers: false,
+
+  // Define custom serializers
+  serializers: {
+    err: pino.stdSerializers.err,
+    req: pino.stdSerializers.req,
+    res: (res) => ({
+      // FIXME: kind of a hack. body should be in in req. but have not being able to do it.
```
nicolasburtey, line 38: we may want to fix this workaround before it's being in too many places in the codebase.
this issue arise before we were using apollo server. We were using another server before (don't recall the name right now) and it's possible now with apollo server this is not an issue any more

src/entrypoint/graphql-admin-server.ts
```diff
@@ -0,0 +1,187 @@
...
+    schema: schemaWithPermissions,
+    playground: process.env.NETWORK !== "mainnet",
+    introspection: process.env.NETWORK !== "mainnet",
+    context: async (context) => {
+      // @ts-expect-error: TODO
+      const token = context.req?.token ?? null
+      const uid = token?.uid ?? null
+      const ip = context.req?.headers["x-real-ip"]
+
+      if (isIPBlacklisted({ ip })) {
```
nicolasburtey, line 91: probably the admin backend should be an "allowlist" as opposed to a "blocklist"?

and as I'm thinking through this, we probably want to use a firewall for this (or a Cloudflare of something like that) instead of trying to do it on our code directly.

src/entrypoint/graphql-admin-server.ts
```diff
@@ -0,0 +1,187 @@
...
+      // currentTime: and(isAuthenticated, isEditor),
+    },
+  },
+  { allowExternalErrors: true },
+)
+
+export async function startApolloServerForAdminSchema() {
+  const app = express()
+
+  const schemaWithPermissions = applyMiddleware(gqlAdminSchema, permissions)
```
nicolasburtey, line 79: beyond the `gqlAdminSchema` parameter, anything else seems to be a copy of the main graphql. 

I'm wondering if you should not just create a function instead that is being used by both the main graphql and the admin graphql to avoid the duplication?

src/AdminOps.ts
```diff
@@ -31,7 +29,7 @@ export const addToMap = async ({
     )
   }
 
-  const user = await User.findByUsername({ username })
+  const user = await User.getUser({ username })
```
nicolasburtey, line 32: we have other way to get the user. by phone number for instance. does this name makes sense as a result?

nicolasburtey (commented):
Didn't go through all the code but have some questions/comments on the code I had looked at.

src/entrypoint/graphql-admin-server.ts
samerbuna: Not sure about this one. I can't determine if it's still used.

src/graphql/core/schema.graphql
```diff
@@ -0,0 +1,9 @@
+type Query {
```
nicolasburtey, line 1: I guess this file is just a placeholder for now?

nicolasburtey (commented):
2 questions pending otherwise this PR looks good to me

nicolasburtey:
tests are not passing. not sure if it needs a rebase?

nicolasburtey:
is the merge from main to this branch intentional?

image

IIUC @bodymindarts was saying this is an anti-pattern

src/entrypoint/graphql-admin-server.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  const schema = applyMiddleware(gqlAdminSchema, permissions)
+  return startApolloServer({ schema, port: 4001 })
+}
+
+setupMongoConnection()
+  .then(async () => {
+    await startApolloServerForAdminSchema()
+  })
+  .catch((err) => graphqlLogger.error(err, "server error"))
```
dolcalmi, line 39: This need to be migrated to the same way used in https://github.com/GaloyMoney/galoy/pull/365 and with the use of node_env validation

src/entrypoint/graphql-core-server.ts
```diff
@@ -0,0 +1,286 @@
...
+        },
+        { username: 1, title: 1, coordinate: 1 },
+      )
+
+      return users.map((user) => ({
+        ...user._doc,
+        id: user.username,
+      }))
+    },
+    usernameExists: async (_, { username }) => usernameExists({ username }),
```
dolcalmi, line 137: possible lint issue with async/await

src/graphql/core/queries.ts
```diff
@@ -0,0 +1,17 @@
+import { GT } from "../index"
+
+import Date from "../types/scalars/date"
+
+const QueryType = new GT.Object({
+  name: "Query",
+  fields: () => ({
+    currentTime: {
```
dolcalmi, line 8: why to create a mutation and a query for current time?

src/graphql/db/index.ts
```diff
@@ -0,0 +1,42 @@
+import { User } from "../../schema"
```
dolcalmi, line 1: I'm not sure about the place of this file, if it's just a first step for the repository refactor that's fine, but I'm not aware if anything else has been defined.
cc @bodymindarts

src/schema.ts
```diff
@@ -360,17 +364,17 @@ UserSchema.statics.getUser = async function ({ username, phone }) {
...
 
-UserSchema.statics.getUserByAddress = async function ({ address }) {
-  return this.findOne({ "onchain.address": address })
-}
-
-// FIXME: Merge findByUsername and getUser
-UserSchema.statics.findByUsername = async function ({ username }) {
-  if (typeof username !== "string" || !username.match(regexUsername)) {
-    return null
+UserSchema.statics.findOrCreate = async function (userInfo) {
```
dolcalmi, line 367: are you sure this not change the current behavior of the app? the tests does not include much about this

test/integration/entrypoint/graphql/admin/mutations.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    if (!data) {
+      throw Error("invalid response")
+    }
+
+    const updatedUser = await User.getUser({ username: "tester" })
+
+    expect(updatedUser.title).toEqual("MapTest")
+    expect(updatedUser.coordinate).toEqual({ longitude: 1, latitude: -1 })
+    expect(data.merchantUpdateMapInfo.userDetails.id).toEqual(updatedUser.id)
```
dolcalmi, line 103: you dont need the if (line 95) just move line 103 to line 93 and destructure the result directly in data

test/integration/entrypoint/graphql/admin/queries.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+        }
+      }
+    `
+
+    const result = await graphql(gqlAdminSchema, query, {})
+    const { data } = result
+
+    if (!data) {
+      throw Error("invalid response")
+    }
```
dolcalmi, line 42: this pattern is not needed, please remove from all tests

dolcalmi (changes requested):

dolcalmi:
@samerbuna @nicolasburtey  Will the port change require something in the fronted?

src/graphql/core/queries.ts
samerbuna: This is just a placeholder starting point or the new core schema

src/graphql/db/index.ts
samerbuna: Yeah, we're gonna need a big refactor here to unify all db calls in one place.

src/schema.ts
samerbuna: We changed all the places that were using this

samerbuna:
> is the merge from main to this branch intentional?
> 
> image
> 
> IIUC @bodymindarts was saying this is an anti-pattern

Rebased.

src/schema.ts
samerbuna: Found a few places where this was in issue. Solved in most recent commit.

samerbuna:
> @samerbuna @nicolasburtey Will the port change require something in the frontend?

Yes, eventually, we'll need to change the admin-panel to use this new graphql endpoint, but the existing endpoint will continue to work for now. No changes are needed in the mobile repo (yet)

src/schema.ts
bodymindarts: Hmm I consider having a `findOrCreate` method somewhat of an anti-pattern. Typically the code that would call this should have enough context to know which use-case is being handled. Also as far as I can tell this method is only called from `test` code and having a function in the production code that is only for testing purposes should also be avoided.

Can we move this to be a test helper?

src/graphql/db/index.ts
bodymindarts: AFAICT this file defines top level functions that get called from mutations.
I like the pattern of how the user is accessed and updated.
I don't like that this file is in the scoped under `graphql/db`. If GraphQL is a layer for interacting with an application then these functions represent the application entrypoints.

This file should be something like:
```
src/app/user.ts
```
or
```
src/usecases/user.ts
```
Actually for the GQL redesign I think it would be great to introduce this kind of structure in general. Functions that get called as app-entrypoints scoped in an expressive folder structure. Should be used both for queries and mutations. Calling the database directly from a graphql resolver should be avoided.

src/graphql/db/index.ts
```diff
@@ -0,0 +1,43 @@
...
+import { baseLogger } from "../../logger"
+import { caseInsensitiveRegex } from "src/utils"
+
+const logger = baseLogger.child({ module: "admin" })
+
+export async function usernameExists({ username }): Promise<boolean> {
+  return Boolean(await User.findOne({ username: caseInsensitiveRegex(username) }))
+}
+
+export const updateUserLevel = async ({ uid, level }) => {
```
bodymindarts, line 12: The existing function has arg validation [here](https://github.com/GaloyMoney/galoy/blob/main/src/AdminOps.ts#L13):
```
  if (Levels.indexOf(level) === -1) {
    const error = `${level} is not a valid user level`
    throw new ValidationError(error, { forwardToClient: true, logger, level: "warn" })
  }
```

bodymindarts:
There are now 4 `*graphql*` files under `entrypoints`. I would have expected there is no need for more than 1:
```
├── graphql-admin-server.ts
├── graphql-core-server.ts
├── graphql-server.ts
├── graphql.ts
```

samerbuna:
> There are now 4 `*graphql*` files under `entrypoints`. I would have expected there is no need for more than 1:
> 
> ```
> ├── graphql-admin-server.ts
> ├── graphql-core-server.ts
> ├── graphql-server.ts
> ├── graphql.ts
> ```

I removed one of them in the latest commit. Nicolas and I reached a conclusion that the admin server should run in its own process rather than be included in the main process. That's why there are 2 entry files. The third file has the common code between the 2 entry ones.

src/graphql/db/index.ts
samerbuna: - I removed `findOrCreate` and used the code directly in the tests for now. I don't like that I had to use that method but the tests were giving me too many troubles otherwise. I'll pair with Juan on making this better moving forward.

- I moved User db methods under `src/db/user.ts` or now. I look at this db directory as the entry point for all the db processes. Let's talk about this a bit more in the next code review.

- The levels check, I left that off as the only place where this function is used is in the GraphQL mutation (and the validation is done on that level). I'm not sure if we have any database-level validation (here and elsewhere) but If we don't, let's start talking about that and how to match these validations in the codebase where needed.

bodymindarts (dismissed):

src/entrypoint/graphql-core-server.ts
```diff
@@ -199,8 +171,8 @@ const resolvers = {
     updateUser: (_, __, { wallet }) => ({
       setUsername: async ({ username }) => await wallet.setUsername({ username }),
       setLanguage: async ({ language }) => await wallet.setLanguage({ language }),
-      updateUsername: async (input) => await wallet.updateUsername(input),
-      updateLanguage: async (input) => await wallet.updateLanguage(input),
+      updateUsername: (input) => wallet.updateUsername(input),
```
krtk6160, line 174: Why are we using a mixed pattern of using await for some functions but not for others. For example line #172 uses `await` but this line has it removed.

bodymindarts (approved):

krtk6160 (approved):

src/utils.ts
```diff
@@ -9,7 +9,7 @@ import axios from "axios"
 import { yamlConfig } from "./config"
 import { baseLogger } from "./logger"
 
-export const isProd = process.env.NODE_ENV === "production"
+export const isDev = process.env.NODE_ENV !== "production"
```
bodymindarts, line 12: This burned us on rollout. We should be more narrow in the definition of `isDev`.
```
process.env.NODE_ENV === "development"
```
perhaps?
Don't want this executing on staging.

samerbuna: I guess I was using that as `!isProd` but I agree that `isDev` and `isTest` are not ideal. How about we introduce some command options like `--disable-helmet`, `--disable-IP-updates`, `--mock-XYZ` with all of them altering the default behavior of production?

PR #349, opened 2021-07-20
Add hedging strategy

sebastienverreault:
Refactor ccxt as an abstract exchange and specialized ftx and okex exchanges


src/dealer/DealerWallet.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    // Testing for 0 is tricky, assuming we wont hedge change
+    if (Math.abs(usdLiability) < 1) {
+      logger.debug({ usdLiability }, "No liabilities to hedge.")
+      return
+    }
+
+    logger.debug("starting with order loop")
+
+    const updatedPosition = await this.strategy.UpdatePosition(
```
nicolasburtey, line 37: is `UpdatePosition` is static method? why the Uppercase otherwise?

src/dealer/DealerWallet.ts
```diff
@@ -0,0 +1,151 @@
...
+
+const activeStrategy = HedgingStrategies.FtxPerpetualSwap
+// const activeStrategy = HedgingStrategies.OkexPerpetualSwap
+
+export class DealerWallet extends OnChainMixin(UserWallet) {
+  strategy: HedgingStrategy
+
+  constructor({ user, logger }: WalletConstructorArgs) {
+    super({ user, logger })
+    this.strategy = createHedgingStrategy(activeStrategy, logger)
```
nicolasburtey, line 19: when passing several arguments to a function, I would typically recommend using an object. ie: 
`createHedgingStrategy({activeStrategy, logger})`. the main advantage is that you don't have to remember the order of the inputs if you use an object

src/entrypoint/dealerEntrypoint.ts
```diff
@@ -0,0 +1,32 @@
+import { setupMongoConnection } from "../mongodb"
```
nicolasburtey, line 1: you're in the entrypoint folder, so `dealer.ts` for the filename is probably good, no need to repeat entry point.

Issue #353, opened 2021-07-20
Replace default bitcoind wallet with multiple purpose specific wallets


jotapea:
The current bitcoind tests implementation use a single "all encompassing" default wallet client which is used for many purposes. The goal is to split this one client/wallet into:

- default client:
  - Only to be used for non-transaction wallet functionality (like: getBlockchainInfo, createWallet, *listWallets*, decodeRawTransaction, loadWallet, unloadWallet, *getBlockCount*?).
    - All code that uses the listWallets call needs to be updated to now consider the addition of new wallets.
    - A single getBlockCount function has been kept for the default wallet ([TODO?](https://github.com/GaloyMoney/galoy/pull/354/files#diff-3787bb1dbeb8d5d5ca497caee5ad3726d7ff527bdc33c92557eee5161e5da8d3R49)).

- outside wallet:
  - Represents the outside world, which is required for simulating transactions in tests.
  - This is replacing all tests from the original "" (unnamed) wallet that do transactions (generate, receive or send coins).

- hot wallet:
  - FUTURE use, will eventually replace the lnd onchain wallet.
  - The scope of this one at this phase is only to be added as another wallet in the tests.

jotapea:
@nicolasburtey PR https://github.com/GaloyMoney/galoy/pull/354 is ready.

PR #354, opened 2021-07-20
Split default wallet

jotapea:
PR for https://github.com/GaloyMoney/galoy/issues/353.

It should be ready to get merged, but some adjustments could still be done.

### DIFF
--- a/src/SpecterWallet.ts
+++ b/src/SpecterWallet.ts
@@ -1,7 +1,7 @@
 import assert from "assert"
 import { createChainAddress, sendToChainAddress } from "lightning"
 import _ from "lodash"
-import { bitcoindDefaultClient, BitcoindClientWallet } from "./bitcoind"
+import { bitcoindDefaultClient, BitcoindWalletClient } from "./bitcoind"
 import {
   bankOwnerMediciPath,
   bitcoindAccountingPath,
@@ -63,7 +63,7 @@ export class SpecterWallet {
 
     this.logger.info({ wallet: specterWallets[0] }, "setting BitcoindClient")
 
-    this.bitcoindClient = new BitcoindClientWallet(specterWallets[0])
+    this.bitcoindClient = new BitcoindWalletClient(specterWallets[0])
 
     return specterWallets[0]
   }
--- a/src/SpecterWallet.ts
+++ b/src/SpecterWallet.ts
@@ -26,6 +26,7 @@ export class SpecterWallet {
   }
 
   async listWallets() {
+    // TODO: include default / outside?
     return await bitcoindDefaultClient.listWallets()
   }
 
--- a/src/SpecterWallet.ts
+++ b/src/SpecterWallet.ts
@@ -28,7 +28,6 @@ export class SpecterWallet {
   }
 
   async listWallets() {
-    // TODO: include default / outside?
     return await bitcoindDefaultClient.listWallets()
   }
 
--- a/src/SpecterWallet.ts
+++ b/src/SpecterWallet.ts
@@ -13,7 +13,7 @@ import { Logger, SpecterWalletConfig, SpecterWalletConstructorArgs } from "./typ
 import { UserWallet } from "./userWallet"
 import { BitcoindWalletClient, bitcoindDefaultClient, btc2sat, sat2btc } from "./utils"
 
-// TODO no longer used? (removing the creation of the default wallet didn't break anything)
+// TODO no longer used in tests, removing the creation of the default wallet didn't break anything
 const staticClient = "default" // was ""
 
 export class SpecterWallet {
--- a/src/SpecterWallet.ts
+++ b/src/SpecterWallet.ts
@@ -63,7 +63,7 @@ export class SpecterWallet {
 
     this.logger.info({ wallet: specterWallets[0] }, "setting BitcoindClient")
 
-    this.bitcoindClient = new BitcoindWalletClient(specterWallets[0])
+    this.bitcoindClient = new BitcoindWalletClient({ walletName: specterWallets[0] })
 
     return specterWallets[0]
   }
@@ -270,17 +270,17 @@ export class SpecterWallet {
       // TODO: won't work automatically with a cold storage wallet
       // make a PSBT instead accesible for further signing.
       // TODO: figure out a way to export the PSBT when there is a pending tx
-      txid = await this.bitcoindClient.sendToAddress(address, sat2btc(sats))
+      txid = await this.bitcoindClient.sendToAddress({ address, amount: sat2btc(sats) })
     } catch (err) {
       const error = "this.bitcoindClient.sendToAddress() error"
       subLogger.error({ txid }, error)
       throw new Error(err)
     }
 
-    const tx = await this.bitcoindClient.getTransaction(
+    const tx = await this.bitcoindClient.getTransaction({
       txid,
-      true /* include_watchonly */,
-    )
+      include_watchonly: true,
+    })
     const fee = btc2sat(-tx.fee) /* fee is negative */
 
     const bankOwnerPath = await bankOwnerMediciPath()
--- a/src/SpecterWallet.ts
+++ b/src/SpecterWallet.ts
@@ -1,6 +1,7 @@
 import assert from "assert"
 import { createChainAddress, sendToChainAddress } from "lightning"
 import _ from "lodash"
+import { bitcoindDefaultClient, BitcoindClientWallet } from "./bitcoind"
 import {
   bankOwnerMediciPath,
   bitcoindAccountingPath,
@@ -11,13 +12,13 @@ import { MainBook } from "./mongodb"
 import { getOnChainTransactions } from "./OnChain"
 import { Logger, SpecterWalletConfig, SpecterWalletConstructorArgs } from "./types"
 import { UserWallet } from "./userWallet"
-import { BitcoindWalletClient, bitcoindDefaultClient, btc2sat, sat2btc } from "./utils"
+import { btc2sat, sat2btc } from "./utils"
 
 // TODO no longer used in tests, removing the creation of the default wallet didn't break anything
 const staticClient = "default" // was ""
 
 export class SpecterWallet {
-  bitcoindClient
+  bitcoindClient // TODO rename?
   readonly logger: Logger
   readonly config: SpecterWalletConfig
 
@@ -62,7 +63,7 @@ export class SpecterWallet {
 
     this.logger.info({ wallet: specterWallets[0] }, "setting BitcoindClient")
 
-    this.bitcoindClient = BitcoindWalletClient({ wallet: specterWallets[0] })
+    this.bitcoindClient = new BitcoindClientWallet(specterWallets[0])
 
     return specterWallets[0]
   }
--- a/src/SpecterWallet.ts
+++ b/src/SpecterWallet.ts
@@ -13,6 +13,7 @@ import { Logger, SpecterWalletConfig, SpecterWalletConstructorArgs } from "./typ
 import { UserWallet } from "./userWallet"
 import { BitcoindWalletClient, bitcoindDefaultClient, btc2sat, sat2btc } from "./utils"
 
+// TODO no longer used? (removing the creation of the default wallet didn't break anything)
 const staticClient = "default" // was ""
 
 export class SpecterWallet {
--- a/src/SpecterWallet.ts
+++ b/src/SpecterWallet.ts
@@ -13,6 +13,9 @@ import { Logger, SpecterWalletConfig, SpecterWalletConstructorArgs } from "./typ
 import { UserWallet } from "./userWallet"
 import { BitcoindClient, bitcoindDefaultClient, btc2sat, sat2btc } from "./utils"
 
+// TODO: Centralizing this special handling. Does not seem to be the "outside" here...
+const specialWalletName = "default" // was ""
+
 export class SpecterWallet {
   bitcoindClient
   readonly logger: Logger
@@ -51,7 +54,7 @@ export class SpecterWallet {
       this.logger.info("specter wallet has not been instantiated")
 
       // currently use for testing purpose. need to refactor
-      return ""
+      return specialWalletName
     }
 
     if (specterWallets.length > 1) {
@@ -87,7 +90,7 @@ export class SpecterWallet {
   async getBitcoindBalance(): Promise<number> {
     if (!this.bitcoindClient) {
       const wallet = await this.setBitcoindClient()
-      if (wallet === "") {
+      if (wallet === specialWalletName) {
         return 0
       }
     }
@@ -100,7 +103,7 @@ export class SpecterWallet {
 
     if (!this.bitcoindClient) {
       const wallet = await this.setBitcoindClient()
-      if (wallet === "") {
+      if (wallet === specialWalletName) {
         return
       }
     }
@@ -179,7 +182,7 @@ export class SpecterWallet {
   async toColdStorage({ sats }) {
     if (!this.bitcoindClient) {
       const wallet = await this.setBitcoindClient()
-      if (wallet === "") {
+      if (wallet === specialWalletName) {
         this.logger.warn("no wallet has been setup")
         return
       }
@@ -233,7 +236,7 @@ export class SpecterWallet {
   async toLndWallet({ sats }) {
     if (!this.bitcoindClient) {
       const wallet = await this.setBitcoindClient()
-      if (wallet === "") {
+      if (wallet === specialWalletName) {
         return
       }
     }
--- a/src/SpecterWallet.ts
+++ b/src/SpecterWallet.ts
@@ -11,7 +11,7 @@ import { MainBook } from "./mongodb"
 import { getOnChainTransactions } from "./OnChain"
 import { Logger, SpecterWalletConfig, SpecterWalletConstructorArgs } from "./types"
 import { UserWallet } from "./userWallet"
-import { BitcoindClient, bitcoindDefaultClient, btc2sat, sat2btc } from "./utils"
+import { BitcoindWalletClient, bitcoindDefaultClient, btc2sat, sat2btc } from "./utils"
 
 const staticClient = "default" // was ""
 
@@ -61,7 +61,7 @@ export class SpecterWallet {
 
     this.logger.info({ wallet: specterWallets[0] }, "setting BitcoindClient")
 
-    this.bitcoindClient = BitcoindClient({ wallet: specterWallets[0] })
+    this.bitcoindClient = BitcoindWalletClient({ wallet: specterWallets[0] })
 
     return specterWallets[0]
   }
--- a/src/SpecterWallet.ts
+++ b/src/SpecterWallet.ts
@@ -13,8 +13,7 @@ import { Logger, SpecterWalletConfig, SpecterWalletConstructorArgs } from "./typ
 import { UserWallet } from "./userWallet"
 import { BitcoindClient, bitcoindDefaultClient, btc2sat, sat2btc } from "./utils"
 
-// TODO: Centralizing this special handling. Does not seem to be the "outside" here...
-const specialWalletName = "default" // was ""
+const staticClient = "default" // was ""
 
 export class SpecterWallet {
   bitcoindClient
@@ -54,7 +53,7 @@ export class SpecterWallet {
       this.logger.info("specter wallet has not been instantiated")
 
       // currently use for testing purpose. need to refactor
-      return specialWalletName
+      return staticClient
     }
 
     if (specterWallets.length > 1) {
@@ -90,7 +89,7 @@ export class SpecterWallet {
   async getBitcoindBalance(): Promise<number> {
     if (!this.bitcoindClient) {
       const wallet = await this.setBitcoindClient()
-      if (wallet === specialWalletName) {
+      if (wallet === staticClient) {
         return 0
       }
     }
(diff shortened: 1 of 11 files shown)


src/SpecterWallet.ts
```diff
@@ -8,6 +8,9 @@ import { getOnChainTransactions } from "./OnChain"
 import { UserWallet } from "./userWallet"
 import { BitcoindClient, bitcoindDefaultClient, btc2sat, sat2btc } from "./utils"
 
+// TODO: Centralizing this special handling. Does not seem to be the "outside" here...
+const specialWalletName = "default" // was ""
```
krtk6160, line 12: Why is it being called `specialWallet`? Perhaps we can find a better (more descriptive) name

src/utils.ts
```diff
@@ -38,7 +38,8 @@ const connection_obj = {
 
 export const BitcoindClient = ({ wallet = "" }) =>
   new bitcoindClient({ ...connection_obj, wallet })
-export const bitcoindDefaultClient = BitcoindClient({ wallet: "" })
+export const bitcoindDefaultClient = BitcoindClient({ wallet: "default" }) // coinless
+export const bitcoindOutsideWalletClient = BitcoindClient({ wallet: "outside" })
```
krtk6160, line 42: Minor nitpick, but we should either have `bitcoindDefaultWalletClient` and `bitcoindOutsideWalletClient` *or* `bitcoindDefaultClient` and `bitcoindOutsideClient`

krtk6160 (commented):
A couple of comments

src/SpecterWallet.ts
jotapea: Agree, that is a temporary name until I am more certain about how to call it. Also, maybe to reconsider how it is used.

src/utils.ts
jotapea: Ok for this one I wanted to be explicit about their difference. The name reflects that the default client should not be used for wallet-specific functionality (transactions). I added the "coinless" comment in there also.

nicolasburtey:
rebase needed

jotapea:
@nicolasburtey Rebase done.

jotapea:
@bodymindarts Rebased. The tests do pass in my machine. And notice the failure in the pipeline:

```
    Expected: 242530719
    Received: 242530718
```

Is only a 1 sat difference. @nicolasburtey Has this kind of issue happened before?

nicolasburtey:
yes things like this happen previously before, the main reasons was rounding error because lnd doesn't use sats but msats or millisats, but our accounting is still done in sats. so there are some inconsistency in it. 

the way we have been managing this is by saying that if the different is less than x sats (x was 5 before) then the test would pass. but @dolcalmi removed this workaround when he did the tests refactoring.

test/helpers/lightning.ts
```diff
@@ -18,9 +18,9 @@ import {
   updateRoutingFees,
 } from "lightning"
 
-import { bitcoindClient, RANDOM_ADDRESS } from "./bitcoinCore"
+import { bitcoindClient, bitcoindOutside, RANDOM_ADDRESS } from "./bitcoinCore"
 
-export * from "lightning"
+export * from "lightning" // TODO?
```
nicolasburtey, line 23: why adding TODO?

jotapea: Right that was something I wanted to ask, but then after listening to @dolcalmi I am more clear about its purpose. Will remove the comment.

src/utils.ts
jotapea: This will be addressed in a future PR more focused on tests, specifically because of the renames that also happen [here](https://github.com/GaloyMoney/galoy/blob/main/test/helpers/bitcoinCore.ts#L4).

test/integration/specter-wallet.spec.ts
```diff
@@ -33,7 +36,9 @@ describe("SpecterWallet", () => {
     }
 
     wallets = await SpecterWallet.listWallets()
-    expect(wallets.length).toBe(2)
+    expect(wallets.length).toBe(4)
+    // expect(wallets.length).toBe(3)
```
nicolasburtey, line 40: we should clean this up

test/integration/specter-wallet.spec.ts
```diff
@@ -21,9 +21,12 @@ afterAll(async () => {
 
 describe("SpecterWallet", () => {
   it("creates wallet", async () => {
-    let wallets = await SpecterWallet.listWallets()
+    let wallets = await SpecterWallet.listWallets() // now ["default", "hot", "outside"]
 
-    if (wallets.length < 2) {
+    // TODO maybe this should be explicitly checking wallet names...
+    if (wallets.length < 4) {
+      // if (wallets.length < 3) {
```
nicolasburtey, line 28: we should clean this up

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
```diff
@@ -14,26 +15,47 @@ import {
...
+      expect(name).toBe("default")
+    } catch (error) {
+      baseLogger.warn({ error }, "bitcoind wallet already exists")
+    }
+    const wallets = await bitcoindClient.listWallets()
+    expect(wallets).toContain("default")
+  })
+
+  // TODO: only to be considered in the rest of the tests
+  it("create hot wallet", async () => {
```
nicolasburtey, line 32: this should be refactored, this is the same code as the `default` wallet above

nicolasburtey: and outside also use the same code

bodymindarts:
The other day we discussed we all like the separation of concerns into `thing that interacts with bitcoind wallet` and `thing that gets information from bitcoind.
But I don't like his way of implementing it.... currently the same class is being used:
```
export const bitcoindDefaultClient = BitcoindClient({ wallet: "default" }) // coinless
export const bitcoindOutsideWalletClient = BitcoindClient({ wallet: "outside" })
```
even though the use-cases and usages of these 2 objects is completely different. I find it too implicit.

Can we move to something like this:

```
export const bitcoindDefaultClient = BitcoindClient()
export const bitcoindWallet = BitcoindWalletClient({ wallet: "outside" })
```
So as to have a clearer and _explicit_ separation?

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
jotapea: Refactor done with createWalletInClient function. And also added an initial test that checks that no wallet is created yet to remove the original try-catch for the error "bitcoind wallet already exists".

jotapea:
@bodymindarts I agree with the separation of concerns. But as mentioned [here](https://github.com/GaloyMoney/galoy/pull/354/files#r675000502), I think we should leave this change to another PR (I will do) that includes changes in [bitcoinCore](https://github.com/GaloyMoney/galoy/pull/354/files#diff-da35e6d51cd3616efe2845c65af00ef79989fc7869a7fdf2e1eec396f349bffb) which also deals with these names.

nicolasburtey (approved):

dolcalmi:
@jotapea did you test it running again `make integration` after first pass? I am asking you because only `lnd-utils.spec.ts` is not idempotent and we need to try to get the others test working without reset docker containers.

When you create additional wallets in bitcoind you can have problems after load them so please check and load or unload them as required like in https://github.com/GaloyMoney/galoy/blob/main/test/integration/specter-wallet.spec.ts#L21

bodymindarts:
I think there has been a misunderstanding in how to split the 2 clients.

`src/utils.ts:41`
```
export const BitcoindClient = () => new bitcoindClient({ ...connection_obj })

export const BitcoindWalletClient = ({ wallet }) =>
  new bitcoindClient({ ...connection_obj, wallet })

// The default client should remain without a wallet (not generate or receive bitcoin)
export const bitcoindDefaultClient = BitcoindClient()
```

You have now created 2 differently named _constructors_ and are consistently naming the variables to reflect wether we are dealing with the wallet client or not.
But I was under the impression from the last discussion we wanted to use 2 different _types_ that explicitly expose only the interfaces needed for the 2 different clients.
The idea here is to leverage the static guarantees that typescript gives us when using the type-system in an explicit way. With this solution a typo (or bad auto-complete) can lead to bugs because the underlying type of the 2 objects - even though consistently named - is identical. We have no static, compile-time mechanism that enforces the segregation of the intended uses - which would guarantee to avoid a whole class of bugs that could come from an accidental confusion of the 2 clients.

jotapea:
@dolcalmi Adding the loading and unloading of wallets per test works fine in the first pass. But when `make integration` is run again, the `resetLnds()` call from test `01-connection` fails. This is just the first test, so some other issues can come up in further ones. And this branch has been focusing on the bitcoind clients, not lnd necessarily...

I think that the scope of this branch is enough as is, and it may be better to leave making the tests idempotent for another branch. Also considering that other branches are blocked waiting for this to get merged.

src/bitcoind.ts
```diff
@@ -1,5 +1,146 @@
...
+
+type VOut = {
+  value: number
+  n: number
+  scriptPubKey: ScriptPubKey
+}
+
+type GetAddressInfoResult = {
+  address: string
+  // TODO
```
jotapea, line 30: Which ones are needed ([reference](https://developer.bitcoin.org/reference/rpc/getaddressinfo.html))?

```
{                                   (json object)
  "address" : "str",                (string) The bitcoin address validated.
  "scriptPubKey" : "hex",           (string) The hex-encoded scriptPubKey generated by the address.
  "ismine" : true|false,            (boolean) If the address is yours.
  "iswatchonly" : true|false,       (boolean) If the address is watchonly.
  "solvable" : true|false,          (boolean) If we know how to spend coins sent to this address, ignoring the possible lack of private keys.
  "desc" : "str",                   (string, optional) A descriptor for spending coins sent to this address (only when solvable).
  "isscript" : true|false,          (boolean) If the key is a script.
  "ischange" : true|false,          (boolean) If the address was used for change output.
  "iswitness" : true|false,         (boolean) If the address is a witness address.
  "witness_version" : n,            (numeric, optional) The version number of the witness program.
  "witness_program" : "hex",        (string, optional) The hex value of the witness program.
  "script" : "str",                 (string, optional) The output script type. Only if isscript is true and the redeemscript is known. Possible
                                    types: nonstandard, pubkey, pubkeyhash, scripthash, multisig, nulldata, witness_v0_keyhash,
                                    witness_v0_scripthash, witness_unknown.
  "hex" : "hex",                    (string, optional) The redeemscript for the p2sh address.
  "pubkeys" : [                     (json array, optional) Array of pubkeys associated with the known redeemscript (only if script is multisig).
    "str",                          (string)
    ...
  ],
  "sigsrequired" : n,               (numeric, optional) The number of signatures required to spend multisig output (only if script is multisig).
  "pubkey" : "hex",                 (string, optional) The hex value of the raw public key for single-key addresses (possibly embedded in P2SH or P2WSH).
  "embedded" : {                    (json object, optional) Information about the address embedded in P2SH or P2WSH, if relevant and known.
    ...                             Includes all getaddressinfo output fields for the embedded address, excluding metadata (timestamp, hdkeypath, hdseedid)
                                    and relation to the wallet (ismine, iswatchonly).
  },
  "iscompressed" : true|false,      (boolean, optional) If the pubkey is compressed.
  "timestamp" : xxx,                (numeric, optional) The creation time of the key, if available, expressed in UNIX epoch time.
  "hdkeypath" : "str",              (string, optional) The HD keypath, if the key is HD and available.
  "hdseedid" : "hex",               (string, optional) The Hash160 of the HD seed.
  "hdmasterfingerprint" : "hex",    (string, optional) The fingerprint of the master key.
  "labels" : [                      (json array) Array of labels associated with the address. Currently limited to one label but returned
                                    as an array to keep the API stable if multiple labels are enabled in the future.
    "str",                          (string) Label name (defaults to "").
    ...
  ]
}
```

src/bitcoind.ts
```diff
@@ -1,5 +1,146 @@
...
+}
+
+type InWalletTransaction = {
+  amount: number
+  fee: number
+  // TODO
```
jotapea, line 36: And here also, which ones are needed ([reference](https://developer.bitcoin.org/reference/rpc/gettransaction.html))?

```
{                                          (json object)
  "amount" : n,                            (numeric) The amount in BTC
  "fee" : n,                               (numeric) The amount of the fee in BTC. This is negative and only available for the
                                           'send' category of transactions.
  "confirmations" : n,                     (numeric) The number of confirmations for the transaction. Negative confirmations means the
                                           transaction conflicted that many blocks ago.
  "generated" : true|false,                (boolean) Only present if transaction only input is a coinbase one.
  "trusted" : true|false,                  (boolean) Only present if we consider transaction to be trusted and so safe to spend from.
  "blockhash" : "hex",                     (string) The block hash containing the transaction.
  "blockheight" : n,                       (numeric) The block height containing the transaction.
  "blockindex" : n,                        (numeric) The index of the transaction in the block that includes it.
  "blocktime" : xxx,                       (numeric) The block time expressed in UNIX epoch time.
  "txid" : "hex",                          (string) The transaction id.
  "walletconflicts" : [                    (json array) Conflicting transaction ids.
    "hex",                                 (string) The transaction id.
    ...
  ],
  "time" : xxx,                            (numeric) The transaction time expressed in UNIX epoch time.
  "timereceived" : xxx,                    (numeric) The time received expressed in UNIX epoch time.
  "comment" : "str",                       (string) If a comment is associated with the transaction, only present if not empty.
  "bip125-replaceable" : "str",            (string) ("yes|no|unknown") Whether this transaction could be replaced due to BIP125 (replace-by-fee);
                                           may be unknown for unconfirmed transactions not in the mempool
  "details" : [                            (json array)
    {                                      (json object)
      "involvesWatchonly" : true|false,    (boolean) Only returns true if imported addresses were involved in transaction.
      "address" : "str",                   (string) The bitcoin address involved in the transaction.
      "category" : "str",                  (string) The transaction category.
                                           "send"                  Transactions sent.
                                           "receive"               Non-coinbase transactions received.
                                           "generate"              Coinbase transactions received with more than 100 confirmations.
                                           "immature"              Coinbase transactions received with 100 or fewer confirmations.
                                           "orphan"                Orphaned coinbase transactions received.
      "amount" : n,                        (numeric) The amount in BTC
      "label" : "str",                     (string) A comment for the address/transaction, if any
      "vout" : n,                          (numeric) the vout value
      "fee" : n,                           (numeric) The amount of the fee in BTC. This is negative and only available for the
                                           'send' category of transactions.
      "abandoned" : true|false             (boolean) 'true' if the transaction has been abandoned (inputs are respendable). Only available for the
                                           'send' category of transactions.
    },
    ...
  ],
  "hex" : "hex",                           (string) Raw data for transaction
  "decoded" : {                            (json object) Optional, the decoded transaction (only present when `verbose` is passed)
    ...                                    Equivalent to the RPC decoderawtransaction method, or the RPC getrawtransaction method when `verbose` is passed.
  }
}
```

src/bitcoind.ts
```diff
@@ -1,5 +1,146 @@
...
+  txid: string
+}
+
+// TODO consistency in params
```
jotapea, line 40: In these newly added classes, some methods use a json object as input while others use direct parameters. Should it be consistent? Which one should we choose?

jotapea:
@dolcalmi Pushed the loading and unloading of wallets, which also allowed to improve another test.

src/bitcoind.ts
```diff
@@ -1,5 +1,146 @@
+import Client from "bitcoin-core"
 import _ from "lodash"
-import { BitcoindClient, bitcoindDefaultClient, btc2sat } from "./utils"
+import { btc2sat } from "./utils"
+
+const connection_obj = {
```
bodymindarts, line 5: this should be constructed in the `config.ts` module and injected. To move forward we could leave it like this and perhaps @vindard could take care of it as part of his ongoing config refactoring.

src/bitcoind.ts
bodymindarts: I think we're converging on `json` as input typed via `XxxConstructorArgs`.

src/bitcoind.ts
```diff
@@ -1,5 +1,146 @@
...
+  amount: number
+  fee: number
+  // TODO
+  txid: string
+}
+
+// TODO consistency in params
+
+export class BitcoindClientNoWallet {
+  client
```
bodymindarts, line 43: This should be `readonly` or?

src/bitcoind.ts
```diff
@@ -1,5 +1,146 @@
...
+type InWalletTransaction = {
+  amount: number
+  fee: number
+  // TODO
+  txid: string
+}
+
+// TODO consistency in params
+
+export class BitcoindClientNoWallet {
```
bodymindarts, line 42: How about:
```
BitcoindClient
```
and 
```
BitcoindWalletClient
```
?

src/bitcoind.ts
```diff
@@ -1,5 +1,146 @@
...
+
+  async getBlockCount(): Promise<number> {
+    return await this.client.getBlockCount()
+  }
+
+  async getBlockchainInfo(): Promise<{ chain: string }> {
+    return await this.client.getBlockchainInfo()
+  }
+
+  async createWallet({
```
bodymindarts, line 57: Since this is the `NoWallet` client we probably shouldn't have a method to `createWallet` here right?

src/bitcoind.ts
```diff
@@ -1,5 +1,146 @@
...
+    wallet_name: string
+  }): Promise<{ name: string; warning: string }> {
+    return await this.client.createWallet({ wallet_name })
+  }
+
+  async listWallets(): Promise<[string]> {
+    return await this.client.listWallets()
+  }
+
+  async loadWallet(filename: string): Promise<{ name: string; warning: string }> {
```
bodymindarts, line 69: Shouldn't the `xxxWallet` related functions be part of the `BitcoindWalletClient`?

src/bitcoind.ts
```diff
@@ -9,12 +150,12 @@ export const getBalancesDetail = async (): Promise<
   const balances: { wallet: string; balance: number }[] = []
 
   for await (const wallet of wallets) {
-    // do not use the default wallet for now (expect for testing).
-    if (wallet === "") {
+    // do not consider the "outside" wallet in tests
+    if (wallet === "outside") {
```
bodymindarts, line 154: Can we remove this conditional that that is only related to testing? Would be better to have no test-related code int he production code.

test/helpers/lightning.ts
```diff
@@ -211,7 +217,7 @@ export const mineBlockAndSync = async ({
   blockHeight?: number
   newBlock?: number
 }) => {
-  await bitcoindClient.generateToAddress(newBlock, RANDOM_ADDRESS)
+  await bitcoindOutside.generateToAddress(newBlock, RANDOM_ADDRESS)
```
bodymindarts, line 220: can we find a better name than `Outside`? Perhaps `nonGaloyBitcoind` or `bitcoindOwnedByThirdParty` or something like that?

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
```diff
@@ -9,31 +10,42 @@ import {
...
-    } catch (error) {
-      baseLogger.warn({ error }, "bitcoind wallet already exists")
-    }
+  it("check no wallet", async () => {
     const wallets = await bitcoindClient.listWallets()
-    expect(wallets).toContain("")
+    expect(wallets.length).toBe(0)
+  })
+
+  it("create outside wallet", async () => {
```
bodymindarts, line 32: This doesn't have an `expect` so it should probably be in a `before` hook or somehow part of test setup code.

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
bodymindarts: this `it` also isn't a test (no `expect`) so needs to be part of setup

src/bitcoind.ts
jotapea: This bypass is in there to be able to test the getBalance function, which is used to check the balances after each test. Maybe it can be abstracted into a "tests bypass" function, but having only a single wallet to bypass it seems ok to me for now.

test/helpers/lightning.ts
jotapea: We also use the outside word for lnd. Is the same context correct?

src/bitcoind.ts
jotapea: I understand what you mean, but I think is more like is the default client's responsibility to know about all the wallets in it.

jotapea:
@bodymindarts most feedback was addressed, except that I think using a `XxxConstructorArgs` for the constructor of BitcoindWalletClient maybe is too much for just a single argument. And also considering we already have multiple galoy-wallet constructor args, which could add confusion.

test/helpers/lightning.ts
```diff
@@ -128,10 +133,11 @@ export const openChannelTesting = async ({
   return { lndNewChannel, lndPartnerNewChannel }
 }
 
+// all the following uses of bitcoind client that send/receive coin must be "outside"
+
 export const fundLnd = async (lnd, amount = 1) => {
   const { address } = await createChainAddress({ format: "p2wpkh", lnd })
-
-  await bitcoindClient.sendToAddressAndConfirm(address, amount)
+  await sendToAddressAndConfirm(bitcoindOutside, address, amount)
```
nicolasburtey, line 140: object based input?

nicolasburtey (dismissed):

test/helpers/lightning.ts
jotapea: Done

bodymindarts (dismissed):
I'm approving it but see 2 follow up tasks:
- remove all references to 'outside' wallet in production code. We should find another way to deal with testing than to have special cases like that.
- put the config objects creation into the `config` module

Can be taken care of in future PRs.

bodymindarts (approved):

nicolasburtey (approved):

PR #355, opened 2021-07-20
Move wait and checkIsBalanced to funds lnd1 test

dolcalmi:


test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
```diff
@@ -53,6 +48,7 @@ describe("Bitcoind", () => {
   })
 
   it("funds lnd1 node", async () => {
+    await waitUntilSyncAll()
```
krtk6160, line 51: Can't we have just `waitUntilSync({lnds: [lnd1]})` here? I guess we can ignore whether the other lnds are synced or not because that's not relevant here. What do you think?

krtk6160: Also I think this should be moved further down to #L63. I don't understand why we want it to be at the beginning of this test. Am I missing something?

dolcalmi: just remove it, and trust in waitUntilBlockHeight and waitUntilSync({ lnds }) of channel setup

PR #357, opened 2021-07-20
removing half developed key send

nicolasburtey:
will be reimplemented later with AMP


src/lndUtils.ts
```diff
@@ -469,17 +457,15 @@ export const validate = async ({
...
     username = params.username
+  }
 
-    const preimage = randomBytes(preimageByteLength)
-    id = createHash("sha256").update(preimage).digest().toString("hex")
-    const secret = preimage.toString("hex")
-    messages = [{ type: keySendPreimageType, value: secret }]
+  if (!!params.amount && !!tokens) {
+    const error = `Invoice contains non-zero amount, but amount was also passed separately`
+    // FIXME: create a new error. this is a not a graphl error.
```
krtk6160, line 465: We already have a `ValidationError` in the `error.ts`. It looks like the wrong one was imported here in this file.

src/lndUtils.ts
```diff
@@ -444,52 +439,38 @@ export const validate = async ({
...
+    // throw new ValidationInternalError(error, {logger})
```
bodymindarts, line 462: This comment is duplicate of the line below??

bodymindarts (approved):
LGTM I'm guessing most of the formatting and adding type information were in order to get rid of linting errors right?

src/lndUtils.ts
nicolasburtey: no the condition are different. it's basically an xor we are doing here, which an explicit error on whether we have too much info or too little

PR #365, opened 2021-07-21
Enforce awaiting on async functions (and lint the test folder)

samerbuna:
- [x] fix all await calls in src/
- [x] get eslint to pass (after including the test folder)


dolcalmi:
About this PR:
@samerbuna found a random fail in the integration test which led us to find a bug related to async/await use. He enabled eslint for test folder and for that rule

PS: We will move "module mapping" to another PR

dolcalmi (approved):

bodymindarts (approved):

PR #366, opened 2021-07-21
Captcha api protection

jotapea:
Adding a captcha check to the requestPhoneCode endpoint. It is currently based on Google's reCaptcha, but the provider might change.


jotapea:
@jrbuhl93 @nicolasburtey

src/schema.graphql
```diff
@@ -253,7 +253,10 @@ enum AccountStatus {
 }
 
 type Mutation {
-  requestPhoneCode(phone: String @pattern(regexp: "^\\+[0-9]+$", policy: THROW)): Success
+  requestPhoneCode(
+    phone: String @pattern(regexp: "^\\+[0-9]+$", policy: THROW)
+    captchaResponse: String @stringLength(min: 10, max: 1000, policy: THROW) # length can vary
```
nicolasburtey, line 258: the response is not a fix-length token?

jotapea: Yes based on experience with recaptcha, but also because we can choose different providers with different lengths.

src/geetest/geetest.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  async register() {
+    const gtLib = new GeetestLib(this.GEETEST_ID, this.GEETEST_KEY)
+    const digestmod = "md5"
+    const userId = "test"
+    const params = {
+      digestmod: digestmod,
+      user_id: userId,
+      client_type: "web",
```
jrbuhl93, line 47: The client type should be `"h5"` not `"web"` based on [this comment](https://github.com/GeeTeam/gt3-server-node-express-sdk/blob/80787f928563c6a5c1d401839bd7dadbb983be24/app.js#L28).

src/geetest/geetest.js
```diff
@@ -0,0 +1,74 @@
...
+      result = await gtLib.register(digestmod, params)
+    } else {
+      result = await gtLib.localRegister()
+    }
+    return JSON.parse(result.data)
+  }
+
+  async validate(challenge, validate, seccode) {
+    const gtLib = new GeetestLib(this.GEETEST_ID, this.GEETEST_KEY)
+    const bypasscache = await this._getBypassStatus() // not a cache
```
jrbuhl93, line 62: Do we want to cache this?

jrbuhl93 (commented):
We need to change all the Chinese comments to English.

~~I believe we also want to add the captcha route to login? Or is it not necessary if requestPhoneCode is protected?~~

src/schema.graphql
```diff
@@ -253,6 +261,13 @@ enum AccountStatus {
 }
 
 type Mutation {
+  requestPhoneCodeGeetest(
```
jotapea, line 264: Are we keeping this approach of a separate mutation, or should the captcha parameters be added to the original requestPhoneCode?

src/text.ts
```diff
@@ -12,11 +12,188 @@ import {
...
+
+  // TODO
+  const captchaRequired = true
+
+  // TODO? any bypass?
+
+  // TODO? ip checks?
+
+  // TODO? any new limiter for the captcha?
+
```
jotapea, line 38: What should be done from the above TODOs?

src/text.ts
```diff
@@ -12,11 +12,188 @@ import {
...
+  logger.info({ phone, ip }, "RequestPhoneCodeGeetest called")
+
+  // const challenge = req.body[GeetestLib.GEETEST_CHALLENGE];
+  // const validate = req.body[GeetestLib.GEETEST_VALIDATE];
+  // const seccode = req.body[GeetestLib.GEETEST_SECCODE];
+
+  // TODO
+  const captchaRequired = captchaChallenge && captchaValidate && captchaSeccode
+
+  // TODO before or after ip?
```
jotapea, line 83: Captcha before or after the ip check?

jotapea:
> We need to change all the Chinese comments to English.

Maybe only focus on some specific comments to change to english? I think is better to try to make as little modifications as possible to this external library (i.e. if it needs to be replaced because it was updated).

dolcalmi (changes requested):

src/geetest/sdk/geetest_lib.js
```diff
@@ -0,0 +1,260 @@
+const cryptoRandomString = require("crypto-random-string") // current dependency
```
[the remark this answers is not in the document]
jotapea: Got it working with @samerbuna's guidance, but thanks!

jotapea:
@dolcalmi The responsibilities point you bring is a good one. I see it in a similar way, the ip and captcha checks purpose is to protect the api, not doing galoy functionality. But maybe this is the right place for these checks? And then the calls to galoy logic be abstracted elsewhere?...

yarn.lock
```diff
@@ -3575,6 +3575,14 @@ astral-regex@^2.0.0:
   resolved "https://registry.yarnpkg.com/astral-regex/-/astral-regex-2.0.0.tgz#483143c567aeed4785759c0865786dc77d7d2e31"
   integrity sha512-Z7tMw1ytTXt5jqMcOP+OQteU1VuNK9Y02uuJtKQ1Sv69jXQKKg5cibLwGJow8yzZP+eAc18EmLGPal0bp36rvQ==
```
jotapea, line 3577: Are we ok with all these added dependencies?

src/servers/graphql-main-server.ts
```diff
@@ -164,8 +169,24 @@ const resolvers = {
...
+      _,
+      { phone, geetestChallenge, geetestValidate, geetestSeccode },
+      { logger, ip },
+    ) => ({
+      success: await requestPhoneCodeGeetest({
+        phone,
+        geetestChallenge,
+        geetestValidate,
+        geetestSeccode,
+        logger,
```
jotapea, line 186: @jrbuhl93 Notice the rename of the parameters, now these start with geetest instead of captcha.

src/graphql/old-schema.graphql
```diff
@@ -253,6 +261,13 @@ enum AccountStatus {
 }
 
 type Mutation {
+  requestPhoneCodeGeetest(
+    phone: String @pattern(regexp: "^\\+[0-9]+$", policy: THROW)
+    geetestChallenge: String @stringLength(min: 10, max: 1000, policy: THROW) # TODO
+    geetestValidate: String @stringLength(min: 10, max: 1000, policy: THROW)
+    geetestSeccode: String @stringLength(min: 10, max: 1000, policy: THROW)
+  ): Success
```
jotapea, line 269: @jrbuhl93 Do you know where the lengths of these is documented? I haven't found it...

src/geetest.js
```diff
@@ -0,0 +1,75 @@
...
+  constructor(GEETEST_ID, GEETEST_KEY) {
+    this.GEETEST_ID = GEETEST_ID
+    this.GEETEST_KEY = GEETEST_KEY
+  }
+
+  // private
+  async _getBypassStatus() {
+    const bypass_status = await sendRequest({ gt: this.GEETEST_ID })
+    return bypass_status
+  }
```
jotapea, line 38: We are currently using the recommended server implementation that includes a [ByPass mode](https://docs.geetest.com/captcha/deploy/bypass). The [gt4](https://github.com/GeeTeam/gt4_node_express_demo) demo does not include this, and is not well documented yet. And it seems to be incompatible with [gt3](https://docs.geetest.com/captcha/deploy/server/node#Resources), because the backend only makes a [single request](https://github.com/GeeTeam/gt4_node_express_demo/blob/main/routes/index.js#L51) GeeTest.

jrbuhl93: If it isn't compatible with gt3 I think it's a no-go for now since I do not see mobile client sdks for gt4.

src/graphql/old-schema.graphql
jrbuhl93: I haven't found it specified either. Do you think this validation is necessary?

jotapea: Yeah we should have some limits, and in this case keep them in a big range just to be safe until we learn their real lengths.

src/geetest.js
jotapea: Something that is left to consider here is if we want to replicate the caching of this value like in the [demo](https://github.com/GeeTeam/gt3-server-node-express-bypass/blob/master/app.js#L12). I don't think is necessary to start, but letting you know.

default.yaml
```diff
@@ -72,6 +72,10 @@ test_accounts:
 
 sms_provider: "twilio"
 
+geetest:
```
jrbuhl93, line 75: I think these should be in [this file](https://github.com/GaloyMoney/galoy/blob/main/charts/galoy/templates/deployment.yaml). @nicolasburtey

yarn.lock
jrbuhl93: We might be able to remove some of these dependencies we don't use. For instance, I don't think this PR uses redis for anything like the example we based the code off of did.

src/graphql/types/mutations/auth-code-request.ts
```diff
@@ -0,0 +1,54 @@
...
+    const {
+      phone,
+      challengeCode: geetestChallenge,
+      validationCode: geetestValidate,
+      secCode: geetestSeccode,
+    } = args.input
+
+    try {
+      // TODO: redo this with use-case and errors pattern
+      await VerifyCaptchaAndReturnOTP({
```
dolcalmi, line 32: maybe this should be `verifyCaptchaAndReturnOTP`

src/services/geetest.ts
```diff
@@ -0,0 +1,67 @@
...
+  let bypass_res
+  try {
+    const res = await axios({
+      url: request_url,
+      method: "GET",
+      timeout: 5000,
+      params: params,
+    })
+    const resBody = res.status === 200 ? res.data : ""
+    console.log(resBody)
```
dolcalmi, line 20: please remove

dolcalmi (changes requested):

dolcalmi (dismissed):

default.yaml
nicolasburtey: yes in addition to what we have here

src/services/geetest.ts
```diff
@@ -0,0 +1,65 @@
+// Based on https://github.com/GeeTeam/gt3-server-node-express-bypass/blob/master/app.js
+
+// https://docs.geetest.com/captcha/apirefer/api/server
+// doing this: "If the storage space is not sufficient: Send request to check bypass status before starting the verfication process."
+
+import axios from "axios"
+import GeetestLib from "gt3-server-node-express-sdk/sdk/geetest_lib" // galoy fork
+
+async function sendRequest(params) {
+  const requestUrl = "http://bypass.geetest.com/v1/bypass_status.php"
```
nicolasburtey, line 10: not https?

yarn.lock
nicolasburtey: agree

nicolasburtey (commented):
need to map environment variable on kubernetes as Justin pointed out before we can merge

samerbuna:
Converted to a local branch, now #458

PR #370, opened 2021-07-22
Refactor 'UserWallet' types/config placement

vindard:
### Description

This is a cleanup PR from some of the work introduced in #321 to bring that work more in line with patterns introduced in #344. Specifically this PR:
- moves the `UserWalletConfig` into a getter function in the `config` module (similar to `SpecterWalletConfig`)

- creates a new `ITransactionLimits` interface to help remove a circular dependency between the `config.ts` and `types.ts` files. This helps to start establishing a **"`types.ts` -> `config.ts` -> `other modules`"** (DAG-like) hierarchy.

### DIFF
--- a/src/walletFactory.ts
+++ b/src/walletFactory.ts
@@ -18,7 +18,6 @@ export const WalletFactory = async ({
   UserWallet.setCurrentPrice(lastPrice)
 
   const transactionLimits = new TransactionLimits({
-    config: yamlConfig.limits,
     level: user.level,
   })
 
--- a/src/walletFactory.ts
+++ b/src/walletFactory.ts
@@ -1,9 +1,9 @@
-import { TransactionLimits, yamlConfig } from "./config"
+import { getUserWalletConfig, yamlConfig } from "./config"
 import { NotFoundError } from "./error"
 import { LightningUserWallet } from "./LightningUserWallet"
 import { getCurrentPrice } from "./realtimePrice"
 import { User } from "./schema"
-import { Logger, UserWalletConfig } from "./types"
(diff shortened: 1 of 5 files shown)


bodymindarts (dismissed):

src/walletFactory.ts
```diff
@@ -1,10 +1,10 @@
-import { TransactionLimits, yamlConfig } from "./config"
+import { getUserWalletConfig, yamlConfig } from "./config"
```
dolcalmi, line 1: yamlConfig use for funder is in the scope of this PR?

krtk6160:
I'm wondering if we should allow passing config as a param, but use yamlConfig as the default value. When #321 was written, @bodymindarts mentioned that if we allow passing config as a param, we would then be able to write unit tests for them without having to change yamlConfig within the tests' scopes.

bodymindarts:
Ah yes I forgot the unit test argument. 

But that still applies since the unit test can just create another instance of this type regardless. No need to reuse the function in this file.

Either way I think we can go with this for now for simplicity and change it later when we actually start using it in unit tests.

bodymindarts:
Yeah I’d consider that out of scope for this

src/walletFactory.ts
bodymindarts: Yeah I’d consider that out of scope for this. Arvin is breaking the latter config refactoring into small steps.

dolcalmi (dismissed):

bodymindarts (approved):

dolcalmi (approved):

Issue #350, opened 2021-07-20
receive USD over lightning (client side)


nicolasburtey:
Currently we can only receive BTC over lightning. 

For the sake of clarification, when mentioning receiving USD over lightning, this means BTC is still the medium of exchange, but USD is the unit of account. 

To be able to do this currency conversion, there needs to be a dynamic hedging on an exchange, which is [work in progress](https://github.com/GaloyMoney/galoy/pull/349). 

This ticket here is relative to the client side: how can a user through our API can received USD instead of BTC.

There are multiples scenarios each with its own specifics in regards to price risk:
- amountless invoice / sendtoUsername
- invoice with amount
- onchain transaction

**amountless invoice / sendtoUsername**
Probably the easiest is the "amountless invoice" or "sendtoUsername" use case. 

It is the easiest is because there is no need to give a "free option" (in the finance sense, ie: give the payer the choice to pay this invoice in the next XX seconds/minutes, knowing that the price could change in the meantime) to the payer. rather the BTC/USD price to be used is determined when the funds are being received.

A small spread should be applied to cover the exchange fees

**invoice with amount**
If the payee is settings an amount, then we should:
- have a higher spread because the payer then can decide to pay or not pay depending of how the price move during the time the invoice is valid
- shorten the validity of the invoice from 1 day to something like 1 minute

eventually the spread should be dynamic depending on the market condition. if this is a very volatile market, we should have a higher spread than if this is a calm market. for now we can probably start with a simple number. 

**onchain transaction**

We can probably do the onchain transaction at a later stage. For onchain transaction, we can also have an amount in the QR code or not have an amount, so the pattern is similar to lightning invoice that can have an amount or be amountless. 

The main different with Onchain is the time to settlement. It can takes days/weeks for a transaction to arrive, so generally I think it's better to apply the price at the time the transaction get mined with X confirmations.

**Other**
[Price](https://github.com/GaloyMoney/price) is what is being used currently to do the USD/BTC conversion, this is a median of 3 exchanges. 

But if we're the counter part of the trade, maybe instead of using a median of 3 exchanges, we should rely on the single price of the exchange we'll be using to do the hedging?

dolcalmi:
1st Stage - Ledger Refactor
- This is a minor refactor and the purpose is to have only one point of access to the ledger
- Some ledger actions are organized like ledger/transaction.ts but we need to unify the others
- MainBook (medici) instance must be only accessed inside ledger
[...]

PR #376, opened 2021-07-24
fix root issue void pending

nicolasburtey:


dolcalmi (approved):

src/Lightning.ts
```diff
@@ -720,17 +715,31 @@ export const LightningMixin = (superclass) =>
...
           if (result.is_confirmed || result.is_failed) {
-            payment.pending = false
-            await payment.save()
+            const resultPendingTrue = await Transaction.updateMany(
+              { _journal: payment._journal, pending: true },
+              { $set: { pending: false } },
+            )
+
+            if (resultPendingTrue.nModified === 0) {
+              // this could happen if dealer and user try to update transaction at the same time
```
bodymindarts, line 729: Does this mean there was a race condition between the dealer code and the 'main' code? Was it because the dealer transactions weren't lock protected?

nicolasburtey: what was not protected is if both the dealer (which is a user) and the user were both trying to update the transaction at the same time.

but the root issue here that we have is that we were doing a `updateOne` (in the form of `payment.save()`) instead of `updateMany`. so we were only updating some entries for the transaction.

dolcalmi: @bodymindarts when we add a payment to the user we create an entry in medici that basically creates at least 2 Transactions (credit and debit) with the same metadata but for different accounts.

So the problem was fixed by `updateMany` because when we update the transactions owned by the user we also update the transaction for lightning assets account (lndAccountingPath)

Also, we went through the code to check if it was happening elsewhere, but everything is fine.

With the refactoring in the first stage of [usd invoices issue](https://github.com/GaloyMoney/galoy/issues/350#issuecomment-885293504) we will be able to tackle these kinds of things easily

bodymindarts:
I think its great you have found the root cause of this issue but I would have 2 suggestions:
- when bug fixing (especially ones that are subtle and have been around a long time) could we  initially add a failing test that documents the bug? This way we will catch if we accidentally introduce a regression - which can easily happen when it is subtle.
- for important bug fixes like this perhaps we could add a few lines to the commit message to add context. I have nothing against 1-liner commit messages in general but I think this is a prime example of a commit that would merit some more information. For example a brief summary the information @nicolasburtey and @dolcalmi have provided as response to me so that another reviewer can understand the `why` better when looking at the diff. I find the additional information that @vindard  puts in his commit messages to be very useful when reviewing (like [here](https://github.com/GaloyMoney/galoy/pull/370) for example)

nicolasburtey:
you're right. time to make commit message more comprehensive!

PR #377, opened 2021-07-24
improve accounting structure

nicolasburtey:
**make liabilities more compact in the db**

the liabilities will only exist for a particular account. so the intermediary "Account" in "Liabilities:Account:uid" is superflous, and make the index bigger/slower. 

**some definition**

for this issue, I'll use wallet as: a set of transactions and a balance that are represented within midici with `Liabilities:${uid}`
and an account: the representation of 1 or n wallets.

**refactor the revenue/fees of the bank account**

first read: first section "A Bank’s Balance Sheet" in 
https://courses.lumenlearning.com/wm-macroeconomics/chapter/banking-profits-and-losses-name/

The first (naive) implementation was to have the "NetWorth" account at the same level of the other wallets. NetWorth of the bank would have been equal to the fees minus the expenses.

fees being:
```
export const lndFeePath = "Expenses:Bitcoin:Fees"
export const bitcoindFeePath = "Expenses:Bitcoin:Fees"
```

revenue being:
```
export const onchainRevenuePath = "Revenue:Bitcoin:Fees"
export const revenueFeePath = "Revenue:Lightning:Fees"
```

Which make the bank account owner a very particular account because it includes wallet that are on a non standard path (not represented by a `Liabilities:uid` wallet path). I now believe this level of abstraction is not the right one. I think it's better to keep a simple structure where there is only Assets (the BTC the bank account) and Liabilities ("wallet") at the medici level. 

this reduce the complexity at the medici level. and will make proof of reserve easier to implement.

There is still a special "bankOwner" wallet that is used for all the fees that needs to be paid for opening channels or doing rebalancing for instance, and that will earn revenue on routing fees/transaction fees charge to the user, but this is now represented in the book like other wallet in `Liabilities:uid` and not by some special `Revenue:...` or `Fee:...` wallets.

We can now have a higher level of abstraction for the NetWorth/Equity of the bank by using a new layer of abstraction, that is being defines as an account that can include several wallets. 

This account abstraction for the bank account could include the "BankOwner" wallet, the "dealer" and maybe some other wallets that the bank manages. 

Other corporate account that are using our product might also want to have several wallets

**note**

- before merging this PR, we will need to update mongodb to align with the new schema. still need to work on a migration script
- instead of using `Liabilities:${uid}`, we may want to to add an abstraction and use `Liabilities:${wid}` (w: for wallet, instead of u for user)

### DIFF
--- a/test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
+++ b/test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
@@ -17,6 +17,14 @@ jest.mock("src/phone-provider", () => require("test/mocks/phone-provider"))
 
 const defaultWallet = ""
 
+it("mandatory users", async () => {
+  // load funder wallet before use it
+  await getUserWallet(4)
+
+  // "bank" user
+  await getUserWallet(14)
+})
+
 describe("Bitcoind", () => {
   it("create default wallet", async () => {
     try {
--- a/test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
+++ b/test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
@@ -17,7 +17,7 @@ jest.mock("src/phone-provider", () => require("test/mocks/phone-provider"))
 
 const defaultWallet = ""
 
-it("mandatory users", async () => {
+beforeAll(async () => {
   // load funder wallet before use it
   await getUserWallet(4)
 
--- a/test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
+++ b/test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
@@ -21,7 +21,7 @@ it("mandatory users", async () => {
   // load funder wallet before use it
   await getUserWallet(4)
 
-  // "bank" user
+  // "bankowner" user
   await getUserWallet(14)
 })
 
--- a/default.yaml
+++ b/default.yaml
@@ -63,8 +63,11 @@ test_accounts:
   code: 321321
 - phone: "+16505554333" # user12
   code: 321321
-- phone: "+12223334444" # user42
+- phone: "+12223334444" # user13
   code: "424242"
+- phone: "+16505554334" # user14
+  code: 321321
+  role: "bank"
 
 sms_provider: "twilio"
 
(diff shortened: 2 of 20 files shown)


test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
```diff
@@ -17,6 +17,14 @@ jest.mock("src/phone-provider", () => require("test/mocks/phone-provider"))
 
 const defaultWallet = ""
 
+it("mandatory users", async () => {
```
bodymindarts, line 20: This should probably be a `before` block and not an `it` block as there is no expect so nothing is actually tested here.

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
```diff
@@ -17,6 +17,14 @@ jest.mock("src/phone-provider", () => require("test/mocks/phone-provider"))
...
+  // load funder wallet before use it
+  await getUserWallet(4)
```
bodymindarts, line 22: These magic numbers are not good. We should not be hard coding IDs.
I think the whole `test_accounts` structure of the `default.yml` could be cleaner but for now perhaps we can simply wrap these in explicit function names:
test/helper/user.ts
```
function getFunderWallet() {
  const testFunderAccountId = 4; # Due to order in default.yml
  return getUserWallet(testFunderAccountId)
}
```

bodymindarts (commented):
I like this simplification a lot! Reviewing this has also given me more context regarding the `dealer negative USD balance` conversation.

> before merging this PR, we will need to update mongodb to align with the new schema. still need to work on a migration script

We need to have a conversation regarding a general strategy for rolling out migrations. AFAICT there is currently no structured way for managing migrations. This refactoring might be a good trigger to introduce a long-term solution for that.

> instead of using Liabilities:${uid}, we may want to to add an abstraction and use Liabilities:${wid} (w: for wallet, instead of u for user)

Strongly agree that we must decouple the user concept from the liabilities. Also the `ratioUsd` / `ratioBTC` fields are candidates that are likely associated with wallet rather than users. I think we are missing an entity here.

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
nicolasburtey: agree but I will left this as an exercise for someone to do another pass on test refactoring. don't want to work on this now because it will delay the merge on this PR , and this PR is a dependency for the USD integration.

bodymindarts (approved):

PR #378, opened 2021-07-24
enabling delete invoice on the fly

nicolasburtey:


test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -457,28 +457,20 @@ describe("UserWallet - Lightning Pay", () => {
...
-        })
+        await userWallet1.updatePendingPayments()
+        const query = {
+          accounts: userWallet1.user.accountPath,
+          type: "payment",
+          pending: true,
+          voided: false,
+        }
+        const count = await Transaction.countDocuments(query)
+        expect(count).toBeTruthy()
```
nicolasburtey, line 470: maybe this should be 0 actually. will need to revisit this.

nicolasburtey:
@krtk6160 do we have any pending migration with lnd chart? this PR will update it.

src/lndUtils.ts
```diff
@@ -184,6 +185,24 @@ export const getRoutingFees = async ({
   return _.map(feePerDate, (v, k) => ({ [k]: v }))
 }
 
+export const getInvoiceAttempt = async ({ lnd, id }) => {
```
dolcalmi, line 188: maybe is better just getInvoice and only returnning the invoice. If return an invoice is true, if is null is false and any other case throw the error (just like now).

```js
try {
  return await getInvoice({ lnd, id })
} catch (err) {
  const invoiceNotFound = "unable to locate invoice"
  if (err.length === 3 && err[2]?.err?.details === invoiceNotFound) {
    return null
  }
  // must be wrapped error?
  throw err
}
```

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
```diff
@@ -52,6 +52,9 @@ describe("Bitcoind", () => {
     const { chain_balance: initialBalance } = await getChainBalance({ lnd: lnd1 })
     const sats = initialBalance + btc2sat(amount)
 
+    // initiate the dealer wallet
+    await getUserWallet(6)
```
dolcalmi, line 56: nothing to be fixed now but maybe we need to add some helpers for funder and dealer wallets

test/integration/02-user-wallet/02-send-lightning.spec.ts
dolcalmi: this needs to use waitFor and must be zero (cancelHodlInvoice sometimes takes a while)

it this pass then the invoices has not been processed

dolcalmi (changes requested):

src/lndUtils.ts
nicolasburtey: this is indeed cleaner

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
nicolasburtey: agree

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -457,28 +457,20 @@ describe("UserWallet - Lightning Pay", () => {
...
-        await waitFor(async () => {
-          try {
-            await cancelHodlInvoice({ id, lnd: lndOutside1 })
-            return true
-          } catch (error) {
-            baseLogger.warn({ error }, "cancelHodlInvoice failed. trying again.")
-            return false
-          }
-        })
+        await cancelHodlInvoice({ id, lnd: lndOutside1 })
```
dolcalmi, line 460: also this needs to be in a waitfor (cancelHodlInvoice sometimes throws connection issues, maybe related to test environment or to slow environments)

dolcalmi (changes requested):

test/integration/lnd-utils.spec.ts
```diff
@@ -11,38 +12,77 @@ import {
   pay,
   waitFor,
 } from "test/helpers"
+import moment from "moment"
```
krtk6160, line 15: If we're adding the `moment` package, we can also use it in other places across the codebase (as part of another PR) where we're relying on less intuitive date manipulation

src/Lightning.ts
```diff
@@ -851,57 +840,43 @@ export const LightningMixin = (superclass) =>
...
           }
 
-          if (!invoiceUser) {
+          if (invoiceUser.paid) {
             lightningLogger.info("invoice has already been processed")
-            // FIXME not sure we should necessarily return true here
             return true
           }
 
+          if (!invoiceUser) {
```
krtk6160, line 875: This should be shifted to above the `invoiceUser.paid` if branch

krtk6160 (changes requested):
One comment but otherwise LGTM

krtk6160:
> @krtk6160 do we have any pending migration with lnd chart? this PR will update it.

Nope

dolcalmi (changes requested):

src/Lightning.ts
```diff
@@ -851,44 +840,26 @@ export const LightningMixin = (superclass) =>
...
-      } else if (invoice.is_confirmed) {
-        const lightningLogger = this.logger.child({
-          hash,
-          user: this.user._id,
-          topic: "payment",
-          protocol: "lightning",
-          transactionType: "receipt",
-          onUs: false,
-        })
+      if (invoice.is_confirmed) {
```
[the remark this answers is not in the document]
nicolasburtey: yes. no more with the new flag.

src/Lightning.ts
```diff
@@ -843,6 +831,7 @@ export const LightningMixin = (superclass) =>
         return false
       }
 
+      // TODO: look if this is this really a condition that can happen?
       if (!invoice) {
         this.logger.warn("received an invalid empty invoice from lnd")
         return false
```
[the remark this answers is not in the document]
krtk6160: Yes we have an identical if check on line 824

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -476,10 +468,12 @@ describe("UserWallet - Lightning Pay", () => {
             voided: false,
           }
           const count = await Transaction.countDocuments(query)
-          const { is_canceled } = await getInvoice({ lnd: lndOutside1, id })
-          return is_canceled && count === 0
+          return count === 0
```
krtk6160, line 471: What are we doing with the value of the return? Where's it being used?

krtk6160 (commented):
Couple more comments

test/integration/02-user-wallet/02-send-lightning.spec.ts
nicolasburtey: this is to make sure we're in sync with the lnd/backend. the goal is to avoid having some long sleeps.

dolcalmi (dismissed):

test/integration/02-user-wallet/02-send-lightning.spec.ts
krtk6160: What I meant was where is the return value being used?

krtk6160 (approved):

PR #379, opened 2021-07-25
make the makefile work with docker compose v2

nicolasburtey:


bodymindarts (changes requested):
I wouldn't want to merge this.
`docker ps -aq | xargs docker rm` will kill _all_ docker containers running on your system. There are many reasons that contributors might have other containers running while working on galoy. I don't think we should introduce commands that could effect containers that are not related to galoy itself.

Perhaps we can use some shell-foo to filter the ids returned from `docker ps` to only kill the galoy related ones.

bodymindarts (approved):
Working for me with docker-compose `v1`

Issue #364, opened 2021-07-21
upgrade our lnds to 0.13.1


nicolasburtey:

dolcalmi:
`checkIsBalanced()` is failing after channel opening in tests, I'm not sure if it's about fees, but if it's not urgent I'll check on the weekend

Image used in docker-compose: `lncm/lnd:v0.13.1-monitoring`

nicolasburtey:
this is not urgent

dolcalmi:
## Premise (according to our code)

- After open the channel: **capacity - commit_transaction_fee - local_balance = 0**


## Problem

we are using commit_transaction_fee to calculate escrow in Lnd but v0.13.1 is returning an inconsistent commit_transaction_fee or there is an additional fee that is not parsed by lightning package (altough I can't see anything new in https://api.lightning.community/#lnrpc-channel)

### lncm/lnd:v0.12.1-monitoring
**capacity - commit_transaction_fee - local_balance = 0**
`1000000 - 9050 - 990950 =` **0** (ok)

<details>
  <summary>data from updateEscrows method in lndUtils.ts</summary>

```json
{
  diff: 9050,
  escrowInLnd: 9050,
  escrowInMongodb: 0,
  channels: [
    {
      "capacity": 1000000,
      "commit_transaction_fee": 9050,
      "commit_transaction_weight": 600,
      "id": "142x1x0",
      "is_active": true,
      "is_anchor": false,
      "is_closing": false,
      "is_opening": false,
      "is_partner_initiated": false,
      "is_private": false,
      "is_static_remote_key": true,
      "is_variable_remote_key": false,
      "local_balance": 990950,
      "local_csv": 144,
      "local_dust": 573,
      "local_given": 0,
      "local_max_htlcs": 50,
      "local_max_pending_mtokens": "990000000",
      "local_min_htlc_mtokens": "1",
      "local_reserve": 10000,
      "partner_public_key": "038e4ed546cf24c19b14a59437ac53e7c89ecd2d3265bd35ab3dcea0856002944d",
      "pending_payments": [],
      "received": 0,
      "remote_balance": 0,
      "remote_csv": 144,
      "remote_dust": 573,
      "remote_given": 0,
      "remote_max_htlcs": 50,
      "remote_max_pending_mtokens": "990000000",
      "remote_min_htlc_mtokens": "1",
      "remote_reserve": 10000,
      "sent": 0,
      "time_offline": 0,
      "time_online": 1000,
[...]

PR #380, opened 2021-07-26
Upgrade lnd in dev env to v0.13.1-monitoring

dolcalmi:
- update docker images to v0.13.1-monitoring
- reorganize lnd.conf file
- add `protocol.no-anchors=true` (to avoid fee issues in tests, this also need to be configured in production)

ref: [issue 364](https://github.com/GaloyMoney/galoy/issues/364)


bodymindarts:
We should keep the dev dependencies == to the versions we are running in production. I will close this PR (to avoid accidental merging) until we have updated to lnd 0.13.1 in prod.

PR #385, opened 2021-07-28
Implement the operations for lnpage

samerbuna:
Updated PR with an implementation following the "[returning exception](https://www.oreilly.com/library/view/programming-typescript/9781492037644/ch07.html)" TypeScript error-handling strategy for type-safety.

This is similar to what we get from `neverthrow` except without needing to wrap things and use a non-idiomatic way.

This is a _basic_ implementation to get the team's input. Things can be extended further and checking can be exhausted (leveraging TS `never` type).

Syntax sugar can also be used for error checking but I'm not in favor of that. The closer we stay to idiomatic and familiar JavaScript code, the better, IMO.


src/graphql/core/schema.graphql
```diff
@@ -5,5 +5,53 @@ type Query {
 scalar Date
 
 type Mutation {
-  currentTime: Date
+  lnNoAmountInvoiceCreateForUser(
```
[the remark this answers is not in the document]
samerbuna: Renamed to on-behalf-of-recipient

src/core/lightning/add-invoice-for-recipient.ts
```diff
@@ -0,0 +1,24 @@
...
+const addInvoiceForRecipient = async ({ recipient, memo }, { logger }) => {
+  const wallet = await getWalletFromUsername({ username: recipient, logger })
+  const paymentRequest = await wallet.addInvoice({
+    selfGenerated: false,
+    memo,
+  })
+
+  const decodedInvoice = decodeInvoice(paymentRequest)
+
+  if (decodedInvoice instanceof Error) return decodedInvoice
```
bodymindarts, line 13: Shouldn’t we be checking against CoreError?

src/core/lightning/payment-status-checker.ts
```diff
@@ -0,0 +1,28 @@
...
+  const { paymentHash, paymentSecret } = decodedInvoice
+
+  // TODO: Improve the following check with a non public payment secret
+  if (paymentSecret !== lookupToken) {
+    return new LnInvoiceLookupError("Invalid invoice data")
+  }
+
+  return {
+    invoiceIsPaid: async (): Promise<boolean | LnInvoiceLookupError> => {
+      const invoiceUser = await InvoiceUser.findOne({ _id: paymentHash })
```
bodymindarts, line 19: What if mongoose throws because there is no connection or something like that?

src/domain/ln-invoice.ts
```diff
@@ -0,0 +1,49 @@
...
+      }
+    }
+    if (tag.tagName === "payment_secret") {
+      if (typeof tag.data === "string") {
+        paymentSecret = tag.data as PaymentSecret
+      } else {
+        return new LnInvoiceDecodeError("Irregular payment_secret")
+      }
+    }
+    if (!paymentHash || !paymentSecret) {
```
bodymindarts, line 39: This looks like it would always fail as there are more tags than those 2. For those to keep looping one of these tags has to be first. Does the spec determine and order here?

src/graphql/types/scalars/ln-invoice-payment-request.ts
```diff
@@ -0,0 +1,24 @@
...
+    if (ast.kind === GT.Kind.STRING) {
+      return validLnInvoicePaymentRequest(ast.value)
+    }
+    return new Error("Invalid type for LnInvoicePaymentRequest")
+  },
+})
+
+function validLnInvoicePaymentRequest(value) {
+  // TODO: verify/improve
+  if (value.match(/^ln[A-Za-z0-9]+$/)) {
```
bodymindarts, line 18: Could we validate by calling the decode domain function?

src/core/lightning/add-invoice-for-recipient.ts
samerbuna: We could if we need to (to something different for `CodeError` for example). Checking against `Error` is kinda the fallback here.

src/core/lightning/payment-status-checker.ts
samerbuna: Ideally, once we have a good entry point for mongoose operations, we can follow this pattern there and catch cases like these. 

We can of course return an error here directly but I think we need to move forward with the high priorities first before we nail every possible error case (that's not currently handled in the existing schema). Same thing goes for tests, let's convert the tests we currently have while we change the outer layer, then go another rounds focusing on coverage.

src/graphql/types/scalars/ln-invoice-payment-request.ts
samerbuna: I guess we could but might be an overkill (during GraphQL parsing). Other input errors might block the operation anyway.

Not sure... @nicolasburtey what do you think?

src/domain/ln-invoice.ts
samerbuna: Good catch (pun intended). That's a duplicate check that's in the wrong place. Removed.

bodymindarts (approved):

vindard (approved):

PR #388, opened 2021-07-28
Fix onchain notification issue with new account structure

dolcalmi:
- Add liabilitiesMainAccount path to ledger
- Add new way to resolve account id (user id) from account path
- Add new unit tests to account id resolver
- Add additional user id validation in onchain "sends a successful payment" test (this prevents future errors)

Ref: [PR 377 - improve accounting structure](https://github.com/GaloyMoney/galoy/pull/377)

### DIFF
--- a/src/entrypoint/trigger.ts
+++ b/src/entrypoint/trigger.ts
@@ -9,6 +9,11 @@ import {
   subscribeToInvoices,
   subscribeToTransactions,
 } from "lightning"
+import {
+  bankOwnerMediciPath,
+  liabilitiesMainAccount,
+  resolveAccountId,
+} from "src/ledger/ledger"
 import { updateUsersPendingPayment } from "../ledger/balanceSheet"
 import { activateLndHealthCheck, lndStatusEvent } from "../lndHealth"
 import { onChannelUpdated } from "../lndUtils"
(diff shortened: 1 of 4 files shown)


src/entrypoint/trigger.ts
```diff
@@ -77,12 +82,29 @@ export async function onchainTransactionEventHandler(tx) {
       { success: true, pending: false, transactionType: "payment" },
       "payment completed",
     )
-    const entry = await Transaction.findOne({
-      account_path: "Liabilities",
+
+    const entries = await Transaction.find({
+      account_path: liabilitiesMainAccount,
```
nicolasburtey, line 87: why not adding something like:
`accounts: { $ne: bankOwnerPath }`
in the query?

src/ledger/ledger.ts
```diff
@@ -10,7 +10,32 @@ export const lndAccountingPath = "Assets:Reserve:Lightning" // TODO: rename to A
 export const escrowAccountingPath = "Assets:Reserve:Escrow" // TODO: rename to Assets:Lnd:Escrow
 
 // liabilities
-export const accountPath = (uid) => `Liabilities:${uid}`
+export const liabilitiesMainAccount = "Liabilities"
+export const accountPath = (uid) => `${liabilitiesMainAccount}:${uid}`
+export const resolveAccountId = (accountPath: string | string[]) => {
```
nicolasburtey, line 15: is this still needed at all?

I don't think we typically want to merge `accountPath` and `accounts` (I think we should eventually have just one of those 2 options in our codebase to avoid the issue)

dolcalmi: I prefer this instead of use index in account path, also it is safer in case of account path changes (it is unit tested)

nicolasburtey: but I'm still not sure to understand why we want the overhead of parsing both a table and a string. this could lead to us mixing types which would be bad IMO

dolcalmi: this is because of medici, account paths are strings (accounts) or arrays of strings (account_path)

Issue #346, opened 2021-07-19
Listen to bitcoind with ZeroMQ


jotapea:
Use [ZeroMQ](https://github.com/bitcoin/bitcoin/blob/master/doc/zmq.md) to listen for wallet transactions and new blocks, replacing lightning's native [event subscriptions](https://github.com/GaloyMoney/galoy/blob/426eba9450963d9e423d9905d2f2a91a5d09319a/src/entrypoint/trigger.ts#L175).

Needed to continue with https://github.com/GaloyMoney/galoy/pull/317

jotapea:
Go with [v6](https://github.com/zeromq/zeromq.js/#zeromqjs-next-generation)? Using v5 for now, as it is still suggested in the [docs](https://zeromq.org/languages/nodejs/).

nicolasburtey:
@jotapea ZeroMQ is already active on our bitcoind I believe. this is also how lnd is getting all its transactions.

jotapea:
@nicolasburtey Thanks for the info (also updated the title).

PR #390, opened 2021-07-29
Zeromq bitcoind subscriptions

jotapea:
PR for [Listen to bitcoind with ZeroMQ #346](https://github.com/GaloyMoney/galoy/issues/346).

### DIFF
--- a/src/bitcoindSubscribers.ts
+++ b/src/bitcoindSubscribers.ts
@@ -1,0 +1,49 @@
+import { Subscriber } from "zeromq"
+import { bitcoindDefaultClient } from "src/bitcoind"
+import { baseLogger } from "src/logger"
+
+export const SUB_ADDR_BLOCK = `tcp://${process.env.BITCOINDADDR}:28332` // TODO? port
+export const SUB_ADDR_TX = `tcp://${process.env.BITCOINDADDR}:28333`
+
+export async function receiveRawBlockSubscriber() {
+  const subscribeTopic = "rawblock" // "rawblock", "hashblock", "rawtx", or "hashtx"
+
+  const sock = new Subscriber()
+  sock.connect(SUB_ADDR_BLOCK)
+  sock.subscribe(subscribeTopic)
+
+  async function worker() {
+    for await (const [topic, msg] of sock) {
+      if (topic.toString() === subscribeTopic) {
+        const rawTx = msg.toString("hex")
+        baseLogger.warn(`rawTx.substring: ${rawTx.substring(0, 64)}`)
+      }
+    }
+  }
+
+  // eslint-disable-next-line require-await
+  worker() // no await
+  return await sock
+}
+
+export async function receiveRawTxSubscriber() {
+  const subscribeTopic = "rawtx"
+
+  const sock = new Subscriber()
+  sock.connect(SUB_ADDR_TX)
+  sock.subscribe(subscribeTopic)
+
+  async function worker() {
+    for await (const [topic, msg] of sock) {
+      if (topic.toString() === subscribeTopic) {
+        const rawTx = msg.toString("hex")
+        const tx = await bitcoindDefaultClient.decodeRawTransaction({ hexstring: rawTx })
+        baseLogger.warn(`tx: ${JSON.stringify(tx)}`)
+      }
+    }
+  }
+
+  // eslint-disable-next-line require-await
+  worker() // no await
+  return await sock
+}
--- a/src/bitcoindSubscribers.ts
+++ b/src/bitcoindSubscribers.ts
@@ -5,7 +5,7 @@ import { baseLogger } from "src/logger"
 export const SUB_ADDR_BLOCK = `tcp://${process.env.BITCOINDADDR}:28332` // TODO? port
 export const SUB_ADDR_TX = `tcp://${process.env.BITCOINDADDR}:28333`
 
-enum EVENTS {
+export enum BITCOIND_EVENTS {
   RAW_BLOCK = "rawblock",
   RAW_TX = "rawtx",
   // HASH_BLOCK = "hashblock",
@@ -15,11 +15,11 @@ enum EVENTS {
 export async function receiveRawBlockSubscriber() {
   const sock = new Subscriber()
   sock.connect(SUB_ADDR_BLOCK)
-  sock.subscribe(EVENTS.RAW_BLOCK)
+  sock.subscribe(BITCOIND_EVENTS.RAW_BLOCK)
 
   async function worker() {
     for await (const [topic, msg] of sock) {
-      if (topic.toString() === EVENTS.RAW_BLOCK) {
+      if (topic.toString() === BITCOIND_EVENTS.RAW_BLOCK) {
         const rawTx = msg.toString("hex")
         baseLogger.warn(`rawTx.substring: ${rawTx.substring(0, 64)}`)
       }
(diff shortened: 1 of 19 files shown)


src/bitcoind.ts
```diff
@@ -87,10 +92,22 @@ export class BitcoindClient {
...
   }
 
+  // default client knows about all transactions of its wallets
+
+  async getTransaction({
+    txid,
+    include_watchonly,
+  }: {
+    txid: string
+    include_watchonly?: boolean
```
bodymindarts, line 106: This shouldn't be nullable. Should always be either `true` or `false`. If you don't always want to pass it best would be to set a default like `include_watchonly = true`

src/bitcoind.ts
```diff
@@ -87,10 +92,22 @@ export class BitcoindClient {
...
+  }): Promise<InWalletTransaction> {
+    return await this.client.getTransaction({ txid, include_watchonly })
```
bodymindarts, line 108: Why are you sometimes using the default client and sometimes `this.client`? I think I'd prefer to always use `this.client` and pass in the `default` client via constructor if that works.

src/bitcoindSubscribers.ts
```diff
@@ -0,0 +1,49 @@
+import { Subscriber } from "zeromq"
+import { bitcoindDefaultClient } from "src/bitcoind"
+import { baseLogger } from "src/logger"
+
+export const SUB_ADDR_BLOCK = `tcp://${process.env.BITCOINDADDR}:28332` // TODO? port
+export const SUB_ADDR_TX = `tcp://${process.env.BITCOINDADDR}:28333`
+
+export async function receiveRawBlockSubscriber() {
+  const subscribeTopic = "rawblock" // "rawblock", "hashblock", "rawtx", or "hashtx"
```
bodymindarts, line 9: can this be an enum?

test/integration/01-setup/01-connection.spec.ts
```diff
@@ -21,6 +27,22 @@ it("connects to bitcoind", async () => {
   expect(chain).toEqual("regtest")
 })
 
+it("detect active bitcoind zeromq notifications", async () => {
+  const notifications = await bitcoindClient.getZmqNotifications()
+  expect(notifications.length).toEqual(4) // 2 per topic
```
bodymindarts, line 32: Where is the code that creates these notifications? This is very implicit and order dependent. Perhaps you could make this idempotent like so:
```

File: src/bitcoindSubscribers.ts [hunk 5]
```diff
@@ -0,0 +1,49 @@
...
+
+  const sock = new Subscriber()
+  sock.connect(SUB_ADDR_TX)
+  sock.subscribe(subscribeTopic)
+
+  async function worker() {
+    for await (const [topic, msg] of sock) {
+      if (topic.toString() === subscribeTopic) {
+        const rawTx = msg.toString("hex")
+        const tx = await bitcoindDefaultClient.decodeRawTransaction({ hexstring: rawTx })
```
bodymindarts, line 40: can we always pass in a client and not ever use the default client except on the top level when we inject it? This would remove coupling to global state and improve testability.

test/integration/01-setup/01-connection.spec.ts
jotapea: Enabling zeromq is set in bitcoind configuration, and it was already enabled. So that call is not being affected by this branch.

jotapea:
@bodymindarts The way I am approaching it is that the wallet-clients are separate from the "loading of" or "total of" loaded wallets in the default (main) client. So, to use a wallet, you actually need a new wallet-client for this. Also the classes are done in a way that their methods don't overlap (at least for now).

src/bitcoind.ts
jotapea: Done

test/integration/01-setup/01-connection.spec.ts
bodymindarts: I don't understand. This is a new test right? So can we try to make any new tests we add be idempotent?

jotapea: It is a newly added test yes, but is not actually testing anything that has a "before" and "after". It is just detecting if the bitcoind client has the zeromq notifications enabled.

jotapea:
@nicolasburtey @bodymindarts I think this is ready to get merged, at least from the perspective of testing. I am not sure where else changes need to be done for production compatibility (i.e. equivalent to the opening of the ports in docker-compose).

src/bitcoind.ts
jotapea: `this.client` is also the instance variable name used in the BitcoindWalletClient because it actually is a client just like in the BitcoindClient class. The only difference is that in BitcoindWalletClient it was instantiated with the `wallet` argument. This is the way for the BitcoindWalletClient instances to contact the same "default" (main) client while being "loaded" and ready to invoke wallet rpc calls.

src/core/specter-wallet.ts
```diff
@@ -267,7 +267,7 @@ export class SpecterWallet {
       throw new Error(err)
     }
 
-    const tx = await this.bitcoindClient.getTransaction({
+    const tx = await bitcoindDefaultClient.getTransaction({
```
bodymindarts, line 270: Can we please avoid coupling to global state / objects. No singletons... use constructor injection.

src/services/bitcoind/subscribers.ts
```diff
@@ -0,0 +1,52 @@
...
+export async function receiveRawTxSubscriber() {
+  const sock = new Subscriber()
+  sock.connect(SUB_ADDR_TX)
+  sock.subscribe(BITCOIND_EVENTS.RAW_TX)
+
+  async function worker() {
+    for await (const [topic, msg] of sock) {
+      if (topic.toString() === BITCOIND_EVENTS.RAW_TX) {
+        const rawTx = msg.toString("hex")
+        const tx = await bitcoindDefaultClient.decodeRawTransaction({ hexstring: rawTx })
```
bodymindarts, line 43: Can we please not reference singletons / depend on global state?

bodymindarts (commented):
The PR looks good but still you are coupling to global state which we are trying to move away from for better testability / code structure / modularisation etc...

jotapea:
@bodymindarts `bitcoindDefaultClient` is now fully removed.

src/core/on-chain/index.ts
```diff
@@ -61,10 +61,14 @@ export const getOnChainTransactions = async ({
 export const OnChainMixin = (superclass) =>
   class extends superclass {
     readonly config: UserWalletConfig
+    // TODO? is this ok here?
+    readonly bitcoindClient
 
     constructor(...args) {
       super(...args)
       this.config = args[0].config
+      // TODO? is this ok here?
```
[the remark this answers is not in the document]
jotapea: @bodymindarts Comment removed, and I assume it is ok to add the bitcoind client as an instance variable to the "mixin", at least for now.

src/services/bitcoind/subscribers.ts
```diff
@@ -0,0 +1,56 @@
...
+import { BitcoindClient } from "./index"
+import { baseLogger } from "@services/logger"
+
+export const SUB_ADDR_BLOCK = `tcp://${process.env.BITCOINDADDR}:28332` // TODO? port
+export const SUB_ADDR_TX = `tcp://${process.env.BITCOINDADDR}:28333`
+
+export enum BITCOIND_EVENTS {
+  RAW_BLOCK = "rawblock",
+  RAW_TX = "rawtx",
+  // HASH_BLOCK = "hashblock",
```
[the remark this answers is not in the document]
jotapea: For this one I am not really sure why we have to remove them. It is there for completeness of the events that bitcoind emits.

src/core/specter-wallet.ts
```diff
@@ -14,23 +14,36 @@ import { btc2sat, sat2btc } from "./utils"
...
 export class SpecterWallet {
-  bitcoindClient // TODO rename?
+  // Needs access to both classes' methods
+  bitcoindClient
+  bitcoindWalletClient
   readonly logger: Logger
   readonly config: SpecterWalletConfig
 
-  constructor({ logger, config }: SpecterWalletConstructorArgs) {
+  // TODO? Add BitcoindClient type to SpecterWalletConstructorArgs
```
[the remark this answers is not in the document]
jotapea: How should I add the BitcoindClient type to the SpecterWalletConstructorArgs? @bodymindarts @samerbuna @dolcalmi

PR #391, opened 2021-07-29
Refactor 'Lightning' types/config placement

vindard:
## Description

This PR removes `yamlConfig` references from the `Lightning.ts` and `lndUtils.ts` files.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -24,7 +24,7 @@ export const yamlConfig = _.merge(defaultConfig, customConfig)
 
 export const getGaloyInstanceName = () => yamlConfig.name
 
-export const getLndParams = (): ILndParams[] => {
+export const getLndParams = (): LndParams[] => {
   const config = yamlConfig.lnds
   return config.map((input) => ({
     cert: process.env[`${input.name}_TLS`] || exit(98),
--- a/src/config.ts
+++ b/src/config.ts
@@ -22,7 +22,7 @@ try {
 
 export const yamlConfig = _.merge(defaultConfig, customConfig)
 
-export const getName = () => yamlConfig.name
+export const getGaloyInstanceName = () => yamlConfig.name
 
 export const getLndParams = (): ILndParams[] => {
   const config = yamlConfig.lnds
--- a/src/config.ts
+++ b/src/config.ts
@@ -2,6 +2,7 @@ import fs from "fs"
 import yaml from "js-yaml"
 import _ from "lodash"
 import { baseLogger } from "./logger"
+import { exit } from "process"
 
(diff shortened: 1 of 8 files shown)


samerbuna (dismissed):

src/lndAuthTypes.d.ts
```diff
@@ -0,0 +1,16 @@
+type nodeType = "offchain" | "onchain"
+
+interface ILndParams {
```
bodymindarts, line 3: I know you're just moving code around... but since you're touching it we may as well improve it. I think pure data types that have no behaviour don't need to be interfaces in general. WDYT?

src/lndUtils.ts
```diff
@@ -444,7 +443,8 @@ export const validate = async ({
     // TODO: if expired_at expired, thrown an error
   } else {
     if (!params.username) {
-      const error = `a username is required for push payment to the ${yamlConfig.name}`
+      const name = getName()
```
bodymindarts, line 446: Here too... may as well clean things up if we are touching this code. Something more descriptive like `galoyInstanceName` might be appropriate?

src/config.ts
```diff
@@ -21,6 +22,21 @@ try {
 
 export const yamlConfig = _.merge(defaultConfig, customConfig)
 
+export const getName = () => yamlConfig.name
```
bodymindarts, line 25: How about `galoyInstanceNames`?

src/lndAuthTypes.d.ts
vindard: I think in this case I chose `interface` so that I could extend it later on (`ILndParamsAuthed` extends from `ILndParams`). Extending felt more correct than declaring it as an " intersection type" (by using `&` to join them), and it felt odd to declare `LndParams` as a type and then extend `ILndParamsAuthed` from it as an interface, so I chose to leave them both as interfaces.

That being said, it could be useful to come up with a general set of expectations around when to use types vs. interfaces. Right off the bat I saw that once a type is created it can't be edited whereas new properties can always be added to an interface. Also interfaces can extend from both types and interfaces whereas types can't extend from anything.

src/lndUtils.ts
vindard: Agreed :+1:

bodymindarts (approved):
I don't like the dependency to `getGaloyInstanceName` in lndUtils but I guess that's a different refactoring.

src/lndAuthTypes.d.ts
bodymindarts: > For e.g., we can reserve interface only for things that we mean to implement?

Yes this is the approach I would take.

samerbuna (approved):
LGTM

PR #393, opened 2021-07-29
Refactor 'TransactionLimits' types/config

vindard:
## Description

This PR refactors the `TransactionLimits` class to instead be a function where the entire limits config is set up on function call instead of on individual method calls. It also splits off a `UserLimits` class to handle _level-specific_ limits. This split was done because there are places in the codebase where properties like `oldEnoughForWithdrawal` from the `TransactionLimits` object are called without needing a `level` value.

These changes are used to remove limits-related `yamlConfig` references from the `schema.ts` & `graphql-core-server.ts` files.

**_Edit 1:_**

_To round off this PR, configs were also created for `yamlConfig.fees` values so the the `yamlConfig` import could be removed entirely from `schema.ts` & `graphql-core-server.ts`._

---
### Open `TODO`s from PR comments:
- [x] ~Rebase to latest tip of `main` after all changes are confirmed to resolve current merge conflicts~ **_[Done](https://github.com/GaloyMoney/galoy/pull/393#issuecomment-891118363)_**
- [x] ~Add time suffix everywhere ([here](https://github.com/GaloyMoney/galoy/pull/393#discussion_r680658651))~ **_Done https://github.com/GaloyMoney/galoy/pull/393/commits/2cb48add8c997ebbff77f711fb0ede1e61374671_**
- [x] ~Add time constant ([here](https://github.com/GaloyMoney/galoy/pull/393#discussion_r680660465))~ **_Done https://github.com/GaloyMoney/galoy/pull/393/commits/be90463873ff2d36b64f55e03c72068638e79d95_**
- [x] ~Add description/type to `toThrow()` to fix integration tests ([see here](https://github.com/GaloyMoney/galoy/pull/393#discussion_r680106515) and fix [here](https://github.com/GaloyMoney/galoy/pull/393#discussion_r680527467))~ **_Done https://github.com/GaloyMoney/galoy/pull/393/commits/4e7ed7d55cf931b2aaa254c963b0fa1c473d04aa_**
- [x] _Remove `Limits` suffix for properties on the `config.limits` object_ (**[open question for reviewers](https://github.com/GaloyMoney/galoy/pull/393#discussion_r681091304)**)

### DIFF
--- a/src/config/app.ts
+++ b/src/config/app.ts
@@ -45,9 +45,11 @@ export const getGenericLimits = (limitsConfig = yamlConfig.limits): GenericLimit
   oldEnoughForWithdrawalMicroseconds: limitsConfig.oldEnoughForWithdrawal,
 })
 
-export const getFees = (feesConfig = yamlConfig.fees): FeeConstants => ({
-  depositFeeRate: feesConfig.deposit,
-  withdrawFeeFlat: feesConfig.withdraw,
+export const getFeeRates = (feesConfig = yamlConfig.fees): FeeRates => ({
+  depositFeeVariable: feesConfig.deposit,
+  depositFeeFixed: 0,
+  withdrawFeeVariable: 0,
+  withdrawFeeFixed: feesConfig.withdraw,
 })
 
 export const getUserLimits = ({
--- a/src/config/app.ts
+++ b/src/config/app.ts
@@ -40,12 +40,12 @@ export const getLndParams = (): LndParams[] => {
   }))
 }
 
-const limitConstants = {
+export const limitConstants: LimitConstants = {
   oldEnoughForWithdrawalHours: yamlConfig.limits.oldEnoughForWithdrawal / MS_IN_HOUR,
   oldEnoughForWithdrawalMicroseconds: yamlConfig.limits.oldEnoughForWithdrawal,
 }
 
-export const selectUserLimits = ({ level }: { level: number }): IUserLimits => {
+export const selectUserLimits = ({ level }: UserLimitsArgs): IUserLimits => {
   const config = yamlConfig.limits
   return {
     onUsLimit: config.onUs.level[level],
@@ -76,16 +76,20 @@ export const getLoginAttemptLimits = () => getRateLimits(yamlConfig.limits.login
 export const getFailedAttemptPerIpLimits = () =>
   getRateLimits(yamlConfig.limits.failedAttemptPerIp)
 
-export const getUserWalletConfig = (user): UserWalletConfig => {
-  const userLimits = selectUserLimits({ level: user.level })
+export const selectTransactionLimits = ({
+  level,
+}: UserLimitsArgs): ITransactionLimits => ({
+  oldEnoughForWithdrawalLimit: limitConstants.oldEnoughForWithdrawalMicroseconds,
+  oldEnoughForWithdrawalLimitHours: limitConstants.oldEnoughForWithdrawalHours,
+  ...selectUserLimits({ level }),
+})
 
+export const getUserWalletConfig = (user): UserWalletConfig => {
+  const transactionLimits = selectTransactionLimits({ level: user.level })
   return {
     name: yamlConfig.name,
     dustThreshold: yamlConfig.onChainWallet.dustThreshold,
-    limits: {
-      oldEnoughForWithdrawalLimit: limitConstants.oldEnoughForWithdrawalHours,
-      ...userLimits,
-    },
+    limits: transactionLimits,
   }
 }
 
--- a/src/config/app.ts
+++ b/src/config/app.ts
@@ -40,7 +40,7 @@ export const getLndParams = (): LndParams[] => {
   }))
 }
 
-export const getLimitConstants = (limitsConfig = yamlConfig.limits): LimitConstants => ({
+export const getGenericLimits = (limitsConfig = yamlConfig.limits): GenericLimits => ({
   oldEnoughForWithdrawalHours: limitsConfig.oldEnoughForWithdrawal / MS_IN_HOUR,
   oldEnoughForWithdrawalMicroseconds: limitsConfig.oldEnoughForWithdrawal,
 })
@@ -87,10 +87,10 @@ export const getTransactionLimits = ({
   level,
   limitsConfig = yamlConfig.limits,
 }: UserLimitsArgs): ITransactionLimits => {
-  const limitConstants = getLimitConstants(limitsConfig)
+  const genericLimits = getGenericLimits(limitsConfig)
   return {
-    oldEnoughForWithdrawalLimit: limitConstants.oldEnoughForWithdrawalMicroseconds,
-    oldEnoughForWithdrawalLimitHours: limitConstants.oldEnoughForWithdrawalHours,
+    oldEnoughForWithdrawalLimit: genericLimits.oldEnoughForWithdrawalMicroseconds,
+    oldEnoughForWithdrawalLimitHours: genericLimits.oldEnoughForWithdrawalHours,
     ...getUserLimits({ level, limitsConfig }),
   }
 }
--- a/src/config/app.ts
+++ b/src/config/app.ts
@@ -40,20 +40,17 @@ export const getLndParams = (): LndParams[] => {
   }))
 }
 
-export class TransactionLimits implements ITransactionLimits {
-  readonly config
-  readonly level
+const limitConstants = {
+  oldEnoughForWithdrawalHours: yamlConfig.limits.oldEnoughForWithdrawal / MS_IN_HOUR,
+  oldEnoughForWithdrawalMicroseconds: yamlConfig.limits.oldEnoughForWithdrawal,
+}
 
-  constructor({ level }) {
-    this.config = yamlConfig.limits
-    this.level = level
+export const selectUserLimits = ({ level }: { level: number }): IUserLimits => {
+  const config = yamlConfig.limits
+  return {
+    onUsLimit: config.onUs.level[level],
+    withdrawalLimit: config.withdrawal.level[level],
   }
-
-  onUsLimit = () => this.config.onUs.level[this.level]
-
-  withdrawalLimit = () => this.config.withdrawal.level[this.level]
-
-  oldEnoughForWithdrawalLimit = () => this.config.oldEnoughForWithdrawal / MS_IN_HOUR
 }
 
 const getRateLimits = (config): IRateLimits => {
@@ -80,14 +77,15 @@ export const getFailedAttemptPerIpLimits = () =>
   getRateLimits(yamlConfig.limits.failedAttemptPerIp)
 
 export const getUserWalletConfig = (user): UserWalletConfig => {
-  const transactionLimits = new TransactionLimits({
-    level: user.level,
-  })
+  const userLimits = selectUserLimits({ level: user.level })
 
   return {
     name: yamlConfig.name,
     dustThreshold: yamlConfig.onChainWallet.dustThreshold,
-    limits: transactionLimits,
+    limits: {
+      oldEnoughForWithdrawalLimit: limitConstants.oldEnoughForWithdrawalHours,
+      ...userLimits,
+    },
   }
 }
 
--- a/src/config/app.ts
+++ b/src/config/app.ts
@@ -40,16 +40,20 @@ export const getLndParams = (): LndParams[] => {
   }))
 }
 
-export const limitConstants: LimitConstants = {
-  oldEnoughForWithdrawalHours: yamlConfig.limits.oldEnoughForWithdrawal / MS_IN_HOUR,
-  oldEnoughForWithdrawalMicroseconds: yamlConfig.limits.oldEnoughForWithdrawal,
-}
+export const generateLimitConstants = (
+  limitsConfig = yamlConfig.limits,
+): LimitConstants => ({
+  oldEnoughForWithdrawalHours: limitsConfig.oldEnoughForWithdrawal / MS_IN_HOUR,
+  oldEnoughForWithdrawalMicroseconds: limitsConfig.oldEnoughForWithdrawal,
+})
 
-export const selectUserLimits = ({ level }: UserLimitsArgs): IUserLimits => {
-  const config = yamlConfig.limits
+export const selectUserLimits = ({
+  level,
+  limitsConfig = yamlConfig.limits,
+}: UserLimitsArgs): IUserLimits => {
   return {
-    onUsLimit: config.onUs.level[level],
-    withdrawalLimit: config.withdrawal.level[level],
+    onUsLimit: limitsConfig.onUs.level[level],
+    withdrawalLimit: limitsConfig.withdrawal.level[level],
   }
 }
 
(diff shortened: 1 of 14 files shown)


vindard:
**Converted to `draft` because I'd like to add some tests for these changes**

src/entrypoint/graphql-core-server.ts
```diff
@@ -150,10 +150,11 @@ const resolvers = {
     },
     getLevels: () => levels,
     getLimits: (_, __, { user }) => {
+      const transactionLimits = selectTransactionLimits({ level: user.level })
       return {
-        oldEnoughForWithdrawal: yamlConfig.limits.oldEnoughForWithdrawal,
-        withdrawal: yamlConfig.limits.withdrawal.level[user.level],
-        onUs: yamlConfig.limits.onUs.level[user.level],
+        oldEnoughForWithdrawal: transactionLimits.oldEnoughForWithdrawalLimit,
```
vindard, line 155: Ideally I'd like to secure this change with a test, but I'm not sure if now is the right time or what the best place would be with the other tests work you all were doing recently? (cc @dolcalmi)

src/config.ts
```diff
@@ -21,20 +21,26 @@ try {
...
+  oldEnoughForWithdrawalHours: limitsConfig.oldEnoughForWithdrawal / MS_IN_HOUR,
+  oldEnoughForWithdrawalMicroseconds: limitsConfig.oldEnoughForWithdrawal,
+})
+
+export const generateFeeConstants = (feesConfig = yamlConfig.fees): FeeConstants => ({
+  depositFeeRate: feesConfig.deposit,
+  withdrawFeeFlat: feesConfig.withdraw,
+})
+
+export const selectUserLimits = ({
```
dolcalmi, line 36: I think is better to use `getUserLimits` like `getRateLimits`, `getUserWalletConfig` ...

src/config.ts
```diff
@@ -60,11 +66,23 @@ export const getLoginAttemptLimits = () => getRateLimits(yamlConfig.limits.login
 export const getFailedAttemptPerIpLimits = () =>
   getRateLimits(yamlConfig.limits.failedAttemptPerIp)
 
-export const getUserWalletConfig = (user): UserWalletConfig => {
-  const transactionLimits = new TransactionLimits({
-    level: user.level,
-  })
+export const selectTransactionLimits = ({
```
dolcalmi, line 69: same here

src/entrypoint/graphql-core-server.ts
dolcalmi: I think you can add extra tests related to this here `test/integration/02-user-wallet/02-index.spec.ts`

dolcalmi: also if you want to test graphql query then add it here `test/integration/entrypoint/graphql.spec.ts`

src/config.ts
```diff
@@ -21,20 +21,26 @@ try {
...
-    this.config = yamlConfig.limits
-    this.level = level
+export const generateLimitConstants = (
+  limitsConfig = yamlConfig.limits,
+): LimitConstants => ({
+  oldEnoughForWithdrawalHours: limitsConfig.oldEnoughForWithdrawal / MS_IN_HOUR,
+  oldEnoughForWithdrawalMicroseconds: limitsConfig.oldEnoughForWithdrawal,
+})
+
+export const generateFeeConstants = (feesConfig = yamlConfig.fees): FeeConstants => ({
```
dolcalmi, line 31: this is a weird name "Generate Constants" :) maybe just use getFees or getFeeRates (although not sure about the use of withdrawFeeFlat)

src/config.ts
```diff
@@ -21,20 +21,26 @@ try {
...
 export const yamlConfig = _.merge(defaultConfig, customConfig)
 
-export class TransactionLimits implements ITransactionLimits {
-  readonly config
-  readonly level
-
-  constructor({ level }) {
-    this.config = yamlConfig.limits
-    this.level = level
+export const generateLimitConstants = (
```
dolcalmi, line 24: the same as generateFeeConstants

test/integration/02-user-wallet/02-send-onchain.spec.ts
```diff
@@ -343,6 +345,9 @@ describe("UserWallet - onChainPay", () => {
       { $group: { _id: null, outgoingSats: { $sum: "$debit" } } },
     ])
     const { outgoingSats } = result || { outgoingSats: 0 }
+
+    // FIXME: 'withdrawalLimit' no longer exists as of
```
dolcalmi, line 349: so the test is passing because a negative amount throws an exception? 

we need to add the description or type to `.toThrow()` and fix te test

dolcalmi (changes requested):

src/config.ts
vindard: Agreed, I think I was trying to use different names based on whether I was passing arguments to select values, or whether the method was doing a straight mapping. Thinking back now, that probably doesn't make a whole lot of sense :sweat_smile:

src/config.ts
vindard: True. I was trying to convey `flat` vs `percentage` fees in the extra `withdrawFeeFlat` naming. What do we think about these instead:
- `depositFeeFixed`
- `depositFeeVariable`
- `withdrawalFeeFixed`
- `withdrawalFeeVariable`

With these new prop names I can also change `FeeConstants` to `FeeRates`.

**All proposed `Fee` changes reflected here: ...**

src/config.ts
vindard: Changed to `getGenericLimits` instead to match up with `getTransactionLimits` and `getUserLimits`

src/config.ts
vindard: **Done https://github.com/GaloyMoney/galoy/pull/393/commits/63ace3b08d777ecc1f3f8acd15c39bad97c35f63**

test/integration/02-user-wallet/02-send-onchain.spec.ts
vindard: I fixed this particular `FIXME` (https://github.com/GaloyMoney/galoy/pull/393/commits/0e48a4cdba5c3934716ae98b6209da4b146407f5) but I suspect the test is still broken because just a few lines above the `userWallet0.accountPath` seems to not exist on that object for example.

This test-fix is probably outside the scope of this PR though... or should I try to fix it here?

src/config.ts
nicolasburtey: unless I missed a recent changes, in our current implementation, deposits are `depositFeeVariable` and withdrawal are `withdrawalFeeFixed`

test/integration/02-user-wallet/02-send-onchain.spec.ts
dolcalmi: https://github.com/GaloyMoney/galoy/blob/main/src/schema.ts#L259

vindard: Ah ok my bad! I was going off a `console.warn` printed value of the `userWallet0` object in the tests lol

src/config.ts
vindard: Funny that we landed on the same names!

I didn't see that naming in my local version of things and couldn't find it in a Github search on here

image

test/integration/02-user-wallet/02-send-onchain.spec.ts
dolcalmi: Ok, I found it, we need to add +1 and use TransactionRestrictedError 

```js
import { TransactionRestrictedError } from "src/error"

....

const amount = userLimits.withdrawalLimit - outgoingSats + 1

await expect(userWallet0.onChainPay({ address, amount })).rejects.toThrow(TransactionRestrictedError)
```

src/config.ts
dolcalmi: @vindard I think you must fix "Conflicting files" (also dont trust github indexation 😅)

src/config.ts
```diff
@@ -21,20 +21,26 @@ try {
...
 export const yamlConfig = _.merge(defaultConfig, customConfig)
 
-export class TransactionLimits implements ITransactionLimits {
-  readonly config
-  readonly level
-
-  constructor({ level }) {
-    this.config = yamlConfig.limits
-    this.level = level
+export const getGenericLimits = (limitsConfig = yamlConfig.limits): GenericLimits => ({
```
dolcalmi, line 24: generic or default limits?

src/Lightning.ts
```diff
@@ -308,7 +308,7 @@ export const LightningMixin = (superclass) =>
             const lightningLoggerOnUs = lightningLogger.child({ onUs: true, fee: 0 })
 
             if (await this.user.limitHit({ on_us: true, amount: tokens })) {
-              const error = `Cannot transfer more than ${this.config.limits.onUsLimit()} sats in 24 hours`
+              const error = `Cannot transfer more than ${this.config.limits.onUsLimit} sats in 24 hours`
```
krtk6160, line 311: Perhaps we should refactor the naming and make it `this.config.limits.onUs` and get rid of the `Limit` suffix. The same can be done for the `withdrawalLimit` and for the generic limits as well since we're already accessing them by the `.limits.*` scope

src/OnChain.ts
```diff
@@ -209,7 +209,7 @@ export const OnChainMixin = (superclass) =>
           onchainLogger = onchainLogger.child({ onUs: false })
 
           if (!this.user.oldEnoughForWithdrawal) {
-            const error = `New accounts have to wait ${this.config.limits.oldEnoughForWithdrawalLimit()}h before withdrawing`
+            const error = `New accounts have to wait ${this.config.limits.oldEnoughForWithdrawalLimitHours}h before withdrawing`
```
krtk6160, line 212: Following on from the previous comment, this could become `oldEnoughForWithdrawalHours`

src/types.d.ts
```diff
@@ -8,12 +8,29 @@ type Primitive = string | boolean | number
...
+
+type UserLimitsArgs = { level: number; limitsConfig? }
+
+interface IUserLimits {
+  onUsLimit: number
+  withdrawalLimit: number
+}
+
+interface ITransactionLimits extends IUserLimits {
+  oldEnoughForWithdrawalLimit: number
```
krtk6160, line 32: If we're adding the time units as suffixes on some places (like in the GenericLimit types), we should add it here as well to avoid confusion.

test/integration/02-user-wallet/02-index.spec.ts
```diff
@@ -36,7 +36,9 @@ describe("UserWallet", () => {
     expect(userWallet2.user.oldEnoughForWithdrawal).toBeFalsy()
 
     // in 6 days:
-    const date = Date.now() + yamlConfig.limits.oldEnoughForWithdrawal - 60 * 60 * 1000
+    const genericLimits = getGenericLimits()
+    const date =
+      Date.now() + genericLimits.oldEnoughForWithdrawalMicroseconds - 60 * 60 * 1000
```
krtk6160, line 41: I know this magic number wasn't added in this PR, but we could have it a const and use it here and in #L55

krtk6160 (commented):
Few comments around naming

bodymindarts:
For graphql v2 we've introduced the `intraLedger` naming to replace the `onUs` concept. Perhaps you could also apply that renaming to here to the code that is effected.

src/config.ts
vindard: image

The whole set of limits are default limits, and then within that I've been thinking of modeling them as:
- `AccessLimits`
  Login attempts, phone code related limits etc.

- `UserLimits`
  Limits that depend on a `level`property

- `GenericLimits`
    Limits that _don't_ depend on any properties

- `TransactionLimits`
    `UserLimits` + `GenericLimits`

src/Lightning.ts
vindard: That makes sense, I'll check it out :+1:

src/config.ts
dolcalmi: so are they defaultTransactionLimits  ? @krtk6160 what do you think?

src/Lightning.ts
bodymindarts: Yeah lets prefer flat structure from the POV of the calling code and also ignore the layout in `default.yml` as that would require changes to the provisioning logic.

dolcalmi:
> For graphql v2 we've introduced the `intraLedger` naming to replace the `onUs` concept. Perhaps you could also apply that renaming to here to the code that is effected.

@vindard be careful with this change, some database operations are linked to it.

@bodymindarts I thought it was just a name for the graphql API, are we re-designing the galoy core based on the graphql design?

bodymindarts:
> @bodymindarts I thought it was just a name for the graphql API, are we re-designing the galoy core based on the graphql design?

Correct - so probably out of scope for this PR. Still I think moving to the `intraLedger` terminology makes sense for the core code base as well and thought we could use this PR to start introducing it.... but you're right there are some risks attached and probably more confusing to have both naming here....
So ignore for now @vindard

src/config.ts
krtk6160: Yeah default is kind of redundant here, I don't think we need to include `default` in the naming itself, because we don't reference any non-default values inside the code.

krtk6160: @bodymindarts What do you think about @vindard's proposed approach of modelling the limits?

dolcalmi: uhmm so Generic limits are the global transaction limits? (not related to the user?)

vindard:
:point_up_2: this [`force-push`](https://github.com/GaloyMoney/galoy/compare/bfb65f9ce4baba038f69000d2d7f28948472865c..a75d13703c2f69d759bf8a00a20fd3b96b2c6b4c) is only to rebase to tip of `main` to incorporate all of @samerbuna's recent folder structure changes

vindard:
> > @bodymindarts I thought it was just a name for the graphql API, are we re-designing the galoy core based on the graphql design?
> 
> Correct - so probably out of scope for this PR. Still I think moving to the `intraLedger` terminology makes sense for the core code base as well and thought we could use this PR to start introducing it.... but you're right there are some risks attached and probably more confusing to have both naming here....
> So ignore for now @vindard

Out of scope, gotcha :+1:

src/Lightning.ts
vindard: I'm just not getting into this refactor and there's one downside I'm seeing where renaming the property to something like `onUs` makes a bit trickier to grep for instances of it in the codebase since `onUS` (and later `intraLedger`) is more generically used in other places.

Is this is a tradeoff we'd like to consider, or should I go ahead with dropping `Limit` from places like these?

---

**Here's the current state of isolating the property for e.g.:**

image

test/integration/02-user-wallet/02-index.spec.ts
vindard: **_Done: https://github.com/GaloyMoney/galoy/pull/393/commits/1d926d4e534f4a0478158b683b28db92cdb6d773_**

src/types.d.ts
vindard: **_Done: https://github.com/GaloyMoney/galoy/pull/393/commits/b01e2283937c65b6fe45bf22442a97c6ba3c466d_**

src/OnChain.ts
vindard: **_Done: https://github.com/GaloyMoney/galoy/pull/393/commits/b01e2283937c65b6fe45bf22442a97c6ba3c466d_**

vindard:
:point_up_2: [`force push`](https://github.com/GaloyMoney/galoy/compare/1d926d4e534f4a0478158b683b28db92cdb6d773..be90463873ff2d36b64f55e03c72068638e79d95) to pull in #404 fixes

src/entrypoint/graphql-core-server.ts
vindard: Removing this planned test from the scope since we have planned broader graphql testing work

vindard:
:point_up_2: [`force push`](https://github.com/GaloyMoney/galoy/compare/4e7ed7d55cf931b2aaa254c963b0fa1c473d04aa..87673a66e2ec40510aed293eb65873dfd783c670) to pull in #411 fixes from tip of `main`

bodymindarts (approved):

vindard:
:point_up_2: [`force push`](https://github.com/GaloyMoney/galoy/compare/223153ff5d270baf79f711e615fb4e4c7eed4e1a..c280e695d38932243a5d7db75e7df1866d39e28d) to tip of `main` to resolve conflicts

src/core/types.d.ts
```diff
@@ -8,12 +8,29 @@ type Primitive = string | boolean | number
 // configs & constructors
 // TODO: clean up this section when "constructor typing" work is
```
dolcalmi, line 9: is this necessary?

src/config/app.ts
```diff
@@ -40,20 +42,26 @@ export const getLndParams = (): LndParams[] => {
...
-    this.level = level
+export const getGenericLimits = (limitsConfig = yamlConfig.limits): GenericLimits => ({
+  oldEnoughForWithdrawalHours: limitsConfig.oldEnoughForWithdrawal / MS_PER_HOUR,
+  oldEnoughForWithdrawalMicroseconds: limitsConfig.oldEnoughForWithdrawal,
+})
+
+export const getFeeRates = (feesConfig = yamlConfig.fees): FeeRates => ({
+  depositFeeVariable: feesConfig.deposit,
+  depositFeeFixed: 0,
+  withdrawFeeVariable: 0,
```
dolcalmi, line 53: @nicolasburtey Should this also be in the config?

dolcalmi (approved):

bodymindarts (approved):

PR #394, opened 2021-07-29
Cleanup liabilities accounting

nicolasburtey:


krtk6160 (approved):

src/migrations/20210726211641-simplify-accounting.ts
```diff
@@ -1,20 +1,21 @@
 module.exports = {
   async up(db) {
-    const ftx_txs = await db
-      .getCollection("medici_transactions")
+    const ftx_txs_custor = await db
```
dolcalmi, line 3: do you mean cursor? please use camel case :)

src/walletFactory.ts
```diff
@@ -37,7 +37,7 @@ export const WalletFromUsername = async ({
   return WalletFactory({ user, logger })
 }
 
-export const getFunderWallet = async ({ logger }) => {
-  const funder = await User.findOne({ username: yamlConfig.funder })
-  return WalletFactory({ user: funder, logger })
+export const WalletFromRole = async ({ logger, role }) => {
```
dolcalmi, line 40: we dont use it as a class so it is a function, please rename to `walletFromRole` or better `getWalletFromRole`

src/entrypoint/exporter.ts
```diff
@@ -1,12 +1,12 @@
 import express from "express"
 import client, { register } from "prom-client"
+import { WalletFromRole } from "src/walletFactory"
```
dolcalmi, line 3: I guess is because new ts build config that the IDE is using `src/` please use relative import

dolcalmi (changes requested):

dolcalmi:
@nicolasburtey please dont forget commit message rule

src/entrypoint/exporter.ts
nicolasburtey: what has been decided for the import (relative/absolute). can we write this down in the repo somewhere?

dolcalmi (approved):

PR #395, opened 2021-07-29
Fix wallet name attempt for production

jotapea:


jotapea:
@nicolasburtey I think this can fix the current issue in production.

PR #398, opened 2021-07-30
Ledger Refactor - 1st Stage USD Invoices

dolcalmi:
It was bigger than expected but I think is a good first step for USD invoices and DB/Medici migration.

**This PR requires a careful review**

- ledger.ts file was renamed to accounts.ts
- The only logic that was changed was the OnUs onchain payments. The change is to use the already implemented OnUsPayment method in transaction.ts
- All `MainBook.balance`, `MainBook.ledger`, `MainBook.entry`, `MainBook.void` were moved to ledger module (ledger folder)
- All `Transaction.updateMany/findOne/find` were moved to ledger module
- The Transaction schema was moved to ledger module. This guaranties the load order that is required for custom Transaction in medici and reduce dependency from medici implementation.
- Now the ledger module only has dependencies to medici and mongoose. All galoy (files, schemas, functions) dependencies were removed (UserWallet and User schema)
- The use of the ledger "instance" was changed to `ledger.<method>`. This helps the devs to know that they are interacting with the ledger and not with other component (easier to read)
- Accounts Paths logic was encapsulated in ledger module (in galoy classes/files only ledger instance is needed)
- Balance sheet was moved to main folder. It is not part of ledger responsibilities

I know that now is easier to implement the ledger as a class but I keep the current way to facilitate the revision of the changes. Now we have a common interface with the module and that change will be easier in the future

Finally, if you have any suggestions for a ledger methods names standard please let me know

Ref: [issue 350](https://github.com/GaloyMoney/galoy/issues/350#issuecomment-885293504)


dolcalmi:
Converted to draft until [PR 394](https://github.com/GaloyMoney/galoy/pull/394) merge

src/Lightning.ts
```diff
@@ -603,7 +599,7 @@ export const LightningMixin = (superclass) =>
             }
 
             // success
-            await Transaction.updateMany({ hash: id }, { pending: false })
+            await ledger.settlePayment(id)
```
nicolasburtey, line 602: I think we should either use pending or settle but not mix them. wdyt?

nicolasburtey (dismissed):

src/Lightning.ts
dolcalmi: what name do you prefer?  I couldn't use confirm because the same code is used for void

samerbuna (dismissed):
Excellent work! I see a lot of really great changes here.

I'm okaying the code changes but someone with more logic knowledge needs to give the final okay. @nicolasburtey @krtk6160

src/Lightning.ts
nicolasburtey: I'm not sure what is better. I like settlePayment, but I think in that case we should have `settle: true` instead of `pending: false`. but this implied a db updates. maybe in another 
PR?

maybe in the meantime `setPending(false)`? not sure really

src/schema.ts
```diff
@@ -2,10 +2,13 @@ import * as _ from "lodash"
 import * as mongoose from "mongoose"
 import { yamlConfig, levels } from "./config"
 import { NotFoundError } from "./error"
-import { accountPath } from "./ledger/ledger"
+import { accountPath } from "./ledger/accounts"
+import { Transaction } from "./ledger/schema"
 import { baseLogger } from "./logger"
 import { caseInsensitiveRegex } from "./utils"
 
+export { Transaction } from "./ledger/schema"
```
nicolasburtey, line 10: what is the goal import/exporting this class?

nicolasburtey: isn't this equivalent to `export Transaction`? because Transaction is already in the scope of this file with the import?

src/ledger/index.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * an accounting reminder:
+ * https://en.wikipedia.org/wiki/Double-entry_bookkeeping
+ */
+
+import * as accounts from "./accounts"
+import * as queries from "./query"
+import * as transactions from "./transaction"
+
+type LoadLedgerOptions = {
```
nicolasburtey, line 10: `LoadLedgerOptions` feels not accurate because it's not an option. maybe `LoadLedgerParams`?

src/ledger/index.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * an accounting reminder:
+ * https://en.wikipedia.org/wiki/Double-entry_bookkeeping
+ */
+
+import * as accounts from "./accounts"
```
nicolasburtey, line 6: `accounts` is plural for the filename. `query` and `transaction` are singular.

dolcalmi: I didnt want to change the `transaction` to transactios because will make it more difficult to review the changes... maybe we can do it in the renaming refactor or.. let me know and I will rename transaction and query

src/ledger/accounts.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  return id
+}
+
+let cacheDealerPath: string
+let cachebankOwnerPath: string
+
+const throwError = (account) => Promise.reject(`Invalid ${account}AccountPath`)
+let bankOwnerResolver = (): Promise<string> => throwError("bankOwner")
+let dealerResolver = (): Promise<string> => throwError("dealer")
```
nicolasburtey, line 43: struggling to understand what this is supposed to do, a promise that returns a Promise.reject?

src/ledger/index.ts
nicolasburtey: ok you just keep the original name here? let's keep it that way then for now

src/ledger/accounts.ts
dolcalmi: this is just to guarantee that there is an implementation outside the ledger like in https://github.com/GaloyMoney/galoy/blob/feat-ledger-refactor/src/mongodb.ts#L12 so if you did not pass the revolvers it will throw an exception

src/schema.ts
dolcalmi: fixed, changed to `export { Transaction }`.

src/ledger/index.ts
dolcalmi: Renamed

src/ledger/index.ts
dolcalmi: ok

src/Lightning.ts
```diff
@@ -718,10 +717,7 @@ export const LightningMixin = (superclass) =>
           }
 
           if (result.is_confirmed || result.is_failed) {
-            const resultPendingTrue = await Transaction.updateMany(
-              { _journal: payment._journal, pending: true },
-              { $set: { pending: false } },
-            )
+            const resultPendingTrue = await ledger.settlePayment(payment.hash)
```
nicolasburtey, line 720: maybe `ledger.settlePayment` should not be returning a mongoose/mongodb object?

ie: the parsing of `resultPendingTrue` with `resultPendingTrue.nModified` feels inappropriate now that we have a wrapper around mongodb

src/OnChain.ts
```diff
@@ -189,10 +187,17 @@ export const OnChainMixin = (superclass) =>
...
-                .commit()
+              const tx = await ledger.addOnUsPayment({
+                description: "",
+                sats,
+                metadata,
+                payerUser: this.user,
+                payeeUser,
+                memoPayer: memo,
+                shareMemoWithPayee: false,
+                lastPrice: UserWallet.lastPrice,
```
nicolasburtey, line 198: this is something we will eventually want to refactor (in later PR) not sure the price should be attached to the wallet any more the way it currently is

src/Lightning.ts
dolcalmi: Fixed. you are right, now it just return true or false if transactions were updated or not

src/OnChain.ts
dolcalmi: yes, maybe the signatures of that methods will change when we add usd balance/invoices

src/SpecterWallet.ts
```diff
@@ -282,13 +271,12 @@ export class SpecterWallet {
...
-    await MainBook.entry(memo)
-      .debit(lndAccountingPath, sats, { ...metadata })
-      .debit(bankOwnerPath, fee, { ...metadata })
-      .credit(bitcoindAccountingPath, sats + fee, { ...metadata })
-      .commit()
+    await ledger.addHotWalletPayment({
+      description: memo,
+      amount: sats,
+      fee,
+      metadata: { hash: txid },
```
nicolasburtey, line 278: maybe better to pass the hash directly?

nicolasburtey: instead of wrapping up in metadata

dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/398/commits/2f06c347efa47e9a003146ea9aa445f754720e90

src/entrypoint/trigger.ts
```diff
@@ -77,20 +72,15 @@ export async function onchainTransactionEventHandler(tx) {
       // transaction has been sent. and this events is trigger before
     }
 
-    await Transaction.updateMany({ hash: tx.id }, { pending: false })
+    await ledger.settlePayment(tx.id)
```
nicolasburtey, line 75: it might not be obvious that settlePayment is for onchain. we should make this more precise?

because it take a (transaction) hash but there could be also a settlePayment for (invoice) hash

dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/398/commits/f4520630c893dc525f8625a7f04e54901943fe17

src/ledger/transaction.ts
```diff
@@ -325,6 +325,14 @@ export const settlePayment = async (hash) => {
   return result.nModified > 0
 }
 
+export const settleLndPayment = (hash) => {
+  return settlePayment(hash)
+}
+
+export const settleOnchainPayment = (hash) => {
+  return settlePayment(hash)
```
nicolasburtey, line 333: I guess both lightning and onchain tx have a hash? let me dive back into it.

dolcalmi: I think its ok, if you see we have differnet methods for onchain and lightning... but let me know how is better for you

nicolasburtey: I didn't had in mind both lightning and onchain are both using a hash and so we could have a unique settlePayment method here. we can leave it as is.
I'm not sure this is a best practice overall, because those hash represent different things. one is hash of a pre-image, the other is hash of a onchain transaction

nicolasburtey (approved):

src/ledger/transaction.ts
dolcalmi: Yes, having consolidated methods in the ledger module helps to see things like this. I think we must keep this in mind if we move from db/medici

PR #411, opened 2021-08-02
Fix test structure

dolcalmi:
- Move test files to match src structure
- Fix update routing fees test to make tests pass (now is idempotent)


test/integration/services/lnd/utils.spec.ts
```diff
@@ -80,16 +83,37 @@ describe("lndUtils", () => {
...
+
+    const endDate = new Date(Date.now() - 864e5)
+    endDate.setUTCHours(0, 0, 0, 0)
+
+    const after = startDate.toISOString()
+    const before = endDate.toISOString()
+
+    const totalFees = (await getForwards({ lnd: lnd1, after, before })).forwards
+      .map((f) => Number(f.fee_mtokens))
+      .reduce((acc, val) => acc + val)
```
samerbuna, line 100: Minor nitpick but I'd generally avoid multiple scanning when possible: 

`.reduce((acc, val) => acc + Number(val.fee_mtokens))`

samerbuna (dismissed):
LGTM

test/integration/services/lnd/utils.spec.ts
dolcalmi: It is not possible do it in that way, in that way acc would be forward type so it crash when you try to sum the fee

dolcalmi: I'll check if setting initial value fixes that, but I tried that way and it didn't work

dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/411/commits/27007f331fb86f9c96147622aa2aefcbfdfe8686 with initial value. `.reduce((acc, val) => acc + Number(val.fee_mtokens), 0)`

samerbuna (approved):

test/integration/services/lnd/utils.spec.ts
```diff
@@ -80,16 +83,38 @@ describe("lndUtils", () => {
...
 
-    baseLogger.debug((await getForwards({ lnd: lnd1 })).forwards, "forwards")
-
     const date = Date.now() + 60 * 60 * 1000 * 24 * 2
     jest.spyOn(global.Date, "now").mockImplementation(() => new Date(date).valueOf())
 
+    const startDate = new Date(0)
+    startDate.setUTCHours(0, 0, 0, 0)
+
+    const endDate = new Date(Date.now() - 864e5)
```
nicolasburtey, line 92: what is `864e5`? magic number should probably be const

PR #415, opened 2021-08-03
Introduce sats threshold for memo sharing as spam workaround 

vindard:
Paired with @samerbuna on this.

### Open `TODO`s based on PR comments
- [x] ~Add test for changes~ https://github.com/GaloyMoney/galoy/pull/415/commits/d975f0ba95a026f6057db7dbf8537be380766c6b
- [x] ~Migration script for `memoFromPayer` changes in transactions (ledger) and schema~
  _decided to undo these changes and tackle separately later https://github.com/GaloyMoney/galoy/pull/415/commits/8b67331530af5d93c784a7b2ea765027eadfbd1d_
- [x] ~Confirm that csv rename changes does not break any integrations (e.g. in dashboard)~
  _decided to undo these changes and tackle separately later https://github.com/GaloyMoney/galoy/pull/415/commits/68b40b01de38339403cdb87cf7849f4fa83bd6e6_

---
### Follow on issues:
- exclude small quiz question payments from the spam filter somehow
- ~change `memoPayer` name to `memoFromPayer` name in codebase, via db migrations, and in any integrations that use the csv-exports~
  _may not be advisable for similar reasons to this: https://github.com/GaloyMoney/galoy/pull/393#issuecomment-891061935_

### DIFF
--- a/src/config/app.ts
+++ b/src/config/app.ts
@@ -10,6 +10,8 @@ export const defaultConfig = yaml.load(defaultContent)
 
 const MS_IN_HOUR = 60 * 60 * 1000
 
+export const MEMO_SHARING_SATS_THRESHOLD = 1000
+
 let customContent, customConfig
 
 try {
--- a/src/config/app.ts
+++ b/src/config/app.ts
@@ -10,8 +10,6 @@ export const defaultConfig = yaml.load(defaultContent)
 
 const MS_IN_HOUR = 60 * 60 * 1000
 
-export const MEMO_SHARING_SATS_THRESHOLD = 1000
-
 let customContent, customConfig
 
 try {
@@ -24,6 +22,8 @@ try {
 }
 
 export const yamlConfig = _.merge(defaultConfig, customConfig)
+
+export const MEMO_SHARING_SATS_THRESHOLD = yamlConfig.limits.memoSharingSatsThreshold
 
 export const getGaloyInstanceName = () => yamlConfig.name
 
--- a/src/core/ledger/schema.ts
+++ b/src/core/ledger/schema.ts
@@ -89,7 +89,7 @@ const transactionSchema = new Schema({
   // for onchain transactions.
   payee_addresses: [String],
 
-  memoFromPayer: String,
+  memoPayer: String,
 
   // not used for accounting but used for usd/sats equivalent
   usd: Number,
--- a/src/core/ledger/schema.ts
+++ b/src/core/ledger/schema.ts
@@ -89,7 +89,7 @@ const transactionSchema = new Schema({
   // for onchain transactions.
   payee_addresses: [String],
 
-  memoPayer: String,
+  memoFromPayer: String,
 
   // not used for accounting but used for usd/sats equivalent
   usd: Number,
--- a/src/core/ledger/transaction.ts
+++ b/src/core/ledger/transaction.ts
@@ -211,7 +211,7 @@ export const addOnUsPayment = async ({
   metadata,
   payerUser,
   payeeUser,
-  memoFromPayer,
+  memoPayer,
   shareMemoWithPayee,
   lastPrice,
 }: IAddTransactionOnUsPayment) => {
@@ -222,13 +222,13 @@ export const addOnUsPayment = async ({
   entry
     .credit(accountPath(payeeUser._id), sats * payeeUser.ratioBtc, {
       ...metadata,
-      memoPayer: shareMemoWithPayee ? memoFromPayer : null,
+      memoPayer: shareMemoWithPayee ? memoPayer : null,
       username: payerUser.username,
       currency: "BTC",
     })
     .debit(payerUser.accountPath, sats * payerUser.ratioBtc, {
       ...metadata,
-      memoPayer: memoFromPayer,
+      memoPayer: memoPayer,
       username: payeeUser.username,
       currency: "BTC",
     })
@@ -251,13 +251,13 @@ export const addOnUsPayment = async ({
     entry
       .credit(accountPath(payeeUser._id), usdEq * payeeUser.ratioUsd, {
         ...metadata,
-        memoPayer: shareMemoWithPayee ? memoFromPayer : null,
+        memoPayer: shareMemoWithPayee ? memoPayer : null,
         username: payerUser.username,
         currency: "USD",
       })
       .debit(payerUser.accountPath, usdEq * payerUser.ratioUsd, {
         ...metadata,
-        memoPayer: memoFromPayer,
+        memoPayer: memoPayer,
         username: payeeUser.username,
         currency: "USD",
       })
--- a/src/core/ledger/transaction.ts
+++ b/src/core/ledger/transaction.ts
@@ -222,13 +222,13 @@ export const addOnUsPayment = async ({
   entry
     .credit(accountPath(payeeUser._id), sats * payeeUser.ratioBtc, {
       ...metadata,
-      memoFromPayer: shareMemoWithPayee ? memoFromPayer : null,
+      memoPayer: shareMemoWithPayee ? memoFromPayer : null,
       username: payerUser.username,
       currency: "BTC",
     })
     .debit(payerUser.accountPath, sats * payerUser.ratioBtc, {
       ...metadata,
-      memoFromPayer,
+      memoPayer: memoFromPayer,
       username: payeeUser.username,
       currency: "BTC",
     })
@@ -251,13 +251,13 @@ export const addOnUsPayment = async ({
     entry
       .credit(accountPath(payeeUser._id), usdEq * payeeUser.ratioUsd, {
         ...metadata,
-        memoFromPayer: shareMemoWithPayee ? memoFromPayer : null,
+        memoPayer: shareMemoWithPayee ? memoFromPayer : null,
         username: payerUser.username,
         currency: "USD",
       })
       .debit(payerUser.accountPath, usdEq * payerUser.ratioUsd, {
         ...metadata,
-        memoFromPayer,
+        memoPayer: memoFromPayer,
         username: payeeUser.username,
         currency: "USD",
       })
--- a/src/core/ledger/transaction.ts
+++ b/src/core/ledger/transaction.ts
@@ -211,7 +211,7 @@ export const addOnUsPayment = async ({
   metadata,
   payerUser,
   payeeUser,
-  memoPayer,
+  memoFromPayer,
   shareMemoWithPayee,
   lastPrice,
 }: IAddTransactionOnUsPayment) => {
@@ -222,13 +222,13 @@ export const addOnUsPayment = async ({
   entry
     .credit(accountPath(payeeUser._id), sats * payeeUser.ratioBtc, {
       ...metadata,
-      memoPayer: shareMemoWithPayee ? memoPayer : null,
+      memoFromPayer: shareMemoWithPayee ? memoFromPayer : null,
       username: payerUser.username,
       currency: "BTC",
     })
     .debit(payerUser.accountPath, sats * payerUser.ratioBtc, {
       ...metadata,
-      memoPayer,
+      memoFromPayer,
       username: payeeUser.username,
       currency: "BTC",
     })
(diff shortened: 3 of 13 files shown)


test/integration/core/ledger/transaction.spec.ts
```diff
@@ -329,6 +330,51 @@ describe("on us payment", () => {
...
+      payerUser: payer,
+      payeeUser: payee,
+      lastPrice: UserWallet.lastPrice,
+      memoFromPayer: spamMsg,
+      shareMemoWithPayee: true,
+    })
+
+    for (const txn of entry.transactions) {
+      console.log(txn)
+      if (txn.credit > 0) {
```
bodymindarts, line 373: This seems like an implicit way to identify this?

src/core/ledger/transaction.types.d.ts
```diff
@@ -5,7 +5,7 @@ type IAddTransactionOnUsPayment = {
   metadata: Record<string, unknown>
   payerUser: UserType
   payeeUser: UserType
-  memoPayer?: string
+  memoFromPayer?: string
```
bodymindarts, line 8: Why nullable? Couldn't we default to empty String `""`?

src/core/ledger/transaction.ts
```diff
@@ -211,24 +212,27 @@ export const addOnUsPayment = async ({
...
   const entry = MainBook.entry(description)
 
+  // This is the current workaround to block spam memos
+  const memoSharingCheck = shareMemoWithPayee && sats >= MEMO_SHARING_SATS_THRESHOLD
+
   entry
     .credit(accountPath(payeeUser._id), sats * payeeUser.ratioBtc, {
       ...metadata,
-      memoPayer: shareMemoWithPayee ? memoPayer : null,
+      memoFromPayer: memoSharingCheck ? memoFromPayer : undefined,
```
bodymindarts, line 229: Why is `undefined` preferred to `null`?

bodymindarts (approved):

src/config/app.ts
```diff
@@ -10,6 +10,8 @@ export const defaultConfig = yaml.load(defaultContent)
 
 const MS_IN_HOUR = 60 * 60 * 1000
 
+export const MEMO_SHARING_SATS_THRESHOLD = 1000
```
dolcalmi, line 13: please use config not a const

src/core/ledger/transaction.ts
dolcalmi: this must not be part of the ledger, please do the validation previous to call this method

src/core/ledger/transaction.ts
```diff
@@ -251,13 +255,13 @@ export const addOnUsPayment = async ({
     entry
       .credit(accountPath(payeeUser._id), usdEq * payeeUser.ratioUsd, {
         ...metadata,
-        memoPayer: shareMemoWithPayee ? memoPayer : null,
+        memoFromPayer: memoSharingCheck ? memoFromPayer : undefined,
```
dolcalmi, line 258: same here

src/core/ledger/transaction.ts
dolcalmi: also I think is better to do the filter in the query, if we change the threshold later all the memos will be lost

test/integration/02-user-wallet/02-index.spec.ts
```diff
@@ -117,7 +117,7 @@ describe("UserWallet", () => {
 
   describe("getStringCsv", () => {
     const csvHeader =
-      "voided,approved,_id,accounts,credit,debit,_journal,book,unix,date,datetime,currency,username,type,hash,txid,fee,feeUsd,sats,usd,memo,memoPayer,meta,pending"
+      "voided,approved,_id,accounts,credit,debit,_journal,book,unix,date,datetime,currency,username,type,hash,txid,fee,feeUsd,sats,usd,memo,memoFromPayer,meta,pending"
```
dolcalmi, line 120: please check with @nicolasburtey if this breaks some integration (dashboard)

test/integration/core/ledger/transaction.spec.ts
```diff
@@ -329,6 +330,51 @@ describe("on us payment", () => {
     await expectBalance({ account: payee.accountPath, currency: "BTC", balance: 0 })
     await expectBalance({ account: payee.accountPath, currency: "USD", balance: 0.1 })
   })
+  it("does not share memo when below sats threshold", async () => {
```
dolcalmi, line 333: the tests must be part of addInvoice not in the ledger

dolcalmi (changes requested):

src/config/app.ts
bodymindarts: But even when moving it to config lets provide a default because needing to provision the value in config will delay rollout.

dolcalmi:
@vindard please keep in mind that this is a business rule and must not be part of the ledger implementation

src/config/app.ts
vindard: Moved the value to `default.yml` file https://github.com/GaloyMoney/galoy/pull/415/commits/7b1982ccb8eded01be8da5d50b08e83483dbd265

test/integration/02-user-wallet/02-index.spec.ts
vindard: This was a question I had as well when making the changes. I think it's probably safer to change the two string values of this back to the original for now.

src/core/ledger/transaction.types.d.ts
vindard: Changed here: https://github.com/GaloyMoney/galoy/pull/415/commits/a3336e2c900468af631065e04b76afcb255cd7e1

src/core/ledger/transaction.ts
vindard: > Why is `undefined` preferred to `null`?

@samerbuna preferred `undefined` in this scenario

nicolasburtey: > also I think is better to do the filter in the query, if we change the threshold later all the memos will be lost

If I'm understanding you correctly @dolcalmi: I would not necessarily do your suggestion here.

I guess it would be more robust to filter on the fly but it would basically change the memo after the fact as perceived by the user. I think we can start with either share/no share the memo at transaction time based on a sats threshold, and that way we make it more "immutable" entry. 

if we could filter it after the fact by changing the threshold, then memo could disappear from the user wallet and that is indeed not a good thing I believe

dolcalmi: query = get transactions in user wallet https://github.com/GaloyMoney/galoy/blob/main/src/core/user-wallet.ts#L119 (we are already doing some mapping there)

nicolasburtey:
@vindard @samerbuna the quiz section currently send few sats/questions. it may be needed to share the memo for this particular user. could be done in a later PR. it could be re-factored in a way where there are trusted user that avoid this rules.

src/core/ledger/transaction.types.d.ts
vindard: I undid the above change because it cascaded into changing a number of other functions where the parameter types would need to be redefined and function calls changed.

src/core/ledger/transaction.ts
vindard: The latest two commits revert the original `addInvoice` method and tries to implement it in that `user-wallet` instead, but I'm not sure that it comprehensively catches things (like can a transaction get to the end user through another pathway that isn't `getTransactions`?).

I'm leaning to undoing those last two commits and putting things back to the original state of the PR with the `addInvoice` changes.

test/integration/core/ledger/transaction.spec.ts
vindard: I think @samerbuna and I were assuming that if a transaction has a memo and a non-null `txn.credit` value then it is a "from someone" transaction and at that point the test checks that the expected `memoFromPayer` value is received based on the `sats` value originally passed into `addInvoice`.

src/core/ledger/schema.ts
```diff
@@ -89,7 +89,7 @@ const transactionSchema = new Schema({
   // for onchain transactions.
   payee_addresses: [String],
 
-  memoPayer: String,
+  memoFromPayer: String,
```
dolcalmi, line 92: if you do this change here you must do a DB migration too

src/core/ledger/transaction.ts
```diff
@@ -222,13 +222,13 @@ export const addOnUsPayment = async ({
   entry
     .credit(accountPath(payeeUser._id), sats * payeeUser.ratioBtc, {
       ...metadata,
-      memoPayer: shareMemoWithPayee ? memoPayer : null,
+      memoFromPayer: shareMemoWithPayee ? memoFromPayer : null,
```
dolcalmi, line 225: same for all ledger metadata changes

src/core/lightning/index.ts
```diff
@@ -369,7 +369,7 @@ export const LightningMixin = (superclass) =>
                 metadata,
                 payerUser: this.user,
                 payeeUser,
-                memoPayer,
+                memoFromPayer,
```
dolcalmi, line 372: if you like you can keep the change at this level but not in medici metadata

src/core/on-chain/index.ts
```diff
@@ -195,7 +195,7 @@ export const OnChainMixin = (superclass) =>
                 metadata,
                 payerUser: this.user,
                 payeeUser,
-                memoPayer: memo,
+                memoFromPayer: memo,
```
dolcalmi, line 198: same here

src/core/user-wallet.ts
```diff
@@ -105,18 +106,28 @@ export abstract class UserWallet {
...
+
       const memoUsername = item.username
-        ? amount > 0
+        ? isCredit
           ? `from ${item.username}`
           : `to ${item.username}`
         : null
 
+      const memoFromPayer = item.memoFromPayer
+        ? isDebit || isValidCreditMemo
```
dolcalmi, line 120: The validation filter is for the receiver, please check.

dolcalmi (changes requested):

src/core/user-wallet.ts
vindard: Yup so the intention here is for the filter to happen in `isValidCreditMemo` and this conditional is saying that we _will add_ the memo if it is either a "debit" or a "valid credit"

src/core/ledger/transaction.ts
vindard: **Update:** Confirmed with @dolcalmi that `getTransactions` is the only entrypoint. Based on discussion on Dev Call we decided to go ahead with these current changes

vindard:
Ready to review again. The only outstanding thing that I'm planning to tackle in a follow-on PR is adding logic to not exclude the memos from small txns for quiz questions.

test/integration/02-user-wallet/02-receive-lightning.spec.ts
```diff
@@ -64,4 +70,33 @@ describe("UserWallet - Lightning", () => {
...
+    await pay({ lnd: lndOutside1, request: invoice })
+    expect(await userWallet1.updatePendingInvoice({ hash })).toBeTruthy()
+
+    // check that spam memo is persisted to database
+    const dbTx = await ledger.getTransactionByHash(hash)
+    expect(dbTx.memo).toBe(memo)
+
+    // check that spam memo is filtered from transaction description
+    const txns = await userWallet1.getTransactions()
+    const spamTxn = txns.filter((txn) => txn.hash === hash)[0]
```
dolcalmi, line 94: you can use `txns.find` to avoid `[0]`

vindard: Didn't know of this method, thanks for the tip :+1: https://github.com/GaloyMoney/galoy/pull/415/commits/9c698c51794249212bc63967ddff5c7f57a85d64

src/core/ledger/transaction.ts
```diff
@@ -228,7 +228,7 @@ export const addOnUsPayment = async ({
     })
     .debit(payerUser.accountPath, sats * payerUser.ratioBtc, {
       ...metadata,
-      memoPayer,
+      memoPayer: memoPayer,
```
dolcalmi, line 231: I dont understand this change

src/core/ledger/transaction.ts
```diff
@@ -257,7 +257,7 @@ export const addOnUsPayment = async ({
       })
       .debit(payerUser.accountPath, usdEq * payerUser.ratioUsd, {
         ...metadata,
-        memoPayer,
+        memoPayer: memoPayer,
```
dolcalmi, line 260: same here

dolcalmi (approved):
just minor changes, LGTM

src/core/ledger/transaction.ts
vindard: The result of one commit changing it to `memoPayer: memoFromPayer` and another changing it back via a search-and-replace 😂

bodymindarts (approved):
Looks good but code checks are failing due to logging sensitive information. Don't think its related to your PR though.

dolcalmi:
> Looks good but code checks are failing due to logging sensitive information. Don't think its related to your PR though.

just did a PR solving a production issue and this warning https://github.com/GaloyMoney/galoy/pull/420

PR #416, opened 2021-08-04
Alt lnpage

bodymindarts:


src/core/payment-status.ts
```diff
@@ -0,0 +1,37 @@
...
+import { toTypedError } from "@domain/utils"
+
+export const MakePaymentStatusChecker = ({
+  paymentRequest,
+}): Result<PaymentStatusChecker, LnInvoiceDecodeError> => {
+  const decodedInvoice = decodeInvoice(paymentRequest)
+  const getStatus = (): ResultAsync<PaymentStatus, PaymentStatusError> => {
+    return decodedInvoice
+      .asyncAndThen(({ paymentHash }) => findUserInvoice(paymentHash))
+      .map((invoice) => ({
```
nicolasburtey, line 13: Is there a need for an array here?

bodymindarts: This `.map` function is being called on the [`ResultAsync`](https://github.com/supermacro/neverthrow#resultasyncmap-method-1) type - its not an array.

src/core/payment-status.ts
```diff
@@ -10,10 +10,7 @@ export const MakePaymentStatusChecker = ({
   const getStatus = (): ResultAsync<PaymentStatus, PaymentStatusError> => {
     return decodedInvoice
```
nicolasburtey: shouldn't it be plural `decodeInvoices`?

bodymindarts: this use case only operates on a _single_ invoice

PR #420, opened 2021-08-05
Fix delete expired invoices

dolcalmi:
- Fix deleteExpiredInvoiceUser query but maybe will not run because `updateUsersPendingPayment` hang in cron
- Fix log security warnings

### DIFF
--- a/src/services/lnd/utils.ts
+++ b/src/services/lnd/utils.ts
@@ -35,7 +35,7 @@ const MS_PER_DAY = 864e5
 export const deleteExpiredInvoiceUser = () => {
   // this should be longer than the invoice validity time
 
-  const delta = 30 // days
+  const delta = 90 // days
 
   const date = new Date()
   date.setDate(date.getDate() - delta)
--- a/src/services/lnd/utils.ts
+++ b/src/services/lnd/utils.ts
@@ -32,7 +32,7 @@ import { FEECAP, FEEMIN, params } from "./auth"
 // milliseconds in a day
 const MS_PER_DAY = 864e5
 
-export const deleteExpiredInvoiceUser = async () => {
+export const deleteExpiredInvoiceUser = () => {
   // this should be longer than the invoice validity time
 
   const delta = 30 // days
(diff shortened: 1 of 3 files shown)


src/services/lnd/utils.ts
```diff
@@ -41,7 +41,7 @@ export const deleteExpiredInvoiceUser = async () => {
   date.setDate(date.getDate() - delta)
   // TODO: assert: only paid: true invoice should be remaining here
   // other invoiceUser should be deleted alongside deletion of lnd invoice
-  await InvoiceUser.deleteMany({ timestamp: { lt: date } })
```
dolcalmi, line 44: seems this never worked, are we sure that we want to delete all invoices (pending and paid)?

src/services/mongodb/index.ts
```diff
@@ -38,7 +38,7 @@ export const setupMongoConnection = async () => {
       useFindAndModify: false,
     })
   } catch (err) {
-    baseLogger.fatal({ err, path }, `error connecting to mongodb`)
+    baseLogger.fatal({ err, user, address, db }, `error connecting to mongodb`)
```
krtk6160, line 41: Is this just so that we don't expose the password in the logs?

test/integration/services/lnd/utils.spec.ts
```diff
@@ -120,4 +124,9 @@ describe("lndUtils", () => {
 
     expect((endBalance - initBalance) * 1000).toBeCloseTo(totalFees, 0)
   })
+
+  it("runs deleteExpiredInvoiceUser without throw an exception", async () => {
+    const result = await deleteExpiredInvoiceUser()
+    expect(result.deletedCount).toBe(0)
```
krtk6160, line 130: Maybe we can mock the date and test whether it actually deletes the invoices?

src/services/mongodb/index.ts
dolcalmi: Yes, it is giving problems with the "GitGuardian Security Checks"

krtk6160 (commented):
Just one comment regarding the new test, otherwise LGTM

test/integration/services/lnd/utils.spec.ts
dolcalmi: Fixed

Issue #229, opened 2021-06-14
Fix cronjob


krtk6160:
It's currently taking 1hr+ on some deployments. One way is to move some functions out of cron.ts into another file and run that as a separate cronjob.

nicolasburtey:
I don't know if this will have any real impact. I think only `updateUsersPendingPayment` is taking a long time. the other tasks should take less than a second.

PR #421, opened 2021-08-06
Fix cronjob - update pending lightning transactions

dolcalmi:
- Solves performance queries for lightning transactions (invoices and payments)
- Adds minor performance query filter for Onchain transactions
- Onchain refactor will be in its own PR

REF: https://github.com/GaloyMoney/galoy/issues/229

### DIFF
--- a/src/core/balance-sheet.ts
+++ b/src/core/balance-sheet.ts
@@ -1,29 +1,53 @@
 import { getBalance as getBitcoindBalance } from "@services/bitcoind"
 import { lndsBalances } from "@services/lnd/utils"
 import { baseLogger } from "@services/logger"
-import { User } from "@services/mongoose/schema"
+import { InvoiceUser, User } from "@services/mongoose/schema"
 import { ledger } from "@services/mongodb"
 
 import { WalletFactory } from "./wallet-factory"
 
 const logger = baseLogger.child({ module: "balanceSheet" })
 
+export const updatePendingLightningTransactions = async () => {
+  let user, userWallet
+
+  // select distinct user ids from pending invoices
+  const usersWithPendingInvoices = InvoiceUser.aggregate([
+    { $match: { paid: false } },
+    { $group: { _id: "$uid" } },
+  ])
+    .cursor({ batchSize: 100 })
+    .exec()
+
+  for await (const { _id } of usersWithPendingInvoices) {
+    logger.trace("updating pending invoices for user %o", _id)
+    user = await User.findOne({ _id })
+    userWallet = await WalletFactory({ user, logger })
+    await userWallet.updatePendingInvoices()
+  }
+
+  const accounts = ledger.getAccountsWithPendingTransactions({ type: "payment" })
+  for await (const account of accounts) {
+    logger.trace("updating pending payments for account %o", account)
+    user = await User.findOne({ _id: ledger.resolveAccountId(account) })
+    userWallet = await WalletFactory({ user, logger })
+    await userWallet.updatePendingPayments()
+  }
+}
+
 export const updateUsersPendingPayment = async ({
   onchainOnly,
 }: { onchainOnly?: boolean } = {}) => {
+  if (!onchainOnly) {
+    await updatePendingLightningTransactions()
+  }
+
   let userWallet
-
-  for await (const user of User.find({})) {
-    logger.trace("updating user %o", user._id)
-
-    // A better approach would be to just loop over pending: true invoice/payment
+  const users = User.find({}).cursor({ batchSize: 100 })
+  for await (const user of users) {
+    logger.trace("updating onchain receipt for user %o", user._id)
     userWallet = await WalletFactory({ user, logger })
-
-    if (onchainOnly) {
-      await userWallet.updateOnchainReceipt()
-    } else {
-      await userWallet.updatePending()
-    }
+    await userWallet.updateOnchainReceipt()
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/core/balance-sheet.ts
```diff
@@ -1,29 +1,55 @@
...
+export const updatePendingLightningTransactions = async () => {
+  let user, userWallet
+
+  // select distinct user ids from pending invoices
+  const usersWithPendingInvoices = InvoiceUser.aggregate([
+    { $match: { paid: false } },
+    { $group: { _id: "$uid" } },
+  ])
+    .cursor({ batchSize: 100 })
+    .exec()
```
dolcalmi, line 20: this query is just distinct

src/core/balance-sheet.ts
```diff
@@ -1,29 +1,55 @@
...
+
+  for await (const { _id } of usersWithPendingInvoices) {
+    logger.trace("updating pending invoices for user %o", _id)
+    user = await User.findOne({ _id })
+    userWallet = await WalletFactory({ user, logger })
+    await userWallet.updatePendingInvoices()
+  }
+
+  const accounts = ledger.getAccountsWithPendingTransactions({ type: "payment" })
```
dolcalmi, line 29: `ledger.getAccountsWithPendingTransactions` is a [generator function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*), internally use a cursor

src/core/ledger/query.ts
```diff
@@ -44,6 +44,19 @@ export const getAccountTransactions = async (account: string, query = {}) => {
   return { query: params, results, total }
 }
 
+export async function* getAccountsWithPendingTransactions(query = {}) {
+  const transactions = Transaction.aggregate([
+    { $match: { "pending": true, "account_path.0": liabilitiesMainAccount, ...query } },
+    { $group: { _id: "$accounts" } },
+  ])
+    .cursor({ batchSize: 100 })
+    .exec()
```
dolcalmi, line 53: this query is just a distinct over liabilities

src/core/balance-sheet.ts
```diff
@@ -1,29 +1,55 @@
...
-    logger.trace("updating user %o", user._id)
+  if (!onchainOnly) {
+    await updatePendingLightningTransactions()
+  }
 
-    // A better approach would be to just loop over pending: true invoice/payment
+  let userWallet
+  const users = User.find({ onchain: { $exists: true, $not: { $size: 0 } } }).cursor({
+    batchSize: 100,
+  })
```
dolcalmi, line 48: @krtk6160 could you please check this query in production? (just to validate if it return less than total Users)

krtk6160: It returns only slightly (1.3%) less than total user count

src/core/ledger/query.ts
```diff
@@ -44,6 +44,19 @@ export const getAccountTransactions = async (account: string, query = {}) => {
   return { query: params, results, total }
 }
 
+export async function* getAccountsWithPendingTransactions(query = {}) {
```
vindard, line 47: Oh this is cool, I didn't realise JS had generator functions too :+1:

src/core/balance-sheet.ts
dolcalmi: :) better than nothing

src/core/balance-sheet.ts
nicolasburtey: I'm not sure to understand what you mean here?

dolcalmi: This is like `select distinct uid from invoiceusers where paid = false`. This query only get the users that have pending invoices

vindard: I had to [look this up](https://codeburst.io/javascript-array-distinct-5edc93501dc4) as well ... I think it means there are no duplicate values in the array(?) that's returned

nicolasburtey: yup ok thanks for clarifying

src/core/balance-sheet.ts
```diff
@@ -1,29 +1,55 @@
...
+
+  // select distinct user ids from pending invoices
+  const usersWithPendingInvoices = InvoiceUser.aggregate([
+    { $match: { paid: false } },
+    { $group: { _id: "$uid" } },
+  ])
+    .cursor({ batchSize: 100 })
+    .exec()
+
+  for await (const { _id } of usersWithPendingInvoices) {
```
nicolasburtey, line 22: does this loop one by one? or 100 by 100 in parrallel? 
I guess it's one by one, but only have 100 InvoiceUser in memory at one particular time?

nicolasburtey: we could speed this up by making those call in parrallel maybe with batch of 5. most of the underlying logic here is waiting for network calls on external service.

dolcalmi: this one `I guess it's one by one, but only have 100 InvoiceUser in memory at one particular time`

src/core/balance-sheet.ts
nicolasburtey: @jrbuhl93 there are some optimization we could do here. only generate a new onchain address is a new user go to the receive bitcoin and swipe right to see the onchain address.

nicolasburtey: it would result in a slightly worse tradeoff because the onchain address would need to be fetch at this time, but we would drastically reduce the number of onchain addresses we have to look up for in the long run

src/core/balance-sheet.ts
nicolasburtey: maybe let's do that in another PR

dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/421/commits/1b4e48a0a66955743533e8717949f8df580aee5d

Issue #363, opened 2021-07-21
sats threshold to share memo


nicolasburtey:

vindard:
We need to apply additional spam checks to the ones implemented in #415 to allow small value transactions from certain senders through the memo filter. The proposed approach is to mark any senders with the `funder` role as valid.

## Feature implementation proposal

### Problem
We need to improve the memo spam filter to allow small value transactions below the `MEMO_SATS_THRESHOLD` value to pass through the filter. Currently all transactions with values below that threshold get filtered. This filter gets applied in the translateDescription method in `tx-history.ts` by call it on each `LedgerTransaction` 'description' as transactions pass through the 'fromLedger' function.

On a `LedgerTransaction` object, I have the 'username' property of a user that I can optionally use to lookup the user's role.

'username' properties are unique for users.


### Proposed solutions
**Option 1**
I can add a check for the payer user's role to see if they are a valid role to allow through the filter. This means adding a db call for each parsed `LedgerTransaction`.

**Option 2**
I can fetch all the usernames from the database with the approved role and then check if the username matches the one on each `LedgerTransaction` as I parse the array.


### Proposed implementations

Spam filtering currently happens internally in the 'fromLedger' method inside the tx-history.ts domain file. It gets applied as a 'translateDescription' function on the 'description' property as transactions are parsed.

**Option 1**
  - implement a new `UsersRepository` method called `addPropertyToTransactions` that:
    - takes a `LedgerTransaction[]` array and a list of property names, and adds the properties to each transaction by querying for the user object in the db using the 'username' property on each `LedgerTransaction`
    - takes an optional filter function that itself takes a `LedgerTransaction` and returns True/False. We can use this to only make db queries for users on transactions that pass the filter.
    - optionally has an ephemeral object to track users already queried to avoid unnecessary db round trips
    - _Note: I believe this ends up working out to a decoupled JOIN on the Users and Transactions tables (collections)_

  - pass `ledgerTransactions` fetched from the ledger through this `addPropertyToTransactions` method before sending that object forward to `WalletTransactionHistory.fromLedger` in `getTransactionsForWallet` function
    - _Note: This potentially changes the type of object that the `fromLedger` method takes as an arg but the return should remain unchanged_

**Option 2**
  - implement a new UsersRepository method that finds all users for a given role
  - call the method and map over the result to isolate just the usernames into an array
[...]

PR #423, opened 2021-08-06
Add spam memo exception for valid payers

vindard:
## Description

This PR builds on #415 and [this comment](https://github.com/GaloyMoney/galoy/pull/415#issuecomment-892184632) about valid transactions that could come in below the 'memo spam threshold'

### Approach

To solve this, @dolcalmi helped me to identify that the user with the `"funder"` role is usually the one sending these valid low-sat transactions.

Our approach was to add an additional step if a transaction is flagged below the memo threshold. We do a db call to lookup the username of the user that has the `role` property of `"funder`, and then compare this retrieved username to the `from` user's `username property gotten from the `transaction` object.

#### Implementation details

An additional db call is added asynchronously in the `getTransactions` method on the `UserWallet` object to retrieve the user with the `role: "funder"` property.

An earlier approach was tried where the additional lookup was included in the `memoSpamFilter` function itself only and only triggered (and then cached for that run of the method) if a txn amount was found _below the memo sats threshold_ to reduce the number of times we would hit the db.

This added more complexity and seemed to be mixing concerns to much, so I abandoned it given our upcoming work to clean things up and separate different concerns.

#### Potential future improvements

- To avoid this extra db call, we can also alternatively add an appropriate additional property (like `user role`) to transactions when they are created. **This would require a migration.**

- While implementing I got stuck with a circular dependency bug I introduced here that was fairly obfuscated for quite a while and didn't cause tests to fail until quite late in the run.
  
  **_Are there any linting or check we can add to catch circular deps at compile time?_**  (cc @samerbuna)
  
  _For e.g. I saw some folks use things like [madge](https://www.npmjs.com/package/madge) for this_

---
## Questions
1. Is the `username` property unique for users? (I assume that it is)

### DIFF
--- a/src/core/user-wallet.ts
+++ b/src/core/user-wallet.ts
@@ -103,6 +103,7 @@ export abstract class UserWallet {
 
   async getTransactions(): Promise<Array<ITransaction>> {
     const rawTransactions = await this.getRawTransactions()
+    const lightningFundingUser = await User.findOne({ role: "funder" })
 
     return rawTransactions.map((item) => {
       const amount = item.credit - item.debit
@@ -115,9 +116,17 @@ export abstract class UserWallet {
         : null
 
       const memoSpamFilter = (memoString) => {
-        const creditBelowThreshold = isCredit && amount < MEMO_SHARING_SATS_THRESHOLD
-        const isValidCreditMemo = !creditBelowThreshold
-        return memoString ? (isValidCreditMemo ? memoString : null) : null
+        const iscreditBelowThreshold = isCredit && amount < MEMO_SHARING_SATS_THRESHOLD
+        const funderPayer =
+          item.username && item.username == lightningFundingUser?.username
+
+        const isCreditWithValidMemo = !iscreditBelowThreshold || funderPayer
+
+        return memoString
+          ? !isCredit || isCreditWithValidMemo
+            ? memoString
+            : null
+          : null
       }
       const memoPayer = memoSpamFilter(item.memoPayer)
       const memo = memoSpamFilter(item.memo)
--- a/src/core/user-wallet.ts
+++ b/src/core/user-wallet.ts
@@ -102,8 +102,12 @@ export abstract class UserWallet {
   }
 
   async getTransactions(): Promise<Array<ITransaction>> {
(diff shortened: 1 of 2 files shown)


dolcalmi:
> 1. Is the `username` property unique for users? (I assume that it is)
@vindard Yes, it is https://github.com/GaloyMoney/galoy/blob/main/src/services/mongoose/schema.ts#L161

src/core/user-wallet.ts
```diff
@@ -102,22 +102,36 @@ export abstract class UserWallet {
...
         : null
 
-      const memoSpamFilter = (memoString) =>
-        memoString ? (isDebit || isValidCreditMemo ? memoString : null) : null
+      const memoSpamFilter = (memoString) => {
+        const creditBelowThreshold = isCredit && amount < MEMO_SHARING_SATS_THRESHOLD
+        const funderPayer =
+          item.username && item.username == lightningFundingUser?.username
+
+        const isCreditWithValidMemo = !creditBelowThreshold || funderPayer
```
dolcalmi, line 127: I think maybe we can have a rule for boolean variables, For ex: must have `is` or `has` prefix so in that way are easier to read and identify the type

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -187,12 +203,30 @@ describe("UserWallet - Lightning Pay", () => {
...
+      memoFunderBelowThreshold,
+    )
+    expect(transaction4FunderBelow).toHaveProperty("username", userWallet0.user.username)
+    expect(transaction4FunderBelow).toHaveProperty(
+      "description",
+      memoFunderBelowThreshold,
+    )
+
     // check contacts being added
+    // TODO: decide if we want to exclude spam senders from contacts
```
dolcalmi, line 229: this is a TODO but not in test, seems to be more a question for #dev channel in slack. Lets try to keep the code as clean as possible.

dolcalmi (changes requested):
Just minor change(s). LGTM

vindard:
This PR will be closed and the new functionality should be implemented based on this proposal: https://github.com/GaloyMoney/galoy/issues/363#issuecomment-900305025

PR #426, opened 2021-08-06
Lightning service

bodymindarts:
## Description

This PR begins the refactoring towards a well-typed and decoupled code-base. @vindard and I have begun by tackling the `addInvoice` use case. The goal is to separate concerns within the method more cleanly into our new planned file layout, and to add types for as many of the elements used as practical. The eventual result will be to have strong runtime guarantees by relying more heavily on the TS compiler, and to get the `addInvoice` method to the point where it can never throw an error.

This work only changes the internal implementation of the `addInvoice` method, and leaves all the same function signatures and returns intact for now.

### Implementation

This PR specifically sets up type definitions and re-implementations for _invoice handling code via lnd_. This is in prep for using the outputs of this code to feed into `InvoiceUser` instantiation which will be tackled next to round out the `addInvoice` refactor.

Specifically, the following were done:

- Fully typed **"LN Invoice"** definitions, accompanying methods that operate on an invoice, unit tests

- Fully typed **"Invoice expiration"** value object that internally has `string` and `Date` representations of its value, method for generating an expiration value

- **"Lightning service"** interface with a maker method that returns implemented service with Lightning methods on it; `registerInvoice` wrapper created for `createInvoice` method

- **Error** types for all error returns so far; **"Unsafe throw"** wrapper that can throw errors for the parts of the function not converted as yet to be able to "never throw"


src/domain/ln-invoice.ts
```diff
@@ -0,0 +1,43 @@
...
+): Result<LnInvoice, LnInvoiceDecodeError> => {
+  let paymentHash: PaymentHash | null = null
+  let paymentSecret: PaymentSecret | null = null
+  const paymentRequest: EncodedPaymentRequest = {
+    inner: bolt11EncodedInvoice,
+  }
+
+  return safeDecode(bolt11EncodedInvoice).andThen(({ tags }) => {
+    tags.forEach((tag) => {
+      if (tag.tagName === "payment_hash") {
```
vindard, line 23: For similar cases like this, could an option be to use a `switch` statement with error returns to remove the "appearance" of nesting?

Example:
```ts
tags.forEach((tag) => {
  const tagError = typeof tag.data != "string"
  switch (tag.tagName) {
    case "payment_hash":
      if (tagError) {
        return err(toLnInvoiceDecodeError({ message: "Irregular payment_hash" }))
      }
      paymentHash = { inner: tag.data }
      break

    case "payment_secret":
      if (tagError) {
        return err(toLnInvoiceDecodeError({ message: "Irregular payment_secret" }))
      }
      paymentSecret = { inner: tag.data }
      break
  }
  if (paymentHash == null || paymentSecret == null) {
    ...
```

bodymindarts: Yup that would be cleaner. Feel free to add a commit

src/domain/ln-invoice.ts
```diff
@@ -0,0 +1,45 @@
+import { ok, err, Result } from "neverthrow"
+import lightningPayReq from "bolt11"
+import { toTypedError } from "./utils"
+
+const toLnInvoiceDecodeError = toTypedError<LnInvoiceDecodeErrorType>({
+  unknownMessage: "Unknown decoding error",
+  _type: "LnInoviceDecodeError",
```
vindard, line 7: Typo here

src/domain/lightning-service.types.d.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+type RegisteredInvoice = {
+  invoice: LnInvoice
+}
+
+interface ILightningService {
+  registerInvoice(
+    registerInvoiceArgs: RegisterInvoiceArgs,
+  ): ResultAsync<RegisteredInvoice, LightningServiceError>
```
vindard, line 19: Making a note to potentially add this line in a declaration file somewhere so that instances like these have access to `ResultAsync`

```ts
type ResultAsync<T, E> = import("neverthrow").ResultAsync<T, E>
```
---
image

package.json
```diff
@@ -47,6 +47,7 @@
     "bip66": "^1.1.5",
     "bitcoin-core": "^3.0.0",
     "bluebird": "^3.7.2",
+    "bolt11": "^1.3.2",
```
dolcalmi, line 50: which is the difference with parsePaymentRequest from "invoices"? if we will use this then maybe we must remove "invoices"

src/core/lightning/index.ts
```diff
@@ -102,31 +83,18 @@ export const LightningMixin = (superclass) =>
         throw new LndOfflineError("no active lnd to create an invoice")
       }
 
-      try {
-        input = {
-          lnd,
-          tokens: value,
-          expires_at,
-        }
+      const lndService = MakeLndService(lnd)
```
dolcalmi, line 86: if we will use this change then lightning service must manage the lnd nodes (this layer must not handle auth lnds initialization)

src/core/lightning/index.ts
```diff
@@ -102,31 +83,18 @@ export const LightningMixin = (superclass) =>
...
+      const result: ResultAsync<RegisteredInvoice, LightningServiceError> =
+        lndService.registerInvoice({
+          description: typeof memo == "string" ? memo : "",
```
dolcalmi, line 89: why do we need to check type here? the type must be validated in the param (IAddInvoiceRequest)

src/core/lightning/index.ts
```diff
@@ -102,31 +83,18 @@ export const LightningMixin = (superclass) =>
...
+          satoshis: toSats(value),
```
dolcalmi, line 90: is not value in sats?

src/domain/primitives/btc.ts
```diff
@@ -0,0 +1,5 @@
+export const toSats = (amount: number): Satoshis => {
+  return {
+    inner: amount,
+  }
+}
```
dolcalmi, line 5: ohh now I see, the name is confusing, with the name I expect some kind of operation and not just a type parser

src/domain/primitives/tx_denomination_currency.types.d.ts
```diff
@@ -0,0 +1 @@
+type TxDenominationCurrency = "USD" | "BTC"
```
dolcalmi, line 1: this cant be a type, fiat currencies cant be hardcoded (this was a "intiial" implementation from nicolas but never was used)

src/domain/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+export const toTypedError = <ErrorType>({
+  unknownMessage,
+  _type,
+}): ErrorConverter<ErrorType> => {
+  return (error: any): ErrorWithMessage<ErrorType> => {
+    if (error.message && typeof error.message == "string") {
+      return { _type, message: error.message as string }
+    } else {
+      return { _type, message: unknownMessage }
+    }
```
dolcalmi, line 12: else block is not needed

src/domain/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+}
+
+export const unsafeThrowErrAsync = async <T, EType>(
```
dolcalmi, line 16: do we need this to make it work with current code? I thought the idea was to get rid of the exceptions.

src/core/lightning/index.ts
dolcalmi: ok.. just found it is a type parser

src/domain/invoice-expiration.ts
```diff
@@ -0,0 +1,20 @@
+import moment from "moment"
+
+const DEFAULT_EXPIRATIONS = {
+  BTC: { value: 1, unit: "days" as moment.unitOfTime.DurationConstructor },
+  USD: { value: 2, unit: "minutes" as moment.unitOfTime.DurationConstructor },
+}
+
+export const invoiceExpirationForCurrency = (
```
dolcalmi, line 8: I think this was just a proof of concept, what you really need to do is query the dealer an get a quote with the time (available) and value in sats for the give amount in fiat currency

src/services/lnd/service.ts
```diff
@@ -0,0 +1,41 @@
+import { decodeInvoice } from "@domain/ln-invoice"
```
dolcalmi, line 1: if this file is an entry point for the whole service then must be exposed in the index.ts file of this folder

src/domain/utils.ts
bodymindarts: Yes we have 2 worlds the "old" try catch code and the "new" `Result` world. We are building the new world and integrating it into the old world. This is the translation back into the try/catch world. The goal is to never use this long term. Its just an indicator of how far the 'new' implementation has come.

package.json
bodymindarts: Its a different library. I wasn't aware that we were already using `invoices`...

src/core/lightning/index.ts
bodymindarts: yes lnd specific auth logic needs to migrate to the lnd service implementation - agreed

src/domain/invoice-expiration.ts
bodymindarts: yes this will be a service call

vindard (approved):

PR #428, opened 2021-08-08
Process only users from block transactions

dolcalmi:
- Change cron workers for updateOnchainReceipt from 10 to 3
- Add new handler to subscribeToBlocks
- Fix query issues in onchainTransactionEventHandler

### DIFF
--- a/src/config/app.ts
+++ b/src/config/app.ts
@@ -26,7 +26,7 @@ try {
 export const yamlConfig = _.merge(defaultConfig, customConfig)
 
 export const MEMO_SHARING_SATS_THRESHOLD = yamlConfig.limits.memoSharingSatsThreshold
-export const ONCHAIN_MIN_CONFIRMATIONS = yamlConfig.onChainWallet.minConfirmations || 2
+export const ONCHAIN_MIN_CONFIRMATIONS = yamlConfig.onChainWallet.minConfirmations
 
 export const getGaloyInstanceName = () => yamlConfig.name
 
--- a/src/config/app.ts
+++ b/src/config/app.ts
@@ -26,6 +26,7 @@ try {
 export const yamlConfig = _.merge(defaultConfig, customConfig)
 
 export const MEMO_SHARING_SATS_THRESHOLD = yamlConfig.limits.memoSharingSatsThreshold
+export const ONCHAIN_MIN_CONFIRMATIONS = yamlConfig.onChainWallet.minConfirmations || 2
 
 export const getGaloyInstanceName = () => yamlConfig.name
 
@@ -107,6 +108,7 @@ export const getUserWalletConfig = (
   return {
     name: yamlConfig.name,
     dustThreshold: yamlConfig.onChainWallet.dustThreshold,
+    onchainMinConfirmations: ONCHAIN_MIN_CONFIRMATIONS,
     limits: transactionLimits,
   }
 }
--- a/src/core/ledger/transaction.ts
+++ b/src/core/ledger/transaction.ts
@@ -321,7 +321,7 @@ export const voidTransactions = (journalId, reason) => {
 }
 
 export const settlePayment = async (hash) => {
-  const result = await Transaction.updateMany({ hash }, { pending: false })
+  const result = await Transaction.updateMany({ hash, pending: true }, { pending: false })
   return result.nModified > 0
 }
 
--- a/src/core/ledger/transaction.ts
+++ b/src/core/ledger/transaction.ts
@@ -321,7 +321,7 @@ export const voidTransactions = (journalId, reason) => {
 }
 
 export const settlePayment = async (hash) => {
-  const result = await Transaction.updateMany({ hash, pending: true }, { pending: false })
+  const result = await Transaction.updateMany({ hash }, { pending: false })
   return result.nModified > 0
 }
 
--- a/src/core/on-chain/index.ts
+++ b/src/core/on-chain/index.ts
@@ -41,15 +41,16 @@ import {
 export const getOnChainTransactions = async ({
   lnd,
   incoming,
+  lookBack,
 }: {
   lnd: AuthenticatedLnd
   incoming: boolean
+  lookBack?: number
 }) => {
   try {
     const { current_block_height } = await getHeight({ lnd })
-    const after = Math.max(0, current_block_height - LOOK_BACK) // this is necessary for tests, otherwise after may be negative
+    const after = Math.max(0, current_block_height - (lookBack || LOOK_BACK)) // this is necessary for tests, otherwise after may be negative
     const { transactions } = await getChainTransactions({ lnd, after })
-
     return transactions.filter((tx) => incoming === !tx.is_outgoing)
   } catch (err) {
     const error = `issue fetching transaction`
--- a/src/core/on-chain/index.ts
+++ b/src/core/on-chain/index.ts
@@ -35,8 +35,10 @@ import {
   btc2sat,
   LoggedError,
   LOOK_BACK,
+  LOOK_BACK_OUTGOING,
   myOwnAddressesOnVout,
 } from "../utils"
+import { transactionNotification } from "@core/notifications/payment"
 
 export const getOnChainTransactions = async ({
   lnd,
@@ -69,7 +71,7 @@ export const OnChainMixin = (superclass) =>
     }
 
     async updatePending(lock): Promise<void> {
-      await Promise.all([this.updateOnchainReceipt(lock), super.updatePending(lock)])
+      await super.updatePending(lock)
     }
 
     async getOnchainFee({
@@ -158,12 +160,9 @@ export const OnChainMixin = (superclass) =>
 
           // on us onchain transaction
           if (payeeUser) {
-            let amountToSendPayeeUser // fee = 0
-            if (!sendAll) {
-              amountToSendPayeeUser = amount
-            }
-            // when sendAll the amount to send payeeUser is the whole balance
-            else {
+            let amountToSendPayeeUser = amount
+            if (sendAll) {
+              // when sendAll the amount to send payeeUser is the whole balance
               amountToSendPayeeUser = balance.total_in_BTC
             }
 
@@ -315,6 +314,7 @@ export const OnChainMixin = (superclass) =>
               const outgoingOnchainTxns = await getOnChainTransactions({
                 lnd,
                 incoming: false,
+                lookBack: LOOK_BACK_OUTGOING,
               })
               const [{ fee: fee_ }] = outgoingOnchainTxns.filter((tx) => tx.id === id)
               fee = fee_
@@ -326,12 +326,9 @@ export const OnChainMixin = (superclass) =>
             fee += this.user.withdrawFee
 
             {
-              let sats // full amount debited from account
-              if (!sendAll) {
-                sats = amount + fee
-              }
-              // when sendAll the amount debited from the account is the whole balance
-              else {
+              let sats = amount + fee
+              if (sendAll) {
+                // when sendAll the amount debited from the account is the whole balance
                 sats = balance.total_in_BTC
               }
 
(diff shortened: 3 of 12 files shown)


src/core/on-chain/index.ts
```diff
@@ -41,15 +41,16 @@ import {
...
+  lookBack,
 }: {
   lnd: AuthenticatedLnd
   incoming: boolean
+  lookBack?: number
 }) => {
   try {
     const { current_block_height } = await getHeight({ lnd })
-    const after = Math.max(0, current_block_height - LOOK_BACK) // this is necessary for tests, otherwise after may be negative
+    const after = Math.max(0, current_block_height - (lookBack || LOOK_BACK)) // this is necessary for tests, otherwise after may be negative
```
dolcalmi, line 52: maybe we need to change LOOK_BACK from 2016 to fewer blocks in getBalance

src/servers/trigger.ts
```diff
@@ -101,7 +106,7 @@ export async function onchainTransactionEventHandler(tx) {
 
     let user
     try {
-      user = await User.findOne({ "onchain.address": tx.output_addresses })
+      user = await User.findOne({ "onchain.address": { $in: tx.output_addresses } })
```
dolcalmi, line 109: the previous query was not working

src/servers/trigger.ts
```diff
@@ -137,6 +142,39 @@ export async function onchainTransactionEventHandler(tx) {
   }
 }
 
+export async function onchainBlockEventhandler({ lnd, height }) {
+  // get incoming txs from the last 36 blocks ~6 hours
+  // we don't use default look back because daily cron job will take care of delayed transactions
+  const lookBack = 6 * 6 // hours * ~ blocks per hour
```
dolcalmi, line 148: is it ok to only check last 6 hours?

src/servers/trigger.ts
```diff
@@ -137,6 +142,39 @@ export async function onchainTransactionEventHandler(tx) {
...
+  }
+
+  await runInParallel({
+    iterator: onchainTxns.values(),
+    logger,
+    workers: 5,
+    processor: async (tx, index) => {
+      logger.trace("updating onchain tx %o in worker %d", tx.id, index)
+
+      const user = await User.findOne({ "onchain.address": { $in: tx.output_addresses } })
```
dolcalmi, line 164: we need to validate performance of this query

src/services/lnd/auth.ts
```diff
@@ -10,6 +10,8 @@ const inputs: LndParams[] = getLndParams()
 // otherwise the tests will have to probe before being able to call lnd
 // TODO: use a mock instead.
 // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+const isSpec = require.main!.filename.indexOf(".spec.") !== -1
```
dolcalmi, line 13: this was needed because test is `trigger.spec.ts`

src/core/on-chain/index.ts
nicolasburtey: could be ~2 days worth of block? cron job run daily. but in case the cron job doesn't run for some reason if we cover 2 days that should work.

src/core/ledger/transaction.ts
```diff
@@ -321,7 +321,7 @@ export const voidTransactions = (journalId, reason) => {
 }
 
 export const settlePayment = async (hash) => {
-  const result = await Transaction.updateMany({ hash }, { pending: false })
+  const result = await Transaction.updateMany({ hash, pending: true }, { pending: false })
```
nicolasburtey, line 324: I guess it was implicit?

nicolasburtey: ie: `nModified` would be 0 if `pending` was already `false`?

src/servers/trigger.ts
nicolasburtey: mmm. I guess this loop was never executed then. 

I guess onchain receipts are validated from the trigger on new blocks. 

there is a flaw in the logic below anyway:
```
    if (tx.is_confirmed === false) {
     // ...
    } else {
     // code logic
    }
```

with `tx.confirmed`, tx is being considered confirm with 1 block finality IIRC. but we have currently by default a 2 blocks confirmation before crediting user balance. so going into 

```
      const wallet = await WalletFactory({ user, logger: onchainLogger })
      await wallet.updateOnchainReceipt()
``` 

is basically a no-op right now.

src/servers/trigger.ts
nicolasburtey: if we have an index that shouldn't be a problem (not sure we have one)

src/servers/trigger.ts
nicolasburtey: why not just do the last block? (not sure how lnd is handling a reorg, maybe last 2 blocks?)

src/core/on-chain/index.ts
nicolasburtey: note: we may not need to do any onchain check within getBalance anymore now? this would help improve the performance of the initial query on the mobile app

src/servers/trigger.ts
```diff
@@ -137,6 +142,39 @@ export async function onchainTransactionEventHandler(tx) {
   }
 }
 
+export async function onchainBlockEventhandler({ lnd, height }) {
+  // get incoming txs from the last 36 blocks ~6 hours
+  // we don't use default look back because daily cron job will take care of delayed transactions
```
nicolasburtey, line 147: not sure to understand the comment about delayed transactions?

the cronjob is here in case the trigger pod is not running while a block get mined.

nicolasburtey (commented):
great PR. this should help on the performance side. couple of questions/comments

src/core/on-chain/index.ts
dolcalmi: ok, moved the default LOOK_BACK for 2 days and add LOOK_BACK_OUTGOING (2 blocks) and LOOK_BACK_CHANNEL_UPDATE (previous value)

dolcalmi: > note: we may not need to do any onchain check within getBalance anymore now? this would help improve the performance of the initial query on the mobile app

done

src/core/on-chain/index.ts
```diff
@@ -655,7 +652,16 @@ export const OnChainMixin = (superclass) =>
...
+
               onchainLogger.info({ success: true, ...metadata })
+
+              await transactionNotification({
+                type,
+                user: this.user,
+                logger: onchainLogger,
+                amount: sats,
+                txid: matched_tx.id,
+              })
```
dolcalmi, line 664: I add the notification here because is the safest place to put it, block handler some times is triggering twice per block (at least in dev environment).

src/servers/trigger.ts
dolcalmi: you are right, just setup pending notification here and add confirmation after `ledger.addOnchainReceipt`

src/core/ledger/transaction.ts
dolcalmi: fixed

src/servers/trigger.ts
dolcalmi: I used onchain confirmations required + 1

src/servers/trigger.ts
dolcalmi: removed

src/config/app.ts
```diff
@@ -26,6 +26,7 @@ try {
 export const yamlConfig = _.merge(defaultConfig, customConfig)
 
 export const MEMO_SHARING_SATS_THRESHOLD = yamlConfig.limits.memoSharingSatsThreshold
+export const ONCHAIN_MIN_CONFIRMATIONS = yamlConfig.onChainWallet.minConfirmations || 2
```
nicolasburtey, line 29: no need to override the default here. the yaml is assumed to always be populated (unless this has been changed recently @vindard ?)

src/core/on-chain/index.ts
nicolasburtey: also in prod

src/services/lnd/auth.ts
```diff
@@ -24,7 +26,7 @@ export const addProps = (array) =>
       // find a way to mock this up for jest
       // for now only trigger is active = false at start
       // because trigger will start listening to lnd event on lnd start/restart
-      active: !isTrigger,
+      active: isSpec || !isTrigger,
```
nicolasburtey, line 29: (I know I already ask the question about the filename but didn't dive into it)

what does spec stands for? test?

src/servers/trigger.ts
```diff
@@ -181,15 +214,18 @@ const listenerOnchain = ({ lnd }) => {
...
   const subBlocks = subscribeToBlocks({ lnd })
-  subBlocks.on("block", () => updateUsersPendingPayment({ onchainOnly: true }))
+  subBlocks.on("block", async ({ height }) => {
+    await onchainBlockEventhandler({ lnd, height })
+  })
 
   subBlocks.on("error", (err) => {
-    console.log({ err }, "error subBlocks")
+    baseLogger.error({ err }, "error subBlocks")
+    subBlocks.removeAllListeners()
```
nicolasburtey, line 228: I'm not sure we want to remove the listener on error? it doesn't mean the app have to shutdown?

src/servers/trigger.ts
```diff
@@ -181,15 +214,18 @@ const listenerOnchain = ({ lnd }) => {
   subTransactions.on("chain_transaction", onchainTransactionEventHandler)
 
   subTransactions.on("error", (err) => {
-    baseLogger.info({ err }, "error subTransactions")
+    baseLogger.error({ err }, "error subTransactions")
     subTransactions.removeAllListeners()
```
nicolasburtey, line 218: maybe the comment about removing listen also apply here

src/config/app.ts
dolcalmi: I thought the same but didn't want to take risks with this, please let me know

src/services/lnd/auth.ts
dolcalmi: yes, test, do you prefer `isTest`?

src/servers/trigger.ts
```diff
@@ -115,26 +121,53 @@ export async function onchainTransactionEventHandler(tx) {
...
-      amount: Number(tx.tokens),
-      txid: tx.id,
-    })
+export async function onchainBlockEventhandler({ lnd, height }) {
+  const lookBack = ONCHAIN_MIN_CONFIRMATIONS + 1
+  const onchainTxns = await getOnChainTransactions({ lnd, lookBack, incoming: true })
+  const hasTransactions = onchainTxns && onchainTxns.length
+
+  if (!hasTransactions) {
+    logger.info(`finish block ${height} handler without transactions`)
```
nicolasburtey, line 148: maybe the wording could be improve here? `no transaction to handle, skipping block ${height}`?

having "finish" implied to me we did some work but we didn't really

src/core/utils.ts
```diff
@@ -10,7 +10,9 @@ import { User } from "@services/mongoose/schema"
 export const isDev = process.env.NODE_ENV !== "production"
 
 // how many block are we looking back for getChainTransactions
-export const LOOK_BACK = 2016
+export const LOOK_BACK = 288
```
nicolasburtey, line 13: let's add a 25% buffer in case block are processing faster? which could be the case over a small sample like 2 days.

src/config/app.ts
vindard: No changes from my end with that expected behavior

src/core/on-chain/index.ts
```diff
@@ -68,7 +71,7 @@ export const OnChainMixin = (superclass) =>
     }
 
     async updatePending(lock): Promise<void> {
```
nicolasburtey, line 73: I'm wondering if we now ever needs this function any more if we're passing awaiting on the parent?

nicolasburtey (commented):
couple more comments but looks good to me otherwise

src/core/on-chain/index.ts
dolcalmi: the order is `OnChainMixin(LightningMixin(UserWallet))` so we need to call the super

src/config/app.ts
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/428/commits/0adda6365e5311a71c03a054caad4adf90046630

src/services/lnd/auth.ts
dolcalmi: changed to isTest https://github.com/GaloyMoney/galoy/pull/428/commits/20668101e2a16ba4ad949eaa1af8980a7bbc26ed

src/servers/trigger.ts
dolcalmi: I removed both https://github.com/GaloyMoney/galoy/pull/428/commits/20668101e2a16ba4ad949eaa1af8980a7bbc26ed

src/servers/trigger.ts
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/428/commits/20668101e2a16ba4ad949eaa1af8980a7bbc26ed

src/servers/trigger.ts
dolcalmi: Changed https://github.com/GaloyMoney/galoy/pull/428/commits/20668101e2a16ba4ad949eaa1af8980a7bbc26ed

src/core/utils.ts
dolcalmi: new value 360 blocks https://github.com/GaloyMoney/galoy/pull/428/commits/20668101e2a16ba4ad949eaa1af8980a7bbc26ed

nicolasburtey (approved):

PR #429, opened 2021-08-08
Invoices repo

bodymindarts:
This PR builds on top of the [LightningService PR](https://github.com/GaloyMoney/galoy/pull/426) and continues the work of introducing patterns for key domain concepts that should be type safe.

Here the `Repository` pattern is introduced.
The `InvoicesRepo` persists a `WalletInvoice`.


bodymindarts:
stale

PR #434, opened 2021-08-09
Use a configuration value to control helmet content policy

samerbuna:
We need this to `true` in staging

### DIFF
--- a/src/servers/graphql-server.ts
+++ b/src/servers/graphql-server.ts
@@ -11,7 +11,7 @@ import PinoHttp from "pino-http"
 import { v4 as uuidv4 } from "uuid"
 import helmet from "helmet"
 
-import { getIpConfig } from "@config/app"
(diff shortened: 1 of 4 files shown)


src/servers/graphql-server.ts
```diff
@@ -108,7 +108,11 @@ export const startApolloServer = async ({
     },
   })
 
-  app.use(helmet({ contentSecurityPolicy: isDev ? false : undefined }))
+  app.use(
+    helmet({
+      contentSecurityPolicy: yamlConfig.helmet.disableContentPolicy ? false : undefined,
```
bodymindarts, line 113: @vindard has been doing so much work to remove all references to this global singleton. Can we not introduce one in the process?
Perhaps rebase off of https://github.com/GaloyMoney/galoy/pull/433 where he removes the other references to `yamlConfig` in this file?

PR #435, opened 2021-08-09
Cleanup and Upgrades

samerbuna:
This PR removes and disallows unnecessary return await, among few other things:

- Remove unused dependencies
- Upgrade dependencies
- Add missing types (and fix a few issues)
- Make GitHub actions run only on PR


.github/workflows/check-code.yml
```diff
@@ -1,6 +1,5 @@
 name: "Check code"
 on:
-  push:
```
dolcalmi, line 3: if we remove this the contributors will not be able to check errors before the PR

src/core/lock.ts
```diff
@@ -45,17 +45,17 @@ export const getResource = (path) => `locks:account:${path}`
...
   if (!!lock && lock.expiration > Date.now()) {
-    return await async_fn(lock)
+    return async_fn(lock)
   }
-  return await using(
+  return using(
     redlockClient.disposer(getResource(path), ttl, errorWrapper({ logger })),
     async (lock) => {
-      return await async_fn(lock)
+      return async_fn(lock)
```
dolcalmi, line 58: I had some issues with this change before (testing Medici void transactions). We can't postpone the execution so the await is necessary

dolcalmi: @nicolasburtey , would you mind taking a look at it?

src/core/on-chain/index.ts
```diff
@@ -145,217 +145,212 @@ export const OnChainMixin = (superclass) =>
...
-        { path: this.user._id, logger: onchainLogger },
-        async (lock) => {
-          const balance = await this.getBalances(lock)
-          onchainLogger = onchainLogger.child({ balance })
-
-          // quit early if balance is not enough
-          if (balance.total_in_BTC < amount) {
-            throw new InsufficientBalanceError(undefined, { logger: onchainLogger })
-          }
+      return redlock({ path: this.user._id, logger: onchainLogger }, async (lock) => {
```
dolcalmi, line 148: not sure why github is not showing correctly the diff because of this ... did you change any logic?

src/core/specter-wallet.ts
```diff
@@ -26,13 +26,11 @@ export class SpecterWallet {
   }
 
   async listWallets() {
```
dolcalmi, line 28: do we need the `async`?

src/core/specter-wallet.ts
```diff
@@ -26,13 +26,11 @@ export class SpecterWallet {
...
-    return await bitcoindDefaultClient.listWallets()
+    return bitcoindDefaultClient.listWallets()
   }
 
   async createWallet() {
```
dolcalmi, line 32: I think is a question for the whole file :)

src/servers/trigger.ts
```diff
@@ -255,7 +255,7 @@ const listenerOffchain = ({ lnd, pubkey }) => {
   })
 }
 
-const main = () => {
+const main = async () => {
```
dolcalmi, line 258: why do we need async here?

src/services/jwt.ts
```diff
@@ -1,10 +1,11 @@
+import { JWT_SECRET } from "@config/app"
 import * as jwt from "jsonwebtoken"
 
 // TODO: replace network by uri of the server
 // the uri will embed the network, ie: graphql.mainnet.server.io
 // and provide more information than just the network
-export const createToken = ({ uid, network }) =>
-  jwt.sign({ uid, network }, process.env.JWT_SECRET, {
+export const createToken = ({ uid, network }) => {
```
dolcalmi, line 7: is this a new rule? should we use return in one line functions?

dolcalmi (changes requested):
there are many methods (in different files) that only return promises and don't need async

.github/workflows/check-code.yml
samerbuna: Confirmed with @bodymindarts that we'd like all work in progress to be in PRs. If it's not ready, make it a draft PR.

bodymindarts: Draft PRs will also get feedback

src/core/lock.ts
samerbuna: If there are issues because of this, does that mean we have missing awaits somewhere?

src/core/specter-wallet.ts
samerbuna: Let's clean that up in another PR by defining the return types for methods that return promises

src/servers/trigger.ts
samerbuna: I guess not. Removed.

src/services/jwt.ts
samerbuna: one liners are okay, but this a multi-line function even though it returns a single thing. I think having return here is clearer.

bodymindarts (approved):

src/core/on-chain/index.ts
samerbuna: No logic changes here. Only removed the return await but I think Prettier is redoing the indentation here.

src/core/lock.ts
dolcalmi: really not sure... I did the same changes trying to debug void transactions but had some problems with concurrency maybe with the new changes is solved but just want to let you know

.github/workflows/check-code.yml
dolcalmi: OK but draft PRs to main (it will not work for PR with other base ). I mention this because you are using that for the lastest changes

dolcalmi (dismissed):

samerbuna:
@bodymindarts I just rebased this branch after merging #448. Please merge.

bodymindarts (approved):

.eslintrc.json
```diff
@@ -34,6 +34,7 @@
     "no-throw-literal": "error",
     "no-void": "error",
     "@typescript-eslint/no-extra-semi": "off", // Prettier work
+    "no-return-await": "error",
```
krtk6160, line 37: From ESLint docs ->
> You can avoid the extra microtask by not awaiting the return value, with the trade off of the function no longer being a part of the stack trace if an error is thrown asynchronously from the Promise being returned. This can make debugging more difficult.

Have we taken into account the "make debugging more difficult" tradeoff? I'm not sure how this changes the stacktrace, so just want to know if this was considered.

krtk6160, line 37: From [ESLint docs](https://eslint.org/docs/rules/no-return-await) ->
> You can avoid the extra microtask by not awaiting the return value, with the trade off of the function no longer being a part of the stack trace if an error is thrown asynchronously from the Promise being returned. This can make debugging more difficult.

Have we taken into account the "make debugging more difficult" tradeoff? I'm not sure how this changes the stacktrace, so just want to know if this was considered.

krtk6160 (commented):
One question regarding the new es lint rule

.eslintrc.json
samerbuna: Yes, I'd rather not have an extra task and stick with the idiomatic way. V8 has some handling of async stack traces so I don't think that tradeoff is significant.

dolcalmi (approved):

PR #437, opened 2021-08-10
Fix errors path

bodymindarts:


src/services/mongoose/invoices.ts
```diff
@@ -1,4 +1,4 @@
-import { UnknownRepositoryError, CouldNotFindError } from "@domain//errors"
```
vindard, line 1: We can probably add a linting rule to catch this

I found [this related issue](https://github.com/import-js/eslint-plugin-import/issues/915#issuecomment-322898231) that says that the `"import/no-useless-path-segments"` rule can fix this. When I tried it though, **it worked** for absolute imports (e.g. `".//schema"` got linted). **It didn't work** for mapped imports though because the double `//` seems to resolve correctly in the checks but doesn't compile correctly when we build? (cc @samerbuna)

Screenshot from 2021-08-10 10-16-05

PR #438, opened 2021-08-10
WIP Bitcoind replace lnd-onchain

jotapea:
New branch, but still a work in progress, for the original https://github.com/GaloyMoney/galoy/pull/317.

### DIFF
--- a/src/core/on-chain/index.ts
+++ b/src/core/on-chain/index.ts
@@ -1,5 +1,5 @@
 import assert from "assert"
-import { GetChainTransactionsResult } from "lightning" // TODO remove eventually?
+import { getChainTransactions, getHeight, GetChainTransactionsResult } from "lightning" // TODO remove eventually?
 import _ from "lodash"
 import moment from "moment"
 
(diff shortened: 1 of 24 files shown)


src/core/on-chain/index.ts
```diff
@@ -84,24 +58,22 @@ export const OnChainMixin = (superclass) =>
...
-        const sendTo = [{ address, tokens: amount ?? defaultAmount }]
         try {
-          ;({ fee } = await getChainFeeEstimate({ lnd, send_to: sendTo }))
+          // (numeric, optional) estimate fee rate in BTC/kB (only present if no errors were encountered)
+          const result = await this.bitcoindClient.estimateSmartFee({
+            conf_target: 1,
+          }) // TODO conf_target
+          const feerate = result.feerate
+          // 1 BTC/kB = 100000 satoshis/byte
+          fee = 100000 * feerate // TODO not sure about this: is fee by size, not a total fee
```
jotapea, line 74: Confirmed that it is different from lightning's getChainFeeEstimate, which does return a total [fee](https://github.com/alexbosworth/ln-service#getchainfeeestimate).

jotapea:
This is still in full WIP mode, so it contains multiple open questions as comments (which will be cleaned up after addressed). The main blocker at the moment is related to the lifecycle of the zeromq socket. I haven't found a way to be able to run multiple tests with it, at the moment in only works in a single test at a time...

src/core/on-chain/index.ts
```diff
@@ -378,20 +356,12 @@ export const OnChainMixin = (superclass) =>
...
-          lnd,
-          format: "p2wpkh",
-        }))
-      } catch (err) {
-        const error = `error getting on chain address`
-        this.logger.error({ err }, error)
-        throw new LoggedError(error)
-      }
+      // TODO pubkey should be known already
+      const { pubkey } = await this.bitcoindWalletClient.getAddressInfo({ address })
```
dolcalmi, line 364: I am not sure if we need this pubkey, the previous pubkey is the pubkey of lnd node that is needed to get chain data in the cron and trigger.

nicolasburtey:
@jotapea you should have the PR as "draft" when they are work in progress

PR #441, opened 2021-08-11
Fix vout validation in amountOnVout and myOwnAddressesOnVout

dolcalmi:


src/core/utils.ts
```diff
@@ -60,17 +60,22 @@ export const addContact = async ({ uid, username }) => {
 export const amountOnVout = ({ vout, addresses }): number => {
   // TODO: check if this is always [0], ie: there is always a single addresses for vout for lnd output
   return _.sumBy(
-    _.filter(vout, (tx) => _.includes(addresses, tx.scriptPubKey.addresses[0])),
+    _.filter(vout, (tx) => {
+      if (tx.scriptPubKey && tx.scriptPubKey.addresses) {
```
nicolasburtey, line 64: `tx.scriptPubKey?.addresses` would do the same as `tx.scriptPubKey && tx.scriptPubKey.addresses`?

dolcalmi: you are right, I am just too used to vanilla js.

PR #446, opened 2021-08-13
Add price server to docker compose

dolcalmi:
- Add price server to dev environment (docker compose)
- Add trigger to Make file
- Update .envrc with price env variables

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,9 +1,14 @@
 version: "2"
 services:
+  price:
+    image: gcr.io/galoyapp/price
+    ports:
+    - "50051:50051"
+    - "9464:9464"
   redis:
     image: bitnami/redis:6.2
     ports:
-    - '6379:6379'
+    - "6379:6379"
     environment:
     - ALLOW_EMPTY_PASSWORD=yes
     - REDIS_DISABLE_COMMANDS=FLUSHDB,FLUSHALL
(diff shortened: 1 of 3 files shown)


jrbuhl93 (dismissed):

docker-compose.yml
```diff
@@ -1,9 +1,14 @@
 version: "2"
 services:
+  price:
+    build: https://github.com/GaloyMoney/price.git#main
```
bodymindarts, line 4: This should be a tagged image not `build: ...#main`. This file is not for developing the images but to provide deps _as they are in production_ to develop _this_ app.

nicolasburtey: we are not doing versioning yet on the price pod

dolcalmi: is there a way to resolve it from chart repository(gcr.io/galoyapp/price)? or it is different?

nicolasburtey: > is there a way to resolve it from chart repository(gcr.io/galoyapp/price with latest tag)? or is it different?

yes you can use the hash. should be shown in the build done by github action

dolcalmi: I tried with `build: gcr.io/galoyapp/price:latest` and `build: gcr.io/galoyapp/price#latest` but is not working

samerbuna (approved):

bodymindarts (approved):

PR #447, opened 2021-08-14
Add all domain elements needed for `getTransactions` use case

bodymindarts:
This PR introduces a consistent structure under the `src/domain` and `src/app` folders.

All domain modules required for replacement of the `getTransactions` call in the existing code have been added.

This includes:
- `domain/accounts`
- `domain/bitcoin/onchain`
- `domain/bitcoin/lightning`
- `domain/ledger`
- `domain/users`
- `domain/wallets`
- `domain/wallet-invoices`

Use cases are exposed via the `src/app` folder like in `src/app/wallets/get-transactions-for-wallet.ts`.

For domain logic and data transformations unit tests were added.

PR #448, opened 2021-08-14
Paginated transactions

bodymindarts:
Builds off of https://github.com/GaloyMoney/galoy/pull/447 to support paginated transactions in the new schema


src/app/wallets/get-transactions-for-wallet.ts
```diff
@@ -0,0 +1,59 @@
...
+  const ledger = MakeLedgerService()
+  const liabilitiesAccountId = toLiabilitiesAccountId(wallet.id)
+  const ledgerTransactions = await ledger.liabilityTransactions(liabilitiesAccountId)
+  if (ledgerTransactions instanceof LedgerError) return ledgerTransactions
+
+  const confirmedHistory = WalletTransactionHistory.fromLedger(ledgerTransactions)
+
+  const onChain = MakeOnChainService(MakeTxDecoder(process.env.NETWORK as BtcNetwork))
+  if (onChain instanceof OnChainError) {
+    return confirmedHistory.transactions
```
nicolasburtey, line 37: may need to add an additional error to let the client know that not all the data has been returned

src/app/wallets/get-transactions-for-wallet.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  const onChainTxs = await onChain.getIncomingTransactions(LOOK_BACK)
+  if (onChainTxs instanceof OnChainError) {
+    return confirmedHistory.transactions
+  }
+
+  const filter = MakeTxFilter({
+    confsLT: ONCHAIN_MIN_CONFIRMATIONS,
```
nicolasburtey, line 46: `confsLT` not clear to me what LT means. not less than I guess?

src/app/wallets/get-transactions-for-wallet.ts
```diff
@@ -0,0 +1,59 @@
...
+    addresses: wallet.onChainAddresses,
+  })
+  const pendingTxs = filter.apply(onChainTxs)
+
+  // TODO should be a service - not a function call
+  let price = await getCurrentPrice()
+  if (typeof price !== "number") {
+    price = 0
```
nicolasburtey, line 54: I think we may never want `price = 0` should be a fallback value. for multiple reasons:
- if we need to do a conversion, we may need to divide by this value. this would lead to further issues
- it could be some form of an attack vector eventually for USD account

nicolasburtey: so I think at this point we either want to have `NaN` or trigger an error.

nicolasburtey: maybe that doesn't matter here because it's only for pending transaction

src/domain/bitcoin/onchain/tx-decoder.ts
```diff
@@ -0,0 +1,31 @@
...
+  return {
+    decode,
+  }
+}
+
+const decodeOutput = (tx, network): TxOut[] => {
+  const format = (out, n, network) => {
+    return {
+      sats: toSats(out.value),
+      n: n as number,
```
nicolasburtey, line 25: may not be needed

src/domain/wallets/tx-history.ts
```diff
@@ -0,0 +1,189 @@
...
+const addPendingIncoming = (
+  confirmedTransactions: WalletTransaction[],
+  pendingTransactions: SubmittedTransaction[],
+  addresses: OnChainAddress[],
+  usdPerSat: UsdPerSat,
+): WalletTransactionHistoryWithPending => {
+  const walletTransactions: WalletTransaction[] = []
+  pendingTransactions.forEach(({ id, rawTx, createdAt }) => {
+    rawTx.outs.forEach(({ sats, address }) => {
+      if (addresses.includes(address)) {
```
nicolasburtey, line 15: may want to use lodash for this one

src/domain/wallets/tx-history.ts
```diff
@@ -0,0 +1,189 @@
...
+          settlementAmount,
+          settlementFee: fee,
+          paymentHash: null,
+          addresses: addresses,
+          recipientId: null,
+          pendingConfirmation,
+          createdAt: timestamp,
+        }
+      }
+      if (type === LedgerTransactionType.IntraLedger) {
```
nicolasburtey, line 86: may need to differentiate between intraledger initiated by username and initiated by lightning invoice

src/graphql/types/object/fiat-wallet.ts
```diff
@@ -0,0 +1,40 @@
...
+    // },
+    transactions: {
+      type: TransactionConnection,
+      args: connectionArgs,
+      resolve: async (source: Wallet, args) => {
+        const transactions = await Wallets.getTransactionsForWallet(source)
+        if (transactions instanceof Error) {
+          throw transactions
+        }
+        console.log("**********", transactions)
```
nicolasburtey, line 33: debug code?

src/domain/wallets/tx-history.ts
bodymindarts: We've been discussing this a lot. Lets finalise our approach with @jrbuhl93 when he arrives

dolcalmi:
please change the base, if we don't merge #447 check this is a duplicate effort

bodymindarts:
> please change the base, if we don't merge #447 check this is a duplicate effort

Yeah we are directly working on this branch... will close the other one

src/domain/ledger/index.types.d.ts
```diff
@@ -0,0 +1,45 @@
...
+type LiabilitiesAccountId = string & { [liabilitiesAccountId]: never }
+
+declare const ledgerTransactionIdSymbol: unique symbol
+type LedgerTransactionId = string & { [ledgerTransactionIdSymbol]: never }
+
+declare const ledgerAccountIdSymbol: unique symbol
+type LedgerAccountId = string & { [ledgerAccountIdSymbol]: never }
+
+type LedgerTransactionType =
+  typeof import("./index").LedgerTransactionType[keyof typeof import("./index").LedgerTransactionType]
```
vindard, line 13: I believe we can type this directly here to get the constant values assigned to each field. Typescript should allow something like `type a = "Alice"` to be type string restricted to single string value `'Alice'`.

```suggestion
type LedgerTransactionType = {
  Invoice: "invoice"
  Payment: "payment"
  IntraLedger: "on_us"
  LnFeeReimbursement: "fee_reimbursement" // lightning
  OnchainReceipt: "onchain_receipt"
  OnchainPayment: "onchain_payment"
  OnchainIntraLedger: "onchain_on_us"
  OnchainDepositFee: "deposit_fee" // onchain
  Fee: "fee"
  Escrow: "escrow"
  RoutingFee: "routing_fee" // channel-related
  ExchangeRebalance: "exchange_rebalance" // send/receive btc from the exchange
  UserRebalance: "user_rebalance" // buy/sell btc in the user wallet
  ToColdStorage: "to_cold_storage"
  ToHotWallet: "to_hot_wallet"
}
```

src/services/mongoose/schema.types.d.ts
```diff
@@ -1 +1,78 @@
...
+  accountPath: string
+  oldEnoughForWithdrawal: boolean
+}
+
+// ?: relocate
+type Account = {
+  level?: number
+  status?: string
+  language?: string
+}
```
vindard, line 78: This is also defined in `domain/accounts/index.types.d.ts`
```suggestion
```

src/domain/accounts/index.types.d.ts
```diff
@@ -0,0 +1,16 @@
+type AccountLevel =
+  typeof import("./index").AccountLevel[keyof typeof import("./index").AccountLevel]
```
vindard, line 2: I believe we can type this directly here to get the constant values assigned to each field, although I'm not sure if we deliberately did it this way so that the type is defined in only one place.

If not, typescript should allow something like `type a = "Alice"` to be type string restricted to single string value `'Alice'`.

_(This comment would be for all places where we use this pattern)_

```suggestion
type AccountLevelType = {
  One: 1,
  Two: 2,
}
```

src/domain/accounts/index.ts
```diff
@@ -0,0 +1,9 @@
+export const AccountLevel = {
+  One: 1,
+  Two: 2,
+} as const
```
vindard, line 4: If we implement the type in the `.d.ts` file, this can probably change. Also, would this be a good candidate for a typescript `enum`?

```suggestion
export const AccountLevel: AccountLevelType = {
  One: 1,
  Two: 2,
}
```

src/domain/wallets/tx-history.ts
```diff
@@ -0,0 +1,189 @@
...
+          settlementAmount,
+          settlementFee: fee,
+          paymentHash: (paymentHash as PaymentHash) || null,
+          recipientId: username || null,
+          addresses: null,
+          pendingConfirmation,
+          createdAt: timestamp,
+        }
+      } else if (addresses && addresses.length > 0) {
+        return {
```
vindard, line 105: Probably doesn't need an `else` here because of early return in previous block and consistency with Ln 86.

```suggestion
      }
      if (addresses && addresses.length > 0) {
        return {
```

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -372,16 +420,29 @@ describe("UserWallet - Lightning Pay", () => {
...
+
+          let txResult = await Wallets.getTransactionsForWalletId({
+            walletId: walletPayee.user.id,
+          })
+          if (txResult.error instanceof Error) {
+            throw txResult.error
+          }
+          const user2Txn = txResult.transactions
           const user2OnUsTxn = user2Txn.filter(matchTx)
-          expect(user2OnUsTxn[0].type).toBe("on_us")
```
vindard, line 379: In places like this, do we also want to check for something like `expect(user2OnUsTxn[0].old.type).toBe("on_us")` since we're doing it in other places?

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -100,25 +127,35 @@ describe("UserWallet - Lightning Pay", () => {
       amount: amountInvoice,
     })
 
+    // WIP FIXME: Fails because it expects GetTransactions... to be
+    //            ordered by latest first in array
```
vindard, line 131: A comment I added that can probably be removed now that it looks like the test works with the new changes.
```suggestion
```

vindard (changes requested):
I added some small cleanup suggested changes and some questions around typing constants.

src/domain/accounts/index.types.d.ts
bodymindarts: What I want is to be able to reference the constants in code without typing the strings:
```
const level = AccountLevelType.One
```

if we put the definition in the `.types.d.ts` file we would have the definition in 2 places or have to hard code values everywhere:
```
const level = 1
```

enums have other drawbacks as they are not native to JS https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums.

All things considered I chose this pattern for collections of constants but perhaps we can discuss it in the team at daily.

test/integration/02-user-wallet/02-send-lightning.spec.ts
bodymindarts: Should probably only check for the consistently `settlementVia` but don't want to clean it all up now.

src/domain/wallets/tx-history.ts
```diff
@@ -64,77 +66,66 @@ export const fromLedger = (
...
-          paymentHash: null,
-          addresses: addresses,
-          recipientId: null,
+          settlementFee: toSats(fee || 0),
           pendingConfirmation,
           createdAt: timestamp,
         }
       }
-      if (type === LedgerTransactionType.IntraLedger) {
+      if (paymentHash) {
```
vindard, line 91: There was some logic on the `main` branch that suggested that `Intraledger` payments could have `paymentHash` values as well and that this was used as the unique identifier for the `invoiceusers` table (see [this test line](https://github.com/GaloyMoney/galoy/blob/548ac0fc673280033fef8f7bff3bbb68e7401baf/test/integration/02-user-wallet/02-send-lightning.spec.ts#L68) for e.g.).

If we changed that behavior then this is ok, but if not then this could potentially give us `Intraledger` transactions as well as `Lightning` transactions I think.

vindard (commented):
~All looks good, just one question before I approve~

src/domain/wallets/tx-history.ts
vindard: Oh nevermind, I just caught the rest of logic just under this... LGTM :+1:

vindard (approved):

src/domain/wallets/tx-history.ts
bodymindarts: yes that is why there is a switch on the `settlementVia` field:
```
      if (paymentHash) {
        return {
          id,
          initiationVia: PaymentInitiationMethod.Lightning,
          settlementVia:
            type === LedgerTransactionType.IntraLedger
              ? SettlementMethod.IntraLedger
              : SettlementMethod.Lightning,
```

src/app/wallets/get-transactions-for-wallet.ts
```diff
@@ -0,0 +1,71 @@
...
+// TODO should be exposed via PriceSerivce / LiquidityProvider
+import { getCurrentPrice } from "@services/realtime-price"
+
+export const getTransactionsForWalletId = async ({
+  walletId,
+}: {
+  walletId: WalletId
+}): Promise<{
+  transactions: WalletTransaction[]
+  error?: ApplicationError
```
dolcalmi, line 20: why this method is using a different pattern for return?

dolcalmi (dismissed):

src/domain/users/index.types.d.ts
```diff
@@ -0,0 +1,18 @@
...
+
+type UserLanguage =
+  typeof import("./index").UserLanguage[keyof typeof import("./index").UserLanguage]
+
+type User = {
+  id: UserId
+  username: Username | null
+  phone: PhoneNumber
+  language: UserLanguage
+  defaultAccountId: AccountId
```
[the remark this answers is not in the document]
bodymindarts: yes that's right

src/app/wallets/get-transactions-for-wallet.ts
bodymindarts: Because it can return partial results. ie getting ledger transactions works then we want to pass those back to the user but also want to return the error to communicate that we couldn't get the pending txs.

src/graphql/types/object/user.ts
```diff
@@ -0,0 +1,42 @@
...
+  fields: () => ({
+    id: { type: GT.NonNullID },
+    username: { type: Username }, // TODO: make username required
+    phone: { type: GT.NonNull(Phone) },
+    language: {
+      type: GT.NonNull(Language),
+      resolve: (source) => source.language,
+    },
+
+    defaultAccount: {
```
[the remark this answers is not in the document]
bodymindarts: I don't understand your question.

A user has a default account (one of many accounts). Account has this type:
```
type Account = {
  id: AccountId
  level: AccountLevel
  status: AccountStatus
  walletIds: WalletId[]
}
```

so to get to your wallet you go through the query:
```
me -> defaultAccount -> wallet
```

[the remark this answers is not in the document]
bodymindarts: All the values for `User`, `Account` and `Wallet` are currently on the same `Users` table of mongoose - but data migration is out of scope of this PR.

src/graphql/types/object/ln-transaction.ts
```diff
@@ -0,0 +1,30 @@
...
+    },
+
+    // TODO: amount/fee/currency
+
+    // memo: {
+    //   type: Memo,
+    // },
+    // status: {
+    //   type: GT.NonNull(TxStatus),
+    // },
```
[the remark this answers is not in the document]
bodymindarts: done

src/graphql/types/object/onchain-transaction.ts
```diff
@@ -0,0 +1,30 @@
...
+    },
+
+    // TODO: amount/fee/currency
+
+    // memo: {
+    //   type: Memo,
+    // },
+    // status: {
+    //   type: GT.NonNull(TxStatus),
+    // },
```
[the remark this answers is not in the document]
bodymindarts: done

src/graphql/types/object/user.ts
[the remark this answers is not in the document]
dolcalmi: thanks

src/app/wallets/get-transactions-for-wallet.ts
dolcalmi: uhmm, so if my inner code can return a partial result then should I use this "pattern"? 

you use a lot of wrappers in other parts of the code, why is this better  than use a wrapper and have consistency in the way that we handle errors?

bodymindarts: > you use a lot of wrappers in other parts of the code, why is this better than use a wrapper and have consistency in the way that we handle errors?

can you show an example here?

bodymindarts: Just discussing this with @samerbuna.
Think we could standardise on returning an array for partial results:
```
[ResultType, ErrorType]
```

dolcalmi:
```js
export const getTransactionsForWalletId = async ({
  walletId,
}: {
  walletId: WalletId
}): Promise<PartialTransactions | ApplicationError >

// or the name you want
type PartialTransactions = {
  transactions: WalletTransaction[]
  isPartial: boolean
  // or use error 
  error?: ApplicationError

  // later we will have to add something like cursor or other related pagination fields
}
```

yes, probably we should do a log if we have an internal error (getting incoming transactions) but we will not stop the flow of the app so we can simply return Transactions (as alternative) and handle the log/notification internally

src/domain/ledger/index.types.d.ts
```diff
@@ -0,0 +1,45 @@
...
+
```
samerbuna, line 14: Extract `typeof import("./index")` into a named type here to simplify?

src/servers/graphql-old-server.ts
```diff
@@ -59,7 +79,16 @@ const resolvers = {
         id: "BTC",
         currency: "BTC",
         balance: async () => (await wallet.getBalances())["BTC"],
-        transactions: () => wallet.getTransactions(),
+        transactions: async () => {
+          const [txs, error] = await Wallets.getTransactionsForWalletId({
+            walletId: wallet.user.id,
+          })
+          if (error instanceof Error) {
```
dolcalmi, line 86: so if we have problem with onchain incoming transactions will we give an error to the user? (we already have transactions from ledger)

src/app/wallets/get-transactions-for-wallet.ts
bodymindarts: I think if both a result and an error is returned its clear that it is a partial result.

src/app/wallets/get-transactions-for-wallet.ts
```diff
@@ -0,0 +1,61 @@
...
+import { WalletTransactionHistory } from "@domain/wallets"
+
+// TODO should be exposed via PriceSerivce / LiquidityProvider
+import { getCurrentPrice } from "@services/realtime-price"
+
+export const getTransactionsForWalletId = async ({
+  walletId,
+}: {
+  walletId: WalletId
+}): Promise<[WalletTransaction[], ApplicationError?]> => {
```
dolcalmi, line 18: I really dont understand why we are over complicating this, just return transactions because we are throwing an error to the user if we have problems with incoming txs

bodymindarts: Right now we are throwing an error to move forward but we want to translate and return the error properly via graphql

src/servers/graphql-old-server.ts
bodymindarts: yes we should then return both the transactions and the error via graphql... handling of error in gql layer is out of scope for this PR

vindard (approved):
The new `PartialResult` code was the only tricky new thing to wrap my head around, but otherwise this looks good from my last review ([relevant diff](https://github.com/GaloyMoney/galoy/compare/5644d7bf0ee93f692ba03d094ced286b59a80f06...31e76ecd0c30e2cc851d2c06eb260bb336f8e4b4))

samerbuna (approved):

PR #449, opened 2021-08-16
Reimplement addInvoice use case with new architecture

vindard:
_Builds on #448_

## Description

This PR implements 4 use-cases under the `app/wallets/add-invoice-for-wallet.ts` file:
- self-generated invoice
- self-generated invoice, no amount
- generated-for-recipient invoice
- generated-for-recipient invoice, no amount

It also adds a new `WalletInvoiceFactory` method and related unit tests for creating `WalletInvoice` objects.


src/app/wallets/add-invoice-for-wallet.ts
```diff
@@ -0,0 +1,100 @@
...
+import { MakeLndService } from "@services/lnd"
+import { MakeWalletsRepository, MakeWalletInvoicesRepository } from "@services/mongoose"
+
+export const addInvoiceForSelf = async ({
+  walletId,
+  memo,
+  amount,
+}: AddInvoiceSelfArgs): Promise<LnInvoice | ApplicationError> => {
+  if (!(amount && amount > 0))
+    return new Error("Incorrect method used for no-amount invoice")
```
vindard, line 13: I wasn't sure what type of `Error` I should use here and which class I should extend from ... would be something like a `BadArgumentError`?

`ApplicationError` is already defined as the union of some `DomainError` sub-types, so it doesn't seem like this would derive from that

vindard:
:point_up_2:  [`force-push`](https://github.com/GaloyMoney/galoy/compare/8ab098faafd63b1602c81180f8a85de0ff7710ca..68ea94691d711dbe8a6aa3b559dc4fbf3c41326f) to bring in latest #448 changes from `main`

src/app/wallets/add-invoice-for-wallet.ts
bodymindarts: I would put this into a `checkedToSats` method that returns `Satoshi | ConversionError`. This method can be called in the gql layer prior to calling the use case.

src/core/lightning/index.types.d.ts
```diff
@@ -19,3 +19,15 @@ interface IPaymentRequest {
   memo?: string
   isReward?: boolean
 }
+
+type AddInvoiceSelfArgs = {
```
bodymindarts, line 23: We're not adding any new code to `core`. Needs to go under `app/wallets/index.types.d.ts`

src/core/lightning/index.types.d.ts
```diff
@@ -19,3 +19,15 @@ interface IPaymentRequest {
...
+  walletId: WalletId
+  memo?: string
```
bodymindarts, line 25: Memo + amount shouldn't be optional. Input validation should happen _before_ calling the use-case method.
Args should not accept undefined should just default to `memo: ""`, and `satoshis: 0` or something like that

bodymindarts: Perhaps the `satoshis: 0` is also its own use case (`AddNoAmountInvoice`). After all it is its own `gql` mutation:
```
  lnNoAmountInvoiceCreateOnBehalfOfRecipient(
    input: LnNoAmountInvoiceCreateOnBehalfOfRecipientInput!
  ): LnNoAmountInvoicePayload!
```

src/domain/wallet-invoices/wallet-invoice-factory.ts
```diff
@@ -0,0 +1,30 @@
+export const WalletInvoiceFactory = (walletId: WalletId): IWalletInvoiceFactory => {
+  const create = ({ registeredInvoice }): WalletInvoice => {
+    return _baseCreate({ registeredInvoice, selfGenerated: true })
```
bodymindarts, line 3: I find the indirection here overkill. Would prefer to see the 2 functions `create` + `createForRecipient` return their own `WalletInvoice` inline. Makes it easier to read on first pass without having to understand the `_baseCreate`.

src/domain/wallet-invoices/wallet-invoice-factory.types.d.ts
```diff
@@ -0,0 +1,8 @@
+interface IWalletInvoiceFactory {
```
bodymindarts, line 1: No need to be an `interface` right? Or are there more than 1 implementations?

src/domain/wallets/index.types.d.ts
```diff
@@ -75,4 +75,5 @@ type Wallet = {
 
 interface IWalletsRepository {
   findById(walletId: WalletId): Promise<Wallet | RepositoryError>
+  walletIdFromUsername(username: Username): Promise<WalletId | RepositoryError>
```
bodymindarts, line 78: This feels misplaced. First of all `Repository` should always return the whole object not just an `id` - so better would be:
```
wallet = repo.findWalletForUser
id = wallet.id
```

But actually the `WalletsRepository` should not know anything about usernames anyway. I think what we need is:
```
user = usersRepo.findByUsername(username)
defaultAccount = accountsRepo.findById(user.defaultAccountId)
walletId = deffaultAccount.walletIds[0]
```

src/services/lnd/index.ts
```diff
@@ -1,11 +1,20 @@
...
 } from "@domain/bitcoin/lightning"
 import { createInvoice } from "lightning"
+import { getActiveLnd } from "./utils"
+
+export const LndService = (): ILightningService => {
+  let lndAuth: AuthenticatedLnd, pubkey: string
+  try {
+    ;({ lnd: lndAuth, pubkey } = getActiveLnd())
+  } catch (err) {
+    throw new LndOfflineError("no active lnd to create an invoice")
```
bodymindarts, line 15: should be returned as error not thrown.
`export const LndService = (): ILightningService | LightningServiceError`

src/services/mongoose/wallets.ts
```diff
@@ -26,7 +26,22 @@ export const WalletsRepository = (): IWalletsRepository => {
     }
   }
 
+  const walletIdFromUsername = async (
```
bodymindarts, line 29: See comment on interface definition

bodymindarts (changes requested):

src/core/lightning/index.types.d.ts
bodymindarts: I see that you are differentiating the use case methods with `NoAmount` so just the `Args` types are not in sync with the methods.

dolcalmi:
Hi guys, I think we must coordinate better this changes, addInvoice is the main method of USD invoices and the first change to allow Justin B to implement changes in mobile app. @vindard do you want to do it?

src/domain/wallet-invoices/wallet-invoice-factory.types.d.ts
vindard: Changed :+1: 

There are a few other interfaces that only have 1 implementation as well. Should we look at changing these too?

image

src/domain/wallets/index.types.d.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/449/commits/1e3fcd2bd79a2ea46b574cadd0119e3b95c120ca

src/services/mongoose/users.ts
```diff
@@ -27,7 +27,28 @@ export const UsersRepository = (): IUsersRepository => {
     }
   }
 
+  const findByUsername = async (username: Username): Promise<User | RepositoryError> => {
+    try {
+      const result = await User.findOne({ username })
```
vindard, line 32: I wasn't sure if to implement this lookup like this or as `user.getUserByUsername(username)` since that schema method can still throw and we haven't started refactoring those as yet

src/services/mongoose/invoices.ts
```diff
@@ -1,7 +1,7 @@
 import { UnknownRepositoryError, CouldNotFindError } from "@domain/errors"
 import { InvoiceUser } from "./schema"
 
-export const InvoicesRepository = () => {
+export const InvoicesRepository = (): IWalletInvoicesRepository => {
```
vindard, line 4: It looks like this entire `invoices.ts` file is a duplicate of [`wallet-invoices.ts`](https://github.com/GaloyMoney/galoy/pull/449/commits/39f7bf4b0aad954c57b42d49f37ae316945978c3#diff-49fbb6ea25d90667e8e063158028c6822c568827ef9ab3221f9626c75fe6054eL1). Do we want both of these for now?

src/core/lightning/index.types.d.ts
bodymindarts: Agree

src/services/mongoose/invoices.ts
bodymindarts: No only wallet-invoices. Delete the other one

src/domain/wallet-invoices/wallet-invoice-factory.types.d.ts
bodymindarts: The reason I think these should be interfaces is because the implementation is in another module - so we are decoupling declaration from implementation.

src/services/mongoose/users.ts
bodymindarts: I like re-implementing it and not using the method that can throw... but you should also bring in the validation logic to be consistent:
```
  if (!username.match(regexUsername)) {
    return null
  }

  const user = await this.findOne({ username: caseInsensitiveRegex(username) })
  ```
  
  In perticular that the lookup is case insensitive is important here.

File: src/services/mongoose/invoices.ts (hunk 10 above)
Comment (bodymindarts) [reply src/services/mongoose/invoices.ts:4]: Looks like @dolcalmi has deleted it in [his PR](https://github.com/GaloyMoney/galoy/pull/456) so you can ignore it here.

File: src/app/wallets/index.types.d.ts [hunk 11]
```diff
@@ -0,0 +1,27 @@
...
+  amount: Satoshis
+  memo?: string
+}
+
+type AddInvoiceNoAmountRecipientArgs = {
+  username: Username
+  memo?: string
+}
+
+type BaseAddInvoiceArgs = {
```
bodymindarts, line 23: I wouldn't declare this here as it is for a function that _isn't_ exposed publicly. Just declare the arg types inline in the function definition. Only declare types here that need exporting.

src/domain/wallet-invoices/wallet-invoice-factory.types.d.ts
```diff
@@ -0,0 +1,8 @@
+type WalletInvoiceFactoryType = {
```
bodymindarts, line 1: Don't think you need the post-fix `Type`. Something with that post fix implies it parameterises another type like:
```
type SomeErrorType
type Error<T>
type SpecificError = Error<SomeErrorType>
```

src/services/mongoose/users.ts
vindard: Agreed on the validation logic, implemented here now: https://github.com/GaloyMoney/galoy/pull/449/commits/7fa8be530da9f637479a70c1d660f2fafa88038c

To clarify, on the "replace + delete" do you mean instantiating a UserRepository and calling the new `findByUsername` method on it in all these places? :point_down: 

image

bodymindarts: Yeah we should avoid having duplicate login so I think you should replace all those calls

src/app/wallets/add-invoice-for-wallet.ts
```diff
@@ -0,0 +1,113 @@
...
+  const walletInvoiceFactory = WalletInvoiceFactory(walletId)
+  return baseAddInvoiceForWallet({
+    amount: toSats(0),
+    memo,
+    walletInvoiceCreateFn: walletInvoiceFactory.create,
+  })
+}
+
+export const addInvoiceForRecipient = async ({
+  username,
```
bodymindarts, line 37: This arg should be `recipient` or `recipientUserName`

src/app/wallets/add-invoice-for-wallet.ts
```diff
@@ -0,0 +1,113 @@
...
+  memo: string
+  walletInvoiceCreateFn: WalletInvoiceFactoryCreateMethod
+}): Promise<LnInvoice | ApplicationError> => {
+  const walletInvoicesRepo = WalletInvoicesRepository()
+  const lndService = LndService()
+  if (lndService instanceof Error) return lndService
+
+  const registeredInvoice = await lndService.registerInvoice({
+    description: memo,
+    satoshis: toSats(amount),
```
bodymindarts, line 100: `toSats` should not be required here as it is already typed when passing in.

src/app/wallets/index.types.d.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+type AddInvoiceRecipientArgs = {
+  username: Username
+  amount: Satoshis
+  memo?: string
+}
+
+type AddInvoiceNoAmountRecipientArgs = {
+  username: Username
```
bodymindarts, line 19: rename to `recipient`

src/services/mongoose/users.ts
```diff
@@ -5,6 +5,8 @@ import {
   RepositoryError,
 } from "@domain/errors"
 import { User } from "@services/mongoose/schema"
+import { caseInsensitiveRegex } from "@core/utils"
```
bodymindarts, line 8: This function is only used by functions that should be part of the `UsersRepository`. I would move it into `mongoose/users.ts` file and out of utils.

src/services/mongoose/users.ts
```diff
@@ -5,6 +5,8 @@ import {
...
+import { getUsernameRegex } from "@config/app"
```
bodymindarts, line 9: This function should probably belong in domain.
We need a `checkedToUsername` that executes this validation  before passing it around. When we have an instance of the `Username` type we should assume it has already been validated.

bodymindarts, line 9: This function should probably belong in domain.
We need a `checkedToUsername` that executes this validation  before passing it around. When we have an instance of the `Username` type we should assume it has already been validated.

bodymindarts (changes requested):

src/app/wallets/add-invoice-for-wallet.ts
vindard: Changing to `recipient` because that's already used in some places

src/services/mongoose/users.ts
vindard: Implemented `checkedToUsername` on another branch where I already started implementing `checkedToSats: https://github.com/GaloyMoney/galoy/pull/455/commits/3b72eb17cfe12f90fd822e9c90d75e97c54d5011

src/services/mongoose/users.ts
vindard: Implementing here: https://github.com/GaloyMoney/galoy/compare/swap-in-new-add-invoice...swap-in-find-by-username

src/domain/user.ts
```diff
@@ -2,7 +2,7 @@ import { User } from "@services/mongoose/schema"
 import { baseLogger } from "@services/logger"
 
 import { NotFoundError } from "@core/error"
-import { caseInsensitiveRegex } from "@core/utils"
+import { caseInsensitiveRegex } from "@services/mongoose"
```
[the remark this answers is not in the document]
dolcalmi: ok just see what we are doing, are you sure that is safe to pass a regex without additional validation directly to mongoose findOne?  

Also, maybe is better to use [exists](https://mongoosejs.com/docs/api.html#model_Model.exists) or count in `usernameExists`

src/services/mongoose/users.ts
```diff
@@ -27,7 +31,33 @@ export const UsersRepository = (): IUsersRepository => {
...
   }
 
+  const findByUsername = async (username: Username): Promise<User | RepositoryError> => {
+    const regexUsername = getUsernameRegex()
+    try {
+      if (!username.match(regexUsername)) {
+        return new CouldNotFindError("Invalid username")
+      }
+
+      const result = await User.findOne({ username: caseInsensitiveRegex(username) })
```
dolcalmi, line 41: do you know what is the performance impact of use this? mongo indexes support this?

test/integration/02-user-wallet/02-invoice.spec.ts
```diff
@@ -13,20 +19,36 @@ beforeAll(async () => {
 
 describe("UserWallet - addInvoice", () => {
   it("adds a self generated invoice", async () => {
-    const request = await userWallet1.addInvoice({ value: 1000 })
+    const lnInvoice = await addInvoiceForSelf({
+      walletId: userWallet1.user.id as WalletId,
+      amount: toSats(1000),
+    })
+    if (lnInvoice instanceof Error) return lnInvoice
+    const { paymentRequest: request } = lnInvoice
```
dolcalmi, line 27: please add and additional validation or test for pubkey

dolcalmi (changes requested):

vindard:
:point_up: [`force-push`](https://github.com/GaloyMoney/galoy/compare/fb37fbef6d94b34c82e168020abfc2732d3ad571..9e58d8aecd9596dbb3b1fba5fc0363905b6c7a7e) to latest tip of `main`

src/domain/user.ts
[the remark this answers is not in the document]
vindard: Not sure about the behavior of this, this was just a simple move of existing code.

Interesting catch though, we can probably investigate this in a separate PR.

src/services/lnd/index.ts
```diff
@@ -2,10 +2,19 @@ import {
...
 } from "@domain/bitcoin/lightning"
 import { createInvoice } from "lightning"
+import { getActiveLnd } from "./utils"
+
+export const LndService = (): ILightningService | LightningServiceError => {
+  let lndAuth: AuthenticatedLnd, pubkey: string
+  try {
+    ;({ lnd: lndAuth, pubkey } = getActiveLnd())
+  } catch (err) {
+    return new LightningServiceError(err)
```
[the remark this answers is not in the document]
vindard: True, I think both of these can be `UnknownLightningServiceError`

src/services/mongoose/users.ts
vindard: https://github.com/GaloyMoney/galoy/pull/449#discussion_r691613163

test/integration/02-user-wallet/02-invoice.spec.ts
vindard: Would checking that it starts with `lnbc1` be enough for a validation?

src/app/wallets/add-invoice-for-wallet.ts
```diff
@@ -0,0 +1,113 @@
...
+export const addInvoiceNoAmountForSelf = async ({
+  walletId,
+  memo = "",
+}: AddInvoiceNoAmountSelfArgs): Promise<LnInvoice | ApplicationError> => {
+  const walletInvoiceFactory = WalletInvoiceFactory(walletId)
+  return baseAddInvoiceForWallet({
+    amount: toSats(0),
+    memo,
+    walletInvoiceCreateFn: walletInvoiceFactory.create,
+  })
```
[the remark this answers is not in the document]
vindard: True, but I think we have some preferences about duplicating code as well. @bodymindarts what do you think of abstracting like this here?

src/app/wallets/add-invoice-for-wallet.ts
```diff
@@ -0,0 +1,113 @@
...
+import { invoiceExpirationForCurrency } from "@domain/bitcoin/lightning"
+import { WalletInvoiceFactory } from "@domain/wallet-invoices/wallet-invoice-factory"
+import { LndService } from "@services/lnd"
+import {
+  WalletInvoicesRepository,
+  UsersRepository,
+  AccountsRepository,
+} from "@services/mongoose"
+
+export const addInvoiceForSelf = async ({
```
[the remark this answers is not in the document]
vindard: I'll apply this change in a later branch so that I don't confuse these with the existing `addInvoice` function

src/domain/users/index.ts
```diff
@@ -2,3 +2,5 @@ export const UserLanguage = {
   EN_US: "en",
   ES_SV: "es",
 } as const
+
+export const getUsernameRegex = () => /(?!^(1|3|bc1|lnbc1))^[0-9a-z_]+$/i
```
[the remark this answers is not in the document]
vindard: I think we're leaning to fetching things with functions when possible as a preference. @bodymindarts would have to advise

src/app/wallets/add-invoice-for-wallet.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  const walletId = defaultAccount.walletIds[0]
+
+  const walletInvoiceFactory = WalletInvoiceFactory(walletId)
+
+  return baseAddInvoiceForWallet({
+    amount: toSats(0),
+    memo,
+    walletInvoiceCreateFn: walletInvoiceFactory.createForRecipient,
+  })
```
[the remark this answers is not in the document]
vindard: https://github.com/GaloyMoney/galoy/pull/449#discussion_r691622004

src/app/wallets/add-invoice-for-wallet.ts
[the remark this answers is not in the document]
bodymindarts: For situations where we are implementing distinct use cases (and here these functions represent 2 different GQL mutations) I prefer to have it be explicit which arguments should be passed and which shouldn't. Its an extra static verification that we are using it correctly.

src/domain/users/index.ts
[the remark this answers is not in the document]
bodymindarts: I agree that this should be a constant defined close to where it is used (ideally only in `checkedToUsername`).

dolcalmi (approved):

src/app/wallets/add-invoice-for-wallet.ts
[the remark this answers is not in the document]
vindard: We decided on call to leave addInvoiceForSelf as is but to change addInvoiceForRecipient since there was a lot of duplicated code there

PR #454, opened 2021-08-17
Fix resolve current price

dolcalmi:
- resolve current price before add it to array
- replace splice for lodash takeright method

this only fix a small gap of the price so still need to figure out how to try to match the resolved `getCurrentPrice` when the invoice is paid

### DIFF
--- a/src/servers/graphql-old-server.ts
+++ b/src/servers/graphql-old-server.ts
@@ -11,6 +11,7 @@ import { and, shield } from "graphql-shield"
 import { makeExecutableSchema } from "graphql-tools"
 import moment from "moment"
 import path from "path"
+import { takeRight } from "lodash"
 
 import { levels, onboardingEarn, getTransactionLimits, getFeeRates } from "@config/app"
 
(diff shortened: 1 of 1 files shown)


src/servers/graphql-old-server.ts
```diff
@@ -11,6 +11,7 @@ import { and, shield } from "graphql-shield"
 import { makeExecutableSchema } from "graphql-tools"
 import moment from "moment"
 import path from "path"
+import { takeRight } from "lodash"
```
samerbuna, line 14: I'd like to eventually get rid of lodash, so please do:

`import takeRight from "lodash.takeright"`

src/servers/graphql-old-server.ts
```diff
@@ -98,16 +99,18 @@ const resolvers = {
...
+      const prices = await getHourlyPrice({ logger })
+      const currentPrice = await getCurrentPrice()
 
       // adding the current price as the lat index array
       // use by the mobile application to convert prices
-      hourly.push({
+      prices.push({
         id: moment().unix(),
-        o: getCurrentPrice(),
+        o: currentPrice,
```
samerbuna, line 109: This is better but I don't think it was a bug since this is fed to GraphQL and GraphQL resolvers are okay with a promise as the resolved value.

samerbuna (approved):
Besides the lodash comment, LGTM.

src/servers/graphql-old-server.ts
dolcalmi: could you please check in your local machine? I was getting the last hour price, if this should work then maybe is something in getCurrentPrice

dolcalmi:
moved to draft for further investigation

src/servers/graphql-old-server.ts
nicolasburtey: maybe your setup didn't fetch `getCurrentPrice()` because it uses another pod and it was not set up correctly on your machine/docker-compose?

PR #455, opened 2021-08-17
Swap in new addInvoice methods throughout codebase

vindard:
_Builds on #449_

---
`TODO`s:
- [x] implement [`checkedToSats` method](https://github.com/GaloyMoney/galoy/pull/449#discussion_r690366413) at graphql layer
- [ ] test the graphql changes somehow


src/services/mongoose/users.ts
```diff
@@ -31,13 +31,12 @@ export const UsersRepository = (): IUsersRepository => {
     }
   }
 
-  const findByUsername = async (username: Username): Promise<User | RepositoryError> => {
-    const regexUsername = getUsernameRegex()
+  const findByUsername = async (
+    username: Username,
+  ): Promise<User | RepositoryError | ValidationError> => {
```
vindard, line 36: Is it ok to return a `ValidationError` in a service method like this?

src/servers/graphql-old-server.ts
```diff
@@ -228,12 +229,24 @@ const resolvers = {
...
-    noauthAddInvoice: async (_, { username, value }, { logger }) => {
-      const wallet = await getWalletFromUsername({ username, logger })
-      return wallet.addInvoice({ selfGenerated: false, value })
+    noauthAddInvoice: async (_, { username, value }) => {
+      const lnInvoice = await addInvoiceForRecipient({
+        recipient: username,
+        amount: value,
+      })
+      if (lnInvoice instanceof Error) throw lnInvoice
+      return lnInvoice.paymentRequest
```
vindard, line 238: @samerbuna @dolcalmi is it ok to return a value here instead of a promise?

I just picked this up on rebase after #435 was merged

src/servers/graphql-old-server.ts
```diff
@@ -228,12 +229,24 @@ const resolvers = {
...
     invoice: (_, __, { wallet }) => ({
-      addInvoice: async ({ value, memo }) => wallet.addInvoice({ value, memo }),
+      addInvoice: async ({ value, memo }) => {
+        const lnInvoice = await addInvoiceForSelf({
+          walletId: wallet.user.id,
+          amount: value,
+          memo,
+        })
+        if (lnInvoice instanceof Error) throw lnInvoice
+        return lnInvoice.paymentRequest
```
vindard, line 248: Same question here -> https://github.com/GaloyMoney/galoy/pull/455/commits/197947d3529f877e74f966cd55e4efeb65f5ca16#r691635094

src/servers/graphql-old-server.ts
vindard: Ah nevermind, resolves in the same way: https://stackoverflow.com/a/42750371

PR #456, opened 2021-08-17
Refactor invoice user

dolcalmi:
this is a pre-requisite to user-account-wallet migration


src/app/lightning/payment-status-checker.ts
```diff
@@ -16,7 +16,7 @@ export const PaymentStatusChecker = ({ paymentRequest, lookupToken }) => {
 
   return {
     invoiceIsPaid: async (): Promise<boolean | RepositoryError> => {
-      const repo = InvoicesRepository()
+      const repo = WalletInvoicesRepository()
       const walletInvoice = await repo.findByPaymentHash(paymentHash)
       if (walletInvoice instanceof RepositoryError) return walletInvoice
```
bodymindarts, line 21: This should also use `isRepoError` or?

src/core/balance-sheet.ts
```diff
@@ -1,36 +1,50 @@
...
 
-export const updatePendingLightningTransactions = async () => {
-  // select distinct user ids from pending invoices
-  const usersWithPendingInvoices = InvoiceUser.aggregate([
-    { $match: { paid: false } },
-    { $group: { _id: "$uid" } },
-  ])
-    .cursor({ batchSize: 100 })
-    .exec()
+const isRepoError = (obj): boolean => obj instanceof RepositoryError
```
bodymindarts, line 14: If you want to use this in multiple places perhaps it should go in `src/domain/utils.ts`

src/core/balance-sheet.ts
```diff
@@ -1,36 +1,50 @@
...
-    processor: async ({ _id }, index) => {
-      logger.trace("updating pending invoices for user %s in worker %d", _id, index)
-      const user = await User.findOne({ _id })
+    processor: async (walletId, index) => {
+      logger.trace(
+        "updating pending invoices for wallet %s in worker %d",
+        walletId,
+        index,
+      )
+      const user = await User.findOne({ _id: walletId })
```
bodymindarts, line 38: Why not use `UsersRepository` here?

src/domain/wallet-invoices/index.types.d.ts
```diff
@@ -8,7 +8,10 @@ type WalletInvoice = {
 
 interface IWalletInvoicesRepository {
   persist: (invoice: WalletInvoice) => Promise<WalletInvoice | RepositoryError>
+
   findByPaymentHash: (
     paymentHash: PaymentHash,
   ) => Promise<WalletInvoice | RepositoryError>
+
+  findWalletsWithPendingInvoices: () => AsyncGenerator<WalletId> | RepositoryError
```
bodymindarts, line 16: The prefix `find` implies a criteria we are looking up. I think better would be `list` or `iter` for this function.

src/servers/trigger.ts
```diff
@@ -177,24 +181,25 @@ export const onInvoiceUpdate = async (invoice) => {
     return
   }
 
-  const invoiceUser = await InvoiceUser.findOne({ _id: invoice.id })
-  if (invoiceUser) {
-    const uid = invoiceUser.uid
-    const hash = invoice.id
+  const walletInvoicesRepo = WalletInvoicesRepository()
+  const invoiceWallet = await walletInvoicesRepo.findByPaymentHash(invoice.id)
```
bodymindarts, line 185: Name should be `walletInvoice`

src/services/mongoose/wallet-invoices.ts
```diff
@@ -49,5 +53,26 @@ export const WalletInvoicesRepository = (): IWalletInvoicesRepository => {
...
+        { $match: { paid: false } },
+        { $group: { _id: "$uid" } },
+      ])
+        .cursor({ batchSize: 100 })
+        .exec()
+    } catch (error) {
+      return new RepositoryError(error)
+    }
+
+    for await (const { _id } of pending) {
```
bodymindarts, line 72: Can't getting the next element ever throw?

dolcalmi: the return pattern only works in in the first part, this is a generator function and return a cursor, if that fails after open the cursor then will throw an exception

src/core/balance-sheet.ts
dolcalmi: done

src/core/balance-sheet.ts
dolcalmi: because this will not be user but wallet (after db migration)

bodymindarts: Ah right then `WalletRepository`. Its there too....

src/app/lightning/payment-status-checker.ts
dolcalmi: I don't want to touch the code that arvin is working on until merge (base branch of this PR is his current working branch), I just removed the duplicated repo by his request

src/core/balance-sheet.ts
dolcalmi: yep :) just re-doing the invoiceuser centralization to be able to add missing tests and then do the migration (more safe in this way), after that I will change related implementations to the new structure

src/core/lightning/index.ts
```diff
@@ -305,14 +311,20 @@ export const LightningMixin = (superclass) =>
...
 
-              const resultUpdate = await InvoiceUser.updateOne(
-                { _id: id },
-                { paid: true },
-              )
-              this.logger.info(
-                { id, user: this.user, resultUpdate },
-                "invoice has been updated from InvoiceUser following on_us transaction",
-              )
+              const isUpdated = await this.invoices.setPaidByPaymentHash(id)
```
bodymindarts, line 314: This is not the right way to use a repository.
It should be:
```
wallet = invoices.findByPaymentHash(hash)
wallet.paid = true
invoices.persist(wallet)
```

What you are doing here is putting use case code into the repository. But repository should only know about persistence details. What kind of data transformation is required to set a wallet as paid is business logic concern that belongs in domain / or if its trivial can be executed directly on domain objects.

dolcalmi: we already have some performance issues. how do you suggest to do it without affect more the performance?

bodymindarts: What are the performance issues?

src/domain/utils.ts
```diff
@@ -0,0 +1,7 @@
+import { RepositoryError, UnknownRepositoryError } from "./errors"
+
+export const isRepoError = (obj): boolean =>
```
bodymindarts, line 3: I have thought about it and don't think we should generally not be checking for specific errors inside the `domain/core` code.
Just:
```
const thing = serviceCall()
if (thing instances Error) {
  return thing
}
```

We should only check for a `SpecificError` when the type is actually relevant. For example when translating the errors back to the user in the GQL layer or something like that.

src/domain/utils.ts
```diff
@@ -0,0 +1,7 @@
...
+  obj instanceof RepositoryError || obj instanceof UnknownRepositoryError
+
+export const isWalletInvoice = (invoice): invoice is WalletInvoice =>
+  (invoice as WalletInvoice).paymentHash !== undefined
```
bodymindarts, line 7: This should not be needed! The type is defined as:
```
type WalletInvoice = {
  paymentHash: PaymentHash
  ...
}
```

We should always assume that when we have a `type` it is valid like the definition. If there is a way we can have a `WalletInvoice` that doesn't have a `paymentHash` then this is a bug and needs to be fixed at the root cause. Adding this method is a hack that masks the root cause. The whole point of using typescript is to avoid these noisy checks.

src/services/mongoose/wallet-invoices.ts
```diff
@@ -10,19 +14,17 @@ export const WalletInvoicesRepository = (): IWalletInvoicesRepository => {
...
     try {
-      await new InvoiceUser({
-        _id: paymentHash,
-        uid: walletId,
-        selfGenerated,
-        pubkey,
-        paid,
-      }).save()
+      const data = { uid: walletId, selfGenerated, pubkey, paid }
+      const doc = await InvoiceUser.findOneAndUpdate({ _id: paymentHash }, data, {
```
bodymindarts, line 18: We shouldn't reuse `persist` to update data. Better to introduce the pattern `persistNew` and `update` to all `Repositories`. This is more explicit. The calling code should always know which method is the right one to call.

src/services/mongoose/wallet-invoices.ts
```diff
@@ -49,5 +51,81 @@ export const WalletInvoicesRepository = (): IWalletInvoicesRepository => {
     }
   }
 
-  return { persist, findByPaymentHash }
+  async function* findPendingByWalletId(
```
bodymindarts, line 54: We have been using the `const functionName = async () => {}` style to define functions. I think we should be consistent.

src/services/mongoose/wallet-invoices.ts
```diff
@@ -49,5 +51,81 @@ export const WalletInvoicesRepository = (): IWalletInvoicesRepository => {
...
+    walletId: WalletId,
+  ): AsyncGenerator<WalletInvoice> | RepositoryError {
+    let pending
+    try {
+      pending = InvoiceUser.find({ uid: walletId, paid: false }).cursor({
+        batchSize: 100,
+      })
+    } catch (error) {
+      return new RepositoryError(error)
```
bodymindarts, line 63: Errors that we catch and wrap always have the `Unknown` prefix. Here it should be `UnknownRepositoryError`.

bodymindarts (changes requested):

src/core/lightning/index.ts
```diff
@@ -829,12 +827,17 @@ export const LightningMixin = (superclass) =>
...
+      if (isRepoError(pendingInvoicesRes)) {
+        this.logger.error(
+          { walletId: this.user._id, error: pendingInvoicesRes },
+          "finish updating pending invoices with error",
+        )
+        return
+      }
 
-      for (const { _id: hash, pubkey: pubkeyCached } of invoices) {
+      const invoices = pendingInvoicesRes as AsyncGenerator<WalletInvoice>
```
bodymindarts, line 839: if this cast is needed there is a problem somewhere.
I think if you use :
```
if (pendingInvoicesRes instanceof Error) {
}
```

instead of `isRepoError` you won't need these casts. We definitely shouldn't use casts this way because it will mask bugs that would be caught statically.

dolcalmi: cast is only for eslint, we cant just repeat code because eslint for TS does not work properly.

dolcalmi: the other solution is worst, declare `pendingInvoicesRes` and then assign (that fix the eslint problem and we dont need cast) do you prefer that?

src/services/mongoose/wallet-invoices.ts
dolcalmi: uhmm really dont know how to implement a generator function with arrow syntax, as far as I know is still a proposal

src/services/mongoose/wallet-invoices.ts
dolcalmi: so this is a general problem, in current use cases you are only validating `RepositoryError` and... `UnknownRepositoryError` and `CouldNotFindError` do not extends from `RepositoryError`  so... what do you prefer? extend repo errors from RepositoryError or use isRepoError to do validations

bodymindarts: Don't use `isRepoError` use `xxx instanceof Error`.
`UnknownRepositoryError` and `CouldNotFindError` should extend from `RepositoryError` if they don't this is a mistake.

Still we should only differentiate which error was being returned if this is meaningful (like translating in the GQL layer or something) otherwise always use `instanceof Error`

src/core/lightning/index.ts
bodymindarts: I don't understand... lets look at code review

src/domain/utils.ts
dolcalmi: I know but is the technique suggested to let TS compiler know/infer the type (after validation), I guess we will not need it because the error extend changes but maybe will be useful later... the syntax is `isWalletInvoice = (invoice): invoice is WalletInvoice =>` (check the weird return syntax)  but you need to check a property inside the method

src/services/mongoose/wallet-invoices.ts
dolcalmi: you suggested the current implementation previously https://github.com/GaloyMoney/galoy/pull/456#discussion_r692580344 

Please confirm that you want to change this

src/domain/utils.ts
dolcalmi: done

src/services/mongoose/wallet-invoices.ts
dolcalmi: done

PR #458, opened 2021-08-18
Captcha api protection

samerbuna:
Converting #366 to a local repo branch


samerbuna:
> Shoot, I just realized that because I integrated the auth code into the last release of the app that we can't make this breaking change to this endpoint anymore. Will need a different endpoint for requesting an authcode with geetest

Renamed

jrbuhl93 (approved):
Make sure we coordinate with @krtk6160 to get the environment variables for GeeTest in. We can used the ones I was using for testing for now, but when I upgrade to a paid plan on GeeTest they will need to be changed.

daviroo:
Hey @samerbuna. I think this change requires one more `yarn install` commit.  When I checked it out locally to work on the mobile app changes it registers a change to the `yarn.lock` file.  Not a massive issue but it will help with idempotency.

samerbuna:
> Hey @samerbuna. I think this change requires one more `yarn install` commit. When I checked it out locally to work on the mobile app changes it registers a change to the `yarn.lock` file. Not a massive issue but it will help with idempotency.

Good catch. Updated.

jrbuhl93 (approved):

src/graphql/main/schema.graphql
```diff
@@ -346,6 +346,8 @@ type Mutation {
   ): LnInvoicePaymentSendPayload!
   onChainAddressCreate(input: OnChainAddressCreateInput!): OnChainAddressPayload!
   onChainAddressCurrent(input: OnChainAddressCurrentInput!): OnChainAddressPayload!
+  captchaCreateChallenge: CaptchaCreteChallengePayload!
```
krtk6160, line 349: Typo. Should be `CaptchaCreateChallengePayload`

krtk6160 (dismissed):

src/graphql/main/schema.graphql
samerbuna: Good catch. Fixed.

samerbuna:
Rebased to resolve conflicts: @jrbuhl93 @krtk6160

src/graphql/root/mutation/captcha-request-auth-code.ts
```diff
@@ -0,0 +1,54 @@
...
+  args: {
+    input: { type: GT.NonNull(CaptchaRequestAuthCodeInput) },
+  },
+  resolve: async (_, args, { logger, ip, geetest }) => {
+    const {
+      phone,
+      challengeCode: geetestChallenge,
+      validationCode: geetestValidate,
+      secCode: geetestSeccode,
+    } = args.input
```
dolcalmi, line 28: does this need validation?
```js
for (const input of [...]) {
      if (input instanceof Error) {
      }
    }
```

src/graphql/root/mutation/captcha-request-auth-code.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    try {
+      // TODO: redo this with use-case and errors pattern
+      await verifyCaptchaAndReturnOTP({
```
dolcalmi, line 32: not all fails throw an exception, if `geetest.validate` fail it returns false

dolcalmi (changes requested):
I could only test `captchaCreateChallenge` because mobile implementation has not been finished.

src/graphql/root/mutation/captcha-request-auth-code.ts
samerbuna: There is an if statement in `verifyCaptchaAndReturnOTP` for that case

dolcalmi: yes, that is the point, it returns false (not an exception) and in that case we are returning `success: true`

dolcalmi (approved):

jrbuhl93 (approved):

PR #460, opened 2021-08-18
Refactor 'UserWallet' class/subclass constructor args typing

vindard:
_Moved from #424_

## Description

Small PR to clean up the constructor args in the following classes:
- `UserWallet`
- `LightningMixin`
- `OnChainMixin`
- `LightningUserWallet`

### DIFF
--- a/src/core/lightning/index.ts
+++ b/src/core/lightning/index.ts
@@ -53,9 +53,9 @@ export const LightningMixin = (superclass) =>
     readonly config: UserWalletConfig
     readonly invoices: IWalletInvoicesRepository
 
-    constructor(...args) {
-      super(...args)
-      this.config = args[0].config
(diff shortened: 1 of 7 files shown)


vindard:
We're getting `check-code` errors from a discrepancy between [`User`](https://github.com/GaloyMoney/galoy/blob/main/src/domain/users/index.types.d.ts#L7) and [`UserType`](https://github.com/GaloyMoney/galoy/blob/main/src/services/mongoose/schema.types.d.ts#L39). What's the difference between these two, and when do we use each?

bodymindarts:
`User` is our domain representation of the user fields. `UserType` data record that comes back from mongoose.

bodymindarts (approved):

dolcalmi (approved):

PR #462, opened 2021-08-19
New onchain adress create

bodymindarts:


src/app/wallets/get-transactions-for-wallet.ts
```diff
@@ -44,20 +42,19 @@ export const getTransactionsForWallet = async (
...
+    addresses,
   })
   const pendingTxs = filter.apply(onChainTxs)
 
-  // TODO should be a service - not a function call
-  let price = await getCurrentPrice()
-  if (typeof price !== "number") {
+  let price = await PriceService().getCurrentPrice()
+  if (price instanceof Error) {
     price = NaN
```
vindard, line 54: Minor nit for the type-checker
```suggestion
  if (price instanceof Error) {
    price = NaN as UsdPerSat
```

test/unit/utils.spec.ts
```diff
@@ -27,198 +21,6 @@ describe("utils.ts", () => {
     })
   })
 
-  describe("amountOnVout", () => {
```
vindard, line 30: Really interesting how we eliminate this entire set of extra logical steps previously needed

vindard (changes requested):
Left some small suggested changes around the type-checker and little typos. Otherwise looks good (& much cleaner) to me.

test/unit/utils.spec.ts
bodymindarts: its a combination of now using `tx-decoder` (which has unit tests) + having a simpler mapping of 1 utxo -> 1 ledger transaction instead of grouping utxos into 1 ledger transaction if they came from the same bitcoin transaction

bodymindarts:
Closing in favour of https://github.com/GaloyMoney/galoy/pull/527

PR #463, opened 2021-08-20
Swap in new 'addInvoice..' methods throughout codebase 

vindard:
## Description
This PR builds on the new `addInvoice..` use-case methods implemented in #449. It swaps in these new instances in favor of the old `addInvoice` implementation since the old implementation is no longer covered in our tests.

---
`TODO`s:
- [x] ~test the `graphql-old-server.ts` changes somehow. These two methods were changed:~
  - ~`noauthAddInvoice`~
  - ~`invoice.addInvoice`~
  _We decided we will try to limit these sorts of changes and merge optimistically with smoke tests and testnet testing to try to catch any inconsistencies._


vindard:
@dolcalmi @bodymindarts I uncovered a potential reason why we don't want to call exported use-case methods from other use-case methods.

I previously added the suggested changes here (https://github.com/GaloyMoney/galoy/pull/449/commits/4ccbc0ae52769dec2fcf7e0fbab62d2ee761ed87) but then after adding use-case validation in this PR (`checkedToSats` method) the tests where we called one use-case from the other one broke.

I instead abstracted the duplicated code into a helper function and reverted the use-case coupling here: https://github.com/GaloyMoney/galoy/commit/c0521f9dcd79f1a22bf05f109dd5cf88408c88ba

Does this sound right to you all as well?

src/servers/graphql-old-server.ts
```diff
@@ -228,12 +229,24 @@ const resolvers = {
         return true
       },
     }),
-    noauthAddInvoice: async (_, { username, value }, { logger }) => {
```
vindard, line 231: Double-checking that the `{ logger }` arg is ok to remove here. I didn't quite understand how this got here from reading the schema, but it's no longer needed in the function body (cc @samerbuna)

vindard:
Wrt `username` -> `walletname` change: To avoid too many conflicts with this work I'm thinking we can review this PR as is and then implement the change after this.

I already have a branch from this with open work on switching over the `getUserByUsername` method. We could potentially do this switch there as well -> https://github.com/GaloyMoney/galoy/compare/swap-in-new-add-invoice...swap-in-find-by-username

src/app/wallets/add-invoice-for-wallet.ts
```diff
@@ -8,73 +9,79 @@ import {
...
 
-export const addInvoiceForSelf = async ({
+export const addInvoice = async ({
   walletId,
   amount,
   memo = "",
-}: AddInvoiceSelfArgs): Promise<LnInvoice | ApplicationError> => {
+}: AddInvoiceArgs): Promise<LnInvoice | ApplicationError> => {
+  const sats = checkedToSats(amount)
+  if (sats instanceof Error) throw sats
```
bodymindarts, line 18: Don't want to throw here. If we start the pattern of using `checkedXxx` methods in the use case functions we need to return the information back to the caller so that GQL can respond back to the user what was wrong.

src/app/wallets/index.types.d.ts
```diff
@@ -1,21 +1,21 @@
-type AddInvoiceSelfArgs = {
+type AddInvoiceArgs = {
```
bodymindarts, line 1: Since we are introducing `checkedXxx` conversion in the user case functions all the methods should accept `amount: number`.

src/servers/graphql-old-server.ts
bodymindarts: you can just remove it: `(_, { username, value })`.

src/servers/graphql-old-server.ts
```diff
@@ -228,12 +229,24 @@ const resolvers = {
...
-      const wallet = await getWalletFromUsername({ username, logger })
-      return wallet.addInvoice({ selfGenerated: false, value })
+    noauthAddInvoice: async (_, { username, value }) => {
+      const lnInvoice = await addInvoiceByUsername({
```
bodymindarts, line 233: I think the function should be called `addInvoiceForRecipient`

test/integration/02-user-wallet/02-invoice.spec.ts
```diff
@@ -43,7 +43,7 @@ describe("UserWallet - addInvoice", () => {
   })
 
   it("adds a public invoice", async () => {
-    const lnInvoice = await addInvoiceNoAmountForRecipient({
+    const lnInvoice = await addInvoiceNoAmountByUsername({
```
bodymindarts, line 46: also not `ByUsername` -> `ForRecipient`

bodymindarts (approved):
lG!

dolcalmi (changes requested):
this changes are returning an error in [lnpage](https://github.com/GaloyMoney/lnpage)

query (http://localhost:3000/user0):
```json
{
   "operationName":"noauthAddInvoice",
   "variables":{
      "username":"user0"
   },
   "query":"mutation noauthAddInvoice($username: String!) {\n  noauthAddInvoice(username: $username)\n}\n"
}
```
response:
```json
{
   "errors":[
      {
         "message":"",
         "locations":[
            {
               "line":2,
               "column":3
            }
         ],
         "path":[
            "noauthAddInvoice"
         ],
         "extensions":{
            "code":"INTERNAL_SERVER_ERROR",
            "exception":{
               "name":"InvalidSatoshiAmount",
               "stacktrace":[
                  "InvalidSatoshiAmount: ",
                  "    at Object.checkedToSats (/galoy/galoy-backend/src/domain/bitcoin/index.ts:8:39)",
                  "    at Object.addInvoiceForRecipient (/galoy/galoy-backend/src/app/wallets/add-invoice-for-wallet.ts:47:16)",
                  "    at noauthAddInvoice (/galoy/galoy-backend/src/servers/graphql-old-server.ts:233:31)",
                  "    at field.resolve (/galoy/galoy-backend/node_modules/@profusion/apollo-validation-directives/build/lib/ValidateDirectiveVisitor.js:322:24)",
                  "    at /galoy/galoy-backend/node_modules/graphql-middleware/src/applicator.ts:35:9",
                  "    at middleware (/galoy/galoy-backend/node_modules/graphql-shield/src/generator.ts:70:22)",
                  "    at processTicksAndRejections (internal/process/task_queues.js:95:5)"
               ]
            }
         }
      }
   ],
   "data":{
      "noauthAddInvoice":null
   }
}
```

dolcalmi:
@jrbuhl93 would you mind check this with mobile app?

src/servers/graphql-old-server.ts
```diff
@@ -228,12 +229,24 @@ const resolvers = {
         return true
       },
     }),
-    noauthAddInvoice: async (_, { username, value }, { logger }) => {
-      const wallet = await getWalletFromUsername({ username, logger })
-      return wallet.addInvoice({ selfGenerated: false, value })
+    noauthAddInvoice: async (_, { username, value }) => {
+      const lnInvoice = await addInvoiceForRecipient({
```
dolcalmi, line 233: I guess this is the problem. `value` can be null/undefined

test/integration/02-user-wallet/02-invoice.spec.ts
```diff
@@ -32,7 +32,7 @@ describe("UserWallet - addInvoice", () => {
   })
 
   it("adds a self generated invoice without amount", async () => {
-    const lnInvoice = await addInvoiceNoAmountForSelf({
+    const lnInvoice = await addInvoiceNoAmount({
```
dolcalmi, line 35: the error was not detected because this was addInvoice, so while we move to the new graphql all previous related tests must be done at graphql level. please add them to `test/integration/servers/graphql-main-server.spec.ts` or add an `addInvoice` wrapper method with the logic to accept null/undefinied amount

dolcalmi (changes requested):

src/servers/graphql-old-server.ts
vindard: Yup looks like it's a problem with how we're testing with use-case separation being incompatible with the old schema.

I can look into a fix for this

test/integration/02-user-wallet/02-invoice.spec.ts
vindard: I suspected this would be a problem, check out the one open task included in this PR's description :sweat_smile:

vindard: Should be fixed here: https://github.com/GaloyMoney/galoy/pull/463/commits/1ab2d83b0a320c03031f0b8f1bb3d0bd522a508e

bodymindarts (approved):

dolcalmi (approved):

PR #464, opened 2021-08-20
re-implement updateOnChainReceipt

bodymindarts:
Bringing `updateOnChainReceipt` into the new application design.


src/services/ledger/index.ts
```diff
@@ -53,12 +56,66 @@ export const LedgerService = (): ILedgerService => {
...
+      const entry = MainBook.entry("")
+        .credit(liabilitiesAccountId, sats - fee, metadata)
+        .debit(lndAccountingPath, sats, metadata)
+
+      if (fee > 0) {
+        const bankOwnerPath = await bankOwnerAccountPath()
+        entry.credit(bankOwnerPath, fee, metadata)
+      }
+
+      await entry.commit()
```
dolcalmi, line 114: is not returning, if we will not migrate this then just return a boolean

src/services/notifications/index.ts
```diff
@@ -0,0 +1,27 @@
...
+export const NotificationsService = (logger: Logger): INotificationsService => {
+  const onChainTransactionReceived = async ({
+    amount,
+    walletId,
+    txId,
+  }: TransactionReceivedArgs) => {
+    try {
+      // work around to move forward before re-wrighting the whole notifications module
+      const user = await User.findOne({ _id: walletId })
+      transactionNotification({
```
dolcalmi, line 15: I think this should be called with await, otherwise we may have strange behavior (finishing after main process finish).

Also applies the same comment of previous method without return

src/services/price/index.ts
```diff
@@ -0,0 +1,15 @@
+import { PriceServiceError } from "@domain/price"
+import { getCurrentPrice } from "@services/realtime-price"
+
+export const PriceService = (): IPriceService => {
+  const fetchPrice = async () => {
+    const price = await getCurrentPrice()
```
dolcalmi, line 6: this method can throw exceptions, please use try/catch and return PriceServiceError

src/core/balance-sheet.ts
```diff
@@ -24,8 +26,13 @@ export const updatePendingLightningTransactions = async () => {
     processor: async ({ _id }, index) => {
       logger.trace("updating pending invoices for user %s in worker %d", _id, index)
       const user = await User.findOne({ _id })
-      const userWallet = await WalletFactory({ user, logger })
-      await userWallet.updatePendingInvoices()
+      const result = await Wallets.updateOnChainReceipt(user.id, logger)
```
dolcalmi, line 29: previously we needed the user because was a param of WalletFactory but with this new structure you can just use `_id` param and remove the extra query `const user = await User.findOne({ _id })`

dolcalmi (changes requested):

src/core/balance-sheet.ts
bodymindarts: The `_id` param is an object. The `id` param is the string.

src/app/wallets/update-on-chain-receipt.ts
```diff
@@ -0,0 +1,90 @@
...
+
+import { redlock } from "@core/lock"
+
+import { LOOK_BACK } from "@core/utils"
+import { ONCHAIN_MIN_CONFIRMATIONS, BTC_NETWORK } from "@config/app"
+
+export const updateOnChainReceipt = async (
+  walletId: WalletId,
+  logger: Logger,
+): Promise<void | ApplicationError> => {
```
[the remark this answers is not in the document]
bodymindarts: Haven't discussed it - should discuss at standup.

src/services/notifications/index.ts
nicolasburtey: this is actually intention. notification can be "trigger and forget", are optional, and running them in parrallel help having a faster api response

src/app/wallets/update-on-chain-receipt.ts
```diff
@@ -0,0 +1,90 @@
...
+    return price
+  }
+  const liabilitiesAccountId = toLiabilitiesAccountId(wallet.id)
+
+  return redlock({ path: wallet.id, logger }, async () => {
+    for (const tx of pendingTxs) {
+      const recorded = await ledger.isOnChainTxRecorded(liabilitiesAccountId, tx.id)
+      if (recorded instanceof Error) {
+        logger.error({ error: recorded }, "Could not query ledger")
+        return recorded
```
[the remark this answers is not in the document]
bodymindarts: Would prefer to have it error completely than partially process.

src/app/wallets/update-on-chain-receipt.ts
```diff
@@ -0,0 +1,90 @@
...
+              txId: tx.id,
+              sats,
+              fee,
+              usd,
+              usdFee,
+              receivingAddress: address,
+            })
+            if (result instanceof Error) {
+              logger.error({ error: result }, "Could not record onchain tx in ledger")
+              return result
```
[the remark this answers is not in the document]
bodymindarts: Would prefer to have it error completely than partially process.

src/core/balance-sheet.ts
dolcalmi: the param of processor `processor: async ({ _id }, index) => {` not user.id what I mean is that you can remove `const user = await User.findOne({ _id })`

src/core/balance-sheet.ts
```diff
@@ -21,11 +23,15 @@ export const updatePendingLightningTransactions = async () => {
   await runInParallel({
     iterator: usersWithPendingInvoices,
     logger,
-    processor: async ({ _id }, index) => {
-      logger.trace("updating pending invoices for user %s in worker %d", _id, index)
-      const user = await User.findOne({ _id })
-      const userWallet = await WalletFactory({ user, logger })
-      await userWallet.updatePendingInvoices()
+    processor: async ({ id }, index) => {
```
dolcalmi, line 26: you cant change to id, the param is _id because is an aggregate function (please check line 18)

dolcalmi (changes requested):

src/core/balance-sheet.ts
bodymindarts: I've taken the commit changing this out of the history. I'd prefer to use the

```
const user = await User.findOne({ _id })
... user.id
```
method for now. I know this is less efficient but I am not sure wether or not passing `_id` as `walletId` will always work. My understanding wast that `_id` was always referring to a reference `Object` and `id` was always a string.

dolcalmi (approved):

vindard (changes requested):
Nothing new to add to [my review](https://github.com/GaloyMoney/galoy/pull/462#pullrequestreview-736273040) of #462 and [this two-dot diff](https://github.com/GaloyMoney/galoy/compare/new-onchain-adress-create..new-update-on-chain-receipt) of the two branches. It looks like some of the comments/suggested changes from there apply here as well.

- https://github.com/GaloyMoney/galoy/pull/462#discussion_r694084413
- https://github.com/GaloyMoney/galoy/pull/462#discussion_r694098055
- https://github.com/GaloyMoney/galoy/pull/462#discussion_r694121133

PR #465, opened 2021-08-21
Format generated SDL schema files with Prettier

samerbuna:


src/services/fs.ts
```diff
@@ -0,0 +1,15 @@
+import fs from "fs/promises"
+import path from "path"
+import prettier from "prettier"
```
bodymindarts, line 3: Is there a way to only include this file when `NODE_ENV=development`? Don't want these dips to be in runtime during prod execution.

samerbuna: The PR is already doing that by importing `@services/fs` only if `isDev`

bodymindarts (approved):

krtk6160 (approved):

PR #467, opened 2021-08-22
Safer translation of lnds returned tx objects to SubmittedTransaction

bodymindarts:
The `createdAt` field was not being set correctly due to usage of casting.

This has triggered me to:
- remove all unused fields of SubmittedTransaction
- add unit tests around the conversion

### DIFF
--- a/src/domain/bitcoin/onchain/tx-filter.ts
+++ b/src/domain/bitcoin/onchain/tx-filter.ts
@@ -16,7 +16,7 @@ export const TxFilter = ({
       }
       if (
         !!addresses &&
-        !outs.some((out) => out.address != null && addresses.includes(out.address))
+        !outs.some((out) => out.address !== null && addresses.includes(out.address))
       ) {
         return false
       }
--- a/src/domain/bitcoin/onchain/tx-filter.ts
+++ b/src/domain/bitcoin/onchain/tx-filter.ts
@@ -4,7 +4,7 @@ export const TxFilter = ({
   addresses,
 }: TxFilterArgs): TxFilter => {
   const apply = (txs: SubmittedTransaction[]): SubmittedTransaction[] => {
-    return txs.filter(({ confirmations, outputAddresses }) => {
+    return txs.filter(({ confirmations, rawTx: { outs } }) => {
       if (
         !!confirmationsGreaterThanOrEqual &&
         confirmations < confirmationsGreaterThanOrEqual
@@ -14,7 +14,10 @@ export const TxFilter = ({
       if (!!confirmationsLessThan && confirmations >= confirmationsLessThan) {
         return false
       }
-      if (!!addresses && !outputAddresses.some((addr) => addresses.includes(addr))) {
+      if (
+        !!addresses &&
+        !outs.some((out) => out.address != null && addresses.includes(out.address))
+      ) {
         return false
       }
       return true
--- a/src/services/lnd/onchain-service.ts
+++ b/src/services/lnd/onchain-service.ts
@@ -31,7 +31,7 @@ export const OnChainService = (
       })
 
       return transactions
-        .filter((tx) => !tx.is_outgoing || !!tx.transaction || !!tx.fee)
+        .filter((tx) => !tx.is_outgoing && !!tx.transaction && !!tx.fee && tx.fee >= 0)
         .map((tx): SubmittedTransaction => {
           return {
             confirmations: tx.confirmation_count || 0,
--- a/src/services/lnd/onchain-service.ts
+++ b/src/services/lnd/onchain-service.ts
@@ -32,17 +32,18 @@ export const OnChainService = (
 
       return transactions
         .filter((tx) => !tx.is_outgoing || !!tx.transaction)
-        .map((tx): SubmittedTransaction => {
-          return {
-            confirmations: tx.confirmation_count || 0,
-            fee: toSats(tx.fee as number),
-            id: tx.id as TxId,
-            outputAddresses: tx.output_addresses as OnChainAddress[],
-            tokens: toSats(tx.tokens),
-            rawTx: decoder.decode(tx.transaction as string),
-            createdAt: new Date(tx.created_at),
-          }
-        })
+        .map(
+          (tx): SubmittedTransaction => {
+            return {
+              confirmations: tx.confirmation_count || 0,
+              id: tx.id as TxId,
+              outputAddresses: tx.output_addresses as OnChainAddress[],
+              tokens: toSats(tx.tokens),
+              rawTx: decoder.decode(tx.transaction as string),
+              createdAt: new Date(tx.created_at),
+            }
+          },
+        )
     } catch (err) {
       return new UnknownOnChainServiceError(err)
     }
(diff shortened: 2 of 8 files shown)


src/services/lnd/onchain-service.ts
```diff
@@ -25,19 +25,23 @@ export const OnChainService = (
...
+        .map((tx): SubmittedTransaction => {
           return {
             confirmations: tx.confirmation_count || 0,
             fee: toSats(tx.fee as number),
             id: tx.id as TxId,
             outputAddresses: tx.output_addresses as OnChainAddress[],
             tokens: toSats(tx.tokens),
             rawTx: decoder.decode(tx.transaction as string),
-          } as SubmittedTransaction
+            createdAt: new Date(tx.created_at),
```
dolcalmi, line 43: please use default values for transaction and fee (as confirmation_count) they are optional too https://github.com/alexbosworth/ln-service#getchaintransactions

dolcalmi (changes requested):

src/services/lnd/onchain-service.ts
bodymindarts: Checking the transaction is in the `filter` statement a few lines above:
```
34       .filter((tx) => !tx.is_outgoing || !!tx.transaction)
```

The `fee` field is never being read so can be ignored - I will remove it to be sure.

dolcalmi: we use it in onChannelUpdated. We need to keep it if we want to re-use this method

bodymindarts: I'm just making a commit that derives as much as possible from the raw transaction hex. That we we do not rely on the return of the lnd api.

dolcalmi: well not sure if i was not clear, the trigger use `onChannelUpdated` to debit fee from ledger `addLndChannelFee`. `onChannelUpdated` use the same code (`getChainTransactions` logic)  to get the fee. If you remove the `fee` then we will need to add it again

bodymindarts: Yes I understood this and tried to derive it from the txHex. But this isn't possible because the utxo value of the previous tx is not encoded so will just ensure `!!tx.fee` in the filter (like with transaction).

src/services/lnd/onchain-service.ts
```diff
@@ -25,19 +25,20 @@ export const OnChainService = (
...
-      const { transactions } = await getChainTransactions({ lnd, after })
+      const { transactions }: GetChainTransactionsResult = await getChainTransactions({
+        lnd,
+        after,
+      })
 
       return transactions
-        .filter((tx) => !tx.is_outgoing || !!tx.transaction)
-        .map((tx) => {
+        .filter((tx) => !tx.is_outgoing || !!tx.transaction || !!tx.fee)
```
dolcalmi, line 34: if this method is only for incoming transactions then filter should be something like `!tx.is_outgoing && (tx.transaction && tx.fee >= 0`

src/services/lnd/onchain-service.ts
```diff
@@ -25,19 +25,20 @@ export const OnChainService = (
...
-      const { transactions } = await getChainTransactions({ lnd, after })
+      const { transactions }: GetChainTransactionsResult = await getChainTransactions({
+        lnd,
+        after,
+      })
 
       return transactions
-        .filter((tx) => !tx.is_outgoing || !!tx.transaction)
-        .map((tx) => {
+        .filter((tx) => !tx.is_outgoing && !!tx.transaction && !!tx.fee && tx.fee >= 0)
```
dolcalmi, line 34: in regards to documentation fee is number so `!!tx.fee` will return false if fee is zero, I know is an exceptional case but happened to me (I setup a receiving address in a mining pool that included its own transactions with zero fee).

Also I just check `null >= 0` in node console and is returning `true` --_-- so we cant use it. So maybe we should use default value before pass it to toSats or use something like `(tx.fee || 0) >= 0`

src/services/lnd/onchain-service.ts
bodymindarts: Added a unit test to handle all cases.

src/services/lnd/onchain-service.ts
bodymindarts: Added a unit test to handle all cases

dolcalmi (dismissed):

src/domain/bitcoin/onchain/tx-filter.ts
```diff
@@ -14,7 +14,10 @@ export const TxFilter = ({
       if (!!confirmationsLessThan && confirmations >= confirmationsLessThan) {
         return false
       }
-      if (!!addresses && !outputAddresses.some((addr) => addresses.includes(addr))) {
+      if (
+        !!addresses &&
+        !outs.some((out) => out.address != null && addresses.includes(out.address))
```
samerbuna, line 19: Minor nitpick: `!==` instead of `!=` (We should add an ESLint rule for this).

Bigger point: the name `outs` is too generic? I'd have no idea what it means reading this code with no context.

src/services/lnd/onchain-service.ts
```diff
@@ -48,3 +44,19 @@ export const OnChainService = (
     getIncomingTransactions,
   }
 }
+
+export const extractIncomingTransactions = (
+  decoder: TxDecoder,
+  { transactions }: GetChainTransactionsResult,
+): SubmittedTransaction[] => {
+  return transactions
+    .filter((tx) => !tx.is_outgoing && !!tx.transaction)
```
samerbuna, line 53: Can we have a descriptive name for this filtering? What dose `!tx.is_outgoing && !!tx.transaction` mean?

Can we also possibly do the filtering/mapping in one loop? Either with `reduce` or `flatMap`

src/services/lnd/onchain-service.ts
```diff
@@ -24,21 +24,17 @@ export const OnChainService = (
...
-            tokens: toSats(tx.tokens),
-            rawTx: decoder.decode(tx.transaction as string),
-          } as SubmittedTransaction
-        })
+      // this is necessary for tests, otherwise after may be negative
+      const after = Math.max(0, current_block_height - scanDepth)
+
+      return extractIncomingTransactions(
+        decoder,
+        await getChainTransactions({
```
krtk6160, line 33: Minor nitpick -> Making the async function call here makes the code less readable than making the call on a separate line, and passing the result here

src/domain/bitcoin/onchain/tx-filter.ts
bodymindarts: Output is the correct name for this data in the context of a bitcoin transaction.

EG:
- [en.bitcoin.it/wiki: output](https://en.bitcoin.it/wiki/Transaction#Output)
- [rust-bitcoin: output](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/src/blockdata/transaction.rs#L271)
- [bitcoinjs-lib: outs](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/types/transaction.d.ts#L26)

All bitcoin transactions are made up of inputs and outputs.

PR #475, opened 2021-08-24
Switch username to walletname in new code

vindard:
## Description

We will be moving the naming reference from the user level to the wallet level. This PR is to switch all references of `username` to `walletname` in the new code, but to leave the old code intact for now so that now functionality changes or migrations are needed.

_**Note:** I didnt remove the `Username` type and `username` property our `User` type because that one goes all the way through to the new graphql schema._

### DIFF
--- a/src/domain/errors.ts
+++ b/src/domain/errors.ts
@@ -12,3 +12,4 @@ export class CouldNotFindError extends DomainError {}
 export class ValidationError extends DomainError {}
 export class InvalidSatoshiAmount extends ValidationError {}
 export class InvalidUsername extends ValidationError {}
+export class InvalidWalletname extends ValidationError {}
--- a/src/domain/errors.ts
+++ b/src/domain/errors.ts
@@ -11,4 +11,4 @@ export class CouldNotFindError extends DomainError {}
 
 export class ValidationError extends DomainError {}
 export class InvalidSatoshiAmount extends ValidationError {}
-export class InvalidWalletname extends ValidationError {}
+export class InvalidWalletName extends ValidationError {}
--- a/src/domain/errors.ts
+++ b/src/domain/errors.ts
@@ -11,5 +11,4 @@ export class CouldNotFindError extends DomainError {}
 
 export class ValidationError extends DomainError {}
 export class InvalidSatoshiAmount extends ValidationError {}
-export class InvalidUsername extends ValidationError {}
 export class InvalidWalletname extends ValidationError {}
--- a/src/domain/ledger/index.types.d.ts
+++ b/src/domain/ledger/index.types.d.ts
@@ -29,7 +29,7 @@ type LedgerTransaction = {
   readonly feeUsd: number
 
   // for IntraLedger
-  readonly walletname?: Walletname
+  readonly walletName?: WalletName
   readonly memoFromPayer?: string
 
   // for ln
--- a/src/domain/ledger/index.types.d.ts
+++ b/src/domain/ledger/index.types.d.ts
@@ -29,7 +29,7 @@ type LedgerTransaction = {
   readonly feeUsd: number
 
   // for IntraLedger
-  readonly username?: Username
+  readonly username?: Walletname
   readonly memoFromPayer?: string
 
   // for ln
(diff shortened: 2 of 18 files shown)


src/domain/errors.ts
```diff
@@ -12,3 +12,4 @@ export class CouldNotFindError extends DomainError {}
 export class ValidationError extends DomainError {}
 export class InvalidSatoshiAmount extends ValidationError {}
 export class InvalidUsername extends ValidationError {}
```
bodymindarts, line 14: Is `InvalidUsername` still used somewhere?

src/domain/primitives/index.types.d.ts
```diff
@@ -13,6 +13,9 @@ type Pubkey = string & { [pubkeySymbol]: never }
 declare const walletIdSymbol: unique symbol
 type WalletId = string & { [walletIdSymbol]: never }
 
+declare const walletnameSymbol: unique symbol
+type Walletname = string & { [walletnameSymbol]: never }
```
bodymindarts, line 17: I'm wondering if we shouldn't use `WalletName` ie camel case

src/services/mongoose/wallets.ts
```diff
@@ -12,27 +13,57 @@ export const WalletsRepository = (): IWalletsRepository => {
...
+    findByWalletname,
+  }
+}
+
+const resultToWallet = (result: UserType): Wallet => {
+  const walletId = result.id as WalletId
+
+  const walletname: Walletname = result.username
+    ? (result.username as Walletname)
+    : ("" as Walletname)
```
bodymindarts, line 48: from talking with @samerbuna about this a while ago I think using `null` for absence of a string is probably better.

bodymindarts (dismissed):
In the name of moving forward I'm approving but have some non-critical suggestions. You should decide wether or not you want to address them / wait for another opinion or merge after another approval.

src/domain/primitives/index.types.d.ts
vindard: It's an easy change if we'd like to do this now

dolcalmi:
did you consider how this change will impact mobile app? if we add a wallet name for each wallet then user/account will have to set it up (will be not only a graphic design problem but also a management problem for corporate accounts)

src/services/mongoose/wallets.ts
vindard: Done here: https://github.com/GaloyMoney/galoy/pull/475/commits/beec2a611a6434636fe0ae7d9ba5f59214b9bec7

src/domain/ledger/index.types.d.ts
```diff
@@ -29,7 +29,7 @@ type LedgerTransaction = {
   readonly feeUsd: number
 
   // for IntraLedger
-  readonly username?: Username
+  readonly walletName?: WalletName
```
dolcalmi, line 32: are you sure if you do this will not affect current data? I guess you will need to add a migration

dolcalmi (changes requested):
@vindard please search for `.username` there are some pending changes

src/domain/ledger/index.types.d.ts
vindard: I was careful to isolation these `LedgerTransaction` changes to one commit (https://github.com/GaloyMoney/galoy/pull/475/commits/f5e7786575da2d77690b704273325b66a674ab36) to double-check that we don't need migrations.

So far, this type is only used in two places:
- input of `fromLedger` where it gets translated to `ConfirmedTransactionHistory` type so `username`-naming is irrelevant
- return from `getLiabilityTransactions`, which only gets used as the input to the `fromLedger` method

vindard:
> @vindard please search for `.username` there are some pending changes

Those are all calls on `UserType` objects that come straight from our db. We haven't changed anything at the db level with this PR, only internal model-level changes. That `UserType` won't exist in the new code when we move all these instances across to using our own `User` types instead, accessed via the various repositories.

bodymindarts (approved):

dolcalmi (approved):

PR #478, opened 2021-08-25
Feature upgrade lnd to 0.13.1

twshelton:
Upgrade lightning package to 0.13.1 and update health checks to use getWalletStatus instead of getWalletInfo.

NOTE:  dev/lnd/lnd.conf required changes to get this branch to compile.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "ioredis": "^4.27.6",
     "js-yaml": "^4.1.0",
     "jsonwebtoken": "^8.5.1",
-    "lightning": "^3.3.14",
+		"lightning": "github:alexbosworth/lightning",
     "ln-service": "^51.8.7",
     "lodash": "^4.17.21",
     "medici": "^4.0.2",
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "ioredis": "^4.27.6",
     "js-yaml": "^4.1.0",
     "jsonwebtoken": "^8.5.1",
-		"lightning": "github:alexbosworth/lightning",
+    "lightning": "^4.1.0",
     "ln-service": "^51.8.7",
     "lodash": "^4.17.21",
     "medici": "^4.0.2",
--- a/src/services/lnd/health.ts
+++ b/src/services/lnd/health.ts
@@ -1,8 +1,21 @@
-import { getWalletInfo } from "lightning"
+import { getWalletStatus } from "lightning"
 
 import { baseLogger } from "@services/logger"
 
 import { params } from "./auth"
+
+/*
+
+	Check the status of the wallet and emit current state
+
+	Wallet can be in the following states:
+		NON_EXISTING
+    LOCKED
+    UNLOCKED
+    RPC_ACTIVE
+    WAITING_TO_START
+
+*/
 
 /* eslint-disable @typescript-eslint/no-var-requires */
 const EventEmitter = require("events")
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -64,7 +64,7 @@
     "ioredis": "^4.27.6",
     "js-yaml": "^4.1.0",
     "jsonwebtoken": "^8.5.1",
-    "lightning": "^3.3.14",
+		"lightning": "github:alexbosworth/lightning",
```
bodymindarts, line 67: Whitespace + don't want to use main. If it is an unreleased version we should at least pin to a commit hash.

test/integration/services/lnd/health.spec.ts
```diff
@@ -0,0 +1,26 @@
+import moment from "moment"
+import { isUp, lndStatusEvent } from "@services/lnd/health"
+import { params } from "@services/lnd/unauth"
+import { baseLogger } from "@services/logger"
+
+beforeAll(done => {
+  done()
```
bodymindarts, line 7: How is this different from default behaviour?

test/integration/services/lnd/health.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+})
+
+afterAll(done => {
+  done()
+})
+
+describe("lndHealth", () => {
+  // this is a test health checks on lnd
+  it('should emit on started', function(){
+    let eventFired = true
```
bodymindarts, line 17: Looks like this variable is initialised to `true` so the test doesn't actually test anything.

src/services/lnd/health.ts
```diff
@@ -1,9 +1,22 @@
...
 
 import { params } from "./auth"
 
+/*
+
+	Check the status of the wallet and emit current state
+
+	Wallet can be in the following states:
+		NON_EXISTING
+    LOCKED
```
bodymindarts, line 13: This should be encoded in a type rather than a comment

src/services/lnd/health.ts
```diff
@@ -13,24 +26,23 @@ const isUpLoop = (param) => setInterval(() => isUp(param), refresh_time)
...
-  if (active && !param.active) {
+  if (state == "RPC_ACTIVE") {
     lndStatusEvent.emit("started", param)
-  }
-
-  if (!active && param.active) {
+  } else {
     lndStatusEvent.emit("stopped", param)
-  }
+	}
```
bodymindarts, line 45: Whitespace... I think you need to check the config of your editor to auto-run prettier on save.

bodymindarts (changes requested):

src/services/lnd/unauth.ts
```diff
@@ -0,0 +1,33 @@
...
+
+      // FIXME: should be inactive first
+      // find a way to mock this up for jest
+      // for now only trigger is active = false at start
+      // because trigger will start listening to lnd event on lnd start/restart
+      active: isTest || !isTrigger,
+    }
+  })
+
+export const params = addProps(_.sortBy(inputs, ["priority"]))
```
bodymindarts, line 33: I understand you duplicated this file from `lnd/auth.ts`. Its really ugly... (not because of you)... perhaps we can keep the ugliness together by also exporting `unauthenticatedParams` from the `lnd/auth.ts` file and add a switch to `addProps` for which auth function to call.

so something like:
```
addProps = (array, authenticated = true) => 
  array.map(() => {
    (...)
    lnd: authenticated ? authenticatedLndGrpc({ ...input, socket }).lnd, : unauthenticatedLndGrpc({ ...input, socket }).lnd
  }
```

That way at least it stays in 1 place.

dolcalmi (dismissed):

bodymindarts (approved):

bodymindarts (dismissed):

dolcalmi (approved):

bodymindarts (approved):

PR #489, opened 2021-08-26
Fix updateOnChainReceipt for multiple outputs in same tx

dolcalmi:
- Add lookback constants to config
- Update updateOnChainReceipt:
  - remove update for wallet id (was unnecessary)
  - update cron implementation: improve performance because now only query lnd once
  - now cron and trigger use the same method without performance impact
  - update test to include multiple outputs validation (different users) in the same blockchain transaction
- add debug script to update last 15 days


bodymindarts (dismissed):

src/app/wallets/update-on-chain-receipt.ts
```diff
@@ -6,14 +6,71 @@ import { NotificationsService } from "@services/notifications"
...
+  let wallets, addresses, walletId, txId, result
+  for (const tx of onChainTxs) {
+    txId = tx.rawTx.id
+
+    addresses = tx.rawTx.outs.reduce<OnChainAddress[]>((a, o) => {
+      if (o.address) a.push(o.address)
+      return a
+    }, [])
+
+    wallets = await walletRepo.findByAddresses(addresses)
```
dolcalmi, line 46: does not need cursor because the amount of wallets updated in a transaction maybe is less than the first batch in a cursor query

src/core/balance-sheet.ts
```diff
@@ -79,27 +79,13 @@ export const updateUsersPendingPayment = async ({
...
-      const result = await Wallets.updateOnChainReceipt(user.id, logger)
-      if (result instanceof Error) {
-        logger.error(
-          { userId: user.id, index },
-          "Could not updateOnChainReceipt from balance-sheet",
-        )
-      }
-    },
-  })
+  const txNumber = await Wallets.updateOnChainReceipt({ logger })
```
dolcalmi, line 82: in this way the lnd load will be reduced at minimum and we will only update users related to the onchain transactions

src/app/wallets/update-on-chain-receipt.ts
```diff
@@ -6,14 +6,71 @@ import { NotificationsService } from "@services/notifications"
...
+    if (wallets instanceof Error) {
+      logError({ walletId: null, txId, error: wallets })
+      continue
+    }
 
-import { LOOK_BACK } from "@core/utils"
-import { ONCHAIN_MIN_CONFIRMATIONS, BTC_NETWORK } from "@config/app"
+    for (const wallet of wallets) {
+      walletId = wallet.id
+      logger.warn({ walletId, txId }, "updating onchain receipt")
```
bodymindarts, line 54: This line makes the integration output very noisy

dolcalmi: yep, missing change to trace before push

bodymindarts (approved):

samerbuna (approved):

PR #492, opened 2021-08-27
Implement user contacts/questions and update mutations

samerbuna:


samerbuna:
> You've added some concepts that we should discuss first - mainly:
> 
> * `MerchantUser`
> * `AccountLevel`
> * `AccountStatus`

These are only in the admin schema (for now)

vindard (approved):
Looks ok code-wise to me.

I had some earlier questions myself on how we'll be handling `UserDetails` and the two map-location fields (`title` and `coordinates`). `MerchantUser` abstraction makes sense for now.

src/domain/users/index.types.d.ts
```diff
@@ -7,11 +7,29 @@ type UserLanguage =
...
+  question: QuizQuestion
+  completed: boolean
+}
+
 type User = {
   id: UserId
   username: Username | null
   phone: PhoneNumber
   language: UserLanguage
+  contacts: UserContact[]
```
dolcalmi, line 31: I have a doubt here, if we have contacts at user level and not at account level.. will we show the transactions for all wallets in contact info? should that screen be redesigned?

src/domain/users/index.types.d.ts
```diff
@@ -7,11 +7,29 @@ type UserLanguage =
...
+  quizQuestions: UserQuizQuestion[]
```
dolcalmi, line 32: same here, to what wallet the rewards will be added? (if the user is in multiple accounts)

src/graphql/admin/schema.graphql
```diff
@@ -1,29 +1,72 @@
...
+  userDetailsByUsername(username: WalletName!): MerchantUser!
 }
 
 enum AccountLevel {
   ONE
   TWO
 }
 
-type UserDetails {
+type MerchantUser {
```
dolcalmi, line 12: what is a merchant user? is there any difference with normal users?

src/domain/users/index.types.d.ts
samerbuna: Not sure I understand. Contact does not expose transactions.

src/domain/users/index.types.d.ts
samerbuna: Good question for when we support multiple wallets, but we'll need to convert the existing API (for now)

src/graphql/admin/schema.graphql
samerbuna: Yes. This is used in the admin schema only. It exposes the leve/status/map-info.

`MerchantUser` is a bad name though. I renamed it to `UserForAdmin` (locally)

dolcalmi (approved):

vindard (approved):

PR #506, opened 2021-08-30
Some fixes for Juan's branch

bodymindarts:
Key fixes:
- introduce `LockService` to wrap when `redlock` throws
- add `uniqueAddresses` as helper method on `SubmittedTransaction`
- fix unrelated typo on `PriceService`
- refactor the `update-on-chain-receipt` function to remove the double nested loop (now single loop)


src/domain/wallets/index.types.d.ts
```diff
@@ -82,5 +82,5 @@ type Wallet = {
 interface IWalletsRepository {
   findById(walletId: WalletId): Promise<Wallet | RepositoryError>
   findByWalletName(walletName: WalletName): Promise<Wallet | RepositoryError>
-  findByAddresses(addresses: string[]): Promise<Wallet[] | RepositoryError>
+  listByAddresses(addresses: string[]): Promise<Wallet[] | RepositoryError>
```
dolcalmi, line 85: in a previous PR you told me that find prefix should be used when we have a search criteria, list prefix when we have no params

src/services/lock/index.ts
```diff
@@ -0,0 +1,19 @@
+import { UnknownLockServiceError } from "@domain/lock"
+import { redlock } from "@core/lock"
```
dolcalmi, line 2: is this temporal right? this creates a circular reference

PR #525, opened 2021-08-31
Implement updateWalletContactAlias use-case

bodymindarts:
Fixing the `WalletContact` types in the new schema and aligning domain with gql types.


samerbuna (commented):
Add a commit (please review) but otherwise LGTM

PR #526, opened 2021-08-31
Implement new `updatePendingPayments` use case

vindard:
## Description

This PR adds a new use-case method to update pending payments. It replaces the existing `updatePendingPayments` method on the `LightningMixin` class that ends up on the `LightningUserWallet` class.

---
### Open `TODO`s:
- [x] Incorporate feedback captured in [this commit](https://github.com/GaloyMoney/galoy/commit/9b0d4a55a6e6337d53004b467c6004bf14110a88)
- [x] swap into tests and pass
- [x] swap into rest of code


vindard:
### Debugging notes:

The tests are failing [at this line](https://github.com/GaloyMoney/galoy/pull/526/files#diff-15e9812156b4ff1c33faaaf99ffcc368a8af2756ccb7a5e81acdcfc46782db9fR597) because the function isn't being picked up as "resolved" for the `waitFor` method to exit its `while` loop.

Inside the `updatePendingPayments` method being run, [at this line](https://github.com/GaloyMoney/galoy/pull/526/files#diff-06f65b85d9680b9109715256f0034490e0e981e8877e3deaff9864a6f0dbc873R31) `updatePendingPayment` runs and as far as I can tell successfully executes once (`pendingPayments` is an array of length 1). It gets the `lnPaymentLookup` successfully [at this line](https://github.com/GaloyMoney/galoy/pull/526/files#diff-06f65b85d9680b9109715256f0034490e0e981e8877e3deaff9864a6f0dbc873R68) and eventually gets to [this final return](https://github.com/GaloyMoney/galoy/pull/526/files#diff-06f65b85d9680b9109715256f0034490e0e981e8877e3deaff9864a6f0dbc873R131) without hitting any of the exit returns.

This is as far as I've gotten in troubleshooting so far.

vindard:
### Testing update:

1st bug was fixed with this [`force-push`](https://github.com/GaloyMoney/galoy/compare/ea670783130762012eebe5188600766a56620d45..e590954112974ea928a4f0e64965fa2c61b7088d) and we now successfully get out of the `waitFor`.

Next issue now is test is failing because the feeDifference (20 sats on 0.02 max_fee) isn't being properly applied

vindard:
### Debugging update 2:

More bugs fixed:
- https://github.com/GaloyMoney/galoy/commit/864daca6cf414de27445bea7d9624301ae823e9f
- https://github.com/GaloyMoney/galoy/commit/095cce7c432f2500a38fc4535323ad8bea6bb352
- https://github.com/GaloyMoney/galoy/commit/5b09fea093f47e92ace49f0a9bc98b13e9f4e7ed

**All direct tests now pass** for new `updatePendingPayments` method

src/app/wallets/update-pending-payments.ts
```diff
@@ -0,0 +1,188 @@
...
+  lock,
+}: {
+  walletId: WalletId
+  logger: Logger
+  lock?: DistributedLock
+}): Promise<void | ApplicationError> => {
+  const liabilitiesAccountId = toLiabilitiesAccountId(walletId)
+
+  const lockService = LockService()
+  await lockService.lockWalletId({ walletId, logger, lock }, async () => {
```
vindard, line 22: @bodymindarts I had to leave in the `walletId` argument to pass it forward to the `redlock`

src/app/wallets/update-pending-payments.ts
```diff
@@ -0,0 +1,187 @@
...
+    transactionType: "payment",
+    onUs: false,
+    payment: paymentLiabilityTx,
+  })
+
+  const lndService = LndService()
+  if (lndService instanceof Error) return lndService
+
+  const { paymentHash, pubkey } = paymentLiabilityTx
+  // If we had PaymentLedgerType => no need for checking the fields
```
vindard, line 56: We can potentially add things like this as `good first issues` to get open contributors familiar with how our typing works

src/core/lightning/index.ts
```diff
@@ -58,8 +53,11 @@ export const LightningMixin = (superclass) =>
           lock,
           logger: this.logger,
         }),
-        this.updatePendingPayments(lock),
-        super.updatePending(lock),
+        Wallets.updatePendingPayments({
```
vindard, line 56: Note: this method can return a `LedgerServiceError` but in instances like this we don't check for it and throw. Is this ok?

src/app/wallets/update-pending-payments.ts
```diff
@@ -0,0 +1,187 @@
...
+        paymentResult: paymentLiabilityTx,
+        feeDifference,
+        maxFee: paymentLiabilityTx.fee,
+        actualFee: roundedUpFee,
+        id: paymentHash,
+      },
+      "logging a fee difference",
+    )
+
+    const price = await PriceService().getCurrentPrice()
```
vindard, line 147: In cases where `updatePendingPayments` is locked and called together with `updatePendingInvoices`, can this price value retrieved diverge within the two implementations, and if yes is that ok?

Our existing implementation has logic for caching the price I believe (`setCurrentPrice` method)

src/core/lightning/index.ts
bodymindarts: When this is being substituted in the old code we should throw if its an error to mirror existing behaviour

src/app/wallets/update-pending-payments.ts
bodymindarts: I don't see this as an issue. In the existing implementation `setCurrentPrice` gets called earlier in the request/response cycle but I cannot see how this should have an effect on behaviour.

src/domain/ledger/fee-reimbursement.ts
```diff
@@ -0,0 +1,13 @@
+import { toSats } from "@domain/bitcoin"
+
+export const FeeReimbursement = (prepaidFee: Satoshis): FeeReimbursement => {
+  const getReimbursement = ({ actualFee }: { actualFee: Satoshis }): Satoshis | null => {
```
dolcalmi, line 4: why to deal with two types when you can just return 0? it does not break any rule and is easier and safe to validate

src/app/wallets/update-pending-payments.ts
```diff
@@ -0,0 +1,187 @@
...
+  if (status === PaymentStatus.Settled || status === PaymentStatus.Failed) {
+    const ledgerService = LedgerService()
+    const settled = await ledgerService.settlePendingLiabilityTransactions(paymentHash)
+    if (settled instanceof Error) {
+      paymentLogger.error({ error: settled }, "we didn't have any transaction to update")
+      return settled
+    }
+
+    // PaymentStatus.Failed transactions have null 'roundedUpFee'
+    if (status === PaymentStatus.Settled && roundedUpFee !== null) {
```
dolcalmi, line 86: same here, why not just set 0 (instead of null) and validate here `roundedUpFee > 0`

src/app/wallets/update-pending-payments.ts
```diff
@@ -0,0 +1,187 @@
...
+  paymentLiabilityTx,
+  lnPaymentLookup,
+  logger,
+}: {
+  paymentLiabilityTx: LedgerTransaction
+  lnPaymentLookup: LnPaymentLookup
+  logger: Logger
+}): Promise<void | ApplicationError> => {
+  const ledgerService = LedgerService()
+  const voided = await ledgerService.voidLedgerTransactionsForJournal(
```
dolcalmi, line 173: Journal is a concept of Medici, do we really want to keep that concept?

src/core/balance-sheet.ts
```diff
@@ -61,7 +61,10 @@ const updatePendingLightningPayments = async () => {
       )
       const user = await User.findOne({ _id: ledger.resolveAccountId(account) })
       const userWallet = await WalletFactory({ user, logger })
-      await userWallet.updatePendingPayments()
+      await Wallets.updatePendingPayments({
+        walletId: userWallet.user.id as WalletId,
```
dolcalmi, line 65: if you only need the wallet Id you dont need line 62, just use `ledger.resolveAccountId(account)` to get the walletId

src/core/lightning/index.ts
dolcalmi: I guess we need to throw an exception, or graphql server do the translation? this can break mobile app behavior. @jrbuhl93 could you confirm how do we handle exceptions in get balance?

src/core/lightning/index.ts
```diff
@@ -604,103 +602,4 @@ export const LightningMixin = (superclass) =>
...
-    async updatePendingPayments(lock) {
-      const query = {
-        type: "payment",
-        pending: true,
-      }
-      const count = await ledger.getAccountTransactionsCount(this.user.accountPath, query)
-
-      if (count === 0) {
-        return
-      }
```
dolcalmi, line 617: this is missing and is needed

src/domain/ledger/index.types.d.ts
```diff
@@ -34,6 +39,8 @@ type LedgerTransaction = {
 
   // for ln
   readonly paymentHash?: PaymentHash
+  readonly pubkey?: Pubkey
+  readonly feeKnownInAdvance?: boolean
```
bodymindarts, line 43: feeKnownInAdvance isn't undefined according to the [schema](https://github.com/GaloyMoney/galoy/blob/main/src/services/ledger/schema.ts#L71)

src/services/ledger/index.ts
```diff
@@ -39,26 +43,24 @@ export const LedgerService = (): ILedgerService => {
...
+      return results.map((tx) => translateToLedgerTx(tx))
+    } catch (err) {
+      return new UnknownLedgerError(err)
+    }
+  }
+
+  const listPendingPayments = async (
+    liabilitiesAccountId: LiabilitiesAccountId,
+  ): Promise<LedgerTransaction[] | LedgerError> => {
+    const type = LedgerTransactionType.Payment
```
bodymindarts, line 55: can be inlined `type: LedgerTransactionType.Payment`

test/helpers/ledger.ts
```diff
@@ -0,0 +1,6 @@
+import { Transaction } from "@services/ledger/schema"
+
+export const getAccountTransactionsCount = (account: string, query = {}) => {
```
bodymindarts, line 3: missing `async` marker I think... its returning a `Promise`

dolcalmi (changes requested):

src/domain/ledger/index.types.d.ts
vindard: There were instances in the db where a value wasn't set. Should we default that to `false` instead in the translation?

src/core/lightning/index.ts
jrbuhl93: Eseentially we get the balance as part of [this query](https://github.com/GaloyMoney/galoy-mobile/blob/2755d29462fdb76b37698efc2cf3ee63b636fc0d/app/graphql/query.ts#L15). If the query doesn't return a new balance, it should continue to just use the cached balance.

src/app/wallets/update-pending-payments.ts
```diff
@@ -0,0 +1,187 @@
...
+  roundedUpFee,
+  logger,
+}: {
+  liabilitiesAccountId: LiabilitiesAccountId
+  paymentLiabilityTx: LedgerTransaction
+  roundedUpFee: Satoshis
+  logger: Logger
+}): Promise<void | ApplicationError> => {
+  const { paymentHash } = paymentLiabilityTx
+  // If we had PaymentLedgerType => no need for checking the fields
```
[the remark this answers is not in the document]
vindard: This was to satisfy the type-checker at `Ln 153`

test/helpers/ledger.ts
vindard: Funnily enough this also worked without the `async` marker. Adding it for completeness though.

src/app/wallets/update-pending-payments.ts
```diff
@@ -0,0 +1,187 @@
...
+  })
+
+  const lndService = LndService()
+  if (lndService instanceof Error) return lndService
+
+  const { paymentHash, pubkey } = paymentLiabilityTx
+  // If we had PaymentLedgerType => no need for checking the fields
+  if (!paymentHash)
+    return new InconsistentDataError("paymentHash missing from payment transaction")
+  if (!pubkey) return new InconsistentDataError("pubkey missing from payment transaction")
```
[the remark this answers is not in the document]
bodymindarts: Agree - can use `assert` or `throw`. Commented on that in other occurrences

src/domain/ledger/fee-reimbursement.ts
bodymindarts: Fine with me

src/app/wallets/update-pending-payments.ts
bodymindarts: a `roundedUpFee` of `0` instead of `undefined` might trigger a false `feeReimbursement`, or?

src/app/wallets/update-pending-payments.ts
bodymindarts: If we can keep it internal to the `ledgerService` that would be great - but how do we then identify the transactions? `PaymentHash`?

src/core/lightning/index.ts
bodymindarts: Why is this needed? If `count === 0` then `ledgerService.listPendingPayments` won't return any payments and we won't do any processing which is the same result? Its a redundant check imo.

src/app/wallets/update-pending-payments.ts
```diff
@@ -0,0 +1,187 @@
...
+  lock,
+}: {
+  walletId: WalletId
+  logger: Logger
+  lock?: DistributedLock
+}): Promise<void | ApplicationError> => {
+  const liabilitiesAccountId = toLiabilitiesAccountId(walletId)
+
+  const lockService = LockService()
+  await lockService.lockWalletId({ walletId, logger, lock }, async () => {
```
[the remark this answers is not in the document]
vindard: Changed to the `return` here https://github.com/GaloyMoney/galoy/pull/526/commits/0eb61e3282d8ceac7b5391597300f2901ecdf04c

> also, in the previous implementation only a lock is done when there is pending transactions, this is specially important because this is used in get balance an lock remains 10 seconds if something goes wrong

`updatePendingPayments` '`lock `' is optional here and only explicitly passed in when called together with `updatePendingInvoices` ([here](https://github.com/GaloyMoney/galoy/pull/526/files#diff-a9beb68fa5b83bb4910b21c35b84a1690884a90dd35209705698f1529b0c6c04R56) & [here](https://github.com/GaloyMoney/galoy/pull/526/files#diff-b5c296b20773b0ef9f3b15889c5f73b142201ae5aeb62961de05340f64e177f4R43)). Is this what you mean?

src/domain/ledger/fee-reimbursement.ts
vindard: I was thinking to return `null` when the checks break (if `feeDifference < 0` or `> max_fee`). Returning `0` when the actual difference isn't 0 might be misleading?

src/core/lightning/index.ts
vindard: Agreed, we'd just iterate over an empty list and return which amounts to the same behavoiur.

src/domain/ledger/fee-reimbursement.ts
bodymindarts: Ah right - this reasoning also makes sense to me. I think either is fine. From the domain perspective `null` is standing in for some kind of `noReimbursementRequiredMarker`. But `0` is also fine to stand in for that because it literaly means `no reimbursement`

src/app/wallets/update-pending-payments.ts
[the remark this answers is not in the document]
vindard: Using `throw` would help with the type checker later in the code as well. `assert` doesn't narrows down types for the checker.

Changed 👍🏾

src/core/lightning/index.ts
vindard: Decided to `throw` to be consistent with the old implementation: https://github.com/GaloyMoney/galoy/pull/526/commits/23705555584cea6bdcf0b924e491641ed2a8fe3e

src/core/lightning/index.ts
dolcalmi: you are locking the account, is less expensive do the count  query. 

did you read this comment?
```
// we only lock the account if there is some pending payment transaction, which would typically be unlikely
// we're doing the the Transaction.find after the lock to make sure there is no race condition
// note: there might be another design that doesn't requiere a lock at the uid level but only at the hash level,
// but will need to dig more into the cursor aspect of mongodb to see if there is a concurrency-safe way to do it.
```

src/app/wallets/update-pending-payments.ts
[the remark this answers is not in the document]
dolcalmi: what I mean is that you are locking the account even when the user has not pending payments,  this is the slower option (and this is used in get balance) and less safe for example when we do Promise.all in updatePending (if cant get the lock it will throw)

src/app/wallets/update-pending-payments.ts
dolcalmi: yep, the solution is use payment hash to query one transaction, get the journalId and do the void. (we can leave this for db migration)

src/core/lightning/index.ts
bodymindarts: Yes this is a valid point - as a performance optimisation we should add the check and only lock if `count > 0`.

vindard: Done: https://github.com/GaloyMoney/galoy/pull/526/commits/2ef5e7d969157bbf501a983bb9fafef21f4054a2

src/app/wallets/update-pending-payments.ts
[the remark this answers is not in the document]
vindard: Done: https://github.com/GaloyMoney/galoy/pull/526/commits/2ef5e7d969157bbf501a983bb9fafef21f4054a2

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -590,12 +591,19 @@ describe("UserWallet - Lightning Pay", () => {
           }
         })
 
-        const userWallet1Path = userWallet1.user.accountPath
-        const query = { type: "payment", pending: true, voided: false }
```
vindard, line 594: I assume the `voided: false` here is redundant because in the new implementation we only void payments after settling (setting `pending: false`). All tests pass.

bodymindarts (approved):

dolcalmi (approved):

PR #527, opened 2021-09-01
Add createOnChainAddress/getLastOnChainAddress  use case

dolcalmi:
- [x] add basic structure
- [x] add tests
- [x] validate use of getActiveOnchainLnd
- [x] add graphql query/mutation


src/services/mongoose/wallets.ts
```diff
@@ -7,6 +7,27 @@ import { User } from "@services/mongoose/schema"
 import { caseInsensitiveRegex } from "./users"
 
 export const WalletsRepository = (): IWalletsRepository => {
+  const persistNewOnChainAddress = async (
```
bodymindarts, line 10: Can we stick to the:
```
wallet = repo.findById(id)
wallet.onChainAddressIdentifieds.push(wallet)
repo.update(wallet)
```
Pattern?

ie Repo is only used for retrieving and persisting _not_ updating specific fields? There are exceptions, like bulk update to avoid n+1 problem, but that doesn't apply here.

bodymindarts (commented):
FYI I have already converted this to the new use-case structure here: https://github.com/GaloyMoney/galoy/pull/462

Just tests aren't passing.

src/services/mongoose/wallets.ts
dolcalmi: if I do this then I will need to use a lock on user  (to avoid addresses lost), also we will loose previous ids and will affect re-indexing

so.. maybe another repo is better just for onchain addresses, what do you think? (we will migrate to this later)

bodymindarts: I like the idea to have a repo just for `onchainAddresses`!

src/app/wallets/get-last-on-chain-address.ts
```diff
@@ -0,0 +1,13 @@
+import { WalletOnChainAddressesRepository } from "@services/mongoose"
+import { createOnChainAddress } from "./create-on-chain-address"
+
+export const getLastOnChainAddress = async (
+  walletId: WalletId,
+): Promise<OnChainAddress | ApplicationError> => {
+  const onChainAddressesRepo = WalletOnChainAddressesRepository()
+  const lastOnChainAddress = await onChainAddressesRepo.findLastByWalletId(walletId)
+
+  if (lastOnChainAddress instanceof Error) return createOnChainAddress(walletId)
```
bodymindarts, line 10: This should only be on `CouldNotFindError` not on _every_ `Error`.

src/servers/graphql-main-server.ts
```diff
@@ -17,6 +17,7 @@ export async function startApolloServerForCoreSchema() {
     {
       Query: {
         me: isAuthenticated,
+        onChainAddressCurrent: isAuthenticated,
```
bodymindarts, line 20: Shouldn't this be scoped under `wallet`?
Like: `me.defaultAccount.wallet.onChainAddressCurrent`? cc @samerbuna

src/services/mongoose/wallet-on-chain-addresses.ts
```diff
@@ -0,0 +1,61 @@
...
+        { _id: walletId },
+        { $push: { onchain: { address, pubkey } } },
+      )
+
+      if (result.n === 0) {
+        return new CouldNotFindError("Couldn't find wallet")
+      }
+
+      if (result.nModified !== 1) {
+        return new RepositoryError("Couldn't add onchain address for wallet")
```
bodymindarts, line 26: So far `RepositoryError` is only used as a base marker and never created directly. Perhaps this could be a `PersistError` or something like that.

bodymindarts (changes requested):

src/servers/graphql-main-server.ts
bodymindarts: Move to mutation

src/app/wallets/get-last-on-chain-address.ts
dolcalmi: fixed

src/servers/graphql-main-server.ts
dolcalmi: fixed

src/services/mongoose/wallet-on-chain-addresses.ts
dolcalmi: fixed

src/graphql/types/payload/on-chain-address.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const OnChainAddressPayload = new GT.Object({
+  name: "OnChainAddressPayload",
+  fields: () => ({
+    errors: {
+      type: GT.NonNullList(IError),
+    },
+    address: {
+      type: GT.NonNull(OnChainAddress),
+    },
```
samerbuna, line 13: Should be nullable

dolcalmi: fixed

dolcalmi:
I added the walletName as input but... 

- I think is inconvenient for the normal/corporate user to setup a walletName (through the app or API is bad user experience) before being able to use the wallet. 
- This will be a problem when we add usd wallets (1 user will have 2 wallets by default).
- We need to decide about this before add the logic to other queries/mutations
- Please keep in mind that walletId is already exposed by `me` query

bodymindarts (approved):

samerbuna (approved):

PR #528, opened 2021-09-01
Pass domain user in context

bodymindarts:
Currently our new use cases always need to first load the user in order to update it.
This PR exposes the `domainUser` in the GQL context so subsequent operations can access it directly.
Accessing the `domainUser` out of scope of this PR.

### DIFF
--- a/src/app/users/index.ts
+++ b/src/app/users/index.ts
@@ -1,5 +1,7 @@
 import { UsersRepository } from "@services/mongoose"
-import { ValidationError } from "@domain/errors"
+import { IpFetcher } from "@services/ipfetcher"
+import { getIpConfig, PROXY_CHECK_APIKEY } from "@config/app"
+import { ValidationError, RepositoryError } from "@domain/errors"
 
 export const getUser = async (userId: UserId): Promise<User | ApplicationError> => {
   const repo = UsersRepository()
@@ -35,4 +37,52 @@ export const updateWalletContactAlias = async ({
   }
 
   return user
+}
+
+export const getUserForLogin = async ({
+  userId,
+  ip,
+}: {
+  userId: string
+  ip?: string
+}): Promise<{ domainUser: User; rawUser: UserType } | ApplicationError> => {
+  const repo = UsersRepository()
+
+  const findResult = await repo.findByIdIncludeRaw(userId as UserId)
+  if (findResult instanceof Error) return findResult
+
+  const { domainUser, raw } = findResult
+  domainUser.lastConnection = new Date()
+
+  // IP tracking logic could be extracted into domain
+  const ipConfig = getIpConfig()
+  if (ipConfig.ipRecordingEnabled) {
+    const lastIP: IPType | undefined = domainUser.lastIPs.find(
+      (ipObject: IPType) => ipObject.ip === ip,
+    )
+    if (lastIP) {
+      lastIP.lastConnection = domainUser.lastConnection
+    } else if (ip && ipConfig.proxyCheckingEnabled) {
+      const ipFetcher = IpFetcher()
+      const ipInfo = await ipFetcher.fetchIPInfo(ip)
+      if (!(ipInfo instanceof Error)) {
+        domainUser.lastIPs.push({
+          ip,
+          ...ipInfo,
+          Type: ipInfo.type,
+          firstConnection: domainUser.lastConnection,
+          lastConnection: domainUser.lastConnection,
+        })
+      }
+    }
+  }
+
+  const updateResult = repo.update(domainUser)
+
+  if (updateResult instanceof RepositoryError) {
+    // Don't understand why this cast is necesairy
+    return updateResult as RepositoryError
+  }
+
+  return { domainUser, rawUser: raw }
 }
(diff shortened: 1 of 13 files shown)


src/app/users/index.ts
```diff
@@ -36,3 +38,51 @@ export const updateWalletContactAlias = async ({
...
+          ...ipInfo,
+          Type: ipInfo.type,
+          firstConnection: domainUser.lastConnection,
+          lastConnection: domainUser.lastConnection,
+        })
+      }
+    }
+  }
+
+  const updateResult = repo.update(domainUser)
```
vindard, line 80:
```suggestion
  const updateResult = await repo.update(domainUser)
```

src/app/users/index.ts
```diff
@@ -36,3 +38,51 @@ export const updateWalletContactAlias = async ({
...
+
+  if (updateResult instanceof RepositoryError) {
+    // Don't understand why this cast is necesairy
+    return updateResult as RepositoryError
```
vindard, line 84: This got correctly linted because we didn't `await` on Ln 80
```suggestion
  if (updateResult instanceof RepositoryError) return updateResult
```

vindard (changes requested):

src/services/mongoose/schema.types.d.ts
```diff
@@ -58,12 +58,12 @@ interface UserType {
   earn?: string[]
   deviceToken?: string[]
   excludeCashback?: boolean
-  contacts: ConcatObjectForUser
-  created_at?: string
+  contacts: ConcatObjectForUser[]
```
vindard, line 61: Is this an old typo coming forward? (`ContactObjectForUser`)

bodymindarts: yup fixed

src/domain/ipfetcher/index.types.d.ts
```diff
@@ -0,0 +1,14 @@
...
+  provider: string
+  country: string
+  region: string
+  city: string
+  type: string
+  status: string
+}
+
+interface IIpFetcherService {
+  fetchIPInfo(ip: string): Promise<IPInfo | IpFetcherServiceError>
```
dolcalmi, line 13: ip should have a custom type? (in regards to yesterday conversation)

src/servers/graphql-server.ts
```diff
@@ -1,4 +1,5 @@
 import { createServer } from "http"
+import { CouldNotFindError } from "@domain/errors"
```
dolcalmi, line 2: please remove

src/servers/graphql-server.ts
```diff
@@ -12,13 +13,14 @@ import { v4 as uuidv4 } from "uuid"
 import helmet from "helmet"
 
 import { getIpConfig, getHelmetConfig, JWT_SECRET } from "@config/app"
+import * as Users from "@app/users"
 
 import { baseLogger } from "@services/logger"
 import { redis } from "@services/redis"
-import { User } from "@services/mongoose/schema"
+import { UsersRepository } from "@services/mongoose/users"
```
dolcalmi, line 20: same here

dolcalmi: also please remove `const ipConfig = getIpConfig()`

src/services/mongoose/users.ts
```diff
@@ -8,99 +8,122 @@ import {
...
-export const caseInsensitiveRegex = (input) => {
+export const caseInsensitiveRegex = (input: string) => {
   return new RegExp(`^${input}$`, "i")
 }
 
-export const UsersRepository = (): IUsersRepository => {
-  const findById = async (userId: UserId): Promise<User | RepositoryError> => {
+export const UsersRepository = () => {
+  // Remove when old GQL api is deprecated => only use domain user
+  const findByIdIncludeRaw = async (
```
dolcalmi, line 17: just leave the method `findById` and move the comment to line 27, but again, it breaks the purpose of clean up

dolcalmi (changes requested):

src/app/users/index.ts
```diff
@@ -1,5 +1,7 @@
 import { UsersRepository } from "@services/mongoose"
-import { ValidationError } from "@domain/errors"
+import { IpFetcher } from "@services/ipfetcher"
+import { getIpConfig, PROXY_CHECK_APIKEY } from "@config/app"
```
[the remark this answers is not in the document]
bodymindarts: Can we increate declared-not-used lint to error status @samerbuna ?

src/app/users/index.ts
```diff
@@ -36,3 +38,50 @@ export const updateWalletContactAlias = async ({
...
+}: {
+  userId: string
+  ip?: string
+}): Promise<{ domainUser: User; rawUser: UserType } | ApplicationError> => {
+  const repo = UsersRepository()
+
+  const findResult = await repo.findByIdIncludeRaw(userId as UserId)
+  if (findResult instanceof Error) return findResult
+
+  const { domainUser, raw } = findResult
```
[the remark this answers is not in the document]
bodymindarts: Agree - the alternative is to have 2 database calls on every login. So far its _only_ for the login use case to expose the raw/old user for the v1 schema.
Alternatively we could keep the use case clean and on successful login make another database call in the context setup code to get the raw user. Would you prefer that?

src/app/users/index.ts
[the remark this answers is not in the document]
samerbuna: Yep. Will do in another branch.

dolcalmi (approved):

PR #529, opened 2021-09-01
Admin schema cleanup

samerbuna:


src/graphql/admin/types/object/user.ts
```diff
@@ -0,0 +1,34 @@
...
+    level: { type: AccountLevel },
+    status: { type: AccountStatus },
+    title: { type: GT.String },
+    coordinates: {
+      type: Coordinates,
+      resolve: (source) => source.coordinate,
+    },
+    createdAt: {
+      type: GT.NonNull(Timestamp),
+      resolve: (source) => source.createdAt ?? source.created_at, // TODO: Get rid of this resolver
```
bodymindarts, line 29: This shouldn't be needed. AFAICT you are always using the `UserType` shape for the user not the domain user (returned by repository).

src/servers/graphql-old-server.ts
```diff
@@ -239,11 +232,11 @@ const resolvers = {
     noauthAddInvoice: async (_, { username, value }) => {
       const lnInvoice =
         value && value > 0
-          ? await addInvoiceForRecipient({
+          ? await Wallets.addInvoiceForRecipient({
```
bodymindarts, line 235: Does switching these out mean we can delete the old methods? Also are we sure the return types line up?

samerbuna: These are the new methods, I just made the imports from wallets standard (using * rather than named imports)

src/graphql/admin/types/object/user.ts
samerbuna: It's still needed in the admin schema but we'll get rid of it soon

bodymindarts (approved):

src/servers/graphql-admin-server.ts
```diff
@@ -18,12 +18,12 @@ export async function startApolloServerForAdminSchema() {
...
         userDetailsByPhone: and(isAuthenticated, isEditor),
-        userDetailsByUsername: and(isAuthenticated, isEditor),
+        userDetailsByWalletName: and(isAuthenticated, isEditor),
       },
       Mutation: {
-        businessUpdateMapInfo: and(isAuthenticated, isEditor),
-        userUpdateLevel: and(isAuthenticated, isEditor),
         userUpdateStatus: and(isAuthenticated, isEditor),
+        userUpdateLevel: and(isAuthenticated, isEditor),
+        businessUpdateMapInfo: and(isAuthenticated, isEditor),
```
dolcalmi, line 26: will we have additional roles? maybe we can only use a fallbackRule

PR #531, opened 2021-09-02
Implement the walletNameAvailable query

samerbuna:


src/app/wallets/wallet-name-available.ts
```diff
@@ -0,0 +1,17 @@
+import { CouldNotFindError } from "@domain/errors"
+import { checkedToWalletName } from "@domain/wallets"
+import { WalletsRepository } from "@services/mongoose"
+
+export const walletNameAvailable = async (
+  walletName: string,
+): Promise<boolean | Error> => {
```
bodymindarts, line 7: For the use-case return types we've been using `Promise<type | ApplicationError>`. This being explicit should make it easier to do an exhaustiveness check if we want to. I will fix and merge.

src/graphql/root/query/wallet-name-available.ts
```diff
@@ -0,0 +1,29 @@
+import { walletNameAvailable } from "@app/wallets/wallet-name-available"
```
bodymindarts, line 1: Here we can also follow the pattern of `import * as Wallets` as in the other files I think.

bodymindarts (dismissed):

Issue #503, opened 2021-08-28
PaymentStatusChecker should be checked in ledger and not in InvoiceUser


dolcalmi:
this should be changed because ledger is the source of truth, also it will allow to check pending payments with the subscription

PR #532, opened 2021-09-02
PaymentStatusChecker must check ledger

bodymindarts:
fixes https://github.com/GaloyMoney/galoy/issues/503

`PaymentStatusChecker` was checking `WalletInvoice` but this is not the point of reference to see if something has been paid or not. It is the ledger!

### DIFF
--- a/src/services/ledger/index.ts
+++ b/src/services/ledger/index.ts
@@ -95,14 +95,16 @@ export const LedgerService = (): ILedgerService => {
     paymentHash: PaymentHash,
   ): Promise<boolean | LedgerServiceError> => {
     try {
-      const result = await Transaction.countDocuments({
+      const { total } = await MainBook.ledger({
+        pending: false,
         hash: paymentHash,
       })
-      return result > 0
+      return total > 0
     } catch (err) {
       return new UnknownLedgerError(err)
     }
   }
+
   const receiveOnChainTx = async ({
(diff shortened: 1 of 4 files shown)


src/services/ledger/index.ts
```diff
@@ -79,6 +79,19 @@ export const LedgerService = (): ILedgerService => {
       return new UnknownLedgerError(err)
     }
   }
+
+  const isLnTxRecorded = async (
+    paymentHash: PaymentHash,
+  ): Promise<boolean | LedgerServiceError> => {
+    try {
+      const result = await Transaction.countDocuments({
+        hash: paymentHash,
```
dolcalmi, line 88: do we have to check with `pending: false`? 🤔

vindard: Also, do we need to check that the journal entry isn't `voided`? Not sure if any other parts of our code take that into account

bodymindarts: I see what you guys mean, but see some issues - lets discuss it at standup.

vindard (approved):

dolcalmi (approved):
did we remove circle ci test? it is not running

vindard:
> did we remove circle ci test? it is not running

It's not running on this one and #557 so far (both rebased to latest tip of `main`)

PR #540, opened 2021-09-02
Implement the intraLedgerPaymentSend mutation

samerbuna:


src/servers/graphql-server.ts
```diff
@@ -116,7 +117,9 @@ export const startApolloServer = async ({
       const reportErrorToCclient =
         ["GRAPHQL_PARSE_FAILED", "GRAPHQL_VALIDATION_FAILED", "BAD_USER_INPUT"].includes(
           err.extensions?.code,
-        ) || isSheildError
+        ) ||
+        isSheildError ||
+        err instanceof ApolloError
```
samerbuna, line 123: Had to do this here because without it, the server 500 on old auth tokens

bodymindarts (approved):

PR #546, opened 2021-09-03
Implement the deviceNotificationTokenCreate mutation

samerbuna:


src/graphql/main/mutations.ts
```diff
@@ -1,5 +1,6 @@
 import { GT } from "@graphql/index"
 
+import DeviceNotificatinoTokenCreateMutation from "@graphql/root/mutation/device-notification-token-create"
```
vindard, line 3: Small type here and in a few other places that use this value
```suggestion
import DeviceNotificationTokenCreateMutation from "@graphql/root/mutation/device-notification-token-create"
```

vindard (changes requested):

bodymindarts (approved):

vindard (approved):

PR #548, opened 2021-09-04
Implement the mobileVersions root query

samerbuna:


vindard (dismissed):
Looks like a straightforward cleanup of the original code

src/graphql/main/schema.graphql
```diff
@@ -309,6 +310,12 @@ type Coordinates {
   latitude: Float!
 }
 
+type MobileVersions {
+  platform: String!
+  currentSupported: Int!
```
bodymindarts, line 315: Should this be a `SemVer` string? Or is it common for mobile to represent versions as simple `Int`? What version is this actually referring too? Should the type be called `SuppertedMobileClientVersions`?

samerbuna: Another thing that's not part of the core, but yeah thees are used in the mobile app to show that's an update is available or force an update.

These values are numbers on the backend.

dolcalmi (approved):
@jrbuhl93 would you mind check this?

jrbuhl93:
> @jrbuhl93 would you mind check this?

The new schema looks like it lines up with the app's [usecase](https://github.com/GaloyMoney/galoy-mobile/blob/2755d29462fdb76b37698efc2cf3ee63b636fc0d/app/screens/move-money-screen/move-money-screen.tsx#L190). But how are these values updated? It looks like their current values are 190 for latest 182 for min on both platforms. I should be updating latest when I release a new build because we are currently on 263.

bodymindarts:
> > @jrbuhl93 would you mind check this?
> 
> The new schema looks like it lines up with the app's [usecase](https://github.com/GaloyMoney/galoy-mobile/blob/2755d29462fdb76b37698efc2cf3ee63b636fc0d/app/screens/move-money-screen/move-money-screen.tsx#L190). But how are these values updated? It looks like their current values are 190 for latest 182 for min on both platforms. I should be updating latest when I release a new build because we are currently on 263.

Good to know - we need to automate this process when new mobile updates get pushed

bodymindarts (approved):

PR #550, opened 2021-09-04
Implement the btcUsdPriceListForGraph qeury

twshelton:
Implement types and mutation for Price List for Graph


src/graphql/root/query/btc-usd-price-list-for-graph.ts
```diff
@@ -0,0 +1,71 @@
...
+import BtcUsdPrice from "@graphql/types/object/btc-usd-price"
+import PriceGraphRange from "@graphql/types/scalar/price-graph-range"
+
+import { getHourlyPrice } from "@services/local-cache"
+import { getCurrentPrice } from "@services/realtime-price"
+
+const rangeStartDateMS: (string) => number | Error = (range) => {
+  const currentTimestamp = Date.now()
+
+  switch (range) {
```
bodymindarts, line 14: This switch is a good place to use static exhaustiveness guarantee

src/graphql/main/schema.graphql
```diff
@@ -309,6 +310,22 @@ type Coordinates {
   latitude: Float!
 }
 
+type BtcUsdPrice {
+  timestamp: Timestamp!
+  price: Float!
```
bodymindarts, line 315: Don't we have a more specific type? Like USDPerSat or something like that? `Float` is not a good type to use.

bodymindarts (changes requested):
Integration tests aren't passing

src/graphql/main/schema.graphql
samerbuna: Changed to use a price object with base/offset

src/graphql/main/schema.graphql
```diff
@@ -317,6 +318,39 @@ type MobileVersions {
...
+}
+
+"""
+The `SafeInt` scalar type represents non-fractional signed whole numeric values. SafeInt can represent values between -(2^53) + 1 and 2^53 - 1
+"""
+scalar SafeInt
+
+enum ExchangeCurrencyUnit {
+  BTCSAT
+  USDCENT
```
jrbuhl93, line 340: This pr looks good to me. Wanted to highlight this for @bodymindarts, and see what he thinks about putting both the currency and unit into one string type like this.

src/graphql/root/query/btc-price-list.ts
```diff
@@ -0,0 +1,87 @@
...
+} from "@graphql/types/scalar/price-graph-range"
+
+import { getHourlyPrice } from "@services/local-cache"
+import { getCurrentPrice } from "@services/realtime-price"
+
+const rangeStartDateMS: (string: typeof priceRangeValues[number]) => number = (range) => {
+  const currentTimestamp = Date.now()
+
+  switch (range) {
+    case "ONE_DAY":
```
dolcalmi, line 18: this should have a default value

src/graphql/root/query/btc-price-list.ts
```diff
@@ -0,0 +1,87 @@
...
+      prices.push({
+        timestamp: price.id,
+        price: {
+          formattedAmount: btcPriceInCents.toString(),
+          base: Math.round(btcPriceInCents * 10 ** 4),
+          offset: 4,
+          currencyUnit: "USDCENT",
+        },
+      })
+    }
```
dolcalmi, line 65: this should be a reduce (or just an splice like in old server) also.. do we have 5 years of hourly prices in cache? (I dont see the filter in getHourlyPrice)

dolcalmi (changes requested):

src/graphql/root/query/btc-price-list.ts
samerbuna: The default value is intentionally left out here for static exhaustiveness

src/graphql/root/query/btc-price-list.ts
samerbuna: I think `for...of` with early `break` here is better until we have an option to query the price data by date.

dolcalmi (approved):

bodymindarts (approved):

PR #552, opened 2021-09-06
Error visibility

samerbuna:
Updated with local test values for APOLLO env vars

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "@grpc/grpc-js": "^1.3.4",
     "@grpc/proto-loader": "^0.6.4",
     "@profusion/apollo-validation-directives": "^2.1.5",
+    "apollo-server-core": "^2.25.2",
     "apollo-server-errors": "^2.5.0",
     "apollo-server-express": "^2.25.2",
     "axios": "^0.21.1",
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "test-local-restart": "./scripts/delete-pods.sh && source ./scripts/export-local.sh && yarn run test-local",
     "test-local-clean": "yarn run restart-charts && yarn jest --clearCache && yarn run test-local",
     "test:unit": ". ./.envrc && LOGLEVEL=warn jest --config ./test/jest-unit.config.js --bail --verbose $TEST | yarn pino-pretty -c -l",
-    "test:integration": ". ./.envrc && LOGLEVEL=warn jest --config ./test/jest-integration.config.js --bail --runInBand --verbose $TEST | yarn pino-pretty -c -l"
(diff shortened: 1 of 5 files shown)


bodymindarts (changes requested):
Integration tests are failing on GitHub actions

package.json
```diff
@@ -30,7 +30,7 @@
     "test-local-restart": "./scripts/delete-pods.sh && source ./scripts/export-local.sh && yarn run test-local",
     "test-local-clean": "yarn run restart-charts && yarn jest --clearCache && yarn run test-local",
     "test:unit": ". ./.envrc && LOGLEVEL=warn jest --config ./test/jest-unit.config.js --bail --verbose $TEST | yarn pino-pretty -c -l",
-    "test:integration": ". ./.envrc && LOGLEVEL=warn jest --config ./test/jest-integration.config.js --bail --runInBand --verbose $TEST | yarn pino-pretty -c -l"
+    "test:integration": ". ./.envrc && LOGLEVEL=warn jest --config ./test/jest-integration.config.js --bail --runInBand --verbose --detectOpenHandles $TEST | yarn pino-pretty -c -l"
```
krtk6160, line 33: Why is `--detectOpenHandles` needed?

samerbuna: Removed

src/graphql/admin/index.ts
```diff
@@ -13,7 +13,7 @@ export const gqlAdminSchema = new GraphQLSchema({
   types: ERROR_INTERFACE_TYPES,
 })
 
-if (isDev) {
+if (!isProd) {
```
dolcalmi, line 16: will this work in test environment? do we have permissions to write?

src/servers/graphql-server.ts
```diff
@@ -42,10 +43,22 @@ export const startApolloServer = async ({
   startSubscriptionServer = false,
 }): Promise<Record<string, unknown>> => {
   const app = express()
+
+  const apolloPulgins = isProd
```
dolcalmi, line 47: did you test this with credentials in your local environment? if this fail we will only find out when it is deployed to production

src/graphql/admin/index.ts
bodymindarts: Testnet environment has `NODE_ENV=production` set so as to be as close to production as possible.

src/servers/graphql-server.ts
bodymindarts: Should definitely verify its working with real creds locally once before deploy - I agree. Has this been checked @samerbuna ?

samerbuna: Yep. However, errors thrown in the context don't seem to trigger it 😞

dolcalmi (approved):

src/servers/graphql-server.ts
bodymindarts: That sucks... guess we can merge this anyhow as it provides value, but still need to find another solution for the `context` errors.

bodymindarts (approved):

PR #553, opened 2021-09-06
Implement the price subscription through trigger/redis

samerbuna:


bodymindarts:
I'm not sure we should increase the dependency on Redis here. Using a pub-sub system for this use-case is certainly the right way to go but I doubt Redis is the right system for pub-sub in the future.
Perhaps its no big deal to switch implementation later... need to discuss

bodymindarts (dismissed):

test/jest-unit.config.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
   },
   testRegex: ".*\\.spec\\.ts$",
   testSequencer: "<rootDir>/test/jest-test-sequencer.js",
+  setupFilesAfterEnv: ["<rootDir>/test/jest.setup.js"],
```
dolcalmi, line 10: We should not have this in unit testing, if this is needed then is something wrong with code/dependencies

dolcalmi (changes requested):

test/jest-unit.config.js
samerbuna: Fixed

src/graphql/root/subscription/price.ts
```diff
@@ -0,0 +1,70 @@
...
+        }),
+      )
+      return pubsub.asyncIterator(eventName)
+    }
+
+    const satUsdPrice = await getCurrentPrice()
+
+    if (satUsdPrice) {
+      setTimeout(
+        () => pubsub.publish(eventName, { satUsdCentPrice: 100 * satUsdPrice }),
```
dolcalmi, line 61: not sure if I am not undestanding but `satUsdCentPrice` should be `price`? (like in trigger)

src/graphql/root/subscription/price.ts
```diff
@@ -0,0 +1,70 @@
...
+  subscribe: async (_, args) => {
+    const { amountCurrencyUnit, priceCurrencyUnit } = args.input
+
+    for (const input of [amountCurrencyUnit, priceCurrencyUnit]) {
+      if (input instanceof Error) {
+        return { errors: [{ message: input.message }] }
+      }
+    }
+
+    const eventName = "SAT-USDCENT-PRICE"
```
dolcalmi, line 45: maybe this should be in a common file/config (to be used in trigger too)

src/graphql/types/scalar/big-int.ts
```diff
@@ -0,0 +1,43 @@
...
+  const num = Number(value)
+  if (num !== num || num > MAX_INT || num < MIN_INT) {
+    throw new Error(
+      "BigInt cannot represent non 53-bit signed integer value: " + String(value),
+    )
+  }
+  const int = Math.floor(num)
+  if (int !== num) {
+    throw new Error("BigInt cannot represent non-integer value: " + String(value))
+  }
```
dolcalmi, line 39: the subscription is working but after a couple of seconds I got "message": "BigInt cannot represent non 53-bit signed integer value: NaN" or maybe is this `num !== num`

price 0.000465380546875

dolcalmi (changes requested):

src/graphql/root/subscription/price.ts
samerbuna: That's actually what's causing the NaN error. Fixed.

src/graphql/types/scalar/big-int.ts
samerbuna: Fixed

src/graphql/root/subscription/price.ts
samerbuna: That's actually what's causing the NaN error. Fixed.

src/graphql/types/scalar/big-int.ts
```diff
@@ -0,0 +1,43 @@
+import { GT } from "@graphql/index"
+
+const MAX_INT = Number.MAX_SAFE_INTEGER
+const MIN_INT = Number.MIN_SAFE_INTEGER
+
+const BigInt = new GT.Scalar({
```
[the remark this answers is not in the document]
samerbuna: Renamed

src/graphql/root/subscription/price.ts
```diff
@@ -0,0 +1,70 @@
...
+      return { errors: source.errors }
+    }
+    const amountPriceInCents = args.input.amount * source.satUsdCentPrice
+    return {
+      errors: [],
+      price: {
+        formattedAmount: amountPriceInCents.toString(),
+        base: Math.round(amountPriceInCents * 10 ** 16),
+        offset: 16,
+        currencyUnit: "USDCENT",
```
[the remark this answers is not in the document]
samerbuna: I've seen prices where we need 12 or more decimal offset to be accurate

src/graphql/types/scalar/safe-int.ts
```diff
@@ -0,0 +1,43 @@
...
+})
+
+function coerceSafeInt(value) {
+  if (value === "") {
+    throw new Error(
+      "SafeInt cannot represent non 53-bit signed integer value: (empty string)",
+    )
+  }
+  const num = Number(value)
+  if (num !== num || num > MAX_INT || num < MIN_INT) {
```
dolcalmi, line 31: `num !== num` should be `num !== value`

src/graphql/root/subscription/price.ts
```diff
@@ -0,0 +1,71 @@
...
+        }),
+      )
+      return pubsub.asyncIterator(eventName)
+    }
+
+    const satUsdPrice = await getCurrentPrice()
+
+    if (satUsdPrice) {
+      setTimeout(
+        () => pubsub.publish(eventName, { satUsdPrice: 100 * satUsdPrice }),
```
dolcalmi, line 62: this should use `satUsdCentPrice` instead of  `satUsdPrice`

dolcalmi (changes requested):

src/graphql/root/subscription/price.ts
samerbuna: Fixed

src/graphql/types/scalar/safe-int.ts
samerbuna: Changed

dolcalmi (approved):

vindard (approved):
No new issues found past the already-resolved comments on here from my end, LGTM 🎉

PR #557, opened 2021-09-07
Add new 'get balances' use case

vindard:
### Description

This PR adds the re-implemented `getBalanceForWallet` method to take the place of the old `getBalances` method.

The re-implementation only focuses on BTC balances. There are two outstanding places where USD balances are used, so the old `getBalances` method hasn't been removed as yet:
- `wallet2` endpoint in `graphql-old-server.ts` file
- `rebalancePortfolio` method in `src/services/ledger/transaction.ts` (with related integration tests)

### DIFF
--- a/src/app/wallets/get-balance-for-wallet.ts
+++ b/src/app/wallets/get-balance-for-wallet.ts
@@ -11,17 +11,19 @@ export const getBalanceForWallet = async ({
   lock?: DistributedLock
   logger: Logger
 }): Promise<Satoshis | ApplicationError> => {
-  Wallets.updatePendingInvoices({
-    walletId: walletId,
-    lock,
-    logger,
-  })
-  const result = await Wallets.updatePendingPayments({
-    walletId: walletId,
-    lock,
-    logger,
-  })
-  if (result instanceof Error) throw result
+  const [, updatePaymentsResult] = await Promise.all([
+    Wallets.updatePendingInvoices({
+      walletId: walletId,
+      lock,
+      logger,
+    }),
+    Wallets.updatePendingPayments({
+      walletId: walletId,
+      lock,
+      logger,
+    }),
+  ])
+  if (updatePaymentsResult instanceof Error) throw updatePaymentsResult
 
   const liabilitiesAccountId = toLiabilitiesAccountId(walletId)
   const balance = await LedgerService().getAccountBalance(liabilitiesAccountId)
--- a/src/app/wallets/get-balance-for-wallet.ts
+++ b/src/app/wallets/get-balance-for-wallet.ts
@@ -1,0 +1,31 @@
+import { toLiabilitiesAccountId } from "@domain/ledger"
+import { LedgerService } from "@services/ledger"
+import * as Wallets from "@app/wallets"
+
+export const getBalanceForWallet = async ({
+  walletId,
+  lock,
+  logger,
+}: {
+  walletId: WalletId
+  lock?: DistributedLock
+  logger: Logger
+}): Promise<Balances | ApplicationError> => {
+  Wallets.updatePendingInvoices({
+    walletId: walletId,
+    lock,
+    logger,
+  })
+  const result = await Wallets.updatePendingPayments({
+    walletId: walletId,
+    lock,
+    logger,
+  })
+  if (result instanceof Error) throw result
+
+  const liabilitiesAccountId = toLiabilitiesAccountId(walletId)
+  const balance = await LedgerService().getAccountBalance(liabilitiesAccountId)
+  if (balance instanceof Error) return balance
+
+  return { BTC: balance, totalInBtc: balance }
+}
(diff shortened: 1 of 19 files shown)


test/helpers/wallet.ts
```diff
@@ -0,0 +1,11 @@
+import * as Wallets from "@app/wallets"
+import { baseLogger } from "@services/logger"
+
+export const getBTCBalance = async (walletId: WalletId): Promise<Satoshis> => {
```
dolcalmi, line 4: maybe just `getBalance`

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -417,12 +418,21 @@ describe("UserWallet - Lightning Pay", () => {
           tokens: amountInvoice,
         })
         await fn(userWallet1)({ invoice: request })
-        const intermediateBalance = await userWallet1.getBalances()
+        const intermediateBalanceSats = await Wallets.getBalanceForWallet({
```
dolcalmi, line 421: why dont you use the helper here??

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -417,12 +418,21 @@ describe("UserWallet - Lightning Pay", () => {
...
+          logger: baseLogger,
+        })
+        if (intermediateBalanceSats instanceof Error) throw intermediateBalanceSats
+
         const result = await fn(userWallet1)({ invoice: request })
         expect(result).toBe("already_paid")
 
-        const finalBalance = await userWallet1.getBalances()
-        expect(finalBalance).toStrictEqual(intermediateBalance)
+        const finalBalanceSats = await Wallets.getBalanceForWallet({
```
dolcalmi, line 430: same here

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -752,7 +762,11 @@ describe("UserWallet - Lightning Pay", () => {
 
     // await sleep(1000)
 
-    await userWallet1.getBalances()
+    const _ = await Wallets.getBalanceForWallet({
```
dolcalmi, line 765: if you remove this the tests fail? I think is just to update payments/invoices

dolcalmi (changes requested):
when I run `make start-old` in this branch I got the next error:

```
OverwriteModelError: Cannot overwrite `Medici_Transaction` model once compiled.
```
I tested in main branch and is working

src/app/wallets/get-balance-for-wallet.ts
```diff
@@ -0,0 +1,31 @@
...
+export const getBalanceForWallet = async ({
+  walletId,
+  lock,
+  logger,
+}: {
+  walletId: WalletId
+  lock?: DistributedLock
+  logger: Logger
+}): Promise<Satoshis | ApplicationError> => {
+  Wallets.updatePendingInvoices({
```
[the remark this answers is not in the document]
vindard: We were wondering about this as well with 2 things:
- lock object locks on two different things (paymentHash vs walletId) for the two calls here, so it might already be inconsistent
- there are places where `getBalanceForWallet` is called without a `lock` arg entirely, so no value is passed forward to the two function calls

We decided to keep the implementation the same as the existing implementation for now, and potentially fix the inconsistencies later on (cc @bodymindarts)

test/helpers/wallet.ts
vindard: There's another `getBalance` still hanging around in the code so I was thinking to leave this differentiated for now to make it easier to remove that function later on

test/integration/02-user-wallet/02-send-lightning.spec.ts
vindard: Good catch! This was a remnant from when we returned an object from the function and needed to compare it with a `toBeStrictEqual` at these 3 places in the tests

test/integration/02-user-wallet/02-send-lightning.spec.ts
vindard: All pass still 👍🏾

vindard:
> when I run `make start-old` in this branch I got the next error:
> 
> ```
> OverwriteModelError: Cannot overwrite `Medici_Transaction` model once compiled.
> ```

I've isolated this issue to commit https://github.com/GaloyMoney/galoy/pull/557/commits/e057cf0bab47124f0e175dc8b54846b4063f7b84 when we call `await LedgerService().getAccountBalance(liabilitiesAccountId)` [here](https://github.com/GaloyMoney/galoy/pull/557/files#diff-25daaabcebc2d70d7745e0212d8ca9b4a558a6f639c9d598197a17f5e5097374R27). This function only calls one meaningful thing, `await MainBook.balance`, but I'm stumped why this would cause this error 🤷🏾‍♂️

dolcalmi (approved):

bodymindarts (approved):

PR #558, opened 2021-09-07
Make login arguments non nullable

krtk6160:


samerbuna (dismissed):

bodymindarts (approved):

krtk6160:
Note: This is a backwards incompatible change. The mobile app versions using the older schema will not work with this change.

samerbuna:
Added guards to the resolvers to prevent the errors.

bodymindarts (approved):

PR #562, opened 2021-09-08
Implement onChain payment send mutations

samerbuna:


src/servers/graphql-main-server.ts
```diff
@@ -34,6 +34,8 @@ export async function startApolloServerForCoreSchema() {
         lnNoAmountInvoicePaymentSend: isAuthenticated,
 
         intraLedgerPaymentSend: isAuthenticated,
+
+        onChainPaymentSendAll: isAuthenticated,
```
dolcalmi, line 38: are we missing `onChainPaymentSend` ?

src/graphql/types/scalar/on-chain-address.ts
```diff
@@ -17,7 +17,7 @@ const OnChainAddress = new GT.Scalar({
 
 function validOnChainAddressValue(value) {
   // TODO: verify/improve. Use bc1/tb1 prefixes?
-  if (value.match(/^[A-Fa-f0-9]+$/i)) {
+  if (value.match(/^[A-Z0-9]+$/i)) {
```
dolcalmi, line 20: did you test this? maybe we should have a helper with some unit tests

dolcalmi (changes requested):

src/graphql/types/scalar/on-chain-address.ts
bodymindarts: We cannot constrain btc addresses to only contain uppercase. Lower case addresses are not only valid but the norm for most wallets.

samerbuna: This the current check we have. It's case insensitive.

src/servers/graphql-main-server.ts
samerbuna: Yep. Fixed.

dolcalmi (dismissed):

src/graphql/types/scalar/on-chain-address.ts
bodymindarts: Ah right `/i` -> case insensitive.

bodymindarts (approved):

vindard (approved):

vindard:
@samerbuna looks like this needs a rebase

bodymindarts:
> @samerbuna looks like this needs a rebase
taken care of - can you re-approve @vindard

bodymindarts (approved):

vindard (approved):

PR #563, opened 2021-09-08
Changed UserQuiz...Input to accept only a single ID instead of list

twshelton:
This PR includes the changes we discussed regarding changing input to single ID instead of list and makes obsolete #549.


samerbuna:
I added this to #549.

PR #565, opened 2021-09-09
Implement 2FA mutations with existing resolvers

samerbuna:


src/graphql/root/mutation/twofa-generate.ts
```diff
@@ -0,0 +1,16 @@
+import { GT } from "@graphql/index"
+import TwoFAGeneratePayload from "@graphql/types/payload/twofa-generate"
+
+const TwoFAGenerateMutation = GT.Field({
+  type: GT.NonNull(TwoFAGeneratePayload),
+  resolve: async (_, __, { wallet }) => {
+    try {
+      const towFA = await wallet.generate2fa()
```
krtk6160, line 8: Typo, should be `twoFA` instead of `towFA`

samerbuna: Fixed.

jrbuhl93:
Is there an equivalent of [this line](https://github.com/GaloyMoney/galoy/blob/29d5ebbe2bcbe3179611c946f446ecb42ffb8454/src/graphql/old-schema.graphql#L126) for the account or wallet in the new schema?

samerbuna:
> Is there an equivalent of [this line](https://github.com/GaloyMoney/galoy/blob/29d5ebbe2bcbe3179611c946f446ecb42ffb8454/src/graphql/old-schema.graphql#L126) for the account or wallet in the new schema?

Added

jrbuhl93 (dismissed):

bodymindarts (dismissed):

samerbuna:
Rebased to resolve conflicts @bodymindarts @jrbuhl93

jrbuhl93 (approved):

bodymindarts (approved):

src/services/mongoose/users.ts
```diff
@@ -48,21 +45,11 @@ export const UsersRepository = (): IUsersRepository => {
         deviceToken: deviceTokens,
         lastConnection,
       }
-      const doc = await User.updateOne({ _id: id }, { $set: data })
-      if (doc.nModified !== 1) {
+      const user: User = await User.findOneAndUpdate({ _id: id }, data)
```
vindard, line 48: I just noticed I missed this change, but would casting the return from `findOneAndUpdate` like this instead of explicitly defining the props bring in extra props that we don't necessarily want to have on an object of `User` type? (cc @samerbuna, @bodymindarts)

samerbuna: Fixed in #579

PR #566, opened 2021-09-10
Add getOnChainFee use case

dolcalmi:
- [x] add basic structure
- [x] add tests
- [x] add graphql query
- [x] add user balance validation to avoid lnd node onchain balance "sniffing" attack


dolcalmi:
- How are we handling custom errors in graphql? Should I use OnChainFeeEstimationError (is in src/core/error.ts)  in the resolver ?
- We don't have expectedConfirmationTime, we use the default behavior in lightning library but I couldn't find with how many blocks the fee is estimated.  Should I add the number of blocks to estimate fee to config?

bodymindarts (commented):
> We don't have expectedConfirmationTime, we use the default behavior in lightning library but I couldn't find with how many blocks the fee is estimated. Should I add the number of blocks to estimate fee to config?

We should certainly discuss this pre-merge.

test/integration/02-user-wallet/02-fees-onchain.spec.ts
```diff
@@ -18,15 +29,35 @@ afterAll(async () => {
...
   it("returns zero for an on us address", async () => {
-    const address = await Wallets.createOnChainAddress(userWallet1.user.id)
+    const address = await Wallets.createOnChainAddress(userWallet1.id)
     if (address instanceof Error) throw address
-    const fee = await userWallet0.getOnchainFee({ address })
+    const fee = await Wallets.getOnChainFee(userWallet0, defaultAmount, address)
     expect(fee).toBe(0)
   })
+
+  it("returns error for dust amount", async () => {
```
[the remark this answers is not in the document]
bodymindarts: Should return `InvalidSatoshiAmount` (if < 0) and `LessThanDustThreshold` (if < dust) or something like that (so caller can differentiate what the issues i)

src/app/wallets/get-on-chain-fee.ts
```diff
@@ -0,0 +1,63 @@
...
+  amount: Satoshis,
+  address: OnChainAddress,
+): Promise<Satoshis | ApplicationError> => {
+  if (amount <= 0) return new InvalidOnChainAmount("Invalid amount")
+
+  const walletsRepo = WalletsRepository()
+  const payeeWallet = await walletsRepo.findByAddress(address)
+  if (payeeWallet instanceof CouldNotFindError) {
+    if (amount < dustThreshold) {
+      return new InvalidOnChainAmount(
```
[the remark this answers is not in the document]
dolcalmi: cant be a early return, if it is our address will be an "on us" transaction and fee must be 0

test/integration/02-user-wallet/02-fees-onchain.spec.ts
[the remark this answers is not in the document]
dolcalmi: is a validation of the returned errors (in the method) and cant be in unit test because it depends on db (only should return the error if it is not an intra ledger tx).

src/app/wallets/get-on-chain-fee.ts
```diff
@@ -0,0 +1,103 @@
...
+    )
+    if (onChainFee instanceof Error) return onChainFee
+
+    return (onChainFee + wallet.withdrawFee) as Satoshis
+  }
+
+  if (payeeWallet instanceof Error) return payeeWallet
+
+  // intraledger tx has zero fee
+  return 0 as Satoshis
```
bodymindarts, line 54: This is business logic and should not be hard coded into the use-case method. Even if its trivial its best to be explicit like with the [DepositFeeCaluclator](https://github.com/GaloyMoney/galoy/blob/main/src/domain/wallets/deposit-fee-calculator.ts)

@vindard did you have some kind of business logic code that handles fee in your `pay` method refactoring?

src/app/wallets/get-on-chain-fee.ts
```diff
@@ -0,0 +1,103 @@
...
+    if (onChainService instanceof Error) return onChainService
+
+    const onChainFee = await onChainService.getOnChainFeeEstimate(
+      amount,
+      address,
+      targetConfirmations,
+    )
+    if (onChainFee instanceof Error) return onChainFee
+
+    return (onChainFee + wallet.withdrawFee) as Satoshis
```
bodymindarts, line 48: This is business logic. Should be encapsulated in some kind of `FeeCalculator` under `domain` - see comment below.

src/graphql/root/query/on-chain-tx-fee-query.ts
```diff
@@ -0,0 +1,62 @@
...
+    for (const input of [walletName, address, amount, targetConfirmations]) {
+      if (input instanceof Error) throw input
+    }
+
+    let fee: Satoshis | Error | null = null
+    if (walletName) {
+      const hasPermissions = await Accounts.hasPermissions(domainUser.id, walletName)
+      if (hasPermissions instanceof Error) throw new ApolloError(hasPermissions.message)
+
+      if (!hasPermissions) throw new ForbiddenError("Invalid walletName")
```
bodymindarts, line 31: Can sending `Forbidden` here leak that the `walletName` existed? Shouldn't we return some kind of `NotFound`?

src/graphql/types/object/onchain-tx-fee.ts
```diff
@@ -1,12 +1,12 @@
...
-import ExpectedConfirmationTime from "../scalar/expected-confirmation-time"
+// import ExpectedConfirmationTime from "../scalar/expected-confirmation-time"
 import SatAmount from "../scalar/sat-amount"
 
 const OnChainTxFee = new GT.Object({
   name: "OnChainTxFee",
   fields: () => ({
     amount: { type: GT.NonNull(SatAmount) },
-    expectedConfirmationTime: { type: GT.NonNull(ExpectedConfirmationTime) },
+    // expectedConfirmationTime: { type: GT.NonNull(ExpectedConfirmationTime) },
```
bodymindarts, line 9: Shouldn't commit commented code. Since moving to the structure with `targetConfirmations` is this file even needed anymore?

src/graphql/root/query/on-chain-tx-fee-query.ts
dolcalmi: we cant use normal not found.. I just returned an apollo error

src/app/wallets/get-on-chain-fee.ts
dolcalmi: Please check and let me know. @vindard if you can check would be great

src/domain/wallets/withdrawal-fee-calculator.ts
```diff
@@ -0,0 +1,10 @@
+import { toSats } from "@domain/bitcoin"
+
+export const WithdrawalFeeCalculator = (wallet: Wallet): WithdrawalFeeCalculator => {
```
vindard, line 3: To match the `DepositFeeCalculator` setup I would probably make this so it can be called like:
```ts
WithdrawalFeeCalculator(onChainFee).onChainWithdrawalFee(wallet.withdrawFee)
```

The pattern here could be that the calculator constructor gets the btc/ln network amount to be operated on and the method takes any wallet-specific params required.

src/graphql/root/query/on-chain-tx-fee-query.ts
```diff
@@ -0,0 +1,65 @@
...
+const OnChainTxFeeQuery = GT.Field({
+  type: GT.NonNull(OnChainTxFee),
+  args: {
+    walletName: { type: WalletName },
+    address: { type: GT.NonNull(OnChainAddress) },
+    amount: { type: GT.NonNull(SatAmount) },
+    targetConfirmations: { type: TargetConfirmations, defaultValue: 3 },
+  },
+  resolve: async (_, args, { domainUser }) => {
+    const { walletName, address, amount, targetConfirmations } = args
```
vindard, line 20: More of a question for my own understanding, when do we use `args` vs. `args.input` in these sorts of spots?

test/integration/02-user-wallet/02-tx-onchain-fees.spec.ts
```diff
@@ -0,0 +1,84 @@
+import * as Wallets from "@app/wallets"
+import { getOnChainWalletConfig } from "@config/app"
+import { InsufficientBalanceError, LessThanDustThresholdError } from "@domain/errors"
+import { bitcoindClient, bitcoindOutside, getUserWallet } from "test/helpers"
+
+jest.mock("@services/realtime-price", () => require("test/mocks/realtime-price"))
+jest.mock("@services/phone-provider", () => require("test/mocks/phone-provider"))
+
+const defaultAmount = 6000 as Satoshis
```
vindard, line 9: We've been using this pattern in different places
```suggestion
const defaultAmount = toSats(6000)
```

test/integration/02-user-wallet/02-tx-onchain-fees.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+describe("UserWallet - getOnchainFee", () => {
+  it("returns a fee greater than zero for an external address", async () => {
+    const address = (await bitcoindOutside.getNewAddress()) as OnChainAddress
+    const fee = await Wallets.getOnChainFee({
+      wallet: userWallet0,
+      amount: defaultAmount,
+      address,
+      targetConfirmations: defaultTarget,
+    })
+    expect(fee).toBeGreaterThan(0)
```
vindard, line 40: We could optionally add an additional `throw` or `not.toBeInstanceOf(Error)` here to get more information from the tests if this method call fails.

Same comment for other places below.

test/integration/02-user-wallet/02-tx-onchain-fees.spec.ts
```diff
@@ -0,0 +1,84 @@
+import * as Wallets from "@app/wallets"
+import { getOnChainWalletConfig } from "@config/app"
+import { InsufficientBalanceError, LessThanDustThresholdError } from "@domain/errors"
+import { bitcoindClient, bitcoindOutside, getUserWallet } from "test/helpers"
+
+jest.mock("@services/realtime-price", () => require("test/mocks/realtime-price"))
+jest.mock("@services/phone-provider", () => require("test/mocks/phone-provider"))
+
+const defaultAmount = 6000 as Satoshis
+const defaultTarget = 3 as TargetConfirmations
```
vindard, line 10: I wonder if we need a similar `toTargetConfs` method here as `toSats`. I'm not sure what our heuristic is for using a method vs. just using `as` to cast the type (cc @bodymindarts)

Would be separate from the already implemented `checkedToTargetConfs` which I believe is only meant to validate value coming from user input.

vindard (changes requested):
Left a few comments and suggestions

src/domain/wallets/withdrawal-fee-calculator.ts
dolcalmi: uhmm we have two cases here, intraledger fee and onchain fee so the param in the constructor cant be onChainFee (the common part should be in the "constructor")

src/app/wallets/get-on-chain-fee.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    const onChainFee = await onChainService.getOnChainFeeEstimate(
+      amount,
+      address,
+      targetConfirmations,
+    )
+    if (onChainFee instanceof Error) return onChainFee
+
+    return withdrawalFeeCalculator.onChainFee(onChainFee)
+  }
```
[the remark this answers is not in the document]
dolcalmi: initially I did something similar but had some TS issues.. I will check again

src/graphql/root/query/on-chain-tx-fee-query.ts
dolcalmi: I had the same question and what I understood (and validate doing some tests) was that makes easier to do changes to input params, i.e, you only need to change the query variables and not the query/mutation definition (for example in mobile app or in tests Ex: `test/integration/servers/graphql-main-server/mutations/ln-invoice-create.gql`). 

Please ask Samer for more details

src/domain/wallets/withdrawal-fee-calculator.ts
vindard: I'm thinking of how to make this consistent with the `DepositFeeCalculator`

That one currently has a unique lifetime for each transaction and its common thing is the `amount` from the ln txn. The methods do operations on the amount using fee-related variables.

For the original comment, I was thinking the `onChainFee` might be the unique txn lifetime value and then methods take fee-related things like the wallet's withdrawFee constant to apply to the `onChainFee` in the constructor.

Having the common thing as the `wallet` could be tricky because we can probably get fee-related values from places like configs as well later on.

But I'm ok with changing this if we think it makes sense to broaden its lifetime. Maybe we can alternatively pass txn-related values to the methods as well and leave the constructor empty?
```ts
WithdrawalFeeCalculator().onChainWithdrawalFee({onChainFee, walletFee: wallet.withdrawFee})
```

dolcalmi: Fixed, will update DepositFeeCalculator in a different PR

vindard (approved):
Looks good, needs rebase with `main`

samerbuna (approved):

vindard (approved):

PR #569, opened 2021-09-13
Replace payment-status db polling with an event

samerbuna:


src/app/wallets/update-pending-invoices.ts
```diff
@@ -143,6 +144,9 @@ const updatePendingInvoice = async ({
         paymentHash,
       })
 
+      const eventName = `LNINVOICE-PAYMENT-STATUS-${paymentHash}`
```
dolcalmi, line 147: should we centralize event names in config?

dolcalmi (changes requested):
also `check code` is failing

src/app/wallets/update-pending-invoices.ts
samerbuna: Updated

vindard (approved):

samerbuna:
> also `check code` is failing

Fixed.

dolcalmi (dismissed):

src/core/lightning/index.ts
```diff
@@ -313,6 +314,9 @@ export const LightningMixin = (superclass) =>
             type: "paid-invoice",
           })
 
+          const eventName = lnPaymentStatusEvent(id)
+          pubsub.publish(eventName, { status: "PAID" })
```
vindard, line 318: Do we need a `publish` for successful `withdrawal` payments as well (since this one is for `onUs`)?

dolcalmi: uhmm maybe for hodl invoices but that is an edge case, in general, when you pay an invoice the status response is immediate

src/services/redis.ts
```diff
@@ -48,8 +48,9 @@ if (process.env.LOCAL === "docker-compose") {
 }
 
 export const redis = new Redis(connectionObj)
-export const redisSub = new Redis(connectionObj)
-export const pubsub = new RedisPubSub({
+
+const redisSub = new Redis(connectionObj)
```
vindard, line 52: `redisSub` needed in `test/jest.setup.js` for tests

```suggestion
export const redisSub = new Redis(connectionObj)
```

vindard (changes requested):

dolcalmi (approved):
I only have doubts with hodl invoices but looks good

vindard (approved):

PR #574, opened 2021-09-16
Add Account Api Keys use case

dolcalmi:
- [X] Add account api key schema
- [X] Add api key creation
- [X] Add get api keys by account
- [X] Disable api key by label
- [x] Add Graphql mutations/queries
- [x] Add middleware and update context
- [ ] ~~Update current mutations/queries to use authenticated account~~ Will be done once we update walletPublicId


src/app/accounts/add-api-key-for-account.ts
```diff
@@ -0,0 +1,25 @@
+import { hashApiKey, randomApiKey } from "@domain/accounts"
+import { AccountApiKeysRepository } from "@services/mongoose"
+
+export const addApiKeyForAccount = async ({
+  accountId,
+  label = "",
```
bodymindarts, line 6: IMO defaults shouldn't be set at the use case level. Client should decide what the default should be.

src/app/accounts/index.types.d.ts
```diff
@@ -0,0 +1,10 @@
+type AddApiKeyForAccountArgs = {
+  accountId: AccountId
+  label?: string
```
bodymindarts, line 3: No optional arguments for use-cases please. Optional arguments on the use-case level are a smell that point to there actually being more than 1 use case in play.

src/domain/accounts/api-keys.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+const hasExpired = (expireAt: Date): boolean => {
+  const now = new Date(Date.now())
+  return expireAt < now
+}
+
+export const randomApiKey = async (
+  expireAt: Date,
+  label?: string,
```
bodymindarts, line 26: Shouldn't be optional - defaults should be set higher up.

test/integration/services/mongoose/account-api-keys.spec.ts
```diff
@@ -0,0 +1,106 @@
+import { hashApiKey, randomApiKey } from "@domain/accounts"
+import { CouldNotFindError } from "@domain/errors"
+import { AccountApiKeysRepository } from "@services/mongoose"
+
+const repo = AccountApiKeysRepository()
+const delta = 30 // days
```
bodymindarts, line 6: how about making the comment part of the variable name: `expiration_in_days` or `delta_days` or something like hat.

bodymindarts (commented):
It lgtm but I'm not sure about the key/secret both being returned. Most API's that I know of just have 1 value as the api secret. I don't really see the use case for having 2.

test/integration/services/mongoose/account-api-keys.spec.ts
dolcalmi: fixed

src/domain/accounts/api-keys.ts
dolcalmi: fixed

samerbuna (dismissed):

vindard (dismissed):
Looks good, needs rebase

samerbuna (dismissed):

samerbuna (approved):

vindard (approved):

PR #575, opened 2021-09-16
activate failed htlc payment pruning

### LINKED ISSUE #573: delete failed payment on lnd
now that we are on lnd 0.13, I think we can activate the deletion of failed payment on the cron job

nicolasburtey:
fix #573 (at least the failed htlc). 

we might want to dive more into the use of the `deletefailedpayments` (instead of `deleteFailedPayAttempts`), because it's possible we need the data about failed payments for support purpose. ie: when we need to do a refund.

the htlc part are safe to delete, and would likely already free the database significantly.

### DIFF
--- a/src/services/lnd/index.ts
+++ b/src/services/lnd/index.ts
@@ -1,19 +1,19 @@
 import { toSats } from "@domain/bitcoin"
 import {
-  decodeInvoice,
   CouldNotDecodeReturnedPaymentRequest,
-  UnknownLightningServiceError,
-  LightningServiceError,
+  decodeInvoice,
   InvoiceNotFoundError,
+  LightningServiceError,
   PaymentStatus,
+  UnknownLightningServiceError,
 } from "@domain/bitcoin/lightning"
 import {
+  cancelHodlInvoice,
   createInvoice,
+  deleteFailedPayAttempts,
   getInvoice,
   getPayment,
-  cancelHodlInvoice,
   GetPaymentResult,
-  deleteFailedPayAttempts,
 } from "lightning"
 import { getActiveLnd, getLndFromPubkey, getLnds } from "./utils"
 
--- a/src/services/lnd/index.ts
+++ b/src/services/lnd/index.ts
@@ -13,6 +13,7 @@ import {
   getPayment,
   cancelHodlInvoice,
   GetPaymentResult,
+  deleteFailedPayAttempts,
 } from "lightning"
 import { getActiveLnd, getLndFromPubkey, getLnds } from "./utils"
 
@@ -137,5 +138,6 @@ export const LndService = (): ILightningService | LightningServiceError => {
     lookupInvoice,
     lookupPayment,
     cancelInvoice,
+    deleteFailedPayAttempts,
   }
 }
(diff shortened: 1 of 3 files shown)


src/services/lnd/index.ts
```diff
@@ -102,5 +108,6 @@ export const LndService = (): ILightningService | LightningServiceError => {
     registerInvoice,
     lookupInvoice,
     lookupPayment,
+    deleteFailedPayAttempts,
```
bodymindarts, line 111: Where is this being called? the `LndService` function should only return an object fulfilling [this](https://github.com/GaloyMoney/galoy/blob/main/src/domain/ledger/index.types.d.ts#L87) interface. Which in turn only defines functions that are being used in `src/app` use-cases. IE its part of the re-worked, clean code.
This is why `check-code` is failing.

If you want to add it here you should be wrapping the code with the same `try/catch` + error pattern we have been using in the other cleaned up code sections - but that really depends on where / how its being called.

bodymindarts:
Change looks simple - has this been tested at all? If so how?

bodymindarts (dismissed):

src/services/lnd/utils.ts
```diff
@@ -53,16 +50,15 @@ export const deleteExpiredInvoiceUser = async () => {
...
-  //     await deleteFailedPayments({lnd})
-  //   }
-  // } catch (err) {
-  //   baseLogger.warn({ err }, "error deleting failed payment")
-  // }
+  const lnds = offchainLnds
+  for (const { lnd, socket } of lnds) {
+    try {
+      baseLogger.info({ socket }, "running deleteFailedPayments")
+      await deleteFailedPayAttempts({ lnd })
```
dolcalmi, line 57: do we want to delete only payment attemps or failed payments too? https://github.com/alexbosworth/ln-service#deletefailedpayments

krtk6160: I have the same question as @dolcalmi. If we're only deleting failed payment attempts then we should also rename the function and change the log message accordingly.

krtk6160 (changes requested):

src/services/lnd/utils.ts
nicolasburtey: @dolcalmi eventually we want to delete all failed payment, but starting with the payment attempt is less risky.

I'm not sure yet if there would be some side effect of deleting failed payment for the refund. but deleting payment attempt should have no effect

I'll do the renaming

dolcalmi (dismissed):

src/services/lnd/utils.ts
```diff
@@ -1,39 +1,36 @@
+import {
```
nicolasburtey, line 1: it feels my vs code wants to change import order. is there a way a setting to fix this @samerbuna ?

bodymindarts (approved):

src/services/lnd/utils.ts
samerbuna: It would be somewhere in your saved settings - cmd+shift+p, open settings(json)

Inspect that for any sort/organize settings for imports

PR #576, opened 2021-09-16
Update concourse pipeline to open PR in charts repo

bodymindarts:


ci/tasks/bump-image-digest.sh
```diff
@@ -3,10 +3,14 @@
 set -eu
 
 digest=$(cat ./edge-image/digest)
+digest=$(cat ./edge-image/digest)
```
krtk6160, line 6: Duplicate

PR #577, opened 2021-09-16
Implement csvTransactions

samerbuna:


src/graphql/types/object/business-account.ts
```diff
@@ -37,10 +40,25 @@ const BusinessAccount = new GT.Object({
...
-        walletIds: {
-          type: GT.NonNullList(GT.ID),
+        walletNames: {
+          type: GT.NonNullList(WalletName),
         },
       },
+      resolve: async (source, args) => {
+        const walletIds: WalletId[] = []
+
+        for (const id of source.walletIds) {
```
bodymindarts, line 50: This should be implemented as a use case method under `app` _not_ directly in the resolver!

src/graphql/types/object/consumer-account.ts
```diff
@@ -43,10 +45,25 @@ const ConsumerAccount = new GT.Object({
...
-        walletIds: {
-          type: GT.NonNullList(GT.ID),
+        walletNames: {
+          type: GT.NonNullList(WalletName),
         },
       },
+      resolve: async (source, args) => {
+        const walletIds: WalletId[] = []
+
+        for (const id of source.walletIds) {
```
bodymindarts, line 55: This looks like duplication - should be calling a use-case method not live in the resolver.

src/core/wallet-factory.ts
```diff
@@ -43,3 +45,11 @@ export const getWalletFromRole = async ({ logger, role }) => {
   const user = await User.findOne({ role })
   return WalletFactory({ user, logger })
 }
+
+export const getStringCsvForWallets = async (walletIds: WalletId[]) => {
```
bodymindarts, line 49: Why is new code being added under `core`? This is fragmenting the code base. The code you have here + under the resolver should be unified under an `app/accounts` use case.

bodymindarts (changes requested):

src/app/accounts/index.ts
```diff
@@ -24,3 +25,26 @@ export const getBusinessMapMarkers = async () => {
...
+  const walletIds: WalletId[] = []
+
+  for (const walletName of walletNames) {
+    const wallet = await wallets.findByWalletName(walletName)
+    if (wallet instanceof Error) {
+      return wallet
+    }
+
+    if (!account.walletIds.includes(wallet.id)) {
+      return new RepositoryError()
```
dolcalmi, line 44: Repository error is an error that only should be used at repository level. Please use a custom error, AuthorizationError, ValidationError

dolcalmi (changes requested):
I did the review assuming that this is not a clean up

src/app/accounts/index.ts
samerbuna: Done

src/graphql/types/object/business-account.ts
samerbuna: Done

src/graphql/types/object/consumer-account.ts
samerbuna: Done

src/core/wallet-factory.ts
samerbuna: Done

dolcalmi (approved):

src/app/accounts/index.ts
```diff
@@ -24,3 +25,26 @@ export const getBusinessMapMarkers = async () => {
   const accounts = AccountsRepository()
   return accounts.listBusinessesForMap()
 }
+
+export const toWalletIds = async (
```
bodymindarts, line 29: Just a minor thing - I think we could find a better name here - perhaps `filteredWalletIds` or something like that.

bodymindarts (approved):
Will approve but have a small rename suggestion - can fix or merge as you want.

src/app/accounts/index.ts
samerbuna: Will include that in the new conversion

PR #568, opened 2021-09-11
WIP: Add new `lnInvoicePaymentSend` use case

vindard:
## Description

This PR introduces the new LN Payment Send use-case method which will eventually replace the existing `pay` method on the `LightningWallet` object.

---
### Open `TODO`s:
- [x] ~Fix `make check-code` errors~ (_Done in [`force-push`](https://github.com/GaloyMoney/galoy/compare/8500bfffeda3e538042e8cce5b98cc9d25ab7346..eacb2525f4e40abfc15f47ef691cc00b420ad149)_)
- [x] ~double-check how `LockService` is applied and remove lock comment~ (_Done in [`force-push`](https://github.com/GaloyMoney/galoy/compare/eacb2525f4e40abfc15f47ef691cc00b420ad149..3c906aeadc5dda5448c935105c863ebd960c88b4)_)
- [x] ~Double-check logic against old `pay` method~ (_Done in `force-pushes`: [`1`](https://github.com/GaloyMoney/galoy/compare/c1afe0fe20d916913913283e6ea0fa0986c13f7d..20a7773dbab785dcc9a3e8675c0566988e606df1), [`2`](https://github.com/GaloyMoney/galoy/compare/20a7773dbab785dcc9a3e8675c0566988e606df1..0ec7b1c2a9c001bc53c36daa2bea4ce4ad0e2ca5)_)
- [x] ~Add new unit tests where appropriate~
  - [x] ~`LimitsChecker`~ (Done: https://github.com/GaloyMoney/galoy/pull/568/commits/77c2ebc724191114873956e4931bba1ad882054d)
  - [x] ~`LnFeeCalculator`~ (Done: https://github.com/GaloyMoney/galoy/pull/568/commits/0698b18bc5920470fe30919da5663f29ac10d31d)
  - [x] ~`LnInvoiceValidator`~ (Done: https://github.com/GaloyMoney/galoy/pull/568/commits/edb73e41e2b9628955122d5795c2e139420cd8da)
  - [x] ~`WalletInvoiceValidator`~ (Done: https://github.com/GaloyMoney/galoy/pull/568/commits/989362a9177217e4b5fe0180ba271389563377f8)
- [x] ~Double-check all new methods have types (not implicit any)~ (Done in [`force-push`](https://github.com/GaloyMoney/galoy/compare/11ed72acf0da10033927dc9de78637f425ccaab5..c944ab7fd2b1a842c8d9f56024ab99c7b642d849))
- [x] ~2FA~
[...]

PR #580, opened 2021-09-17
Refactor reimburseFee method nested control flow

vindard:
### Description

This PR removes a level of indentation in the `reimburseFee` method by adding an early return for a conditional check in the method. It also moves the method to its own module to be re-used by the upcoming `send-payment` use-case.

### DIFF
--- a/src/app/wallets/index.ts
+++ b/src/app/wallets/index.ts
@@ -7,6 +7,7 @@ export * from "./update-pending-invoices"
 export * from "./update-pending-payments"
 export * from "./add-invoice-for-wallet"
 export * from "./wallet-name-available"
+export * from "./reimburse-fee"
 
 import { WalletsRepository } from "@services/mongoose"
 
--- a/src/app/wallets/index.ts
+++ b/src/app/wallets/index.ts
@@ -7,7 +7,6 @@ export * from "./update-pending-invoices"
 export * from "./update-pending-payments"
 export * from "./add-invoice-for-wallet"
 export * from "./wallet-name-available"
-export * from "./reimburse-fee"
 
 import { WalletsRepository } from "@services/mongoose"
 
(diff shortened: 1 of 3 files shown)


src/app/wallets/index.ts
```diff
@@ -7,6 +7,7 @@ export * from "./update-pending-invoices"
 export * from "./update-pending-payments"
 export * from "./add-invoice-for-wallet"
 export * from "./wallet-name-available"
+export * from "./reimburse-fee"
```
bodymindarts, line 10: Should this really be exposed outside of the module? It seems like this is a 'helper' use-case that is only called internally or?

src/app/wallets/reimburse-fee.ts
```diff
@@ -0,0 +1,57 @@
...
+  actualFee: Satoshis
+  logger: Logger
+}): Promise<void | ApplicationError> => {
+  const feeReimbursement = FeeReimbursement(maxFee)
+  const feeDifference = feeReimbursement.getReimbursement({
+    actualFee,
+  })
+  if (feeDifference === null) {
+    logger.warn(
+      `Invalid reimbursement fee for ${{
```
nicolasburtey, line 26: logger should be use with: `logger.warn({var}, "message with no variable")`. this make searching/parsing messages in the log much easier in production

src/app/wallets/index.ts
vindard: It also gets called from `send-payment.ts` in the `@app/lightning` module ([see here](https://github.com/GaloyMoney/galoy/pull/568/files#diff-d555811974fb87c7d88c2f0b8cb302056376a294117f01ef0a35620cb2e69ab0R383))

Do we prefer to call it more directly instead? Maybe like:
```
import { reimburseFee } from "@app/wallets/reimburse-fee"
```

src/app/wallets/reimburse-fee.ts
vindard: Sure can do, would it be like this?:
```ts
logger.warn(
  {
    maxFee,
    actualFee,
  },
  `Invalid reimbursement fee`,
)
```

vindard: Done: https://github.com/GaloyMoney/galoy/pull/580/commits/63eb1d1891e6843e386c0ea12fad485083d639d1

bodymindarts (approved):

samerbuna (approved):

PR #581, opened 2021-09-17
Add `isLocal` & `delete..` lightning service methods

vindard:
## Description

Add two new needed lightning service methods to codebase:
- `isLocal`
- `deleteUnpaidInvoice`

### DIFF
--- a/src/core/lightning/index.ts
+++ b/src/core/lightning/index.ts
@@ -330,7 +330,7 @@ export const LightningMixin = (superclass) =>
             try {
               const lndService = LndService()
               if (lndService instanceof Error) return lndService
-              const deleteResult = lndService.deleteUnpaidInvoice({
+              const deleteResult = lndService.cancelInvoice({
                 pubkey,
                 paymentHash: id,
               })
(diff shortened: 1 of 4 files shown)


bodymindarts (approved):

src/core/lightning/index.ts
```diff
@@ -318,9 +323,13 @@ export const LightningMixin = (superclass) =>
             // if we failed to do it, the invoice would still be present in InvoiceUser
             // in case the invoice were to be paid another time independantly (unlikely outcome)
             try {
-              const { lnd } = getLndFromPubkey({ pubkey })
-
-              await cancelHodlInvoice({ lnd, id })
+              const lndService = LndService()
+              if (lndService instanceof Error) return lndService
+              const deleteResult = lndService.deleteUnpaidInvoice({
```
nicolasburtey, line 328: I think Unpaid in `deleteUnpaidInvoice` may be a bit misleading. the invoice was paid if we are in this branch. just not through lnd but with intra ledger.

vindard: Ok I can maybe use something like `cancelInvoice` instead?

vindard: Done here: https://github.com/GaloyMoney/galoy/pull/581/commits/99d27842d1cd175ebee5580928804bde6e5bf0ca

bodymindarts (approved):

bodymindarts (approved):

samerbuna (approved):
💪🏽 💪🏽 💪🏽  So much more readable

PR #582, opened 2021-09-17
Add new `send...Tx` methods to Ledger service

vindard:
## Description

Add new ledger service methods:
- `sendLnTx` to replace `addLnPayment`
- `sendIntraledgerTx` to replace `addOnUsPayment`

### Open `TODO`s:
- [x] Add/confirm support for `onChainPay` usage
- [x] swap into places that use `addLnPayment` and `addOnUsPayment`

### DIFF
--- a/src/core/lightning/index.ts
+++ b/src/core/lightning/index.ts
@@ -5,7 +5,7 @@ import { verifyToken } from "node-2fa"
 
 import * as Wallets from "@app/wallets"
 import { TIMEOUT_PAYMENT } from "@services/lnd/auth"
-import { WalletInvoicesRepository } from "@services/mongoose"
+import { WalletInvoicesRepository, WalletsRepository } from "@services/mongoose"
 import { getActiveLnd, getLndFromPubkey, validate } from "@services/lnd/utils"
 import { ledger } from "@services/mongodb"
 import { redis } from "@services/redis"
@@ -29,6 +29,11 @@ import { addContact, isInvoiceAlreadyPaidError, timeout } from "../utils"
 import { lnPaymentStatusEvent } from "@config/app"
 import pubsub from "@services/pubsub"
 import { LndService } from "@services/lnd"
+import { PriceService } from "@services/price"
+import { LedgerService } from "@services/ledger"
+import { toSats } from "@domain/bitcoin"
+import { toLiabilitiesAccountId } from "@domain/ledger"
+import { CouldNotFindError } from "@domain/errors"
 
 export type ITxType =
   | "invoice"
@@ -298,19 +303,40 @@ export const LightningMixin = (superclass) =>
             })
           }
 
-          await lockExtendOrThrow({ lock, logger: lightningLoggerOnUs }, async () => {
-            const tx = await ledger.addOnUsPayment({
-              description: memoInvoice,
-              sats,
-              metadata,
-              payerUser: this.user,
-              payeeUser,
-              memoPayer,
-              shareMemoWithPayee: isPushPayment,
-              lastPrice: UserWallet.lastPrice,
-            })
-            return tx
-          })
+          const price = await PriceService().getCurrentPrice()
+          if (price instanceof Error) throw price
+          const lnFee = toSats(0)
+          const usd = sats * price
+          const usdFee = lnFee * price
+
+          const payerWallet = await WalletsRepository().findById(this.user.id)
+          if (payerWallet instanceof CouldNotFindError) throw payerWallet
+          if (payerWallet instanceof Error) throw payerWallet
+          const recipientWallet = await WalletsRepository().findById(payeeUser.id)
+          if (recipientWallet instanceof CouldNotFindError) throw recipientWallet
+          if (recipientWallet instanceof Error) throw recipientWallet
+
+          const journal = await lockExtendOrThrow(
+            { lock, logger: lightningLoggerOnUs },
+            async () => {
+              return LedgerService().sendLnIntraledgerTx({
+                liabilitiesAccountId: toLiabilitiesAccountId(this.user.id),
+                paymentHash: id,
+                description: memoInvoice,
+                sats,
+                fee: lnFee,
+                usd,
+                usdFee,
+                pubkey,
+                recipientLiabilitiesAccountId: toLiabilitiesAccountId(payeeUser.id),
+                payerWalletName: payerWallet.walletName,
+                recipientWalletName: recipientWallet.walletName,
+                memoPayer: memoPayer || null,
+                shareMemoWithPayee: isPushPayment,
+              })
+            },
+          )
+          if (journal instanceof Error) throw journal
 
           transactionNotification({
             amount: sats,
--- a/src/core/lightning/index.ts
+++ b/src/core/lightning/index.ts
@@ -319,7 +319,7 @@ export const LightningMixin = (superclass) =>
           const journal = await lockExtendOrThrow(
             { lock, logger: lightningLoggerOnUs },
             async () => {
-              return LedgerService().sendLnIntraledgerTx({
+              return LedgerService().addLnIntraledgerTxSend({
                 liabilitiesAccountId: toLiabilitiesAccountId(this.user.id),
                 paymentHash: id,
                 description: memoInvoice,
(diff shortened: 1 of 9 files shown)


vindard:
**Note:** I can't remove `addOnUsPayment` or `addLndPayment` methods as yet because they're used in the tests in places that test the USD dealer

bodymindarts (approved):

samerbuna:
We'll need to either merge this first and rebase #592 or merge the latter first and rebase this one.

vindard:
> We'll need to either merge this first and rebase #592 or merge the latter first and rebase this one.

True, would be easier to do this one first from a merge-conflict-resolution standpoint

src/core/on-chain/index.ts
```diff
@@ -199,19 +205,40 @@ export const OnChainMixin = (superclass) =>
...
+          const onChainFee = toSats(0)
+          const usd = sats * price
+          const usdFee = onChainFee * price
+
+          const payerWallet = await WalletsRepository().findById(this.user.id)
+          if (payerWallet instanceof CouldNotFindError) throw payerWallet
+          if (payerWallet instanceof Error) throw payerWallet
+          const recipientWallet = await WalletsRepository().findById(payeeUser.id)
+          if (recipientWallet instanceof CouldNotFindError) throw recipientWallet
+          if (recipientWallet instanceof Error) throw recipientWallet
```
samerbuna, line 219: These are repeated lightning, I wonder if we should extract them into a new method.

samerbuna (changes requested):
Branch needs a rebase now

src/core/lightning/index.ts
```diff
@@ -451,17 +477,27 @@ export const LightningMixin = (superclass) =>
...
+          const price = await PriceService().getCurrentPrice()
+          if (price instanceof Error) throw price
+          const lnFee = toSats(fee)
+          const usd = sats * price
+          const usdFee = lnFee * price
+
           entry = await lockExtendOrThrow({ lock, logger: lightningLogger }, async () => {
             // reduce balance from customer first
-            const tx = await ledger.addLndPayment({
+            return LedgerService().sendLnTx({
```
dolcalmi, line 488: `send` makes sense in the app context but in this case you are adding records to the ledger.

src/core/lightning/index.ts
```diff
@@ -293,19 +298,40 @@ export const LightningMixin = (superclass) =>
...
-              shareMemoWithPayee: isPushPayment,
-              lastPrice: UserWallet.lastPrice,
-            })
-            return tx
-          })
+          const price = await PriceService().getCurrentPrice()
+          if (price instanceof Error) throw price
+          const lnFee = toSats(0)
+          const usd = sats * price
+          const usdFee = lnFee * price
```
dolcalmi, line 305: this is alwasy 0

src/core/lightning/index.ts
```diff
@@ -293,19 +298,40 @@ export const LightningMixin = (superclass) =>
...
+          if (payerWallet instanceof CouldNotFindError) throw payerWallet
+          if (payerWallet instanceof Error) throw payerWallet
+          const recipientWallet = await WalletsRepository().findById(payeeUser.id)
+          if (recipientWallet instanceof CouldNotFindError) throw recipientWallet
+          if (recipientWallet instanceof Error) throw recipientWallet
+
+          const journal = await lockExtendOrThrow(
+            { lock, logger: lightningLoggerOnUs },
+            async () => {
+              return LedgerService().sendLnIntraledgerTx({
```
dolcalmi, line 317: same here, we are not `sending` we are adding a record to the ledger

src/core/on-chain/index.ts
```diff
@@ -199,19 +205,40 @@ export const OnChainMixin = (superclass) =>
...
-              shareMemoWithPayee: false,
-              lastPrice: UserWallet.lastPrice,
-            })
-            return tx
-          })
+          const price = await PriceService().getCurrentPrice()
+          if (price instanceof Error) throw price
+          const onChainFee = toSats(0)
+          const usd = sats * price
+          const usdFee = onChainFee * price
```
dolcalmi, line 212: always 0

dolcalmi (changes requested):
we need to update/add ledger tests for this changes

src/core/on-chain/index.ts
vindard: Potentially, it's done twice so far. I've been using a heuristic of not abstracting into methods until we repeat more than twice (guided by @bodymindarts on some "duplicated code" discussions we've had) or where the logic is potentially tricky (like with [this `checkAndVerifyTwoFA` helper method](https://github.com/GaloyMoney/galoy/pull/583/files#diff-8d34911386ecfd5c7a8de035b7d60a6981ac3911029b4e530b503f5009db7f70R35) I made)

src/core/on-chain/index.ts
vindard: Yup, but I was thinking to leave it in to signal that it's dependent on `onChainFee` & `price` (in case we change the `onChainFee` value to be non-zero).

We can also toy with abstracting this differently maybe?

src/core/lightning/index.ts
vindard: This make sense. Change all send/receive references in the LedgerService to `add...`: https://github.com/GaloyMoney/galoy/pull/582/commits/e03d7bee78d2c789f9baece3cfd03ac695c425cf

src/core/lightning/index.ts
vindard: https://github.com/GaloyMoney/galoy/pull/582#discussion_r715630944

vindard:
> we need to update/add ledger tests for this changes

Done here: https://github.com/GaloyMoney/galoy/pull/582/commits/4eb3214cc5ba90ab3fc7039f7b3bfd2799520b4f

src/core/on-chain/index.ts
dolcalmi: the two ifs are unnecessary, CouldNotFindError extends Error so you only need it once
```
if (payerWallet instanceof Error) throw payerWallet
```
and
```
if (recipientWallet instanceof Error) throw recipientWallet
```

src/services/ledger/index.types.d.ts
```diff
@@ -0,0 +1,37 @@
...
+  hash: PaymentHash
+  pubkey: Pubkey
+  memoPayer: string | null
+  username: WalletName | null
+}
+
+type AddOnChainIntraledgerTxSendMetadata = TxMetadata & {
+  payee_addresses: OnChainAddress[]
+  sendAll: boolean
+  memoPayer: string | null
```
dolcalmi, line 31: dont we use `memoPayer?: string` because we need to persist `memoPayer` attribute? or because we need null instead undefined?

src/services/ledger/index.types.d.ts
```diff
@@ -0,0 +1,37 @@
...
+  username: WalletName | null
```
dolcalmi, line 32: same question here

dolcalmi (changes requested):

src/core/on-chain/index.ts
vindard: @dolcalmi yup but I wanted to leave it in until PR #588 in case we decide to add some separate logic to the `CouldNotFindError` branch instead of simply a return

src/services/ledger/index.types.d.ts
vindard: We've been discussing how to represent the absence of an expected value like this (`undefined`, `""`, or `null`) and I believe the general preference is to use `null` wherever possible. @bodymindarts or @samerbuna would have to confirm

dolcalmi (approved):

src/services/ledger/index.types.d.ts
bodymindarts: Yes I think we should converge on a consistent representation of `absence of an expected value`. I don't really mind which we use as long as we are consistent. The discussion once came up with @samerbuna wether nor not to use `""` or `null` to represent `absence of a string value` and he said `null` was more idiomatic.

Either way I think `optional?` arguments should be avoided as I find this point towards there actually  being multiple  different use-cases rather than just the absence of a value.

bodymindarts (approved):

src/core/on-chain/index.ts
vindard: Example of separate logic [here](https://github.com/GaloyMoney/galoy/pull/602/files#diff-e43ce1074d48454f455720e70079bd3d8c6832daeb324523f6fddd5ba4cc7e81R14)

PR #583, opened 2021-09-17
Add `LimitsChecker` object

vindard:
## Description

This PR adds a `LimitChecker` to aggregate and replace the various limit checks we run in the existing code. It also implements a new `txVolumeFor` method in the Ledger service that can be used to fetch the TxVolume input for the checker methods.

Old methods being replaced:
- `remainingWithdrawalLimit`
- `remainingOnUsLimit`
- `remainingTwoFALimit`

### Open `TODO`s
- [x] swap into places that use the existing implementation

### DIFF
--- a/src/core/lightning/index.ts
+++ b/src/core/lightning/index.ts
@@ -227,7 +227,7 @@ export const LightningMixin = (superclass) =>
       const twoFACheck = twoFA?.secret
         ? await checkAndVerifyTwoFA({
             amount: toSats(tokens),
-            twoFAToken: twoFAToken as TwoFAToken,
+            twoFAToken: twoFAToken ? (twoFAToken as TwoFAToken) : null,
             twoFASecret: twoFA.secret,
             limitsChecker: twoFALimitsChecker,
           })
--- a/src/core/lightning/index.ts
+++ b/src/core/lightning/index.ts
@@ -231,7 +231,7 @@ export const LightningMixin = (superclass) =>
             twoFASecret: twoFA.secret,
             limitsChecker: twoFALimitsChecker,
           })
-        : null
+        : true
       if (twoFACheck instanceof TwoFANewCodeNeededError)
         throw new TwoFAError("Need a 2FA code to proceed with the payment", {
           logger: lightningLogger,
--- a/src/core/lightning/index.ts
+++ b/src/core/lightning/index.ts
@@ -251,18 +251,22 @@ export const LightningMixin = (superclass) =>
         })
         if (balanceSats instanceof Error) throw balanceSats
 
+        const limitsChecker = await getLimitsChecker(this.user.id)
+        if (limitsChecker instanceof Error) throw limitsChecker
+
         // On us transaction
         const lndService = LndService()
         if (lndService instanceof Error) throw lndService
         if (lndService.isLocal(destination) || isPushPayment) {
           const lightningLoggerOnUs = lightningLogger.child({ onUs: true, fee: 0 })
 
-          const remainingOnUsLimit = await this.user.remainingOnUsLimit()
-
-          if (remainingOnUsLimit < tokens) {
-            const error = `Cannot transfer more than ${this.config.limits.onUsLimit} sats in 24 hours`
-            throw new TransactionRestrictedError(error, { logger: lightningLoggerOnUs })
-          }
+          const intraledgerLimitCheck = limitsChecker.checkIntraledger({
+            amount: tokens,
+          })
+          if (intraledgerLimitCheck instanceof Error)
+            throw new TransactionRestrictedError(intraledgerLimitCheck.message, {
+              logger: lightningLoggerOnUs,
+            })
 
           let payeeUser, pubkey, payeeInvoice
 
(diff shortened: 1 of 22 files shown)


src/core/lightning/index.ts
```diff
@@ -227,16 +237,20 @@ export const LightningMixin = (superclass) =>
...
         if (balanceSats instanceof Error) throw balanceSats
 
+        const limitsChecker = await getLimitsChecker(this.user.id)
+        if (limitsChecker instanceof Error) throw limitsChecker
+
         // On us transaction
         if (isMyNode({ pubkey: destination }) || isPushPayment) {
           const lightningLoggerOnUs = lightningLogger.child({ onUs: true, fee: 0 })
 
-          const remainingOnUsLimit = await this.user.remainingOnUsLimit()
```
vindard, line 234: `TODO` double-check that this logic makes sense

There seems to be a code-path where an `onUs` payment can be successfully added but then this limit check can cause the method to throw an error without the `onUs` payment being reverted. To determine if:
- this check **is not** needed here since  it's **_settled intraledger_**, or
- the check **is** needed and the transaction should be reverted because it's **_initiated via lightning_**

src/core/lightning/index.ts
```diff
@@ -201,19 +203,27 @@ export const LightningMixin = (superclass) =>
...
+      const { twoFA } = user
 
-        if (!verifyToken(this.user.twoFA.secret, twoFAToken)) {
-          throw new TwoFAError(undefined, { logger: lightningLogger })
-        }
-      }
+      const twoFACheck = twoFA?.secret
+        ? await checkTwoFA({
+            amount: toSats(tokens),
+            twoFAToken: twoFAToken as TwoFAToken,
```
samerbuna, line 216: `towFAToken` can be undefined, the casting here hides that from the type system

vindard: Fixed: https://github.com/GaloyMoney/galoy/pull/583/commits/5b24d043aa744a41b3f6343a1d5baa8d9a2cbbd6

src/domain/accounts/index.types.d.ts
```diff
@@ -37,9 +37,23 @@ type BusinessMapMarker = {
   mapInfo: BusinessMapInfo
 }
 
+type LimitsChecker = {
+  checkTwoFA({ pendingAmount }: { pendingAmount: Satoshis }): true | LimitsExceededError
```
dolcalmi, line 41: didn't know we can use true as a type, this is a bit weird and maybe shoud be a custom type or just boolean.

vindard: Recommendation from @samerbuna

src/services/ledger/index.ts
```diff
@@ -104,6 +104,41 @@ export const LedgerService = (): ILedgerService => {
     }
   }
 
+  const txVolumeSince = async ({
+    liabilitiesAccountId,
+    timestamp,
+  }: {
+    liabilitiesAccountId: LiabilitiesAccountId
+    timestamp: UnixTimeMs
```
dolcalmi, line 112: we are using Date and unixtime, maybe we should unify how we handle dates/times cc @samerbuna @bodymindarts

src/domain/accounts/index.types.d.ts
```diff
@@ -37,9 +37,23 @@ type BusinessMapMarker = {
...
+  checkIntraledger({
+    pendingAmount,
```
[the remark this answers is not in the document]
vindard: True, maybe just `amount`?

src/services/mongoose/schema.types.d.ts
```diff
@@ -9,8 +9,8 @@ type IPType = {
   lastConnection: Date
 }
 
-type TwoFA = {
-  secret: string | undefined
+type TwoFAForUser = {
+  secret?: TwoFASecret
```
dolcalmi, line 13: should be `secret: TwoFASecret | null` (maybe should be confirmed in code review)

src/core/lightning/index.ts
dolcalmi: @nicolasburtey do we have this in case of fraud attemps?

dolcalmi (commented):
No changes are needed, but some of the comments need to be clarified.

src/services/ledger/index.ts
bodymindarts: I don't mind having multiple different representations.
There are always 3 layers of data representation
- how is it persisted?
- what is the domain type?
- what does the external caller want?

What layer are you suggesting to unify here?

vindard: I was just looking it over again and I think in this specific instance the suggestion makes sense since we convert the unix timestamp [back to a `Date` object](https://github.com/GaloyMoney/galoy/blob/4b5353eb09f353ed0dc6df5d4fb9bf46541a1a79/src/services/ledger/index.ts#L121) when passing it to the Mongoose method.

Done here: https://github.com/GaloyMoney/galoy/pull/583/commits/63bfa7a9617ee483ab34fb2d2589a1ca18e56cde

src/services/mongoose/schema.types.d.ts
vindard: Looking through the codebase I think this `twoFA.secret` point can be a separate cleanup issue because there's some inconsistent logic (sometimes `undefined`, sometimes strict check for `null`) in different places.

bodymindarts (approved):

dolcalmi (approved):

PR #585, opened 2021-09-17
Add `addNewContact` use-case method

vindard:
## Description

Implement `addContactByWalletName` method on `UsersRepository` and use it in new use-case. This new implementation deviates from the old  one in that a user can add a wallet contact, but we can't do the same for a recipient if we only have their wallet info and no user details.

This use-case will later be called from the `lnSendAmount` use-cases.


src/domain/users/index.types.d.ts
```diff
@@ -45,4 +45,11 @@ type User = {
 interface IUsersRepository {
   findById(userId: UserId): Promise<User | RepositoryError>
   update(user: User): Promise<User | RepositoryError>
+  addContactByWalletName({
```
bodymindarts, line 48: This should not be on the repository - adding a contact is business logic.

Should be:
```
let user = UsersRepository.findByWalletName(walletName)
user.addContact(contact)
UsersRepository.update(user)
```

Similar to how we add `lastIp` in [this use-case](https://github.com/GaloyMoney/galoy/blob/main/src/app/users/index.ts#L51-L78)

bodymindarts (commented):
Fine except you have moved business logic into the repository. Thats no good unless as an optimisation (eg for batch updates) - but should really be avoided!

src/domain/users/index.types.d.ts
vindard: I missed this one, thanks!

Refactored now to place/implement this logic appropriately.

src/core/lightning/index.ts
```diff
@@ -344,9 +345,11 @@ export const LightningMixin = (superclass) =>
...
-            await addContact({ uid: this.user._id, username: payeeUser.username })
-          }
+          const addContactResult = await addNewWalletContact({
+            userId: this.user.id,
+            contactWalletId: payeeUser.id,
+          })
+          if (addContactResult instanceof Error) throw addContactResult
 
           // adding contact for the payee
           if (this.user.username) {
```
vindard, line 355: If it's ok to delete this no-longer-supported direction here, I can also delete the soon-to-be-redundant `addContact` method

vindard:
[Moved back to Draft](https://github.com/GaloyMoney/galoy/pull/585#event-5338395212) until we resolve open `Username`/`WalletName`/`WalletPublicId`/`handle` discussion

vindard:
Closing in favor of re-implemented use-case using `Username` instead of `WalletName` paradigm

PR #586, opened 2021-09-19
Wallet invoice validator

vindard:
## Description

This PR adds a `walletInvoiceValidator` to check wallet invoices for lightning initiated payments that will be settled intraledger.

_**Note:** This was not swapped into any parts of the existing codebase because the logic implemented here is strung out across multiple `if` blocks and relies on objects that are poorly typed. We will have to be mindful of capturing the existing errors when using this in the new `send-payment` use-case._

### DIFF
--- a/src/domain/wallet-invoices/index.types.d.ts
+++ b/src/domain/wallet-invoices/index.types.d.ts
@@ -6,6 +6,10 @@ type WalletInvoice = {
   paid: boolean
 }
 
+type WalletInvoiceValidator = {
+  validateToSend({ fromWalletId: WalletId }): void | Error
+}
+
 interface IWalletInvoicesRepository {
   persistNew: (invoice: WalletInvoice) => Promise<WalletInvoice | RepositoryError>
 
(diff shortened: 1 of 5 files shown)


src/domain/wallet-invoices/index.types.d.ts
```diff
@@ -6,6 +6,10 @@ type WalletInvoice = {
   paid: boolean
 }
 
+type WalletInvoiceValidator = {
+  validateToSend({ fromWalletId: WalletId }): void | Error
+}
+
```
samerbuna, line 12: I don't like having this as a void function. Should we make a `isValidToSend` that returns `true | Error`?

vindard: We were discussing the same today. I think this is `true | Error` is the right answer given that two persons have noticed it.

There's a few other places with this pattern that I can change as well.

vindard: Fixed here: https://github.com/GaloyMoney/galoy/pull/586/commits/ab6ace841bc33df931e0f3cd32c29a7ac5141baa

And also fixed in `LimitsChecker` and `TwoFA.verify` methods in #583 for consistency

samerbuna (approved):

bodymindarts (approved):

PR #589, opened 2021-09-21
Introduce tsc-new-check in app/domain/graphql

samerbuna:
- Convert default config yaml file into json
- Standardize custom error aguments
- Fix a few type problems


nicolasburtey:
> Convert default config yaml file into json
why? 😟

PR #594, opened 2021-09-22
Add new Routes repository

vindard:
## Description

Add a repository for fetching, persisting, deleting calculated routes.

### Open `TODO`s
- [x] Swap new repo methods into existing places in the codebase

### DIFF
--- a/src/domain/routes/index.types.d.ts
+++ b/src/domain/routes/index.types.d.ts
@@ -11,15 +11,18 @@ type Route = { roundedUpFee: Satoshis; roundedDownFee: Satoshis }
 
 type CachedRoute = { pubkey: Pubkey; route: RawRoute }
 
-declare const cacheKeySymbol: unique symbol
-type CacheKey = string & { [cacheKeySymbol]: never }
+declare const cachedRouteLookupKeySymbol: unique symbol
+type CachedRouteLookupKey = string & { [cachedRouteLookupKeySymbol]: never }
 
 type CachedRouteKeyGenerator = {
-  generate(args: { paymentHash: PaymentHash; milliSats: MilliSatoshis }): CacheKey
+  generate(args: {
+    paymentHash: PaymentHash
+    milliSats: MilliSatoshis
+  }): CachedRouteLookupKey
 }
 
-interface IRoutesRepository {
-  persist: ({
+interface IRoutesCache {
+  store: ({
     key,
     routeToCache,
   }: {
@@ -27,7 +30,7 @@ interface IRoutesRepository {
     routeToCache: CachedRoute
   }) => Promise<CachedRoute | RepositoryError>
 
-  findByKey: (key: CacheKey) => Promise<CachedRoute | RepositoryError>
+  findByKey: (key: CachedRouteLookupKey) => Promise<CachedRoute | RepositoryError>
 
-  deleteByKey: (key: CacheKey) => Promise<true | RepositoryError>
+  deleteByKey: (key: CachedRouteLookupKey) => Promise<true | RepositoryError>
 }
--- a/src/domain/routes/index.types.d.ts
+++ b/src/domain/routes/index.types.d.ts
@@ -1,66 +1,15 @@
-type RawProbedHop = {
-  /** Standard Format Channel Id */
-  channel: string
-  /** Channel Capacity Tokens */
-  channel_capacity: number
-  /** Fee */
-  fee: number
-  /** Fee Millitokens */
-  fee_mtokens: string
-  /** Forward Tokens */
-  forward: number
-  /** Forward Millitokens */
-  forward_mtokens: string
-  /** Public Key Hex */
-  public_key?: string
-  /** Timeout Block Height */
-  timeout: number
-}
+type ProbeForRouteArgs = import("lightning").ProbeForRouteArgs
+type RawHopWithNumbers = NonNullable<ProbeForRouteArgs["routes"]>[number][number]
 
-type RawProbedRoute = {
-  /** Route Confidence Score Out Of One Million Number */
-  confidence?: number
-  /** Route Fee Tokens Rounded Down Number */
-  fee: number
-  /** Route Fee Millitokens String */
-  fee_mtokens: string
-  hops: RawProbedHop[]
-  messages?: {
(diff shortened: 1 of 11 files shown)


src/domain/routes/index.types.d.ts
```diff
@@ -0,0 +1,58 @@
...
+  forward: number
+  /** Forward Millitokens */
+  forward_mtokens: string
+  /** Public Key Hex */
+  public_key?: string
+  /** Timeout Block Height */
+  timeout: number
+}
+
+type RawRoute = {
```
vindard, line 20: #### Open question (cc @bodymindarts)

I'm having problems reconciling this type. The `RawRoute` type (I made up):
- first gets received from calling `lnService.probeForRoute` (no in-code type found, documented types [here](https://github.com/alexbosworth/ln-service#probeForRoute))
- gets persisted to redis cache along with the pubkey (this becomes `CachedRoute` defined [lower down here](https://github.com/GaloyMoney/galoy/pull/594/files#diff-89760a3bcdcc20c6dd12e0ffd947b3dccccb9b75164c1da95eef1e3a2f604edeR40) by us)
- gets read from cache as part of the `CachedRoute` type
- eventually gets passed back to the `payViaRoutes` method which expects a slightly different type (defined in-code [here](https://github.com/alexbosworth/lightning/blob/9a4d0c726a6dbe455108bc8cb80e8b6cfb47d9c6/lnd_methods/offchain/pay_via_routes.d.ts#L12), documented [here](https://github.com/alexbosworth/ln-service#payViaRoutes))

The type is inconsistent when comparing the `probeForRoute` return and the `payViaRoutes` arg, but the arg is a subset of the return type since we pass the return value directly to the `payViaRoutes` method via the redis cache in the existing codebase.

The type also exists as part of other types when it exists in the library, but not separately as itself.

**The question:** is it ok what I'm trying to do here by trying to isolate the smallest subset of common things from the `probeForRoute` return to pass forward to the `payViaRoutes` method, even though it breaks some of our naming conventions? Do we need some sort of translation for this instead (I suspect this is the case)?

dolcalmi: maybe we should update the package, I had to do it for admin queries. you are referencing the last code/docs from https://github.com/alexbosworth/ln-service#probeForRoute or did you check previous versions?

src/domain/routes/index.types.d.ts
```diff
@@ -0,0 +1,27 @@
...
+type Route = { roundedUpFee: Satoshis; roundedDownFee: Satoshis }
+
+type CachedRoute = { pubkey: Pubkey; route: RawRoute }
+
+declare const cacheKeySymbol: unique symbol
+type CacheKey = string & { [cacheKeySymbol]: never }
+
+type CachedRouteKeyGenerator = {
+  generate(milliSatsAmounts: MilliSatoshis): CacheKey
+}
```
dolcalmi, line 12: has this the same problem discussed for fee calculators?

src/domain/routes/index.types.d.ts
```diff
@@ -0,0 +1,27 @@
...
+
+interface IRoutesRepository {
+  persist: ({
+    key,
+    routeToCache,
+  }: {
+    key
+    routeToCache: CachedRoute
+    time?: Seconds
```
dolcalmi, line 21: missing in params or is not necessary?

src/services/redis/routes.ts
```diff
@@ -0,0 +1,49 @@
...
+    try {
+      const value = JSON.stringify(routeToCache)
+      await redis.set(key, value, "EX", time)
+      return routeToCache
+    } catch (err) {
+      return new UnknownRepositoryError(err)
+    }
+  }
+
+  const findByKey = async (key: CacheKey): Promise<CachedRoute | RepositoryError> => {
```
dolcalmi, line 24: if we use Cache/Cached prefixes maybe the repo should be CacheRouteRepository and not RoutesRepository.

dolcalmi (changes requested):

src/domain/routes/index.types.d.ts
vindard: I put it as an optional param with a default value in the function definition... not sure if that's ok?

src/domain/routes/index.types.d.ts
vindard: Ah do you mean we should also pass the `paymentHash` into the method instead of the constructor... I don't mind that change if yes

src/domain/routes/index.types.d.ts
vindard: I was able to reconcile this by making the types [like this](https://github.com/GaloyMoney/galoy/pull/594/files#diff-89760a3bcdcc20c6dd12e0ffd947b3dccccb9b75164c1da95eef1e3a2f604edeR1)

I was thinking about doing a PR, but I'm not sure I'd push this back up to the library because it may not be a clean change there 🤔

src/domain/routes/index.types.d.ts
vindard: Changed here: https://github.com/GaloyMoney/galoy/pull/594/commits/6efb0abebf31ef6c97c4c57affc220429ba266d2

src/services/redis/routes.ts
vindard: Hm, potentially because the `CacheKey` would usually be the unique identifier for items in the repository but we construct/handle it differently because we're using redis (implementation detail).

Would be interested in hearing @bodymindarts' take on this whole `RoutesRepository` abstraction

vindard: Changed a few names and changed the repo to `RoutesCache` ([changes](https://github.com/GaloyMoney/galoy/compare/e184bf8...9f403b0))

src/domain/routes/index.types.d.ts
vindard: Removed: https://github.com/GaloyMoney/galoy/pull/594/commits/1939ad2700f915e8ef53253e7f44a2484fd01d04

dolcalmi (approved):

bodymindarts (approved):

bodymindarts (approved):

PR #597, opened 2021-09-22
Make quiz question return value and naming singular like the input

jrbuhl93:
This pr isn't strictly necessary because the code is functional as is, but I think these changes make sense considering we made the input a single quiz question id. I wanted to put it out there as an option before integrating the mobile app.


src/graphql/types/payload/user-quiz-questions-update-completed.ts
```diff
@@ -15,4 +15,4 @@ const UserQuizQuestionsUpdateCompletedPayload = new GT.Object({
   }),
 })
 
-export default UserQuizQuestionsUpdateCompletedPayload
+export default UserQuizQuestionUpdateCompletedPayload
```
samerbuna, line 18: Rename this file too, and line 13: `type: UserQuizQuestion`

samerbuna (changes requested):

src/graphql/types/payload/user-quiz-questions-update-completed.ts
jrbuhl93: done

samerbuna (approved):

vindard (approved):

PR #588, opened 2021-09-20
WIP: Add new `lnInvoicePaymentSend` use case

vindard:
## Description

This PR will eventually zero in on the changes for implementing the various Ln send payment use-case methods.

It will start with the same diff from #568 and then commits will progressively drop from the diff as the smaller PRs from [this listing](https://github.com/GaloyMoney/galoy/pull/568#issuecomment-921983040) get merged.

Starting diff is: https://github.com/GaloyMoney/galoy/compare/878525022e4df58026989af771c22d1db78d8c64..e15dcceae57966ebda43c48907ba72f00084b278

PR #598, opened 2021-09-23
Add `..Pubkey` & `payInvoice` lightning service methods

vindard:
## Description

This PR adds the following lightning service methods in prep for implementing the new `send-payment` use-cases:
- `defaultPubkey`
- `lndFromPubkey`
- `payInvoice`


src/core/lightning/index.ts
```diff
@@ -612,19 +602,22 @@ export const LightningMixin = (superclass) =>
...
               id,
               related_journal: entry.journalId,
             })
           }
 
           lightningLogger.info(
-            { success: true, paymentResult, ...metadata },
+            {
+              success: true,
+              paymentResult: { safe_fee: result.roundedUpFee },
```
dolcalmi, line 618: with the previous code we got the full response so not sure if is necessary do the wrapping here, maybe just the payment id and preimage

src/domain/bitcoin/lightning/index.types.d.ts
```diff
@@ -61,30 +80,59 @@ type LnFeeCalculator = {
   max(amount: Satoshis): Satoshis
 }
 
+type PayInvoiceResult = {
+  roundedUpFee: Satoshis
```
dolcalmi, line 84: thinking more about this, is weird that we didn't save the preimage (paymentSecret) in the ledger `ledger.settleLndPayment(id)`. I think should be showed in the app because is the way to do support requests to external services that do not handle the payments correctly. We have had issues with strike, chivo and opennode and the users need to request (to our support chat) the secret/preimage to submit a support case

src/domain/bitcoin/lightning/index.types.d.ts
```diff
@@ -61,30 +80,59 @@ type LnFeeCalculator = {
...
+}
+
 interface ILightningService {
   isLocal(pubkey: Pubkey): boolean | LightningServiceError
+
+  defaultPubkey(): Pubkey
```
dolcalmi, line 90: is this method used?

src/domain/bitcoin/lightning/index.types.d.ts
```diff
@@ -61,30 +80,59 @@ type LnFeeCalculator = {
...
+    maxFee,
+    timeoutMs,
+  }: {
+    paymentHash: PaymentHash
+    rawRoute: RawRoute | null
+    lndAuthForRoute: AuthenticatedLnd | null
+    decodedInvoice: LnInvoice
+    milliSatsAmount: MilliSatoshis
+    maxFee: Satoshis
+    timeoutMs?: MilliSeconds
```
dolcalmi, line 136: I guess this was done previous to optional params conversation, maybe we just need a default value here

src/domain/bitcoin/lightning/ln-invoice.ts
```diff
@@ -27,17 +27,19 @@ export const decodeInvoice = (
...
-    decodedInvoice.routes.forEach((route) =>
-      routeHints.push({
-        baseFeeMTokens: route.base_fee_mtokens,
-        channel: route.channel,
-        cltvDelta: route.cltv_delta,
-        feeRate: route.fee_rate,
-        nodePubkey: route.public_key as Pubkey,
-      }),
-    )
+    decodedInvoice.routes.forEach((rawRoute) => {
```
dolcalmi, line 32: isn't this a reduce?

src/domain/routes/index.types.d.ts
```diff
@@ -0,0 +1,30 @@
...
+type RawHopWithStrings = NonNullable<PayViaPaymentDetailsArgs["routes"]>[number][number]
+
+type PayViaRoutesArgs = import("lightning").PayViaRoutesArgs
+type RawRouteFromProbe = PayViaRoutesArgs["routes"][number]
+type RawRoute = { safe_fee: number } & RawRouteFromProbe
+type Route = { roundedUpFee: Satoshis; roundedDownFee: Satoshis }
+
+type CachedRoute = { pubkey: Pubkey; route: RawRoute }
+
+interface IRoutesRepository {
```
dolcalmi, line 14: now I have doubts with this service, although we use redis is part of `bitcoin/lightning` domain/service

src/domain/routes/index.types.d.ts
```diff
@@ -0,0 +1,30 @@
...
+    milliSatsAmounts: MilliSatoshis
+  }) => Promise<CachedRoute | RepositoryError>
+
+  deleteByPaymentHash: ({
+    paymentHash,
+    milliSatsAmounts,
+  }: {
+    paymentHash: PaymentHash
+    milliSatsAmounts: MilliSatoshis
+  }) => Promise<void | RepositoryError>
```
dolcalmi, line 29: in regards to last conversation is better use `Promise<true | RepositoryError>`?

dolcalmi (changes requested):

src/domain/bitcoin/lightning/index.types.d.ts
vindard: Yup, removed now

src/domain/bitcoin/lightning/index.types.d.ts
vindard: It is in the next PR #588

src/core/lightning/index.ts
vindard: Yup this is one of the open questions that I swapped in here because this is how it's being done currently in the new use-case based on how we're doing the types.

Do you mean we should add the payment id and preimage to the `PayInvoiceResult` type too?

src/domain/bitcoin/lightning/ln-invoice.ts
vindard: I think it's more of a mapping... it's getting at the hop objects inside a `route` array which is collected inside a `routes` array, and remapping the property names (and casting types where appropriate).

src/core/lightning/index.ts
vindard: I just realised, should we be logging things like the `preimage` at all?

src/domain/bitcoin/lightning/index.types.d.ts
vindard: Good idea... it might be easier to make this change directly in the new use-case branch

src/domain/routes/index.types.d.ts
vindard: Hmm I'm not sure. It maybe comes closer to the lightning domain but I'm not sure it quite falls under the lnd `service`.

Are you thinking to have the caching method definitions under the `ILightningService` interface, or to simply move all the files from the `domain/routes` dir to be in the `domain/bitcoin/lightning` dir?

src/core/lightning/index.ts
```diff
@@ -129,7 +132,7 @@ export const LightningMixin = (superclass) =>
 
       const { lnd, pubkey } = getActiveLnd()
 
-      const milliSats = toMilliSats(parseFloat(mtokens))
+      const milliSats = toMilliSats(parseInt(mtokens))
```
bodymindarts, line 135: The call to `parseInt` should be inside the `toMilliSats` function (as it can error and we want to return a validation error if it does).
If we need both conversion from `string` and `number` then we need 2 `toMilliSats` functions. `toMilliSatsFromStr` / `toMilliSatsFromNumber`.

src/core/lightning/index.ts
```diff
@@ -566,37 +568,25 @@ export const LightningMixin = (superclass) =>
...
+            const decodedInvoice = decodeInvoice(params.invoice as EncodedPaymentRequest)
+            if (decodedInvoice instanceof Error) throw decodedInvoice
+
+            const lndService = LndService()
+            if (lndService instanceof Error) throw lndService
+
+            result = await lndService.payInvoice({
+              paymentHash: decodedInvoice.paymentHash,
+              rawRoute: route,
+              lndAuthForRoute: lnd,
```
bodymindarts, line 582: lnd should not be part of the interface of `ILightningService`. Its an interface precisely to hide the implementation details. That we are using lnd is an implementation detail. Maybe its just naming - or perhaps there is a boundary issue. Not sure.

src/domain/bitcoin/lightning/index.types.d.ts
```diff
@@ -67,30 +86,57 @@ type LnFeeCalculator = {
...
+type PayInvoiceResult = {
+  roundedUpFee: Satoshis
+}
+
 interface ILightningService {
   isLocal(pubkey: Pubkey): boolean | LightningServiceError
+
+  defaultPubkey(): Pubkey
+
+  lndFromPubkey(pubkey: Pubkey): AuthenticatedLnd | LightningServiceError
```
bodymindarts, line 98: Tho whole client-side load-balancing / pub key stuff needs to be handled internally to the service! We should not expose any of it to the outside layers.

src/services/lnd/index.ts
```diff
@@ -36,6 +47,15 @@ export const LndService = (): ILightningService | LightningServiceError => {
       .find((item) => item == pubkey)
   }
 
+  const lndFromPubkey = (pubkey: Pubkey): AuthenticatedLnd | LightningServiceError => {
```
bodymindarts, line 50: No method in this service should return `AuthenticatedLnd` - auth is an internal detail of this service.

PR #601, opened 2021-09-27
Change geetest secret key names

krtk6160:


bodymindarts (approved):
Also need to change in `charts` repo.

PR #602, opened 2021-09-27
Use walletPublicId/username instead of walletName

samerbuna:
- In the GrahpQL API, the new field is exposed as WalletId
- Remove the WalletName field and restore the Username field where appropriate
- Add 2 new queries, usernameAvailable and userWalletId, the latter
  to be used to get the public wallet ID from a username


src/app/wallets/add-invoice-for-wallet.ts
```diff
@@ -102,12 +103,64 @@ const registerAndPersistInvoice = async ({
...
   const walletsRepo = WalletsRepository()
-  const wallet = await walletsRepo.findByWalletName(walletName)
+  const wallet = await walletsRepo.findByUsername(username)
   if (wallet instanceof Error) return wallet
 
   return wallet.id
 }
+
+// TODO: Remove. Used in v1 only
+export const addInvoiceForUsername = async ({
```
vindard, line 128: Making a note here about the tradeoffs of this approach (I'm assuming it's simplicity) of going straight from `username` to `walletId` in this separate `addInvoiceForUsername` method vs. having a `username` resolver that can get us a `walletPublicId` and then using that resolved `walletPublicId` to call the `addInvoiceForRecipient` method.

I'm thinking of this within the context of if we decide to go with the `handle` approach we were talking about. I'm ok with this change here for now though since it's only a v1 thing.

src/app/wallets/add-invoice-for-wallet.ts
```diff
@@ -102,12 +103,64 @@ const registerAndPersistInvoice = async ({
...
+  username,
+  amount,
+  memo = "",
+}): Promise<LnInvoice | ApplicationError> => {
+  const walletPublicId = checkedToWalletPublicId(username)
```
vindard, line 133: This feels a little odd (but may be ok in practice?).

On reading I would either expect `username` to be:
- checked user [`checkedToUsername`](https://github.com/GaloyMoney/galoy/pull/602/files#diff-c1ccf97093bb2237b9e83707b603c5c9594131c667291779378b17888e931a6dR10) method
- named/typed as `walletPublicId`/`WalletPublicId`
- be used in an intermediary step to retrieve the `walletPublicId` from the user and then used as `walletPublicId` from here

src/domain/users/index.types.d.ts
```diff
@@ -31,7 +31,9 @@ type UserQuizQuestion = {
 
 type User = {
   readonly id: UserId
-  readonly contacts: WalletContact[]
+  readonly username: Username
+  readonly walletPublicId: WalletPublicId
```
vindard, line 35: This starts to get a little confusing for me. My current understanding is that an `Account` can have a default `WalletPublicId` and a `User` can have a default `AccountId` that can be used to get to a default `WalletPublicId`.

src/domain/users/index.types.d.ts
```diff
@@ -44,5 +46,7 @@ type User = {
 
 interface IUsersRepository {
   findById(userId: UserId): Promise<User | RepositoryError>
+  findByUsername(username: Username): Promise<User | RepositoryError>
+  findByWalletPublicId(walletPublicId: WalletPublicId): Promise<User | RepositoryError>
```
vindard, line 50: Same comment as question above about where `WalletPublicId` props live.

vindard (changes requested):
Finished my first read-through of non-gql code ... left some small comments/questions

src/graphql/types/index.ts
```diff
@@ -2,7 +2,7 @@
 import BTCWallet from "./object/btc-wallet"
 import ConsumerAccount from "./object/consumer-account"
 // import FiatWallet from "./object/fiat-wallet"
-import WalletNameTransaction from "./object/wallet-name-transaction"
+import UsernaemTransaction from "./object/wallet-id-transaction"
```
vindard, line 5: It looks like this was changed to [`IntraLedgerTransaction`](https://github.com/GaloyMoney/galoy/pull/602/files#diff-62e0e652988ce1a59d766870da93ce8f1f3817333e2fbaa5d9493ea085efa6d7R13)

```suggestion
import IntraLedgerTransaction from "./object/wallet-id-transaction"
```

src/graphql/types/index.ts
```diff
@@ -19,7 +19,7 @@ export const ALL_INTERFACE_TYPES = [
   // BusinessAccount,
   BTCWallet,
   // FiatWallet,
-  WalletNameTransaction,
+  UsernaemTransaction,
```
vindard, line 22:
```suggestion
  IntraledgerTransaction,
```

vindard (changes requested):

src/graphql/types/object/wallet-id-transaction.ts
```diff
@@ -48,12 +48,12 @@ const WalletNameTransaction = new GT.Object({
...
     // Non-interface fields
     recipient: {
-      type: WalletName,
-      description: `Settlement destination:
-  Could be null when originalDestination is onChain/LN
-  and the payeee does not have a WalletName`,
+      type: WalletId,
+      description:
+        "Settlement destination: Could be null when originalDestination is onChain/LN" +
+        " and the payeee does not have a username",
```
vindard, line 54:
```suggestion
        " and the payee does not have a username",
```

src/domain/wallets/index.types.d.ts
```diff
@@ -24,28 +24,28 @@ type BaseWalletTransaction = {
   readonly deprecated: Deprecated
 }
 
-type WalletNameTransaction = BaseWalletTransaction & {
-  readonly initiationVia: "walletname"
+type IntraLedgerTransaction = BaseWalletTransaction & {
+  readonly initiationVia: "walletid"
   readonly settlementVia: "intraledger"
-  readonly recipientId: WalletName
+  readonly recipient: Username
```
vindard, line 30: Small nit, `recipient` feels like it could sometimes refer to an entire `user` object while something like `recipientName` or `recipientUsername` feels clearer as a single property.

I just had this question when I came across it being used as `tx.recipient` in the gql code. Just an optional comment though, we don't necessarily need to change this.

dolcalmi (changes requested):
still pending gql review from client

src/graphql/root/mutation/user-contact-update-alias.ts
```diff
@@ -0,0 +1,47 @@
...
+  resolve: async (_, args, { uid }) => {
+    const { username, alias } = args.input
+
+    for (const input of [username, alias]) {
+      if (input instanceof Error) {
+        return { errors: [{ message: input.message }] }
+      }
+    }
+
+    const user = Users.updateContactAlias({
```
vindard, line 30:
```suggestion
    const user = await Users.updateContactAlias({
```

src/graphql/root/mutation/user-contact-update-alias.ts
```diff
@@ -0,0 +1,47 @@
...
+      alias,
+    })
+
+    if (user instanceof Error) {
+      return { errors: [{ message: user.message }] }
+    }
+
+    return {
+      errors: [],
+      user,
```
vindard, line 42: This currently errors if the `createdAt` property is included in the request because our internal `User` model is missing the property. I can do a separate PR for that that we can rebase this on top of.

src/app/index.types.d.ts
```diff
@@ -11,3 +11,8 @@ type ApplicationError =
   | PriceServiceError
   | ValidationError
   | LockServiceError
+
+type ToWalletIdsFunction = (args0: {
```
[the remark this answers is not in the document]
samerbuna: We can go with args

src/app/wallets/add-invoice-for-wallet.ts
samerbuna: I moved them under core

src/app/wallets/add-invoice-for-wallet.ts
samerbuna: Fixed

src/domain/wallets/index.types.d.ts
samerbuna: Renamed

vindard:
`$ make check-code` currently failing... it looks like we're still trying to call `walletName` on the `Wallet` object in a few places

vindard (changes requested):
`$ make check-code` currently failing... it looks like we're still trying to call `walletName` on the `Wallet` object in a few places

src/graphql/types/scalar/wallet-id.ts
```diff
@@ -0,0 +1,27 @@
...
+    return new UserInputError("Invalid type for WalletId")
+  },
+})
+
+function validWalletIdValue(value) {
+  const checkedWalletId = checkedToWalletPublicId(value)
+  if (checkedWalletId instanceof Error) {
+    return checkedWalletId
+  }
+  return new UserInputError("Invalid value for WalletId")
```
vindard, line 24: It looks like this returns an error here on the happy path

src/graphql/root/mutation/ln-invoice-create-on-behalf-of-recipient.ts
```diff
@@ -21,15 +21,15 @@ const LnInvoiceCreateOnBehalfOfRecipientMutation = GT.Field({
...
+    const { recipientWalletId, amount, memo } = args.input
+    for (const input of [recipientWalletId, amount, memo]) {
       if (input instanceof Error) {
         return { errors: [{ message: input.message }] }
       }
     }
 
     const result = await addInvoiceForRecipient({
-      recipient,
+      recipientWalletPublicId: recipientWalletId,
```
vindard, line 32: Unrelated to this change (which looks good), it looks like this mutation can have a `satoshis` property in the return under the `LnInvoice` type but we don't add one here so the mutation always returns a `null` value for that property

```gql
mutation lnInvoiceCreateOnBehalfOfRecipient($input: LnInvoiceCreateOnBehalfOfRecipientInput!) {
  lnInvoiceCreateOnBehalfOfRecipient(input: $input) {
    errors {
      message
    }
    invoice {
      paymentRequest
      paymentHash
      paymentSecret
      satoshis
    }
  }
}
```

Note: you may need to change the `recipientWalletId` value below based on local setup

```json
{"input": {"recipientWalletId": "60b8b880-f7ba-41e9-a4de-ea7cca19912b", "amount": "1000", "memo": "This is a test"}}
```

src/app/users/update-contact-alias.ts
```diff
@@ -0,0 +1,31 @@
...
+  const user = await repo.findById(userId)
+  if (user instanceof Error) {
+    return user
+  }
+
+  const found = user.contacts.find((contact) => contact.username === username)
+  if (!found) {
+    return new ValidationError(`User doesn't have contact ${username}`)
+  }
+  found.alias = alias as ContactAlias
```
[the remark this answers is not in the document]
vindard: We mentioned it, but no further decision was made. This casting here is probably ok to work with for now though, since that's what we're doing in most other places.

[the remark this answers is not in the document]
vindard: On looking at this again though, I know we had said sometime before that we should ideally be checking user inputs before passing them forward. Would `alias` be a candidate for this here?

As far as I can tell no checks were done before it gets to this `persist` point (unless we are doing things in the GraphQL layer somewhere).

test/integration/core/specter-wallet.spec.ts
```diff
@@ -7,7 +7,7 @@ import { getSpecterWalletConfig } from "@config/app"
 import { bitcoindClient, getChainBalance, mineBlockAndSyncAll } from "test/helpers"
 
 const { lnd } = getActiveOnchainLnd()
-const specterWalletName = "specter/coldstorage"
+const specterUsername = "specter/coldstorage"
```
dolcalmi, line 10: this should be `specterWalletName`

dolcalmi (changes requested):

src/graphql/types/scalar/wallet-id.ts
samerbuna: Pushed a fix to this. Please verify.

src/graphql/root/mutation/ln-invoice-create-on-behalf-of-recipient.ts
samerbuna: I don't see the mutation exposing that field. Should it?

vindard: I don't think we need it, we can keep it consistent with the `lnNoAmountInvoiceCreateOnBehalfOfRecipient` mutation which returns a `LnNoAmountInvoice` with no `satoshis` property.

But I guess my question is more should we be able to enter `satoshis` as a property in the GraphQL interface for `lnInvoiceCreateOnBehalfOfRecipient` given that it's a valid property on the `LnInvoice` type returned from that mutation's definition even though we haven't implemented `satoshis` in the mutation implementation?

For context, the difference between the 2 right now is that the [`LnInvoice` type](https://github.com/GaloyMoney/galoy/blob/749544702c63e6e19ddaf0f89aec0be4faabd1c0/src/graphql/main/schema.graphql#L587) in the `LnInvoicePayload` has a `satoshis` property and the [`LnNoAmountInvoice` type](https://github.com/GaloyMoney/galoy/blob/749544702c63e6e19ddaf0f89aec0be4faabd1c0/src/graphql/main/schema.graphql#L611) in the `LnNoAmountInvoicePayload` doesn't.

src/graphql/types/scalar/wallet-id.ts
vindard: Confirmed, works 👍🏾 

---

```gql
mutation lnInvoiceCreateOnBehalfOfRecipient($input: LnInvoiceCreateOnBehalfOfRecipientInput!) {
  lnInvoiceCreateOnBehalfOfRecipient(input: $input) {
    errors {
      message
    }
    invoice {
      paymentRequest
      paymentHash
      paymentSecret
			satoshis
    }
  }
}

{"input": {"recipientWalletId": "60b8b880-f7ba-41e9-a4de-ea7cca19912b", "amount": "100", "memo": "This is a test"}}

```

vindard:
I'm collecting a few related comments here about figuring out how we treat with the `user` -> `account` -> `wallet` relationships. These comments all seem to come back to the question:
> _Can a `user` have context about a `wallet` or should it always refer to an `account` which would have the `wallet` context?_

---
- https://github.com/GaloyMoney/galoy/pull/602#discussion_r718506512

- https://github.com/GaloyMoney/galoy/pull/602#discussion_r718770649
- https://github.com/GaloyMoney/galoy/pull/602#discussion_r718840593
- https://github.com/GaloyMoney/galoy/pull/602#discussion_r718507362
- https://github.com/GaloyMoney/galoy/pull/602#discussion_r718741085
- https://github.com/GaloyMoney/galoy/pull/602#discussion_r718745591
- https://github.com/GaloyMoney/galoy/pull/602#discussion_r718798163
- https://github.com/GaloyMoney/galoy/pull/602#discussion_r718802101

bodymindarts:
I agree not having a direct relationship from user <-> account.

We are aiming for
`user 1-* account 1-* wallet` and should refactor accordingly - even if it is only logical for the time being. IE in code but not reflected (yet) in the db schema.

dolcalmi (approved):

vindard (approved):
Finished testing on my end, all looks good functionality-wise 👍🏾 

If we're doing the user- > account -> wallet changes in a separate PR then this one's good to go

PR #612, opened 2021-10-01
Refactor build version numbers

samerbuna:


dolcalmi (approved):

jrbuhl93 (approved):

src/services/local-cache.ts
```diff
@@ -3,21 +3,30 @@ import { Price } from "@core/price-impl"
 import { DbMetadata } from "./mongoose/schema"
 export const mainCache = new NodeCache()
 
-export const getMinBuildNumber = async () => {
+type BuildNumbers = {
+  minBuildNumber: number
+  lastBuildNumber: number
+}
+
+export const getBuildVersionNumbers = async (): Promise<BuildNumbers | Error> => {
```
vindard, line 11: We've also been using an overall function `try...catch` pattern with an `Unknown...ServiceError` in the `catch` in other services ([e.g.](https://github.com/GaloyMoney/galoy/blob/0bce8ab92dfde8773752db7d5d76eb29af7f9620/src/services/lnd/index.ts#L39)). This might be a good candidate for that pattern too.

vindard (changes requested):
Looks good, just one suggested change

PR #613, opened 2021-10-03
Add lightning payment query to admin schema

dolcalmi:
This is part of admin graphql PRs related to admin support queries

PR #614, opened 2021-10-03
Add lightning invoice query to admin schema

dolcalmi:
This is part of admin graphql PRs related to admin support queries


src/graphql/admin/schema.graphql
```diff
@@ -35,6 +35,7 @@ type Query {
   allLevels: [AccountLevel!]!
   userDetailsByPhone(phone: Phone!): UserDetails!
   userDetailsByUsername(username: Username!): UserDetails!
+  lightningInvoice(hash: String!): LightningInvoice!
```
vindard, line 38: Same as https://github.com/GaloyMoney/galoy/pull/613#discussion_r721673640

```suggestion
  lightningInvoice(hash: LnPaymentHash!): LightningInvoice!
```

vindard (changes requested):
Tested and works well, left one nit about a GraphQL type

src/graphql/admin/schema.graphql
dolcalmi: fixed https://github.com/GaloyMoney/galoy/pull/614/commits/cd91fa92216584ce60db666c43e11a7a7ff26c50

src/app/lightning/lookup-invoice-by-hash.ts
```diff
@@ -0,0 +1,21 @@
...
+import { offchainLnds } from "@services/lnd/utils"
+import { InvoiceNotFoundError } from "@domain/bitcoin/lightning"
+
+export const lookupInvoiceByHash = async (
+  paymentHash: PaymentHash,
+): Promise<LnInvoiceLookup | ApplicationError> => {
+  const lndService = LndService()
+  if (lndService instanceof Error) return lndService
+
+  for (const { pubkey } of offchainLnds) {
```
bodymindarts, line 11: This should not be exposed here. That the `LndService` is doing client-side load balancing is an implementation detail. The loop should be hidden from the caller.

bodymindarts (changes requested):

PR #617, opened 2021-10-04
Enforce more specific node version

RonanJackson:
I was having difficulties setting up my environment. One of the reasons was having the exact version of node working in the terminal at the time of installing packages and running the `make` commands.

I have updated the `package.json` to enforce this a little better to save the next person from some trouble!

Unsure if this engine should have an upper bound.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "test:integration": ". ./.envrc && LOGLEVEL=warn jest --config ./test/jest-integration.config.js --bail --runInBand --verbose $TEST | yarn pino-pretty -c -l"
   },
   "engines": {
-    "node": "14"
+    "node": ">=14.17 <15"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


samerbuna (dismissed):

package.json
```diff
@@ -33,7 +33,7 @@
     "test:integration": ". ./.envrc && LOGLEVEL=warn jest --config ./test/jest-integration.config.js --bail --runInBand --verbose $TEST | yarn pino-pretty -c -l"
   },
   "engines": {
-    "node": "14"
+    "node": ">=14.17"
```
samerbuna, line 36: I think: `>=14.17 <15`

RonanJackson: Done!

samerbuna (approved):

dolcalmi (approved):

PR #624, opened 2021-10-06
make faster onchain confirmation

nicolasburtey:
@vindard @samerbuna @dolcalmi there is no place in the yaml to change that?


samerbuna:
> @vindard @samerbuna @dolcalmi there is no place in the yaml to change that?

I see `minConfirmations` but nothing about default

nicolasburtey:
ok maybe something to add @vindard
I'll push that for now

PR #647, opened 2021-10-15
Refactor limit checks 

vindard:
## Description (see #646)

This PR fixes the mis-allocated limit checks in the different pay functions. It does this by refactoring the `LimitsChecker` object to take the `walletVolume` argument via its sub-methods instead of in its top-level constructor. This facilitates getting the different `walletVolume` values for each of the 3 checks.

### Testing

The existing bug was that **withdrawal** checks were referencing **_intraledger_** volumes and never being properly updated.

Some simple tests were attempted for this but I ran into 2 problems:
- we increment the global time in tests by one week for `oldEnoughForWithdrawal` checks which is messing up `walletVolume` checks which check transactions in the past day
- we'd need to undo transactions (or have fresh transaction set) to test each check individually against its transaction volume while the other volumes remain at 0

Given the complexity of this, I'm opting to not include these new tests until we merge #588 which removes `oldEnoughForWithdrawal` checks entirely and makes some other changes to the tests and `pay` implementation.

### DIFF
--- a/src/services/ledger/index.ts
+++ b/src/services/ledger/index.ts
@@ -129,10 +129,10 @@ export const LedgerService = (): ILedgerService => {
       liabilitiesAccountId,
       timestamp,
       txnTypes: [
-        { type: LedgerTransactionType.IntraLedger },
-        { type: LedgerTransactionType.OnchainIntraLedger },
-        { type: LedgerTransactionType.Payment },
-        { type: LedgerTransactionType.OnchainPayment },
+        LedgerTransactionType.IntraLedger,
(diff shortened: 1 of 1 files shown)


bodymindarts:
> > I think `Intraledger` should be `IntraLedger` - other than that it looks good!
> 
> I noticed this inconsistency as well, but there's 47 different places that use `Intraledger` in the codebase so I opted to leave in for now. Same thing for `Onchain` vs. `OnChain`.

We should converge on `OnChain` and `IntraLedger` - could be a follow up PR

PR #648, opened 2021-10-17
Fix missing onchain transaction hash

dolcalmi:
Fix to issues reported by users in onchian txs


dolcalmi:
> I thought e are using the payment hash field also for on chain up to now

I did some history check and paymentHash has been only in WalletLnTransaction for a while https://github.com/GaloyMoney/galoy/blame/16ebbc1d203170a0f9ffc24eb0f2cadaa94d35e3/src/domain/wallets/index.types.d.ts#L41

PR #658, opened 2021-10-21
Upgrade mongoose to fix mpath vulnerability

krtk6160:
Fix [mpath vulnerability](https://github.com/GaloyMoney/galoy/security/dependabot/yarn.lock/mpath/open)


krtk6160:
> Seems like the migrate-mongo types are not compatible with MongoDB 6.x

Yep, and the problem isn't just with migrate-mongo types, it's with the mongoose types as well.

PR #662, opened 2021-10-22
Refactor price service

dolcalmi:
- Refactor price service
  - current price: now service implementation has all the logic related to local cache and query to grpc
  - history: now range and interval grouping is done in db
- Add price unit/integration tests
  - Add initial setup with 1 year of price data. To test price query please run `TEST="01" make reset-integration`
- Remove old real time price mock
- Fix tests open handles (writeSDLFile and httpServers)
- Remove price-impl class
  - updatePriceHistory was only moved and not refactored because will be changed according to https://github.com/GaloyMoney/galoy/issues/683


dolcalmi:
closed in favor of https://github.com/GaloyMoney/galoy/pull/726

Issue #605, opened 2021-09-29
add payment metadata to medici_transactions


nicolasburtey:
lnd 0.14 will have the capability to delete individual payments.

in the interest of making our node database smaller, we should take advance of this features and delete payment after they complete. 

that said, we will loose some data that we need to store beforehands.

Currently, for data analysis (ie: to know where payment are most frequently sent to), we are getting the individual payment using [trackpayment](https://api.lightning.community/#trackpayment) (using [listpayments](https://github.com/lightningnetwork/lnd/issues/5658) would get better but our node is currently failing with that)

now if we want to delete succesful payments, we need to store this metadata somewhere. I'm suggesting to store it alongside the medici transactions. we could store the json as a blob, and will extract/handle the data manually at a later stage in a data pipeline/data exploration environment.

I think both postgres and mongodb accept having a field as a json (mongodb even more so than postgres. the point is we don't need a structured schema for it)

nicolasburtey:
maybe also @vindard or @dolcalmi can look over this

vindard:
Adding implementation notes here, I believe we'd be persisting values from the returns of these two functions:
- https://github.com/alexbosworth/ln-service#payViaRoutes
- https://github.com/alexbosworth/ln-service#payviapaymentdetails

dolcalmi:
> Adding implementation notes here, I believe we'd be persisting values from the returns of these two functions:
> 
> * https://github.com/alexbosworth/ln-service#payViaRoutes
> * https://github.com/alexbosworth/ln-service#payviapaymentdetails

these methods not always return a final state so we need to use getPayment (for ex in trigger), the only concern about this is that lightning library not always return all attributes from lnd node

vindard:
Adding some other things we'd like to persist as well:
- [ ] Add preimage to `PayInvoiceResult` type and persist in medici (see [comment](https://github.com/GaloyMoney/galoy/pull/598#discussion_r721499238))

nicolasburtey:
this is done

nicolasburtey:
@vindard : "I'll double-check that cron deletes these from lnd"

vindard:
> @vindard : "I'll double-check that cron deletes these from lnd"

It does, all good

PR #665, opened 2021-10-25
Persist Lightning payment data in medici transactions

vindard:
### Description

This PR is to add relevant lightning payment info from lnd to medici transactions in the database (see #605).

A new `LnPayments` collection was set up to persist these entries and they can be uniquely referenced by their `paymentHash` property to tie to transactions we have in medici.

The new `payment` property in the database looks like:

image


dolcalmi (changes requested):

src/services/lnd/index.ts
```diff
@@ -305,28 +305,34 @@ const lookupPaymentByPubkeyAndHash = async ({
...
-        roundedUpFee: toSats(payment.safe_fee),
+        createdAt: new Date(payment.created_at),
+        destination: payment.destination,
+        milliSatsFee: toMilliSatsFromString(payment.fee_mtokens),
+        paymentHash: payment.id,
         milliSatsAmount: toMilliSatsFromString(payment.mtokens),
-        secret: payment.secret as PaymentSecret,
-        status,
+        paths: payment.paths || [],
+        paymentRequest: payment.request,
```
dolcalmi, line 332: lightning library is not returning `payment.request`, you mentioned that there was a pending library update but I dont see updates to package or yarn.lock

dolcalmi (changes requested):

src/services/ledger/index.ts
```diff
@@ -550,13 +551,17 @@ export const LedgerService = (): ILedgerService => {
...
+    payment,
+  }: {
+    paymentHash: PaymentHash
+    payment: LnPaymentLookup
+  }): Promise<boolean | LedgerServiceError> => {
     try {
       const result = await Transaction.updateMany(
         { hash: paymentHash },
-        { pending: false },
+        { pending: false, payment },
```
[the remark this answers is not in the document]
vindard: This one is query by `hash` and then update `pending` and `payment` properties on found objects (at least I believe this is how it works).

src/services/lnd/index.ts
vindard: This always comes back as `null` so far but the property exists in the `getPayment` type ([see here](https://github.com/alexbosworth/lightning/blob/7ca4a1116e74f1bd5d6b233a066fe6153953c670/lnd_methods/offchain/get_payment.d.ts#L96)) so I assumed it returns sometime. But this makes sense if it doesn't ever return and the library type is somehow wrong right now?

vindard:
Latest commits here implement a separate `LnPayments` collection with a `_payment` property in the relevant transactions in the `medici_transactions` collection that has an `ObjectId` object with the id of the payment object in `LnPayments`

For migration, I'm planning to query medici transactions in our db for ones without the `_payment` property and to use the `hash` on these to ping lnd for payment info to populate `LnPayments` collection.

src/services/ledger/schema.ts
```diff
@@ -110,6 +111,10 @@ const transactionSchema = new Schema({
     type: Number,
     min: 0,
   },
+  _payment: {
```
vindard, line 114: Naming this as `_payment` to be consistent with other places where we cross-reference collections (`_journal`, `_transactions`) by their mongoDB ObjectId

src/app/wallets/ln-send-payment.ts
```diff
@@ -483,7 +493,16 @@ const executePaymentViaLn = async ({
         })
     if (payResult instanceof LnPaymentPendingError) return PaymentSendStatus.Pending
 
-    const settled = await ledgerService.settlePendingLnPayments(paymentHash)
+    const payment = await lndService.lookupPayment({ pubkey, paymentHash })
+    if (payment instanceof Error) return payment
+    payment.paymentRequest = payment.paymentRequest || paymentRequest
```
nicolasburtey, line 498: still not clear to me why we're doing this

nicolasburtey: oh nevermind I remember

src/app/wallets/ln-send-payment.ts
```diff
@@ -483,7 +493,16 @@ const executePaymentViaLn = async ({
...
+
+    const persistedPayment = await LnPaymentsRepository().update(payment)
+    if (persistedPayment instanceof Error) return persistedPayment
+    const settled = await ledgerService.settlePendingLnPayments({
```
nicolasburtey, line 502: is it related to the metadata we want to store? or this is core to the payment logic?

vindard: This was previous code we had to settle pending transactions by update the `pending` property on the transactions to `true`.

This change is to also include an update for the `_payment` property to add the `ObjectId` value for the payment in the new payments collection.

nicolasburtey:
> Latest commits here implement a separate `LnPayments` collection with a `_payment` property in the relevant transactions in the `medici_transactions` collection that has an `ObjectId` object with the id of the payment object in `LnPayments`
> 
> For migration, I'm planning to query medici transactions in our db for ones without the `_payment` property and to use the `hash` on these to ping lnd for payment info to populate `LnPayments` collection.
> 
> Planning to do this in a follow-up PR.

then a follow-up to this follow-up PR can be that we delete the [whole payment history, instead of just the payment failure](https://github.com/GaloyMoney/galoy/blob/main/src/servers/cron.ts#L25). this would help reduce lnd bbolt size

src/domain/ledger/index.types.d.ts
```diff
@@ -217,7 +218,13 @@ interface ILedgerService {
...
   ): Promise<LedgerJournal | LedgerServiceError>
 
-  settlePendingLnPayments(paymentHash: PaymentHash): Promise<boolean | LedgerServiceError>
+  settlePendingLnPayments({
+    paymentHash,
+    paymentId,
+  }: {
+    paymentHash: PaymentHash
+    paymentId: PaymentId
+  }): Promise<boolean | LedgerServiceError>
```
nicolasburtey, line 227: I guess `boolean` means `success` here? and this can be only be true? is it a pattern we have adopted? wouldn't it be better if it was `true` instead of `boolean` since it can't be false?

(I know it's not part of the PR so no need to fix it here, just trying to understand the new code modifications)

src/app/wallets/ln-send-payment.ts
```diff
@@ -483,7 +493,16 @@ const executePaymentViaLn = async ({
         })
     if (payResult instanceof LnPaymentPendingError) return PaymentSendStatus.Pending
```
nicolasburtey, line 494: what is the flow is the payment return an error here? (not the case of a pending error but the case the payment simply failed)

src/app/wallets/ln-send-payment.ts
```diff
@@ -483,7 +493,16 @@ const executePaymentViaLn = async ({
...
 
-    const settled = await ledgerService.settlePendingLnPayments(paymentHash)
+    const payment = await lndService.lookupPayment({ pubkey, paymentHash })
+    if (payment instanceof Error) return payment
```
nicolasburtey, line 497: do we always go through this route if the payment had already failed?

src/app/wallets/ln-send-payment.ts
nicolasburtey: is this to tie the payment from medici to the new `LnPayment` collection? in that case, assuming we have a 1-1 mapping, I would recommend having the `_journal.id` to `LnPayment` instead: this is metadata, so it's better not to overload the medici table with more data, because at scale the more data there is on this collection, the slower the requests like `getBalance()` will run.

src/services/mongoose/ln-payments.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+  }
+
+  return {
+    update,
+  }
+}
+
+const lnPaymentFromRaw = (result: LnPaymentType): LnPayment => ({
+  id: result.id as PaymentId,
```
nicolasburtey, line 27: shouldn't this be `._id`?

nicolasburtey: I guess mongoose might provide an alias for `id`

src/domain/ledger/index.types.d.ts
vindard: That's usually right if the function can only return `true` or an error, but in this case that method returns `result.nModified > 0` which can also be false if no records are modified (I believe).

src/app/wallets/ln-send-payment.ts
vindard: The conditional would fail and this return ignored. All other errors not explicitly handled would then be caught at [`Ln 508`](https://github.com/GaloyMoney/galoy/blob/persist-ln-payment-in-medici/src/app/wallets/ln-send-payment.ts#L508).

src/app/wallets/ln-send-payment.ts
vindard: As it is now yes. The intention is that if the payment fails, we can still get back a `status: "failed"` object from lnd with useful metadata.

src/app/wallets/ln-send-payment.ts
nicolasburtey: are you sure the errors from the `lightning` package would be caught correctly?

lightning doesn't return an `Error` instance, but an array (the custom way to doing error reporting was not intuitive to me at first)

therefore I would not think `if (payResult instanceof Error) {` would be trigger with a payment error because payResult would be an `Array`, not an `Error`

src/domain/ledger/index.types.d.ts
nicolasburtey: in that case we want want to trigger an error instead of `false`? 

none of the place where `settlePendingLnPayments` is being called really check for the value of the boolean currently

src/app/wallets/ln-send-payment.ts
vindard: It looks like we handle the errors we know about that get thrown from lightning (see [here](https://github.com/GaloyMoney/galoy/blob/041fd608e022d32fcf512463766f51ea3f967f9f/src/services/lnd/index.ts#L169)) and then for the ones we don't know about we wrap the array in an `UnknownLightningServiceError` object and bubble it up to be handled by calling code.

So in this case as I understand now, any error that isn't an `isInvoiceAlreadyPaidError` error would get put into that wrapper and would be caught by the `if (payResult instanceof Error) {` line

src/domain/ledger/index.types.d.ts
vindard: We potentially can. I didn't include this initially to reflect the original behaviour of this method [here](https://github.com/GaloyMoney/galoy/blob/aa00a0a2e55e147dbb2e4f0a6c6c95aaa57608e5/src/core/lightning/index.ts#L516) & [here](https://github.com/GaloyMoney/galoy/blob/aa00a0a2e55e147dbb2e4f0a6c6c95aaa57608e5/src/core/lightning/index.ts#L545) where we didn't do anything with the return (`true` or `false`), but where if an error happened the entire method would throw.

In the new code we still don't do anything with the return (`true`/`false`), but we return an error where before the method would have just thrown it. 

Do we want to change the behaviour of this function so that `false` is an error as well. I haven't tested for any cases as yet where that could potentially happens and what the implications could be.

vindard:
Temporarily reverting this to _**draft**_ until I add the integration tests I'm working on

src/app/wallets/ln-send-payment.ts
vindard: This was actually just an extra thing and isn't necessary since we already have the mapping via the payment hash. I can remove this to make it simpler.

vindard:
Added tests and reverted the extra property were adding to medici transactions. This should be good to go for review again.

src/app/wallets/ln-send-payment.ts
```diff
@@ -481,9 +491,18 @@ const executePaymentViaLn = async ({
...
+
+    payment.paymentRequest = payment.paymentRequest || paymentRequest
+    payment.paymentHash = payment.paymentHash || paymentHash
+    const persistedPayment = await LnPaymentsRepository().update(payment)
+    if (persistedPayment instanceof Error) return persistedPayment
+
     if (payResult instanceof LnPaymentPendingError) return PaymentSendStatus.Pending
 
-    const settled = await ledgerService.settlePendingLnPayments(paymentHash)
+    const settled = await ledgerService.settlePendingLnPayments({ paymentHash })
```
jireva, line 505: Very possible I'm missing something here but I think we decided to drop the braces for single arguments.

jireva (commented):
This will be great to have.

src/services/lnd/index.ts
```diff
@@ -305,28 +305,34 @@ const lookupPaymentByPubkeyAndHash = async ({
       : PaymentStatus.Pending
 
     let paymentLookup: LnPaymentLookup = {
-      amount: toSats(0),
-      createdAt: new Date(0),
```
vindard, line 309: I changed this here to default to `undefined` instead because 1970 dates in the db was looking confusing for null values. Is there any reason why we initially put this to be `new Date(0)` @dolcalmi?

dolcalmi: yes, created_at is not optional https://github.com/alexbosworth/ln-service#getpayment

src/app/wallets/ln-send-payment.ts
```diff
@@ -481,9 +491,18 @@ const executePaymentViaLn = async ({
           milliSatsAmount: toMilliSatsFromNumber(amount * 1000),
           maxFee,
         })
+
+    const payment = await lndService.lookupPayment({ pubkey, paymentHash })
+    if (payment instanceof Error) return payment
+
+    payment.paymentRequest = payment.paymentRequest || paymentRequest
```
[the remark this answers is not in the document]
dolcalmi: probably is a lightning library issue but sometimes it does not return it. We need to keep this

src/app/wallets/update-pending-payments.ts
```diff
@@ -89,14 +90,24 @@ const updatePendingPayment = async ({
...
         )
         return settled
       }
+      const persistedPayment = await LnPaymentsRepository().update(lnPaymentLookup)
+      if (persistedPayment instanceof Error) {
+        paymentLogger.error(
+          { error: lnPaymentLookup },
+          "we couldn't update payment data to our database",
+        )
+        return persistedPayment
```
dolcalmi, line 107: we should only log the update error but we cant compromise reimburseFee or revert transaction. if you want to keep the return error please move after reimburseFee and revertTransaction

src/app/wallets/ln-send-payment.ts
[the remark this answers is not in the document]
vindard: Fixed, this change got caught up in that in-between period between the dev call and code review 🙈  https://github.com/GaloyMoney/galoy/pull/665/commits/21646d3bcee1290ce843fdf08006c7acb32ed1e2

src/app/wallets/ln-send-payment.ts
```diff
@@ -481,6 +491,15 @@ const executePaymentViaLn = async ({
...
+    payment.paymentHash = payment.paymentHash || paymentHash
+    const persistedPayment = await LnPaymentsRepository().update(payment)
+    if (persistedPayment instanceof Error) return persistedPayment
```
dolcalmi, line 501: please move this after settle or void, in this point payment has been made and we should have settle or void as close as possible

src/services/lnd/schema.ts
```diff
@@ -0,0 +1,75 @@
...
+const paymentSchema = new Schema<LnPaymentType>({
+  status: { type: String, enum: ["settled", "failed", "pending"], required: true },
+  confirmedAt: Date,
+  createdAt: Date,
+  destination: String,
+  milliSatsFee: {
+    type: Number,
+    default: 0,
+  },
+  paymentHash: String,
```
dolcalmi, line 57: please add an index to payment hash

dolcalmi (changes requested):

src/app/wallets/ln-send-payment.ts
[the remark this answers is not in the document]
vindard: I just ran our test suite and for 17 payments added to the `lnpayments` collection they all never return a value for `request`.

But yes the library is inconsistent because it also doesn't return a `pending` array as defined in its documentation [here](https://github.com/alexbosworth/lightning/blob/e948ebd7fca54a04b2b0d3715dd3bc9063aef57f/lnd_methods/offchain/get_payment.js#L20) when status is `pending`.

We would like to persist the payment request somehow though, so instead of including it separately I chose to overwrite the place we would expect to see it from the library return if that value is undefined. We can change this behaviour to something else though if this isn't appropriate for here.

Note: there's a similar thing we do with the `paymentHash` property since for pending transactions lnd doesn't give us any info at all even though there should be a `pending` property.

```ts
// Object returned from lib
{
  failed: undefined,
  is_confirmed: false,
  is_failed: false,
  is_pending: true,
  payment: undefined
}
```

src/app/wallets/update-pending-payments.ts
vindard: Removed the error and opted to log only. We can probably double-check that this is added correctly in whichever job we will be deleting payments from lnd's database from

src/app/wallets/ln-send-payment.ts
vindard: Same as https://github.com/GaloyMoney/galoy/pull/665#discussion_r746697848

src/services/lnd/index.ts
vindard: Right, but the entire `payment` object is optional so there's cases where this isn't present

src/app/wallets/ln-send-payment.ts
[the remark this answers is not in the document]
vindard: Just got confirmation, it looks like `request` is never returned for the `getPayment` method: https://github.com/alexbosworth/lightning/issues/50#issuecomment-965593923

vindard:
[`force-push`](https://github.com/GaloyMoney/galoy/compare/24f6a9d96a457fe82f8097a7c3342f4383c683ee..e1dae3662914dd7b63cfa1d205267a6c6ea8610c) to tip of main

src/services/lnd/schema.ts
vindard: Done https://github.com/GaloyMoney/galoy/pull/665/commits/4643aa6da1b3f17daf754e164bb07ba51d8dcf95

src/services/lnd/schema.types.d.ts
```diff
@@ -0,0 +1,16 @@
+interface LnPaymentType {
+  id: string
+  _id: string
```
nicolasburtey, line 3: same as above? maybe no need to expose both?

src/domain/bitcoin/lightning/index.types.d.ts
```diff
@@ -8,6 +8,9 @@ type InvoiceExpiration = Date & { [invoiceExpirationSymbol]: never }
 declare const encodedPaymentRequestSymbol: unique symbol
 type EncodedPaymentRequest = string & { [encodedPaymentRequestSymbol]: never }
 
+declare const paymentIdSymbol: unique symbol
+type PaymentId = string & { [paymentIdSymbol]: never }
```
nicolasburtey, line 12: I feel `PaymentId` might be too vague given the thing it can mean multiple things. Would `PaymentLedgerId` be a better name?

nicolasburtey (dismissed):
some mostly cosmetic feedback. PR looks good otherwise.

src/domain/bitcoin/lightning/index.types.d.ts
vindard: Yea that makes sense

src/services/lnd/schema.types.d.ts
vindard: I wasn't sure if mongo would need both... was following example from other schema types we have

src/services/lnd/schema.ts
```diff
@@ -46,15 +46,24 @@ const pathSchema = new Schema({
...
   destination: String,
   milliSatsFee: {
     type: Number,
     default: 0,
   },
-  paymentHash: String,
+  paymentHash: {
+    type: String,
+    index: true,
+    unique: true,
```
nicolasburtey, line 64: is this unique? what happen when there are different payment attempt for the same hash?

vindard: Based on what Alex said, these attempts are stored separately internally to `lnd` but are only ever exposed as a single object with latest state when accessed via the api. He was saying too that they had some sort of an attempts arrays which I was experimenting with a few weeks back but couldn't find.

Decided to make this unique because either way it's only one object that would ever get returned from `lnd` based on what they said and we've observed so far.

nicolasburtey: so this means that we only get the last payment from the api? not the history?

nicolasburtey (approved):

src/services/lnd/schema.ts
vindard: As far as I know so far yes

vindard: Some feedback from Alex on this: https://github.com/alexbosworth/lightning/issues/50#issuecomment-965833080

src/services/lnd/index.ts
vindard: Looks like it doesn't return this property when using the `getPayment` method: https://github.com/alexbosworth/lightning/issues/50#issuecomment-965593923

vindard:
Moving to draft to test issue in update-payments flow

vindard:
Ready for review again.

I changed the structure of the persisted data so that the `status`, `paymentHash` and `paymentRequest` properties are at the top-level, and the rest of data is placed in a `paymentDetails` object that gets updated if the payment moves from `pending` to `settled`, but doesn't get erased if the payment eventually fails.

package.json
```diff
@@ -74,7 +74,7 @@
     "ioredis": "^4.27.6",
     "js-yaml": "^4.1.0",
     "jsonwebtoken": "^8.5.1",
-    "lightning": "^4.10.6",
+    "lightning": "^4.14.2",
```
dolcalmi, line 77: latest version now is 5.1.0

src/services/mongoose/ln-payments.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  const update = async (
+    payment: LnPaymentLookup,
+  ): Promise<LnPayment | RepositoryError> => {
+    try {
+      const result = await LnPayment.findOneAndUpdate(
+        { paymentHash: payment.paymentHash },
+        payment,
+        { upsert: true, new: true },
```
dolcalmi, line 30: this does not insert default values, if in this case we want to use them then we should use:
```
{ upsert: true, new: true, setDefaultsOnInsert: true },
```

dolcalmi (changes requested):

nicolasburtey:
@vindard this needs a rebase

src/services/lnd/schema.ts
```diff
@@ -0,0 +1,88 @@
...
+  },
+  forward_mtokens: String,
+  public_key: String,
+  timeout: {
+    type: Number,
+    min: 0,
+  },
+})
+
+const pathSchema = new Schema({
```
nicolasburtey, line 28: it should work for MPP right?
There can be many paths with many hops?

vindard: Yes exactly, we haven't been able to trigger/test mpp as yet but from the object you sent when we were testing with Loop this is exactly where all the paths end up

vindard:
@jireva can you run `TEST="01|02" make reset-integration` and take a look at the objects in new `lnpayments` collection in mongo to see if what we're doing here to persist payments looks ok or can be improved?

nicolasburtey:
the second row doesn't have a `createdAt` for some reason:

image

src/app/wallets/update-pending-payments.ts
```diff
@@ -59,21 +61,45 @@ const updatePendingPayment = async ({
...
-      protocol: "lightning",
-      transactionType: "payment",
-      onUs: false,
-    })
-    lightningLogger.error({ err: lnPaymentLookup }, "issue fetching payment")
+    lightningLogger.error(
+      { err: lnPaymentLookup },
+      "issue fetching payment from database",
+    )
     return lnPaymentLookup
```
nicolasburtey, line 93: I'm not sure we should fail if there is an issue with `lnPaymentLookupPromise`, ie: `LnPaymentsRepository().findByPaymentHash(paymentHash)`. 

this is a non mandatory metadata, so we could just log the error and continue?

src/graphql/error-map.ts
```diff
@@ -80,6 +80,10 @@ export const mapError = (error: ApplicationError): CustomApolloError => {
       message = "Invoice is missing its 'payment secret' value"
       return new InvoiceDecodeError({ message, logger: baseLogger })
 
+    case "CouldNotFindLnPaymentFromHashError":
+      message = `Ln Payment does not exist for payment hash ${error.message}`
```
nicolasburtey, line 84: should not be sent to client if we make this error not flaw back to the client because it's not critical?

or is it required because of the Typescript rules?

nicolasburtey:
one last set of comments, but looks good for me otherwise

jireva:
> the second row doesn't have a `createdAt` for some reason:

It seems that when a payment fails, sometimes `paymentDetails` is set to `null` while other times it get's created but is missing `confirmedAt` and `secret`.  
If you dump the `lnpayments` collection:
```
$ mongoexport --uri="mongodb://testGaloy:password@localhost/galoy" --collection=lnpayments --jsonArray > lnpayments
```

You can use jq to compare two failed payments, one of which is missing the `paymentDeatils` field:
```
$ jq '.[0,7]' lnpayments
{
  "_id": {
    "$oid": "61a7c893c7f362f7473a08a1"
  },
  "paymentHash": "ded83fcb8cd2944a715adde74d275a1e8d141212207c0581943057f86c57b633",
  "__v": 0,
  "paymentDetails": null,
  "paymentRequest": "lnbcrt15m1ps60jynpp5mmvrljuv622y5u26mhn56f66r6x3gysjyp7qtqv5xptlsmzhkcesdqqcqzpuxqr23ssp5vkar0rfxagl6msvn84cnmjrsjc7j7vtf397qcu952tr8c6y0e5qq9qyyssq52qwfdtfeawtwlkc7tfp2mfap6aum4vkyzc0f32tchv8jdl9km042tpxdq57num282d92m5w5cn5vn6fk3f9rvhhehzzg7z2wsmkquqqyd4q7d",
  "status": "failed"
}
{
  "_id": {
    "$oid": "61a7c8a5c7f362f7473a0b4f"
  },
  "paymentHash": "f1330670b47c8215f7031624cfadc0832ad10c9332b0505bee88a7be0a798533",
  "__v": 0,
  "paymentDetails": {
    "milliSatsFee": 0,
    "roundedUpFee": 0,
    "_id": {
      "$oid": "61b2548c082b7fede7ba6885"
    },
    "createdAt": {
      "$date": "2021-12-01T19:10:26.076Z"
    },
    "destination": "0362bf71f3ff53aa074be9a19da6c7e49a50739dbf03cefaff8c4dc73d2aac6079",
    "milliSatsAmount": 1000000,
    "paths": [
      {
        "fee": 0,
        "_id": {
          "$oid": "61b2548c082b7fede7ba687f"
        },
        "fee_mtokens": "0",
        "hops": [
          {
            "fee": 0,
            "_id": {
              "$oid": "61b2548c082b7fede7ba6880"
            },
            "channel": "148x1x0",
            "channel_capacity": 1000000,
            "fee_mtokens": "0",
            "forward": 1000,
            "forward_mtokens": "1000000",
            "public_key": "0362bf71f3ff53aa074be9a19da6c7e49a50739dbf03cefaff8c4dc73d2aac6079",
            "timeout": 270
          }
        ],
        "mtokens": "1000000",
        "payment": "08004f3d447164fd2343b95d717ac28ecd2c632f4c7782f18b6b9b67648430e8",
        "timeout": 270,
        "tokens": 1000,
        "total_mtokens": "1000000"
      }
    ],
    "amount": 1000
  },
  "paymentRequest": "lnbcrt10u1ps60j9ppp57yesvu950jpptacrzcjvltwqsv4dzrynx2c9qklw3znmuznes5esdqqcqzpuxqyz5vqsp5pqqy702yw9j06g6rh9whz7kz3mxjcce0f3mc9uvtdwdkweyyxr5q9qyyssq3ankzemmgsjrxyyaqadgq8mtpequnh4m6l7fq3973204fullz058htveakpnnuwde7agdaakt7340fqcyy3awqf6nun8dnhy56uflkgq7unl42",
  "status": "failed"
}
```

Is the difference that some payments are pending and then fail while others never become pending?

vindard:
> the second row doesn't have a `createdAt` for some reason:
> 
> image

Is this the latest version of the code? There's now a nested `paymentDetails` object that holds all the properties that can change as a payment moves from `pending` to `settled`/`failed`.

The `createdAt` prop is currently in there but I can pull it up to the top level of the object and assign it a value in the case of the api not returning one (when payments `fail` without being `pending`)

src/app/wallets/update-pending-payments.ts
vindard: Sure that makes sense. We could optionally always retrieve this later on as well from lnd.

vindard: Done https://github.com/GaloyMoney/galoy/pull/665/commits/b79ea2f740c6169dde892407e1e079b32a0614b5

jireva (dismissed):

vindard:
> Is the difference that some payments are pending and then fail while others never become pending?

Yes exactly, `null` is when the payment immediately fails and we get back no information for those fields from `lnd` whereas we can sometimes have information from a `pending` state that eventually fails. Is this the best way to handle this, or would some other way of reflecting this be better?

vindard:
> the second row doesn't have a `createdAt` for some reason:
> 
> image

Fixed here: https://github.com/GaloyMoney/galoy/pull/665/commits/4e0a6550c56c0cae4cf90739980e1fc59ed728cd

src/app/wallets/update-pending-payments.ts
```diff
@@ -59,21 +61,45 @@ const updatePendingPayment = async ({
...
   }
-  const { status, roundedUpFee } = lnPaymentLookup
+
+  lnPaymentLookup.status = lnPaymentLookupFromLightning.status
+  if (lnPaymentLookupFromLightning.status !== PaymentStatus.Failed) {
+    lnPaymentLookup.paymentDetails = lnPaymentLookupFromLightning.paymentDetails
+  }
+
+  const { status, paymentDetails } = lnPaymentLookup
+  const roundedUpFee = paymentDetails?.roundedUpFee || toSats(0)
```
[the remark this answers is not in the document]
vindard: It's because I changed the interface to accommodate for status `failed` payments that don't have this property in the object returned

src/graphql/error-map.ts
vindard: We need to put something for the typescript switch, but it currently never gets to the user because we no longer return that error.

src/app/wallets/update-pending-payments.ts
```diff
@@ -59,21 +61,45 @@ const updatePendingPayment = async ({
...
+  const lnPaymentLookupPromise = LnPaymentsRepository().findByPaymentHash(paymentHash)
+
+  const lightningLogger = logger.child({
+    topic: "payment",
+    protocol: "lightning",
+    transactionType: "payment",
+    onUs: false,
+  })
+
+  const [lnPaymentLookupFromLightning, lnPaymentLookup] = await Promise.all([
```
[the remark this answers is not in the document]
vindard: Done https://github.com/GaloyMoney/galoy/pull/665/commits/afcc66aae77cebcca72cb0ca0660b01100ebf1f9

nicolasburtey (dismissed):

jireva:
> Yes exactly, `null` is when the payment immediately fails and we get back no information for those fields from `lnd` whereas we can sometimes have information from a `pending` state that eventually fails. Is this the best way to handle this, or would some other way of reflecting this be better?
> 
> The idea behind the `paymentDetails` object as well is to have values that can change in there, vs. other top-level properties which stay fixed during the lifetime of the payment

Yes that makes sense, then it's definitely good to move createdAt to the top level.  I'm not sure what circumstances would lead to `lnd` failing without returning any information but it might be useful to also know as much information as possible about these intended payments (e.g. the intended destination).

Would it be possible to simply add a field like `failedImmediately` to `paymentDetails`?

vindard:
> I'm not sure what circumstances would lead to `lnd` failing without returning any information

The hint is in [this type](https://github.com/alexbosworth/lightning/blob/master/lnd_methods/offchain/get_payment.d.ts#L11) on the library we use. The library actually returns no information whenever it fails whether it was immediate or after a pending state (see [`failed`](https://github.com/alexbosworth/lightning/blob/master/lnd_methods/offchain/get_payment.d.ts#L12) property). What I'm doing is persisting the transition state (from [`pending`](https://github.com/alexbosworth/lightning/blob/master/lnd_methods/offchain/get_payment.d.ts#L108) property) internally if we see a pending state first that eventually switches to failed.

For additional context, if a payment succeeds either eventually or immediately, we always overwrite with the data we get back from the [`payment`](https://github.com/alexbosworth/lightning/blob/master/lnd_methods/offchain/get_payment.d.ts#L28) property.

>  but it might be useful to also know as much information as possible about these intended payments (e.g. the intended destination)

This is all already encoded in the top-level `paymentRequest` property.

> Would it be possible to simply add a field like failedImmediately to paymentDetails?

Based on what you're saying here, is it that you'd like us to decode whatever we can get from the `paymentRequest` and add it to `paymentDetails` and then also add something like a `failedImmediately` property to signal that immediate fail instead of simply leaving `paymentDetails` as `null`?

src/services/mongoose/ln-payments.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/665/commits/bd9233cf74c4e0b00758585e4e40a118c1b61cc8

jireva:
> > I'm not sure what circumstances would lead to `lnd` failing without returning any information
> 
> The hint is in [this type](https://github.com/alexbosworth/lightning/blob/master/lnd_methods/offchain/get_payment.d.ts#L11) on the library we use. The library actually returns no information whenever it fails whether it was immediate or after a pending state (see [`failed`](https://github.com/alexbosworth/lightning/blob/master/lnd_methods/offchain/get_payment.d.ts#L12) property). What I'm doing is persisting the transition state (from [`pending`](https://github.com/alexbosworth/lightning/blob/master/lnd_methods/offchain/get_payment.d.ts#L108) property) internally if we see a pending state first that eventually switches to failed.
> 
> For additional context, if a payment succeeds either eventually or immediately, we always overwrite with the data we get back from the [`payment`](https://github.com/alexbosworth/lightning/blob/master/lnd_methods/offchain/get_payment.d.ts#L28) property.
> 
> > but it might be useful to also know as much information as possible about these intended payments (e.g. the intended destination)
> 
> This is all already encoded in the top-level `paymentRequest` property.
> 
> > Would it be possible to simply add a field like failedImmediately to paymentDetails?
> 
> Based on what you're saying here, is it that you'd like us to decode whatever we can get from the `paymentRequest` and add it to `paymentDetails` and then also add something like a `failedImmediately` property to signal that immediate fail instead of simply leaving `paymentDetails` as `null`?

Thank you for the link to the library, I think I understand the issue much better now and I think it actually makes more sense to keep `paymentDetails` null for failed payments (it should also be set to null for payments that were pending but eventually failed too!).  The reason for this is that `paymentDetails` only describes a single payment attempt which is canonical for successful payments but not for failed payments.

With regards to decoding the `paymentRequest`, I don't think it's necessary to do that here.  It could be decoded as part of whatever downstream data analysis we want to do.

I think we should also persist data on failed payments (and failed attempts of successful payments) but, since it's of quite a different structure, maybe it needs to be a separate collection and PR.

vindard:
## Update

Based on a call we had a little while back, the open things left to do on this are:
- move the update process to cronjob and use `getPayments` method instead for full update
- edit the update we do in `updatePayments` to be as light as possible (can even remove if we think that's appropriate, or add an `isPersisted` flag to indicate that cronjob hasn't persisted as yet)

**In a follow-up PR**
- add a delete-from-lnd step in the cronjob for payments already persisted

vindard:
Closing in favour of PR #848

PR #678, opened 2021-10-29
Limit instrumentation + make GQL errors searchable

bodymindarts:
This PR removes some of the auto-instrumentation libraries to try and reduce the number of events we are generating.

It also adds information about GQL errors as span attributes via a hook to make them more searchable.


src/services/tracing.ts
```diff
@@ -2,20 +2,96 @@ import {
...
+      }
+    })
+  }
+}
+
 registerInstrumentations({
-  instrumentations: getNodeAutoInstrumentations(),
+  instrumentations: [
+    new HttpInstrumentation({
+      ignoreIncomingPaths: ["/healthz"],
```
vindard, line 85: `"/healthz"` is for some sort of testing I assume?

vindard (approved):

src/services/tracing.ts
bodymindarts: its a health check that the k8s runtime executes at regular intervals. In terms of tracing it creates a bunch of events/minute that don't add a lot of value.

I'm looking into filtering the events at another component level soon - so we can remove this - this is just for now.

dolcalmi (approved):

src/services/tracing.ts
nicolasburtey: @vindard it's a k8s convention: https://kubernetes.io/docs/reference/using-api/health-checks/

src/services/tracing.ts
```diff
@@ -2,20 +2,96 @@ import {
...
+  Span,
+  SpanAttributes,
+  SpanStatusCode,
+} from "@opentelemetry/api"
 import { tracingConfig } from "@config/app"
 
 propagation.setGlobalPropagator(new W3CTraceContextPropagator())
 
+const gqlResponseHook = (span: Span, data: graphqlTypes.ExecutionResult) => {
+  if (data.errors && data.errors.length > 0) {
```
nicolasburtey, line 30: this could be squeezed to `data?.errors.length > 0` I think.

nicolasburtey: `errors` is always an array?

src/services/tracing.ts
```diff
@@ -2,20 +2,96 @@ import {
...
+    span.recordException({
+      name: "graphql.execution.error",
+      message: JSON.stringify(data.errors),
+    })
+    span.setStatus({
+      code: SpanStatusCode.ERROR,
+    })
+    const firstErr = data.errors[0]
```
nicolasburtey, line 38: don't quite get why there is a separation on how the first error is being handle and the rest of the array with the loop line 58

nicolasburtey (commented):
some question/comments but LGTM otherwise

src/services/tracing.ts
bodymindarts: According to the type definition yes... there will be a follow up PR shortly. Will fix it.

bodymindarts: shouldn't it be `data.errors?.length`?

src/services/tracing.ts
bodymindarts: Its adding the first error twice - once as `graphql.error.type` and once as `graphql.error.0.type`. This is to make the most common case (only 1 error) nicer to search for - but also allows representing the whole array of errors incase there are more.

PR #682, opened 2021-10-30
Add graphql values to instrumentation

bodymindarts:
This PR adds all input values to the graphql instrumentation by setting:
```
allowValues: true
```
in the constructor.

Unfortunately while testing this I encountered a bug in the instrumentation library.

I have fixed the issue on a fork and manually published the fix to npm.
You can see the details of the fix in the upstream PR https://github.com/open-telemetry/opentelemetry-js-contrib/pull/720.

As this change exposes all input - including login credentials - I've added a processor that redacts the critical values to our infrastructure before it gets shipped to honeycomb https://github.com/GaloyMoney/galoy-infra/commit/bb2367ddfabc766efcb6fd212fefb0ecc36815bc.


dev/otel-agent-config.yaml
```diff
@@ -1,10 +1,16 @@
 receivers:
   jaeger:
     protocols:
+      thrift_compact: # on port 6831
```
nicolasburtey, line 4: what is the purpose of exposing an empty object?

nicolasburtey (approved):
LGTM. we should probably open a ticket to reminder ourselves to get back to the upstream once your fix has been merged

dev/otel-agent-config.yaml
bodymindarts: Empty adds the default listener.
Its equivalent to the following structure:
```
thrift_compact:
  endpoint: 0.0.0.0:6831
```

vindard (approved):

PR #688, opened 2021-11-01
Add address limit

vindard:


dolcalmi (dismissed):

nicolasburtey (dismissed):

src/app/wallets/create-on-chain-address.ts
```diff
@@ -30,3 +39,17 @@ export const createOnChainAddressByWalletPublicId = async (
   if (wallet instanceof Error) return wallet
   return createOnChainAddress(wallet.id)
 }
+
+const checkOnChainAddressWalletIdLimits = async (
```
bodymindarts, line 43: Is this part of our app or part of graphql adapter layer?

This PR is inconsistent - OTOH this function is defined under `app/wallets` indicating its part of our app - but in that case the GQL layer shouldn't call it directly - it should be called from within the `createOnChainAddress` use case.

If the rate limiting is not a part of our application in general but just part of the graphql layer then it should not be defined under `app/` but under `graphql/`.


IMO its part of the app so the former approach should be used - the use-case functions must check the rate limit NOT the GQL resolvers.

bodymindarts (changes requested):

bodymindarts (approved):

dolcalmi (approved):

PR #690, opened 2021-11-01
App tracing

bodymindarts:
This PR begins to add some app-specific tracing.
In particular instrumentation is added to the pay-via-lightning use cases using helpers exposed in the `services/tracing.ts` module.

The diff is much smaller than it seems. Mostly code has just been indented und runs under a callback (similar like in the redis lock pattern we already have).

EDIT - a bit more context:
there are 2 _different_ helpers that 'wrap' functions - one sets attributes that should be added to _all_ child spans like:
```
 addAttributesToCurrentSpanAndPropagate(
      {
        [SemanticAttributes.ENDUSER_ID]: domainUser?.id,
        [ENDUSER_ALIAS]: domainUser?.username,
        [SemanticAttributes.HTTP_CLIENT_IP]: ip,
      },
```

This doesn't create its own span - it just makes sure it populates some context for all children to pick up. It should be called in _all_ GQL resolvers.

and then there is:
```
  asyncRunInSpan(
    "app.lnInvoicePaymentSendWithTwoFA",
    {
      [SemanticAttributes.CODE_FUNCTION]: "lnInvoicePaymentSendWithTwoFA",
    }
```
this is called from _within_ the app use case functions - and probably also from within the service functions when we get to that layer.

### DIFF
--- a/src/app/wallets/ln-send-payment.ts
+++ b/src/app/wallets/ln-send-payment.ts
@@ -1,5 +1,11 @@
 import { getBalanceForWallet } from "@app/wallets"
+import { PaymentInitiationMethod, SettlementMethod } from "@domain/wallets"
 import { toMilliSatsFromNumber, toSats } from "@domain/bitcoin"
+import {
+  asyncRunInSpan,
+  SemanticAttributes,
+  addAttributesToCurrentSpan,
+} from "@services/tracing"
 import {
   decodeInvoice,
   LnPaymentPendingError,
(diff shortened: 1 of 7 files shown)


src/graphql/root/mutation/ln-invoice-payment-send.ts
```diff
@@ -18,31 +23,39 @@ const LnInvoicePaymentSendMutation = GT.Field({
...
   },
-  resolve: async (_, args, { user, wallet, logger }) => {
-    const { paymentRequest, memo } = args.input
+  resolve: async (_, args, { ip, domainUser, wallet, logger }) =>
+    addAttributesToCurrentSpanAndPropagate(
+      {
+        [SemanticAttributes.ENDUSER_ID]: domainUser?.id,
+        [ENDUSER_ALIAS]: domainUser?.username,
+        [SemanticAttributes.HTTP_CLIENT_IP]: ip,
+      },
```
dolcalmi, line 32: can't be added in context?

src/app/wallets/ln-send-payment.ts
```diff
@@ -45,66 +51,80 @@ export const lnInvoicePaymentSendWithTwoFA = async ({
...
-    ? await checkAndVerifyTwoFA({
-        amount: lnInvoiceAmount,
-        twoFAToken: twoFAToken ? (twoFAToken as TwoFAToken) : null,
-        twoFASecret: twoFA.secret,
+}: LnInvoicePaymentSendWithTwoFAArgs): Promise<PaymentSendStatus | ApplicationError> =>
+  asyncRunInSpan(
+    "app.lnInvoicePaymentSendWithTwoFA",
+    {
+      [SemanticAttributes.CODE_FUNCTION]: "lnInvoicePaymentSendWithTwoFA",
+    },
```
dolcalmi, line 59: I think is not good to add this in app layer, cannot be done in graphql layer?

dolcalmi (changes requested):

src/app/wallets/ln-send-payment.ts
bodymindarts: If this is added by the parent then it has to be put everywhere this function is called. Imo the spans that are for the app layer need to be put inside the function call.
But we can discuss this in code review.

src/graphql/root/mutation/ln-invoice-payment-send.ts
bodymindarts: This is not a in the call stack of context - so no it cannot.

You can only influence things that are in the direct call graph.

It would only work if :
```
context() {
  resolver()
}
```

But actually its:
```
someOuterFunction() {
  context() // cannot influence attributes set in resolver()
  resolver()
}
```

nicolasburtey:
are we using the payment under `app/` yet? or only `core/`? my assumption is that we're doing the later until we switch the mobile app to v2. this would means we would have no tracing yet?

src/servers/graphql-old-server.ts
```diff
@@ -297,33 +310,41 @@ const resolvers = {
...
+                paymentRequest: invoice,
+                memo,
+                walletId: wallet.user.id as WalletId,
+                userId: wallet.user.id as UserId,
+                logger,
+              })
+              if (status instanceof Error) throw mapError(status)
+              return status.value
+            }
+            const status = await lnNoAmountInvoicePaymentSend({
```
bodymindarts, line 336: Its being called here @nicolasburtey - that was the giant PR vindard was working on for a month. It was merged last week.

nicolasburtey: thanks. I thought it would have been used only with api v2. 

we need to get rid of `core/` after v2 is been merged.

src/graphql/root/query/me.ts
```diff
@@ -1,17 +1,23 @@
...
 const MeQuery = GT.Field({
   type: UserWithAccounts,
-  resolve: async (_, __, { uid }) => {
-    const user = await Users.getUser(uid)
-    if (user instanceof Error) {
-      throw user
-    }
-    return user
-  },
+  resolve: async (_, __, { ip, domainUser }) =>
```
nicolasburtey, line 12: domainuser is the user from `await Users.getUser(uid)`? 

this one is actually an optimization where prior to this we were calling `getUser` twice?

bodymindarts:
It’s the response from getUserForLogin in the context - used for v2.

bodymindarts:
Yes it’s an optimization

nicolasburtey (dismissed):
It's unclear to me what those traces add versus what we have by default but this can be done at code review.

this is the type of PR that would benefits from manual testing on staing before rolling on to bbw

vindard (commented):
Just one question about functionality & naming.

For these two methods, is it that?:
- `addAttributesToCurrentSpanAndPropagate` adds attributes to all child spans
- `addAttributesToCurrentSpan` adds attributes to the "parent" span it sits within

If this is the case, I'm wondering if we can reflect this distinction in the naming of those methods as well somehow.

bodymindarts:
> It's unclear to me what those traces add versus what we have by default but this can be done at code review.
> 
> this is the type of PR that would benefits from manual testing on staing before rolling on to bbw

You are right - I forgot to add the use-case specific fields. Added here https://github.com/GaloyMoney/galoy/pull/690/commits/990b26b63e12f392c53346d76b4382beb7c03774

bodymindarts:
> Just one question about functionality & naming.
> 
> For these two methods, is it that?:
> 
> * `addAttributesToCurrentSpanAndPropagate` adds attributes to all child spans
> * `addAttributesToCurrentSpan` adds attributes to the "parent" span it sits within
> 
> If this is the case, I'm wondering if we can reflect this distinction in the naming of those methods as well somehow.

Both add the attributes to the current span they sit within - the `AndPropagate` one also adds it to all child spans - should be used for fields that describe the high-level context for easier searching.

vindard (approved):

src/graphql/root/query/me.ts
bodymindarts: It’s the response from getUserForLogin in the context - used for v2. - Yes it’s an optimization

nicolasburtey (dismissed):

nicolasburtey (approved):

vindard (approved):

PR #692, opened 2021-11-02
Implement the updateUsername mutation

samerbuna:
Includes switching the updateLanguageMutation to use the new app use-case

### DIFF
--- a/src/graphql/main/schema.graphql
+++ b/src/graphql/main/schema.graphql
@@ -459,6 +459,7 @@ type Mutation {
   accountApiKeyCreate(input: AccountApiKeyCreateInput!): AccountApiKeyPayload!
   accountApiKeyDisable(input: AccountApiKeyDisableInput!): SuccessPayload!
   userUpdateLanguage(input: UserUpdateLanguageInput!): UserUpdateLanguagePayload!
+  userUpdateUsername(input: UserUpdateUsernameInput!): UserUpdateUsernamePayload!
   userContactUpdateAlias(
     input: UserContactUpdateAliasInput!
   ): UserContactUpdateAliasPayload!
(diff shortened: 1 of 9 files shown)


src/graphql/root/mutation/user-update-language.ts
```diff
@@ -15,24 +15,22 @@ const UserUpdateLanguageMutation = GT.Field({
...
 
-    // TODO: redo this with app use-case
-    try {
-      user.language = language
-      await user.save()
-    } catch (err) {
-      return { errors: [{ message: err.message }] }
+    const result = wallet.updateLanguage({ language })
+
+    if (result instanceof Error) {
```
dolcalmi, line 27: if you dont call updateLanguage with await then result will be always a promise

src/graphql/root/mutation/user-update-username.ts
```diff
@@ -0,0 +1,38 @@
...
+  resolve: async (_, args, { wallet }) => {
+    const { username } = args.input
+
+    if (username instanceof Error) {
+      return { errors: [{ message: username.message }] }
+    }
+
+    const result = wallet.updateUsername({ username })
+
+    if (result instanceof Error) {
```
dolcalmi, line 27: same here, result is always a promise

src/graphql/main/schema.graphql
```diff
@@ -567,30 +568,49 @@ input AccountApiKeyDisableInput {
 
 type UserUpdateLanguagePayload {
   errors: [Error!]!
-  user: UserDetails
+  user: UserLanguageDetails
```
dolcalmi, line 571: not sure if user is the proper name for this attribute. UserLanguageDetails only has id and language

src/graphql/types/payload/user-update-username.ts
```diff
@@ -0,0 +1,26 @@
...
+  }),
+})
+
+const UserUpdateUsernamePayload = new GT.Object({
+  name: "UserUpdateUsernamePayload",
+  fields: () => ({
+    errors: {
+      type: GT.NonNullList(IError),
+    },
+    user: {
```
dolcalmi, line 20: same here, is this really user?

dolcalmi (changes requested):

src/graphql/root/mutation/user-update-language.ts
samerbuna: Fixed

src/graphql/root/mutation/user-update-username.ts
samerbuna: Fixed

src/graphql/main/schema.graphql
samerbuna: It's a partial record for a user, so I think that's ok

dolcalmi (approved):

bodymindarts (approved):

PR #693, opened 2021-11-02
README and misc update

nicolasburtey:


src/servers/graphql-server.ts
```diff
@@ -37,7 +37,7 @@ export const isAuthenticated = rule({ cache: "contextual" })((parent, args, ctx)
 
 export const isApiKeyAuthenticated = rule({ cache: "contextual" })(
   (_parent, _args, ctx) => {
-    return ctx.account !== null ? true : "NOT_AUTHENTICATED"
+    return ctx.account !== null ? true : "NOT_AUTHENTICATED_APIKEY"
```
krtk6160, line 40: What's the context around this change?

krtk6160 (commented):
One question otherwise LGTM

src/servers/graphql-server.ts
nicolasburtey: there are issues related to nginx in prod. and the string is `NOT_AUTHENTICATED`. so making this string unique with the previous condition will help to understand where the issue is

krtk6160 (changes requested):
Line 153 in `graphql-server.ts` needs to be changed also I think. The isShieldError function.

bodymindarts (dismissed):

dolcalmi (dismissed):

krtk6160 (approved):

bodymindarts (approved):

PR #695, opened 2021-11-02
Added documentation

daviroo:
Myself and @nicolasburtey added some docs for the graphql.  A long way to go but it's a start.  There are some questions we had while writing the docs which I have marked with `TODO`.

### DIFF
--- a/src/graphql/admin/schema.graphql
+++ b/src/graphql/admin/schema.graphql
@@ -132,12 +132,32 @@ type TransactionEdge {
   cursor: String!
 }
 
+"""
+Give details about an individual transaction.
+Galoy have a smart routing system which is automatically settling intraledger when both the payer and payee use the same wallet - therefore it's possible the transactions is being initiated onchain or with lightning but settled intraledger.
+"""
 interface Transaction {
   id: ID!
+
+  """
+  From which protocol the payment has been initiated.
+  """
   initiationVia: PaymentInitiationMethod!
+
+  """
+  To which protocol the payment has settled on.
+  """
   settlementVia: SettlementMethod!
+
+  """
+  Amount of sats paid by the user.
+  """
   settlementAmount: SatAmount!
   settlementFee: SatAmount!
+
+  """
+  Price in USDCENT/SATS at time of settlement.
+  """
   settlementPrice: Price!
   direction: TxDirection!
   status: TxStatus!
@@ -237,10 +257,26 @@ type BTCWallet implements Wallet {
 
 type IntraLedgerTransaction implements Transaction {
   id: ID!
+
+  """
+  From which protocol the payment has been initiated.
+  """
   initiationVia: PaymentInitiationMethod!
+
+  """
+  To which protocol the payment has settled on.
+  """
   settlementVia: SettlementMethod!
+
+  """
+  Amount of sats paid by the user.
+  """
   settlementAmount: SatAmount!
   settlementFee: SatAmount!
+
+  """
+  Price in USDCENT/SATS at time of settlement.
+  """
   settlementPrice: Price!
   direction: TxDirection!
   status: TxStatus!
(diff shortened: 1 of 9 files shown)


src/graphql/admin/schema.graphql
```diff
@@ -237,10 +257,26 @@ type BTCWallet implements Wallet {
 
 type IntraLedgerTransaction implements Transaction {
   id: ID!
+
+  """
+  From which protocol the payment has been initiated.
```
bodymindarts, line 262: is `protocol` really the right name?

src/graphql/types/object/wallet-contact.ts
```diff
@@ -9,11 +9,12 @@ import * as Accounts from "@app/accounts"
...
 import { checkedToUsername } from "@domain/users"
 
+//FIXME: Discrepancy with filename
 const UserContact = new GT.Object({
   name: "UserContact",
   fields: () => ({
-    username: { type: GT.NonNull(Username) },
-    alias: { type: ContactAlias },
+    username: { type: GT.NonNull(Username), description: "Actual identifier of the contact." },
+    alias: { type: ContactAlias, description: "Alias the user can set for this contact.\nOnly the user can see the alias attached to their contact." },
```
bodymindarts: This line is a bit long - shouldn't get by linting or?
Isn't there a better way to do multi-line string that respects our line width settings?

src/graphql/types/object/user.ts
```diff
@@ -15,15 +15,17 @@ import { UnknownClientError } from "@core/error"
...
 const mainUserFields = () => ({
   id: { type: GT.NonNullID },
-  phone: { type: GT.NonNull(Phone) },
-  username: { type: Username },
+  phone: { type: GT.NonNull(Phone), description: "Phone number with international calling code." },
+  username: { type: Username, description: "Optional immutable user friendly identifier." },
   language: {
     type: GT.NonNull(Language),
     resolve: (source) => source.language,
+    description: "Preferred language for user.\nWhen value is 'default' the intent is to use preferred language from OS settings."
```
bodymindarts: Can we use multiline string instead of `\n`?

src/graphql/admin/schema.graphql
nicolasburtey: do you have a better name for describing onchain/lightning/intraledger?

src/graphql/types/abstract/transaction.ts
```diff
@@ -34,10 +39,14 @@ export const transactionInterfaceFields = {
       return {
         formattedAmount: settlementUsdPerSatInCents.toString(),
         base: Math.round(settlementUsdPerSatInCents * 10 ** 4),
+        //TODO: What is offset?  What is difference between offset + sats/usd?
         offset: 4,
+        //TODO: Is current offset 0.000001 to the dollar which would be 4 for offset and 2 for usd/cent?
+        // Why not use offset 6 and USD?
```
nicolasburtey, line 45: the reason we are using cents and not dollar is to work around the limitation of graphql int. ie: to not overflow.

src/graphql/types/object/wallet-contact.ts
nicolasburtey: good idea. use: ``

src/graphql/admin/schema.graphql
krtk6160: `bitcoin layer` perhaps?

bodymindarts: Hmm its probably fine - can't think of something better.

src/graphql/types/object/wallet-contact.ts
daviroo: 👍

src/graphql/types/object/user.ts
daviroo: I have switched out all strings which use new lines to use multi line strings.

samerbuna (approved):

src/graphql/types/abstract/transaction.ts
```diff
@@ -25,11 +25,9 @@ export const transactionInterfaceFields = {
   },
   settlementAmount: {
     type: GT.NonNull(SatAmount),
-    //TODO: Does the amount include fees.
```
nicolasburtey: this would still be relevant information to give to the user as part of the documentation

nicolasburtey (approved):

PR #697, opened 2021-11-02
Add honeycomb config to local dev environment

bodymindarts:
By setting the following keys:

```
export HONEYCOMB_API_KEY=<api-key>
export HONEYCOMB_DATASET=<your-dataset>
```

in the terminal you run `make reset-deps` in you will get local traces in honeycomb.io


vindard (approved):
Was able to see some of my local data in a new dataset on honeycomb

dolcalmi (approved):

PR #699, opened 2021-11-02
Validate block subscription height

dolcalmi:
avoid lock in wallet id when same onchain block event is triggered multiple times

### DIFF
--- a/test/integration/servers/trigger.spec.ts
+++ b/test/integration/servers/trigger.spec.ts
@@ -85,8 +85,10 @@ describe("onchainBlockEventhandler", () => {
     const initialBlock = await bitcoindClient.getBlockCount()
     const subBlocks = subscribeToBlocks({ lnd: lnd1 })
     subBlocks.on("block", async ({ height }) => {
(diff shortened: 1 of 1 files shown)


vindard (approved):

samerbuna (approved):

test/integration/servers/trigger.spec.ts
```diff
@@ -85,8 +85,10 @@ describe("onchainBlockEventhandler", () => {
     const initialBlock = await bitcoindClient.getBlockCount()
     const subBlocks = subscribeToBlocks({ lnd: lnd1 })
     subBlocks.on("block", async ({ height }) => {
-      await onchainBlockEventhandler({ height })
-      lastHeight = height > lastHeight ? height : lastHeight
+      if (height > lastHeight) {
```
nicolasburtey, line 88: I would add a TODO to check how this handle a 1 or 2 block reorg (don't think we have to spend time on this right now because this is a low likely hood so we can dive into the potential reorg issue later)

PR #700, opened 2021-11-03
Remove getCurrentPrice from transactionNotification

dolcalmi:
Reviewing https://github.com/GaloyMoney/galoy/pull/662 and pairing with Arvin, we decided to split it in different PRs

- The main purpose of this PR is remove getCurrentPrice from transactionNotification
- Remove direct calls to transactionNotification
- Although we have some tests for notifications, we need to do final tests in staging

### DIFF
--- a/src/services/notifications/index.ts
+++ b/src/services/notifications/index.ts
@@ -1,25 +1,33 @@
-import { LedgerTransactionType } from "@domain/ledger"
-import { NotificationsServiceError } from "@domain/notifications"
+import { NotificationsServiceError, NotificationType } from "@domain/notifications"
 import { User } from "@services/mongoose/schema"
 import { transactionNotification } from "./payment"
 
 export const NotificationsService = (logger: Logger): INotificationsService => {
-  const onChainTransactionReceived = async ({
+  const sendOnChainNotification = async ({
+    type,
     amount,
     walletId,
     txId,
-  }: OnChainTxReceivedArgs) => {
+    usdPerSat,
+  }: {
+    type: NotificationType
+    walletId: WalletId
+    amount: Satoshis
+    txId: TxId
+    usdPerSat?: UsdPerSat
+  }): Promise<void | NotificationsServiceError> => {
     try {
       // work around to move forward before re-wrighting the whole notifications module
       const user = await User.findOne({ _id: walletId })
 
       // Do not await this call for quicker processing
       transactionNotification({
-        type: LedgerTransactionType.OnchainReceipt,
+        type,
         user,
-        logger: logger,
+        logger,
         amount,
         txid: txId,
+        usdPerSat,
       })
       return
     } catch (err) {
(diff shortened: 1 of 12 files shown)


src/services/notifications/payment.ts
```diff
@@ -23,13 +21,12 @@ export const transactionNotification = async ({
...
+  usdPerSat,
 }: IPaymentNotification) => {
-  const satsPrice = await getCurrentPrice()
-
   let title
 
-  if (satsPrice) {
-    const usd = (amount * satsPrice).toFixed(2)
+  if (usdPerSat) {
+    const usd = (amount * usdPerSat).toFixed(2)
```
bodymindarts, line 29: toFixed is display logic - should probably be implemented client side or wrapped via a 'CurrencyFormatter' object - not critical as I can see in the diff this wasn't changed. Just an observation

src/services/notifications/payment.ts
```diff
@@ -1,19 +1,17 @@
-import { getCurrentPrice } from "@services/realtime-price"
-
 import { sendNotification } from "./notification"
 
 export const getTitle = {
-  "paid-invoice": ({ usd, amount }) => `+$${usd} | ${amount} sats`,
-  "onchain_receipt": ({ usd, amount }) => `+$${usd} | ${amount} sats`,
-  "onchain_receipt_pending": ({ usd, amount }) => `pending +$${usd} | ${amount} sats`,
-  "onchain_payment": ({ amount }) => `Sent onchain payment of ${amount} sats confirmed`,
+  paid_invoice: ({ usd, amount }) => `+$${usd} | ${amount} sats`,
```
bodymindarts, line 4: Why are these snake case and not idiomatic `camelCase`?

dolcalmi: thanks for pointing this out, these keys are used in mobile app. I had to rollback the paid-invoice change.

src/services/notifications/payment.ts
dolcalmi: this value is shown in the notification, the mobile app cant update it

src/services/notifications/index.ts
```diff
@@ -1,54 +1,111 @@
...
     walletId,
     paymentHash,
+    usdPerSat,
   }: LnPaymentReceivedArgs) => {
     try {
       // work around to move forward before re-wrighting the whole notifications module
       const user = await User.findOne({ _id: walletId })
 
       // Do not await this call for quicker processing
       transactionNotification({
```
vindard, line 91: How is the `sendOnChainNotification` helper different from what's here (can we include the abstraction here as well for consistency)?

vindard (changes requested):
Looks good, just one suggested change

src/services/notifications/index.ts
dolcalmi: `hash: paymentHash,`

vindard (approved):

dolcalmi:
> Is there tracing we can add to make the testing on staging better?

the request is done through `firebase-admin` package so not sure if we can intercept that requests like with grpc to price server.

this is the code https://github.com/GaloyMoney/galoy/blob/main/src/services/notifications/notification.ts#L60

bodymindarts (approved):

bodymindarts:
Here we can identify calls to firebase https://ui.honeycomb.io/galoy/datasets/galoy-bbw?query=%7B%22time_range%22%3A86400%2C%22granularity%22%3A0%2C%22breakdowns%22%3A%5B%22name%22%5D%2C%22calculations%22%3A%5B%7B%22op%22%3A%22COUNT%22%7D%5D%2C%22filters%22%3A%5B%7B%22column%22%3A%22http.host%22%2C%22op%22%3A%22contains%22%2C%22value%22%3A%22googleapis%22%7D%5D%2C%22filter_combination%22%3A%22AND%22%2C%22orders%22%3A%5B%7B%22op%22%3A%22COUNT%22%2C%22order%22%3A%22descending%22%7D%5D%2C%22havings%22%3A%5B%5D%2C%22limit%22%3A1000%7D

PR #701, opened 2021-11-03
Add CONTRIBUTING.md

vindard:
### Description

Adding documentation to help explain some of the finer points of how we add code in this repo to help potential contributors.

### Additional points to add here in follow-on PRs:
- our approach to de-duping and use-cases (3 instance where it's ok to de-dupe, and how we go about de-duping. Reasoning that it's to avoid branching/re-merging of logic and entrypoint ambiguity)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,226 @@
+# Contributing
+
+This repo is built and maintained by the Galoy team. We welcome and appreciate new contributions and encourage you to check out the repo and [join our community Slack](https://join.slack.com/t/galoymoney-workspace/shared_invite/zt-rvnhsdb5-72AZCD_jzw6_Q05aCs0SgA) to get started.
+
+To help get you started, we will explain a bit about how we've laid out the code here and some of the practices we use. Our layout details fall mostly into two categories:
+- [Working with Types](#working-with-types)
+- [Understanding our Architecture](#architecture)
+
+---
+
+## Working with Types
+This codebase is implemented using a well-typed approach where we rely on Typescript for catching & enforcing certain kinds of checks we would like to perform.
+
+This approach helps to strenghthen guarantees that bugs/issues don't get to runtime unhandled if our types are implemented properly.
+
+### Our Approach
+We define all our types in declaration files (`*.d.ts`) where typescript automatically adds any types to the context without needing to explicitly import. Declaration files have limitations on what can be imported & instantiated within them, so in the few cases where types derive from instances of things we would usually have to use special tricks to derive them.
+
+The majority of our types are defined in the `domain` layer (see Architecture section below). We generally create types for the following scenarios:
+
+#### Symbol types
+These are unique types that are alternatives to generically typing things as Typescript's primitive types (e.g. `string`, `number`, `boolean`). Doing this helps us to add context to primitive types that we pass around and it allows the type checked to distinguish between different kinds of primitives throughout the code.
+
+For example, an onchain address and a lightning network payment request are both strings, but they aren't interchangeable as a data type. Instead of using `string` type for these types we would define as follows using a "unique symbol":
+
+```
+declare const encodedPaymentRequestSymbol: unique symbol
+type EncodedPaymentRequest = string & { [encodedPaymentRequestSymbol]: never }
+
+declare const onChainAddressSymbol: unique symbol
+type OnChainAddress = string & { [onChainAddressSymbol]: never }
+
+```
+
+#### Error types
+
+These are types mostly used in function signature type definitions. They derive from implemented error classes and need to be imported in a special way to their type declaration file.
+
+For example, an error may be defined in a module's `error.ts` file like this:
+```
+export class LightningError extends Error {
+  name = this.constructor.name
+}
+```
+
+and then it is imported to its `index.types.d.ts` declaration file and made into a type like this:
+```
+type LightningError = import("./errors").LightningError
+```
+
+
+#### Imported Library types
+
+In the places where we would like to work with types defined in imported libraries, we have two options. When we would like to use them directly, we can simply import them. If we would like to re-use a type to create our own types, we would re-import to get those types into our declaration files.
+
+For example, for a result type from the `lightning` library, we would import and re-use it like:
+```
+type GetPaymentResult = import("lightning").GetPaymentResult
+type RawPaths = NonNullable<GetPaymentResult["payment"]>["paths"]
+```
+
+In this example, we need the type definition for the `payment.paths` property of the `GetPaymentResult` type from the library that we otherwise would not have direct access to.
+
+
+#### Function signatures and argument types
+Whenever we implement a new function, the argument and return types are assigned at the point of the function's implementation.
+
+For example:
+```ts
+const myFunction = async (myArg: MyArg): Promise<ThingResult> => {
+    return doThing(myArg)
+}
+
+const myOtherFunction = async ({
+  myFirstArg,
+  mySecondArg,
+}: {
+  myFirstArg: MyFirstArg
+  mySecondArg: MySecondArg
+}): Promise<ThingResult> => {
+  return doThing(myFirstArg, mySecondArg)
+}
+
+```
+
+In cases where the function has a complex set of arguments, the argument type can be defined in a declaration file and then assigned at the point of function implementation.
+
+```ts
+// in '.d.ts' file
+type MyFuncArgs = {
+    myFirstArg: MyFirstArg
+    mySecondArg: MySecondArg
+}
+
+// in other file
+const myFunction = async ({
+  myFirstArg,
+  mySecondArg,
+}: MyFuncArgs): Promise<ThingResult> => {
+  return doThing(myFirstArg, mySecondArg)
+}
+```
+
+#### Defining objects with methods
+We use functional constructors to define certain types of objects that we can call method on. The intention here is to instantiate the object first and then call methods on that object with method-specific args to execute some functionality.
+
+For objects like these, the interface for the object is defined using a `type` declaration and methods are typed at this point.
+
+For example, our fee calculator is typed as follows:
+```ts
+// In '.d.ts' file
+type DepositFeeCalculator = {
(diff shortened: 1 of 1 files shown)


vindard:
## Notes of things to add here

- `exports` at top of file ([here](https://github.com/GaloyMoney/galoy/pull/730/files#r749014653))

CONTRIBUTING.md
```diff
@@ -0,0 +1,223 @@
...
+type EncodedPaymentRequest = string & { [encodedPaymentRequestSymbol]: never }
+
+declare const onChainAddressSymbol: unique symbol
+type OnChainAddress = string & { [onChainAddressSymbol]: never }
+
+```
+
+#### Error types
+
+These are types mostly used in function signature type definitions. They derive from implemented error classes and need to be imported in a special way to thet type declaration file.
```
krtk6160, line 37: *their

nicolasburtey:
```suggestion
These are types mostly used in function signature type definitions. They derive from implemented error classes and need to be imported in a special way to their type declaration file.
```

nicolasburtey: @krtk6160 : https://haacked.com/archive/2019/06/03/suggested-changes/

nicolasburtey: learning about it. @vindard have been using it for quite some time. it's really nice.

krtk6160: @nicolasburtey Thanks! Had no idea this existed

CONTRIBUTING.md
```diff
@@ -0,0 +1,223 @@
...
+This repo is built and maintained by the Galoy team. We welcome and appreciate new contributions and encourage you to check out the repo and [join our community Slack](https://join.slack.com/t/galoymoney-workspace/shared_invite/zt-rvnhsdb5-72AZCD_jzw6_Q05aCs0SgA) to get started.
+
+To help get you started, we will explain a bit about how we've laid out the code here and some of the practices we use. Our layout details fall mostly into two categories:
+- [Working with Types](#working-with-types)
+- [Understanding our Architecture](#architecture)
+
+---
+
+## Working with Types
+This codebase is implemented using a well-typed approach where we rely heavily on the Typescript compiler for catching & enforcing certain kinds of checks we would like to perform.
```
nicolasburtey, line 12:
```suggestion
This codebase is implemented using a well-typed approach where we rely on Typescript for catching & enforcing certain kinds of checks we would like to perform.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,223 @@
...
+
+This approach helps to strenghthen guarantees that bugs/issues don't get to runtime unhandled if our types are implemented properly.
+
+### Our Approach
+We define all our types in declaration files (`*.d.ts`) where typescript automatically adds any types to the context without needing to explicitly import. Declaration files have limitations on what can be imported & instantiated within them, so in the few cases where types derive from instances of things we would usually have to use special tricks to derive them.
```
nicolasburtey, line 17: why are we using this approach?

reading on https://stackoverflow.com/questions/21247278/about-d-ts-in-typescript, it seems the approach of using `.d.ts` is for library that are not typescript native

CONTRIBUTING.md
```diff
@@ -0,0 +1,223 @@
...
+```
+declare const encodedPaymentRequestSymbol: unique symbol
+type EncodedPaymentRequest = string & { [encodedPaymentRequestSymbol]: never }
+
+declare const onChainAddressSymbol: unique symbol
+type OnChainAddress = string & { [onChainAddressSymbol]: never }
+
+```
+
+#### Error types
```
nicolasburtey, line 35: probably make sense to document the different error type we have in the codebase

CONTRIBUTING.md
```diff
@@ -0,0 +1,223 @@
...
+
+and then it is imported to its `index.types.d.ts` declaration file and made into a type like this:
+```
+type LightningError = import("./errors").LightningError
+```
+
+
+#### Imported Library types
+
+In the places where we would like to work with types defined in imported libraries, we have two options. When we would like to use them directly, we can simply import them. If we would like to re-use a type to create our own types, we would have to use our import trick to get those types into our declaration files.
```
nicolasburtey, line 54:
```suggestion
In the places where we would like to work with types defined in imported libraries, we have two options. When we would like to use them directly, we can simply import them. If we would like to re-use a type to create our own types, we would re-import to get those types into our declaration files.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,223 @@
...
+
+In this example, we need the type definition for the `payment.paths` property of the `GetPaymentResult` type from the library that we otherwise would not have direct access to.
+
+
+#### Function signatures and argument types
+Whenever we implement a new function, the argument and return types are assigned at the point of the function's implementation.
+
+For example:
+```ts
+const myFunction = async (myArg: MyArg): ThingResult => {
```
nicolasburtey, line 70: I think of tend to disagree on adding `: ThingResult` is typescript can automatically infer the type itself. 

otherwise the function can have an implementation issue and effectively return a different type, but forcing the return type is the equivalent of a cast IIUC. 

don't know if you have any input on this: @samerbuna @dolcalmi @bodymindarts ?

CONTRIBUTING.md
```diff
@@ -0,0 +1,223 @@
...
+export const AccountStatus = {
+  Locked: "locked",
+  Active: "active",
+} as const
+
+// In declaration file
+type AccountStatus =
+  typeof import("./index").AccountStatus[keyof typeof import("./index").AccountStatus]
+```
+## Architecture
```
nicolasburtey, line 149: we could use # for this title, and get one level above for all the below title (## instead of ### and so on)

CONTRIBUTING.md
vindard: @samerbuna advocated for this. The reason was so that we don't have to explicitly import types everywhere I believe. It works well from a dev flow approach so far in my experience.

CONTRIBUTING.md
```diff
@@ -0,0 +1,223 @@
...
+
+### Our Approach
+We define all our types in declaration files (`*.d.ts`) where typescript automatically adds any types to the context without needing to explicitly import. Declaration files have limitations on what can be imported & instantiated within them, so in the few cases where types derive from instances of things we would usually have to use special tricks to derive them.
+
+The majority of our types are defined in the `domain` layer (see Architecture section below). We generally create types for the following scenarios:
+
+#### Symbol types
+These are unique types that are alternatives to generically typing things as Typescript's primitive types (e.g. `string`, `number`, `boolean`). Doing this helps us to add context to primitive types that we pass around and it allows the type checked to distinguish between different kinds of primitives throughout the code.
+
+For example, an onchain address and a lightning network payment request are both strings, but they aren't interchangeable as a data type. Instead of using `string` type for these types we would define as follows using a "unique symbol" trick:
```
nicolasburtey, line 24:
```suggestion
For example, an onchain address and a lightning network payment request are both strings, but they aren't interchangeable as a data type. Instead of using `string` type for these types we would define as follows using a "unique symbol":
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,223 @@
...
+    return doThing(myArg)
+}
+
+const myOtherFunction = async ({
+  myFirstArg,
+  mySecondArg,
+}: {
+  myFirstArg: MyFirstArg
+  mySecondArg: MySecondArg
+}): ThingResult => {
```
nicolasburtey, line 80: if the function is async should probable be `Promise<ThingResult>` for the type in return

CONTRIBUTING.md
```diff
@@ -0,0 +1,223 @@
...
+
+Note that the top-level function arguments (for `DepositFeeCalculator`) would still be typed at the point of implementation like with any other function, and it is only the method signatures that are included in declaration files.
+
+#### Interfaces
+We use the `interface` keyword to define objects with methods that are intended to be implemented outside of the domain. This most often happens with service implementations like our `ILightningService` interface that gets implemented as the `LndService` function.
+
+Everything else about how we go about typing and implementing things with the `interface` keyword is the same as with how we handle "objects with methods" in our domain as described above.
+
+#### "Enums" via object constants
+Typescript's `enum` keyword has drawbacks that don't fully meet our typing needs. To get around this, we use the trick where we define our intended enum as a standard object and then import it into our type declarations.
```
nicolasburtey, line 135: I don't understand why enum doesn't fit our typing needs. some explanation would be useful

CONTRIBUTING.md
```diff
@@ -0,0 +1,223 @@
...
+  Locked: "locked",
+  Active: "active",
+} as const
+
+// In declaration file
+type AccountStatus =
+  typeof import("./index").AccountStatus[keyof typeof import("./index").AccountStatus]
+```
+## Architecture
+We use hexagonal architecture pattern (some link).
```
nicolasburtey, line 150: still need some link :)

CONTRIBUTING.md
```diff
@@ -0,0 +1,223 @@
...
+#### Examples
+- Access to external resources (database, redis, bitcoind, lnd)
+- Consumption of external services/APIs (twilio, geetest, price, hedging)
+
+#### Dependencies
+- Internal: Domain Layer
+- External: all required dependencies
+
+
+### Application layer
```
nicolasburtey, line 186: `Application access layer` versus `Application layer` is not clear to me the difference -> naming can probably be improved

CONTRIBUTING.md
vindard: I can't the remember the full context for this choice as well. @samer or @bodymindarts may be able to help.

CONTRIBUTING.md
vindard: Sure, this whole section is probably incomplete though. I believe our intention here is to commit what we have for now and then come back to finishing off this document instead of leaving it as a pending PR.

bodymindarts (approved):
IMO this is valuable to give a general overview as is. Can always continue improving.

dolcalmi (approved):

PR #702, opened 2021-11-03
Add github action to generate docs

krtk6160:


spectaql-config.yml
```diff
@@ -0,0 +1,20 @@
+spectaql:
```
dolcalmi, line 1: should this file be in dev folder? (like otel-agent-config.yaml)

dolcalmi (changes requested):
looks good, just a minor suggestion

spectaql-config.yml
krtk6160: The dev folder seems more like a place to keep files involved in the docker-compose dev flow. Not sure if moving the file there makes sense. Wdyt @bodymindarts ?

nicolasburtey (dismissed):

dolcalmi (approved):

bodymindarts (approved):

sandipndev (approved):

PR #706, opened 2021-11-04
upgrade last mongodb version docker-compose

nicolasburtey:
(last version for the v4 branch)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,7 +21,7 @@ services:
     - ALLOW_EMPTY_PASSWORD=yes
     - REDIS_DISABLE_COMMANDS=FLUSHDB,FLUSHALL
   mongodb:
-    image: bitnami/mongodb:4.4.6-debian-10-r0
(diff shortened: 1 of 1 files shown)


bodymindarts:
Have you verified this is the same container as the chart we are running on prod?

nicolasburtey:
> Have you verified this is the same container as the chart we are running on prod?

nope. will do.

nicolasburtey:
I guess it was already up to date from bitnami?

https://github.com/bitnami/charts/blob/master/bitnami/mongodb/values.yaml#L102

nicolasburtey:
actually we may not have the last version of the charts. 

I have hard time finding the mapping charts --> mongodb version mapping

bodymindarts (approved):

krtk6160:
@nicolasburtey  [This PR](https://github.com/GaloyMoney/charts/pull/189) will upgrade our mongodb chart to latest so that we use the latest mongodb as well

krtk6160 (approved):

PR #708, opened 2021-11-04
make ip check asynchronous

nicolasburtey:
also some refactoring to only fetch the user info once from mongodb from the `context` function


bodymindarts:
This is breaking some boundaries would like to keep. Will comment in more detail soon.

src/servers/graphql-server.ts
```diff
@@ -92,31 +92,42 @@ export const startApolloServer = async ({
...
+          [SemanticAttributes.HTTP_CLIENT_IP]: ip,
+        },
         async () => {
-          if (uid) {
-            const loggedInUser = await Users.getUserForLogin({ userId: uid, ip })
-            if (loggedInUser instanceof Error)
+          if (userId) {
+            const lastConnection = new Date()
+
+            const user = await UsersRepository().findByIdAndUpdateLastConnectionDate(
```
bodymindarts, line 110: This function is not appropriate for a `Repository`. Repositories are for retrieving data (`findXxx`) and storing data (aka `persist`) NOT for updating data.
This is (among other things) to prevent coupling of changes of state with the database implementation.
Business logic for around updating state should be implemented on the entities in the domain layer.

src/servers/graphql-server.ts
```diff
@@ -92,31 +92,42 @@ export const startApolloServer = async ({
...
               )
-            domainUser = loggedInUser
-            user = await User.findOne({ _id: uid })
+            domainUser = user
+
             wallet =
-              !!user && user.status === "active"
-                ? await WalletFactory({ user, logger })
+              !!domainUser && domainUser.status === "active"
+                ? await WalletFactory({ user: domainUser, logger })
```
bodymindarts, line 127: This is broken `domainUser` is not the correct type for the `WalletFactory` - getting rid of the `WalletFactory` is an ongoing cleanup that Arvin is working on.

src/servers/graphql-server.ts
```diff
@@ -92,31 +92,42 @@ export const startApolloServer = async ({
...
                 : null
+
+            Users.updateIpInfo({ userId, iPs: user.lastIPs, ip, lastConnection })
```
bodymindarts, line 130: Again a function that should not be on the repository as it is mutating.

src/services/mongoose/users.ts
```diff
@@ -29,6 +27,29 @@ export const UsersRepository = (): IUsersRepository => {
     }
   }
 
+  const findByIdAndUpdateLastConnectionDate = async (
```
bodymindarts, line 30: See explanation above - not appropriate for Repository.

src/services/mongoose/users.ts
```diff
@@ -57,13 +78,23 @@ export const UsersRepository = (): IUsersRepository => {
     }
   }
 
+  const updateIps = async (id, ips) => {
```
bodymindarts, line 81: See above

src/app/users/get-user.ts
```diff
@@ -1,65 +1,53 @@
...
 import { UsersRepository } from "@services/mongoose"
-import { IpFetcher } from "@services/ipfetcher"
 import { getIpConfig } from "@config/app"
-import { RepositoryError } from "@domain/errors"
+import { IpFetcher } from "@services/ipfetcher"
 
 const users = UsersRepository()
 
-export const getUserForLogin = async ({
+export const updateIpInfo = async ({
```
bodymindarts, line 8: Extracting an `updateIpInfo` from the `getUserForLogin` and calling it asynchronously makes sense IMO - but it should still probably be called as part of the `getUserForLogin` use case. Having the outer adapter layer interact with the `UserRepository` and this use case in different ways is much too granular and introduces back a lot of the coupling and hard to test structure that we are working to get away from.

src/app/users/get-user.ts
```diff
@@ -1,65 +1,53 @@
...
-}): Promise<User | ApplicationError> =>
+  userId: UserId
+  iPs: IPType[]
+  ip?: Ip
+  lastConnection: Date
+}): Promise<void> =>
   asyncRunInSpan(
     "app.getUserForLogin",
-    { [SemanticAttributes.CODE_FUNCTION]: "getUserForLogin" },
+    { [SemanticAttributes.CODE_FUNCTION]: "addIp" },
```
bodymindarts, line 21: If you want to change this you should also change the name of the span `app.getUserForLogin`.

bodymindarts (commented):
This PR breaks many of the principles we are working towards.
Hexagonal architecture prescribes a structure which isolates the IO layer (GQL for us) from the domain layer and the domain layer from the services + persistence layer.

The interaction is typically as follows:
graphql-layer: incoming http call
graphql-layer:  calls 1 use case function
use-case-function: retrieves Entity from Repo
use-case-function: executes business logic (ie calls a function on the entity)
domain: checks invariants
domain: mutates internal state of entity (or returns a new one with some fields changed)
use-case-function: persist Entity back to Repo
use-case-function: returns result of operation
graphql-layer: translates returned value into graphql response

src/servers/graphql-server.ts
```diff
@@ -30,6 +29,7 @@ import {
   addAttributesToCurrentSpan,
   ENDUSER_ALIAS,
 } from "@services/tracing"
+import { UsersRepository } from "@services/mongoose"
```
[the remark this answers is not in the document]
nicolasburtey: that makes sense

src/servers/graphql-server.ts
nicolasburtey: the previous design requires 2 iteration to mongodb, which is adding latency. 

we don't want to load the state just to update `lastConnection`. `lastConnection` in the context of grqphql is a write only operation, and this can be done in single pass to mongodb.

How would you do that to not incur the additional network latency?

For payment applications, latency is a key element, and requiring additional back and forth to a database to enforce some pattern might not fit the requirement of having fast query.

src/servers/graphql-server.ts
```diff
@@ -92,31 +92,42 @@ export const startApolloServer = async ({
...
               throw new ApolloError(
                 "Invalid user authentication",
                 "INVALID_AUTHENTICATION",
                 {
-                  reason: loggedInUser,
+                  reason: user,
                 },
               )
-            domainUser = loggedInUser
-            user = await User.findOne({ _id: uid })
```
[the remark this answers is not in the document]
nicolasburtey: what is this breaking? v1?

src/servers/graphql-server.ts
nicolasburtey: what is this breaking? v1?

src/servers/graphql-server.ts
```diff
@@ -138,10 +149,10 @@ export const startApolloServer = async ({
           return {
             ...context,
             logger,
-            uid,
+            uid: userId,
             wallet,
             domainUser,
-            user,
+            user: domainUser,
```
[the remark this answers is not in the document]
nicolasburtey: what is this breaking? v1?

src/services/mongoose/users.ts
nicolasburtey: how to deal with that then? we don't want to retrieve this lastConnection to the local state because this is a write only operation so there is no need for prior state here. 

So the most efficient way to do it is just to update the value while we're getting the last state of the User.

src/services/mongoose/users.ts
nicolasburtey: we only want to update the ip array.because we run this function asynchronously, we don't want to the other elements of the user. with the different promise running in parrallel, we would be at a risk of writing stale data if we were to use the more generic `update` function below

src/servers/graphql-server.ts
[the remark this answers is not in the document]
bodymindarts: Yes the v1 gql resolvers rely on this user being present.

src/servers/graphql-server.ts
bodymindarts: Any calls to the wallet factory that call the internal user reference. The object schema of `user` and `domainUser` are not identical.

src/servers/graphql-server.ts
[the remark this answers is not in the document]
bodymindarts: yes

bodymindarts:
> the previous design requires 2 iteration to mongodb, which is adding latency.

 we don't want to load the state just to update lastConnection. lastConnection in the context of grqphql is a write only operation, and this can be done in single pass to mongodb.

> How would you do that to not incur the additional network latency?

> For payment applications, latency is a key element, and requiring additional back and forth to a database to enforce some pattern might not fit the requirement of having fast query.

A few observations as to the trade of clean code vs performance.

In general I would err on the side of clean code even at the cost of some higher latency. And its not about enforcing a pattern for the sake of it - its about the direct business value that is derived from the end-goals the pattern is intended to achieve.

That being said of course performance is also important - but when well designed this doesn't have to be in conflict.

First of all I would define clear performance related SLOs and then do a performance analysis to see wether or not we are achieving the goals or not. If not I would further analyse where the bottlenecks are (taking an end-to-end systems level view) and discover what the biggest wins would be in terms of effort / reward.

Why this approach? In my past experience designing performance critical applications I have learnt to mistrust my intuition when it comes to performance. For sure avoiding unnecessary round trips will improve performance but at what cost? And how can I be sure that I wouldn't get greater performance benefits by putting my effort somewhere else? Currently I only put like to put effort into performance issues when I
- have proof that the optimisation I am making will have a big impact
- the solution I have chosen is not trading off overall design and bringing in a lot of unwanted side-effects
And in general if we are achieving SLO's or don't have any defined I would not do any of this to begin with - as we have no bench mark to measure our progress against.

Now lets go down a hypothetical path - lets say we had defined SLOs and are not reaching them and analysed that solving exactly this additional round trip to mongo DB is the biggest pay-off in the entire end-to-end system in regards to performance.

Getting high performance without compromising application design requires a clear breakdown of data access patterns (as data access and especially mutation are often the bottlenecks).

Here I would say we actually have 2 use cases:
1) Fetch user data for downstream GQL resolvers
- requires low latency
- read only

2) Update metadata about the user for internal purposes
- has no latency requirements
- mutation

Given the different use cases and requirements I would say the problem here is the coupling of the data.
The solution would be to segregate the data by use-case and extract the connection related stuff into its own table.
That way we could:
- fetch the user as quickly as possible
- update the connection data asynchronously
without a performance or design degradation.

Now refactoring to this is more effort than what this PR proposes but is IMO the correct longterm approach. Is it worth it? I'm not sure - without SLO's or performance analysis its impossible to say.

bodymindarts:
Also keep in mind that one we’ve migrated to v2 there will be many round trips saved. I wouldn’t begin performance analysis until that is done.

nicolasburtey:
I'll make the ip collection independent of the user collection that should solve the issue

src/app/users/get-user.ts
```diff
@@ -1,65 +1,86 @@
...
+        iPs.push(ipInfo)
       }
+      const result = await users.updateIps(userId, iPs)
 
-      const updateResult = await users.update(user)
-
-      if (updateResult instanceof RepositoryError) {
-        return updateResult
+      if (result instanceof Error) {
+        logger
```
dolcalmi, line 82: is this line doing something?

src/services/mongoose/users.ts
```diff
@@ -57,13 +78,25 @@ export const UsersRepository = (): IUsersRepository => {
...
+    try {
+      const result = await User.updateOne({ _id: id }, { lastIps: ips })
```
dolcalmi, line 83: this is not working, should be `lastIPs`

src/domain/users/index.types.d.ts
```diff
@@ -47,7 +48,14 @@ type User = {
 
 interface IUsersRepository {
   findById(userId: UserId): Promise<User | RepositoryError>
+  updateIps(userId: UserId, iPs: IPType[]): Promise<null | RepositoryError>
```
dolcalmi, line 51: Null is not a good return here, please use `true` or `void`

src/services/mongoose/users.ts
```diff
@@ -29,6 +27,29 @@ export const UsersRepository = (): IUsersRepository => {
...
+      const result = await User.findOneAndUpdate(
+        {
+          _id: userId,
+        },
+        {
+          lastConnection,
+        },
+      )
+      if (!result) {
+        return new CouldNotFindUserFromIdError(userId)
```
dolcalmi, line 44: should be something like 
```
      if (result.n === 0) {
        return new CouldNotFindUserError()
      }

      if (result.nModified !== 1) {
        return new CouldNotUpdateLastConnectionError()
      }
```

dolcalmi (changes requested):

src/domain/users/index.types.d.ts
nicolasburtey: this has been deleted

src/app/users/get-user.ts
```diff
@@ -1,65 +1,99 @@
...
 
 const users = UsersRepository()
+const usersIp = UsersIpRepository()
 
 export const getUserForLogin = async ({
   userId,
   ip,
+  logger,
 }: {
-  userId: string
```
bodymindarts, line 18: We're passing in a string here because this is the boundary to the 'type-safe' world. Conversion of primitive types to domain types happen within the `app` layer as it may require domain specific checks that would be reported back to the outer layer.

src/app/users/get-user.ts
```diff
@@ -1,65 +1,99 @@
...
-  ip?: string
+  userId: UserId
+  ip?: Ip
```
bodymindarts, line 17: Same here - should be a primitive type as input. Convert to domain type via `checkedToIp` which can make sure it really is an ip.

src/app/users/get-user.ts
```diff
@@ -1,65 +1,99 @@
...
     { [SemanticAttributes.CODE_FUNCTION]: "getUserForLogin" },
     async () => {
-      const user = await users.findById(userId as UserId)
+      const user = await users.findById(userId)
+
       if (user instanceof Error) {
         return user
       }
-      addAttributesToCurrentSpan({
-        [ENDUSER_ALIAS]: user.username,
```
bodymindarts, line 30: Why is recording the `user.username` deleted?

src/app/users/get-user.ts
```diff
@@ -1,65 +1,99 @@
...
+  userId,
+  ip,
+  logger,
+}: {
+  userId: UserId
+  ip?: Ip
+  logger: Logger
+}): Promise<void> =>
+  asyncRunInSpan(
+    "app.addIp",
```
bodymindarts, line 47: This should be the same as the name of the function `app.updateUserIpInfo`

src/app/users/get-user.ts
```diff
@@ -1,65 +1,99 @@
...
+    { [SemanticAttributes.CODE_FUNCTION]: "addIp" },
```
bodymindarts, line 48: Also wrong name.

src/app/users/get-user.ts
```diff
@@ -1,65 +1,99 @@
...
-        } else if (ipConfig.proxyCheckingEnabled) {
+
+      const lastConnection = new Date()
+
+      const userIp = await usersIp.findById(userId)
+
+      if (userIp instanceof RepositoryError) throw userIp
+
+      if (!ip || !ipConfig.ipRecordingEnabled) {
+        const result = await usersIp.update(userId, lastConnection)
```
bodymindarts, line 59: `XxxRepository.update` should receive the entity - destructuring should happen behind the repo interface.
In this case that would mean:
`const result = await usersIp.update(userIp)`

src/app/users/get-user.ts
```diff
@@ -1,65 +1,99 @@
...
+
+        if (result instanceof Error) {
+          logger.error(
+            { result, userId, ip },
+            "impossible to update user last connection",
+          )
+        }
+
+        return
```
bodymindarts, line 68: Why aren't you returning the error here? Returning it will mean it automatically gets reported via OpenTelemetry `span.recordException` in the wrapper function `runInSpan`... eating it means we loose that information in honeycomb.

src/app/users/get-user.ts
```diff
@@ -1,65 +1,99 @@
...
         }
+        userIp.lastIPs.push(ipInfo)
       }
+      const result = await usersIp.update(userId, lastConnection, userIp.lastIPs)
 
-      const updateResult = await users.update(user)
-
-      if (updateResult instanceof RepositoryError) {
-        return updateResult
+      if (result instanceof Error) {
```
bodymindarts, line 94: Again don't eat the error. Always return it - it will automatically be reported via OpenTelemetry.

src/app/users/get-user.ts
```diff
@@ -1,65 +1,99 @@
...
-            })
+          const ipFetcherInfo = await ipFetcher.fetchIPInfo(ip as IpAddress)
+
+          if (!(ipFetcherInfo instanceof Error)) {
+            ipInfo = { ...ipInfo, ...ipFetcherInfo }
           }
         }
+        userIp.lastIPs.push(ipInfo)
       }
+      const result = await usersIp.update(userId, lastConnection, userIp.lastIPs)
```
bodymindarts, line 92: Again bad usage of repo.update - see below.

src/domain/users-ip/index.types.d.ts
```diff
@@ -0,0 +1,15 @@
+type UserIp = {
```
bodymindarts, line 1: Not sure I like the name - how about `UserConnectionsRecord` or at least it the `Ip` should probably be plural `UserIPs` right?

src/domain/users-ip/index.types.d.ts
```diff
@@ -0,0 +1,15 @@
...
+  readonly id: UserId
+  lastIPs: IPType[]
```
bodymindarts, line 3: Why is this called `lastIPs`? Is the array pruned?

src/domain/users-ip/index.types.d.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+interface IUsersIpRepository {
+  update(
```
bodymindarts, line 7: This function should just receive 1 argument `update(userIp: UserIp)`.

src/domain/users/index.types.d.ts
```diff
@@ -37,11 +37,10 @@ type User = {
   readonly quizQuestions: UserQuizQuestion[]
   readonly defaultAccountId: AccountId
   readonly deviceTokens: DeviceToken[]
-  readonly lastIPs: IPType[]
   readonly createdAt: Date
+  readonly status: AccountStatus
```
bodymindarts, line 41: Why is this added as part of this PR? Is it related?

src/servers/graphql-server.ts
```diff
@@ -82,7 +82,7 @@ export const startApolloServer = async ({
       const apiKey = context.req?.apiKey ?? null
       // @ts-expect-error: TODO
       const apiSecret = context.req?.apiSecret ?? null
-      const uid = token?.uid ?? null
+      const userId = (token?.uid as UserId) ?? null
```
bodymindarts, line 85: Type casting to domain types should happen within the app layer to guarantee application of any needed preconditions / validation.

src/servers/trigger.ts
```diff
@@ -95,7 +95,7 @@ export async function onchainTransactionEventHandler(tx) {
       return
     }
 
-    const user = await User.findOne({ _id: userId })
+    const user = await User.findOne({ _id: userId }, { lastIPs: 0, lastConnection: 0 })
```
bodymindarts, line 98: Can you explain why this change is needed?

src/services/mongoose/users-ip.ts
```diff
@@ -0,0 +1,60 @@
...
+import {
+  CouldNotFindError,
+  CouldNotFindUserFromIdError,
+  PersistError,
+  RepositoryError,
+  UnknownRepositoryError,
+} from "@domain/errors"
+
+export const UsersIpRepository = (): IUsersIpRepository => {
+  const update = async (
```
bodymindarts, line 11: See other comments - signature should be:
```
const update = async (
  {
  ...
  }: UserIp): Promise<true | RepositoryError>
```

src/services/mongoose/users-ip.ts
```diff
@@ -0,0 +1,60 @@
...
+  return {
+    update,
+    findById,
+  }
+}
+
+const userFromRaw = (result: UserIpType): UserIp => {
+  return {
+    id: result.id as UserId,
+    lastIPs: (result.lastIPs || []) as IPType[],
```
bodymindarts, line 58: Is this cast really necessary?

src/services/mongoose/users-ip.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }
+
+  return {
+    update,
+    findById,
+  }
+}
+
+const userFromRaw = (result: UserIpType): UserIp => {
```
bodymindarts, line 55: Should probably be called `userIpFromRaw` as `userIp` is the entity here.

src/services/mongoose/users.ts
```diff
@@ -97,41 +102,38 @@ export const UsersRepository = (): IUsersRepository => {
...
-    defaultAccountId: result.id as AccountId,
-    deviceTokens: (result.deviceToken || []) as DeviceToken[],
-    lastConnection: result.lastConnection,
-    lastIPs: (result.lastIPs || []) as IPType[],
-    createdAt: new Date(result.created_at),
-  }
-}
+const userFromRaw = (result: UserType): User => ({
+  id: result.id as UserId,
+  status: result.status as AccountStatus,
```
bodymindarts, line 107: Why is status added as part of this PR?

src/services/mongoose/schema.ts
```diff
@@ -408,7 +408,7 @@ UserSchema.virtual("userIsActive").get(async function (this: typeof UserSchema)
 })
 
 UserSchema.statics.getUserByPhone = async function (phone: string) {
```
bodymindarts, line 410: This should be moved to `UsersRepository` - its only used in admin panel and would be a quick change - but can be done in another PR.

bodymindarts (changes requested):
This iteration is much more in line with the design we're going for.

Only thing is the `Repo.update` function doesn't have the right signature - and I'm not sure about the naming of the entity.

Also I don't understand why you added:
```
{ lastIPs: 0, lastConnection: 0 }
```
To all the other queries in trigger - I'm sure there is a good explanation, just don't understand.

src/app/users/get-user.ts
nicolasburtey: I'm not sure to understand. which layer is typesafe, and which one is not?

graphql is not type safe? app should not be as well? or app is where the conversion is happening?

src/app/users/get-user.ts
nicolasburtey: do you have an example to follow this somewhere else in the code?

src/app/users/get-user.ts
nicolasburtey: some DRY would be great here. not sure if that could work [Function.name](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name) as it's called within the function

src/app/users/get-user.ts
bodymindarts: Conversation happens in app layer. Instructed user input gets checked / validated and if all is good converted to correct type. Internally we trust our types because they are checked as a first thing inside the app boundary.

src/app/users/get-user.ts
nicolasburtey: that makes sense

src/app/users/get-user.ts
bodymindarts: here for example - https://github.com/GaloyMoney/galoy/blob/main/src/app/wallets/get-on-chain-fee.ts#L65

Its not 100% consistent but it is what we have decided to converge to. And if a type got returned from app then that is also trusted as long as conversion happened in app layer.

src/app/users/get-user.ts
```diff
@@ -1,65 +1,99 @@
...
-        )
-        if (lastIP) {
-          lastIP.lastConnection = user.lastConnection
-        } else if (ipConfig.proxyCheckingEnabled) {
+
+      const lastConnection = new Date()
+
+      const userIp = await usersIp.findById(userId)
+
+      if (userIp instanceof RepositoryError) throw userIp
```
nicolasburtey, line 56: this should be return here not throw right? @bodymindarts @dolcalmi ?

src/app/users/get-user.ts
nicolasburtey: I didn't returned it because this is a promise that is not being awaited and there will be not action done whether this promise succeed or not.

if I'm return (or throwing?) an error this would have an effect on honeycomb?

src/app/users/get-user.ts
bodymindarts: Agree - I think @dolcalmi wanted to think about a good way to add tracing to app + service layer via wrapping our index files anyway. Until he's spent some effort here I wouldn't change anything.

src/app/users/get-user.ts
bodymindarts: Correct in app layer always return the error

src/domain/users-ip/index.types.d.ts
nicolasburtey: not as of now but eventually it could/should

src/domain/users-ip/index.types.d.ts
nicolasburtey: moving to `UserIPs` for now

src/app/users/get-user.ts
bodymindarts: Errors should always be returned from app layer. Its up to the client what to do with it (including silently ignore).

Apart from that general approach to errors there is also the added benefit that since you are wrapping the function with `runInSpan` it will automatically report the error to honeycomb - but regardless you should always return it and let the client decide to ignore or not.

Also IMO logging should happen at a higher level where we decide if the error is unrecoverable or returned to user or just eaten etc. That decision should happen at the top of the call stack - outside of app layer and that's where we should log the decision IMO. Passing around the logger down into the call stack is a lot of boilerplate without much added value. But its not a big deal - as some of our monitoring relies on logging we haven't completely cleaned it up yet.

src/servers/trigger.ts
nicolasburtey: it's a performance optimization. ips array can be quite large currently because there is no pruning, and it's used by no application expect the one at the context. this would dramatically reduce the need of sending data to the wire between nodejs and mongodb for frequent users

src/services/mongoose/schema.ts
nicolasburtey: keeping that for another PR

src/domain/users-ip/index.types.d.ts
bodymindarts: I prefer to use names that reflect the _is_ rather than the _should be_ but its not a big deal.

src/services/mongoose/users-ip.ts
nicolasburtey: I'm not sure? I basically copied what was in the `mongoose/user.ts` here. still not up to speed on best practice with typing on Typescript (specially wrt our codebase)

bodymindarts: If a cast es necessary to make the type checker happy then this is the right place for it. Still want to avoid it as much as possible and since `IPType` is a type defined on schema I thought perhaps the compiler would be happy without it here.

src/services/mongoose/users.ts
nicolasburtey: it's not an `update` query, it's a `findBy`.

src/services/mongoose/users.ts
nicolasburtey: I seen as missing, but I guess that might be intentional? it's currently only used from the admin panel AFAIR

bodymindarts: Aha admin panel is still using the old user type (coming from mongoose) not this one. No need to add it now.

bodymindarts: The admin panel user could and probably should be its own type with its own repository - as it serves different use cases than the user we are returning here.

src/services/mongoose/users-ip.ts
nicolasburtey: an empty array would not be a `IPType[]`

src/app/users/get-user.ts
nicolasburtey: there is basically no check being done here at the app level previously. 

doing a `await users.findById(userId as UserId)` is basically doing the same thing?

src/app/users/get-user.ts
nicolasburtey: > but regardless you should always return it and let the client decide to ignore or not.

maybe you miss my point above: we're not bubbling up this function

nicolasburtey: > Also IMO logging should happen at a higher level where we decide if the error is unrecoverable or returned to user or just eaten etc. That decision should happen at the top of the call stack - outside of app layer and that's where we should log the decision IMO. Passing around the logger down into the call stack is a lot of boilerplate without much added value. But its not a big deal - as some of our monitoring relies on logging we haven't completely cleaned it up yet.

your assumption is that you only want to log the critical issue. I guess my assumption is that we can log more lower level call (typically called "info", "debug" or "trace")

src/app/users/get-user.ts
bodymindarts: Correct - but still its happening at the correct layer. If we wanted to add a check we wouldn't have to investigate all the client code to make sure the check is added everywhere before calling into app.

src/app/users/get-user.ts
bodymindarts: > maybe you miss my point above: we're not bubbling up this function

Ah do you mean because its not `export`ed there is no external client? In many languages (ie rust, Golang) its still common to return the error unless it's a top level function (like `main`) so it's just something I consider normal. I guess its not idiomatic in JS due to `try/catch` semantics but still it would be good here so that honeycomb will receive it.

> your assumption is that you only want to log the critical issue. I guess my assumption is that we can log more lower level call (typically called "info", "debug" or "trace")

Yeah I have no problem with that. We'll see how it evolves with the usage of the OTEL libraries - how we want to approach logging vs tracing etc in the future.

src/app/users/get-user.ts
nicolasburtey: it's also happening at the correct layer now? the `string` is passed from graphql, and casted as `UserId` when entering this function

bodymindarts: AFAICT this function receives a `userId: UserId` - it should be casted _within_ this function.

src/app/users/get-user.ts
nicolasburtey: btw, we should consider honeycomb as a complement: we can't rely on this for our tech stack as the sole way to debug our application because this is not FOSS. so saying "this works with honeycomb" is not something we should emphasize if we/other people using the stack don't have another solution to debug issues

nicolasburtey: > Ah do you mean because its not exported there is no external client

no because this function is called as a "fire and forget". this is the whole point of this whole PR: run this function asynchronously from the rest of the graphql flow.

bodymindarts: OTEL is open source - if we don't want to rely on honeycomb we can deploy a jaeger UI. We have nothing honeycomb specific in our codebase. The only thing related to honeycomb is where our central `opentelemetry-collector` forwards the traces for visualising. All comments I said above regarding returning errors you can replace the name `honeycomb` with `opentelemetry-collector`.

bodymindarts: Any tracing visualiser can be used to help debugging. That we ship to honeycomb is an implementation detail, but not of our app - rather of how we've configured the collector.

If you don't want to return the error than you can also do the following:
```
const span = trace.getSpan(context.active())
span.recordException(error)
return
```

This is the equivalent code like adding the log.

But this is boilerplate you can avoid because of https://github.com/GaloyMoney/galoy/blob/main/src/services/tracing.ts#L159-L162
This has nothing to do with honeycomb.

bodymindarts (dismissed):
LGTM - approving with minor comment.

vindard (changes requested):
Left some small comments for practices we've been using to get the most out of the type-checker. Otherwise this looks ok and I can approve.

src/app/users/get-user.ts
```diff
@@ -5,61 +5,113 @@ import {
...
+
       if (user instanceof Error) {
         return user
       }
       addAttributesToCurrentSpan({
         [ENDUSER_ALIAS]: user.username,
       })
-      user.lastConnection = new Date()
+      // this routing run asynchrously, to update metadata on the background
+      updateUserIPsInfo({ userId, ip, logger } as {
```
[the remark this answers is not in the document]
nicolasburtey: maybe you can sync with @bodymindarts on this because he told me this is the thing __not__ to do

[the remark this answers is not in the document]
bodymindarts: @vindard  cast should happen 1 time only at the top of the function. This is what we have been doing with the checkedTo methods. We should be doing the same with the UserId type of types. 

Arvin is right though that we shouldn’t cast more than one at the top of the function

src/app/users/get-user.ts
```diff
@@ -5,61 +5,113 @@ import {
...
 }: {
   userId: string
   ip?: string
+  logger: Logger
 }): Promise<User | ApplicationError> =>
   asyncRunInSpan(
     "app.getUserForLogin",
     { [SemanticAttributes.CODE_FUNCTION]: "getUserForLogin" },
     async () => {
       const user = await users.findById(userId as UserId)
```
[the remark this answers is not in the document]
nicolasburtey: maybe you can sync with @bodymindarts on this because he told me this is the thing not to do

src/domain/users-ips/index.types.d.ts
```diff
@@ -0,0 +1,11 @@
...
+  readonly id: UserId
+  lastIPs: IPType[]
+}
+
+interface IUsersIPsRepository {
+  update(userIp: UserIPs): Promise<true | RepositoryError>
+  findById(userId: UserId): Promise<UserIPs | RepositoryError>
+}
+
+type Ip = string
```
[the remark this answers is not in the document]
nicolasburtey: what does `type Ip = string` don't do that `type Ip = string & { [ipSymbol]: never }` do?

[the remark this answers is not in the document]
bodymindarts: The former makes any string be a valid value of the type. It’s like an alias. The later will only compile with an explicit cast. It is a me unique type. Not just an alias. More safe and what we want

src/app/users/get-user.ts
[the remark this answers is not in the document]
bodymindarts: Contd once a tour has been returned from the app layer we can then pass it on directly (not via the underlying primitive) but the initial conversion should happen in app. Cc @vindard

src/domain/users-ips/index.types.d.ts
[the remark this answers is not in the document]
nicolasburtey: ok make sense. thanks

nicolasburtey:
I'm suggesting to keep this typing like this for now. we can work on better type translation on the app layer on a more dedicated PR. @bodymindarts @vindard

vindard (dismissed):

vindard (approved):

dolcalmi (approved):

PR #709, opened 2021-11-04
Disable playground polling setting

samerbuna:


bodymindarts:
I don’t like the implicit coupling to the btc network. We may want to have a testing environment in mainnet. Can we have an injectable setting in the yaml?

bodymindarts:
“playground: true” could also be responsible for the helmet setting.

dolcalmi (approved):

default.yaml
```diff
@@ -146,8 +146,8 @@ onChainWallet:
 # TODO
 # carrierRegexFilter: ""
 
-helmet:
-  disableContentPolicy: false
+apollo:
+  playground: false
```
nicolasburtey, line 150: I would have this on by default. wdyt?

we can disable it on the charts so that it's by default off in production

dolcalmi: if you decide to set to true by default please add additional validation for tests in writeSDLFile to avoid open handles

samerbuna: I'm actually in favor of removing defaults all together. Every environment should have its own configurations. 

Having defaults carries the risk of one environment using a default value when it shouldn't. Setting things closer to production is a safer bet for now (it's still risky too).

bodymindarts: I like having explicit settings per environment to a certain extent we already have this via the `charts` + `deployments` repos the only thing that needs specifying in this repo are the dev settings. How would you make this explicit in the file structure so that the workflow is streamlined?

bodymindarts (approved):
I'm approving but think we should certainly clean up the way we provide config for dev environment. Its not natural to have to change a setting to get to what most ppl would expect.

PR #711, opened 2021-11-05
Update ip info asynchronously

nicolasburtey:


src/app/users/get-user.ts
```diff
@@ -23,43 +22,40 @@ export const getUserForLogin = async ({
...
+                ip,
+                ...ipInfo,
+                Type: ipInfo.type,
+                firstConnection: user.lastConnection,
+                lastConnection: user.lastConnection,
+              })
+            }
           }
         }
+        await users.update(user)
```
nicolasburtey, line 54: this is not concurrency safe. 

if the user get updated in the meantime (because `fetchIPInfo` is slow for instance, which is frequently the case), then we would overwrite the change.

cc @bodymindarts

dolcalmi: If I remember corretly we had the same discussion with addresses. The solution was to create its own repo https://github.com/GaloyMoney/galoy/blob/main/src/services/mongoose/wallet-on-chain-addresses.ts

bodymindarts: Exactly need to segregate the update paths either in the database level or logically via a repository. 

For the address situation we had to do the update database side because of the atomicity.

bodymindarts: Thinking more about it - other than segregating on DB level - this is probably the best intermediate solution.

Have a `UserConnectionsRepository` with a `persistNew` function that is atomic like the on-chain-addresses example.

We should remove `lastIPs` and `lastConnection` from the [user type](https://github.com/GaloyMoney/galoy/blob/main/src/domain/users/index.types.d.ts#L32) and remove those fields from the [`UsersRepository.update` function](https://github.com/GaloyMoney/galoy/blob/main/src/services/mongoose/users.ts#L60) so that there is no risk of concurrent update.

Like that we can continue to postpone the refactoring of the db schema itself until later. Thats the approach we've been taking so far.

PR #718, opened 2021-11-06
no need for looping through invoices for admin panel

nicolasburtey:
this is different for payments.

note: there are a aside case for payment not covered here

if we were to have 2 lnd, and the payment would have been sent through the first lnd, and failed. then send again with the second lnd and succeed.

we may only show the failed payment where in fact the payment would have succeed.

I think we should rely on medici also for the payments

### DIFF
--- a/src/app/lightning/lookup-invoice-by-hash.ts
+++ b/src/app/lightning/lookup-invoice-by-hash.ts
@@ -1,13 +1,21 @@
 import { LndService } from "@services/lnd"
+import { WalletInvoicesRepository } from "@services/mongoose"
+import { RepositoryError } from "@domain/errors"
 
-export const lookupInvoiceByHashAndPubkey = async ({
+export const lookupInvoiceByHash = async ({
   paymentHash,
-  pubkey,
 }: {
   paymentHash: PaymentHash
-  pubkey: Pubkey
 }): Promise<LnInvoiceLookup | ApplicationError> => {
   const lndService = LndService()
   if (lndService instanceof Error) return lndService
+
+  const invoicesRepo = WalletInvoicesRepository()
+  const walletInvoice = await invoicesRepo.findByPaymentHash(paymentHash)
+
+  if (walletInvoice instanceof RepositoryError) return walletInvoice
+
+  const { pubkey } = walletInvoice
+
   return lndService.lookupInvoice({ paymentHash, pubkey })
 }
--- a/src/app/lightning/lookup-invoice-by-hash.ts
+++ b/src/app/lightning/lookup-invoice-by-hash.ts
@@ -1,9 +1,13 @@
 import { LndService } from "@services/lnd"
 
-export const lookupInvoiceByHash = async (
-  paymentHash: PaymentHash,
-): Promise<LnInvoiceLookup | ApplicationError> => {
+export const lookupInvoiceByHashAndPubkey = async ({
+  paymentHash,
+  pubkey,
+}: {
+  paymentHash: PaymentHash
+  pubkey: Pubkey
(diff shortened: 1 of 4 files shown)


src/graphql/admin/root/query/lightning-invoice.ts
```diff
@@ -11,7 +13,19 @@ const LightningInvoiceQuery = GT.Field({
   resolve: async (_, { hash }) => {
     if (hash instanceof Error) throw hash
 
-    const lightningInvoice = await lookupInvoiceByHash(hash)
+    const invoicesRepo = WalletInvoicesRepository()
+    const walletInvoice = await invoicesRepo.findByPaymentHash(hash)
```
dolcalmi, line 17: we cant do this, we are cleaning userinvoices and the purpose of this query (this is only a support query) is to look for a hash directly on lnd, should not be related to business logic.

nicolasburtey: > Edit: graphql query should call only use cases in app layer, this should be moved to a new use case (maybe inside wallet)

yup ok

nicolasburtey: > we cant do this, we are cleaning userinvoices (cron) and the purpose of this query (this is only a support query) is to look for a hash directly on lnd, should not be related to business logic.

we will also delete payment on lnd after this has been merged: https://github.com/GaloyMoney/galoy/pull/665

so really we should look into lnd and medici. but looping over lnd to find for a hash is not the best way to move forward. we know which lnd the hash belongs to, and the invoiceUser is currently only deleted after the invoice has expired ( > 90 days ).

bodymindarts (approved):

src/app/lightning/lookup-invoice-by-hash.ts
```diff
@@ -1,9 +1,23 @@
 import { LndService } from "@services/lnd"
+import { WalletInvoicesRepository } from "@services/mongoose"
+import { RepositoryError } from "@domain/errors"
 
-export const lookupInvoiceByHash = async (
-  paymentHash: PaymentHash,
-): Promise<LnInvoiceLookup | ApplicationError> => {
+export const lookupInvoiceByHash = async ({
+  hash,
```
bodymindarts, line 6: why did you rename this argument? We've been consistently naming it `paymentHash` everywhere. If you don't rename it you can also delete line 19-20...

src/app/lightning/lookup-invoice-by-hash.ts
```diff
@@ -1,9 +1,21 @@
 import { LndService } from "@services/lnd"
+import { WalletInvoicesRepository } from "@services/mongoose"
+import { RepositoryError } from "@domain/errors"
 
-export const lookupInvoiceByHash = async (
-  paymentHash: PaymentHash,
-): Promise<LnInvoiceLookup | ApplicationError> => {
+export const lookupInvoiceByHash = async ({
+  paymentHash,
```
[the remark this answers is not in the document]
nicolasburtey: we may want to have a standard for this. I just revert the change for now

vindard (approved):
This looks ok to me now. My only question would be if there could be any edge cases where we try to call this method for a payment that's already been deleted from the `invoiceusers` collection.

bodymindarts (approved):

nicolasburtey:
> This looks ok to me now. My only question would be if there could be any edge cases where we try to call this method for a payment that's already been deleted from the invoiceusers collection.

This would return an error?

but the assumption is the payment would be deleted by lnd also anyhow. 

a future iteration would be to fetch the value also in medici

PR #720, opened 2021-11-06
better default (IMO)

nicolasburtey:
so there is no need for dirty state on dev. this is override in prod with helm.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -56,10 +56,7 @@ $ make watch
 
 ### Using GraphiQL
 
-You can load GraphiQL, a web GUI for GraphQL.
-
-First, `apollo.playground` should be set to `true` in `default.yaml`
-Then, restart the server and open the following url:
+You can load GraphiQL, a web GUI for GraphQL. Start the server and open the following url:
 
 - http://localhost:4000/graphql (old API - deprecated)
 - http://localhost:4001/graphql (admin API)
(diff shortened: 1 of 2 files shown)


bodymindarts:
Needs to be overwritten https://github.com/GaloyMoney/charts/blob/main/charts/galoy/values.yaml (You're PR disabled it in the _dev_ settings where it should actually also be active).

bodymindarts (approved):

dolcalmi:
please dont forget to resolve writeSDLFile open handle if this is set to true by default.

nicolasburtey:
@dolcalmi not sure to completely understand your point? do you wan to make a commit on my PR if there's something to fix?

samerbuna:
> please dont forget to resolve writeSDLFile open handle if this is set to true by default. I fixed it here
> 
> https://github.com/GaloyMoney/galoy/blob/2d3a3452d083927b0876161017c03d1ab2965ab7/src/graphql/admin/index.ts#L16
> 
> but after rebase will not be validated when server runs from tests

Maybe we can mock that value in jest?

dolcalmi:
> @dolcalmi not sure to completely understand your point? do you wan to make a commit on my PR if there's something to fix?

I dont know if this will be approved or not, just letting you know that this will create an open handle in tests (I had fixed it in another PR)

nicolasburtey:
@dolcalmi can you approve this PR if this is good on your end? I'm not sure what is the status of the other PR you are referring to

dolcalmi (approved):
please dont forget update prod

PR #725, opened 2021-11-07
Use only 1 type for the User object

samerbuna:
Using multiple types mess with client caching

This PR is a requirement for https://github.com/GaloyMoney/galoy-mobile/pull/283


src/graphql/root/mutation/user-contact-update-alias.ts
```diff
@@ -27,19 +27,19 @@ const USerContactUpdateAliasMutation = GT.Field({
...
-    if (user instanceof Error) {
-      return { errors: [{ message: user.message }] }
+    if (contact instanceof Error) {
+      return { errors: [{ message: contact.message }] }
     }
 
     return {
       errors: [],
-      user,
+      contact,
```
vindard, line 42: `SuccessPayload` type has `success` key here instead I believe

src/graphql/types/payload/user-contact-update-alias.ts
```diff
@@ -1,15 +1,16 @@
 import { GT } from "@graphql/index"
+
 import IError from "../abstract/error"
-import { UserDetails } from "../object/user"
+import UserContact from "../object/wallet-contact"
 
 const UserContactUpdateAliasPayload = new GT.Object({
```
vindard, line 6: Is this still needed now given that we're using `SuccessPayload`?

vindard (commented):
Looks like a straight switchover to using our `User` type everywhere instead of having partial `User` types. I don't know enough about the tradeoffs of that with graphql so this looks ok code-structure wise.

Just left two small comments that may lead to potential changes.

src/graphql/root/mutation/user-contact-update-alias.ts
samerbuna: Rebased to fix this and merge with new description fields.

I made the payload here to be a contact user.

bodymindarts (dismissed):

vindard (dismissed):

vindard (approved):

bodymindarts (approved):

PR #704, opened 2021-11-04
Remove price mock + Price history setup

dolcalmi:
- Remove price mock
- Fix tests open handles (writeSDLFile and httpServers)
- Add price history (1 year)

This is the 2nd split of Price refactor https://github.com/GaloyMoney/galoy/pull/662

Issue #683, opened 2021-10-31
move ccxt/price history out of the main repo


nicolasburtey:
now that we now longer have ccxt dependency in the main repo, we should have price history in an independant repo, so that we don't have to embedded ccxt in our main logic. 

- ccxt is a large repo (which could become a supply chain attack vector) and here we're only using it for fetching ohlcv. we could just use the api of the exchange directly
- we don't need to use mongodb for the price history. we could just have a simple pod and disk storage for it.

the data could be sent to the graphql pod either with GRPC, or just point to the disk storage.

this can be done after https://github.com/GaloyMoney/galoy/pull/662 is been merged. there is no urgency to do it.

dolcalmi:
should we include price histoy in https://github.com/GaloyMoney/price/ ?

nicolasburtey:
> should we include price history in https://github.com/GaloyMoney/price/ ?

we could. but ultimately they serve different purpose so this could be another pod.

PR #726, opened 2021-11-08
Refactor price use cases

dolcalmi:
- Add price use case. 
- Remove price query from internal services https://github.com/GaloyMoney/galoy/pull/700
- Add price unit/integration tests (implemented in https://github.com/GaloyMoney/galoy/pull/704)
  - Add initial setup with 1 year of price data. To test price query please run `TEST="01" make reset-integration`
- Remove old real time price mock
- Fix tests open handles (writeSDLFile and httpServers)
- Remove price-impl class
  - updatePriceHistory was only moved and not refactored because will be changed according to https://github.com/GaloyMoney/galoy/issues/683
- Refactor price service
  - get real time price
  - history: now range and interval grouping is done in db
  - Schema moved to price service (in preparation to remove it from backend https://github.com/GaloyMoney/galoy/issues/683)


src/domain/cache/index.types.d.ts
```diff
@@ -0,0 +1,14 @@
+type LocalCacheServiceError = import("./errors").LocalCacheServiceError
+
+type CacheKeys =
+  typeof import("./index").CacheKeys[keyof typeof import("./index").CacheKeys]
+
+interface ILocalCacheService {
+  set<T>(
+    key: CacheKeys | string,
+    value: T,
+    ttlSecs: number,
```
vindard, line 10: We have a `Seconds` type we can potentially use here

src/graphql/root/query/btc-price-list.ts
```diff
@@ -35,39 +46,32 @@ const BtcPriceListQuery = GT.Field({
...
 
-    for (const price of hourlyPrices.reverse()) {
-      if (1000 * price.id < rangeStart) {
-        break
-      }
-      const btcPriceInCents = price.o * 100 * 10 ** 8
-      prices.unshift({
-        timestamp: price.id,
+    const prices: PricePointType[] = hourlyPrices.map(({ date, price }) => {
+      const btcPriceInCents = price * 100 * 10 ** 8
```
vindard, line 59: Should we make `10 ** 8` a constant like we have with different millisecond values (maybe `SATS_PER_BTC`)?

We also use `Math.pow(10, 8)` in other places.

src/servers/trigger.ts
```diff
@@ -166,19 +164,18 @@ export const onInvoiceUpdate = async (invoice) => {
...
 
 const updatePriceForChart = () => {
-  const price = new Price({ logger: baseLogger })
   const interval = 1000 * 30
   return setInterval(async function () {
     try {
-      await price.update()
-      publishCurrentPrice()
+      await updatePriceHistory()
+      await publishCurrentPrice()
```
vindard, line 178: Just for my knowledge, why did we change to this to explicit `await`?

src/services/cache/index.ts
```diff
@@ -0,0 +1,45 @@
...
+    value: T,
+    ttlSecs: number,
+  ): Promise<T | LocalCacheServiceError> => {
+    try {
+      const res = localCache.set<T>(key, value, ttlSecs)
+      if (res) return Promise.resolve(value)
+    } catch (err) {
+      return Promise.resolve(new UnknownLocalCacheServiceError(err))
+    }
+    return Promise.resolve(new LocalCacheNotAvailableError())
```
vindard, line 22: For readability, these might be better closer to the `if` conditional that they're related to (at `Ln18`)

src/services/price/index.ts
```diff
@@ -1,19 +1,97 @@
...
 
 export const PriceService = (): IPriceService => {
-  const fetchPrice = async () => {
+  const pair = "BTC/USD"
+  const exchange = "bitfinex"
+
+  const getRealTimePrice = async (): Promise<UsdPerSat | PriceServiceError> => {
+    try {
+      const { price } = await getPrice({})
+      if (price > 0) return (price / Math.pow(10, 8)) as UsdPerSat
```
vindard, line 29: Constant comment

src/services/price/update-price-history.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    try {
+      for (const value of ohlcv) {
+        // FIXME inefficient
+        if (doc.pair.exchange.price.find((obj) => obj._id.getTime() === value[0])) {
+          continue
+        }
+
+        baseLogger.debug({ value0: value[0] }, "adding entry to our price database")
+        doc.pair.exchange.price.push({ _id: value[0], o: value[1] / Math.pow(10, 8) })
```
vindard, line 63: Constant comment

vindard (changes requested):
New PriceService abstraction and related use-cases look good to me. Also noting that we intend to move `update-price-history` out of this repo so code was simply moved and not refactored.

Just left some small nits otherwise.

src/servers/trigger.ts
dolcalmi: this is in trigger so we dont need fire and forget, also this makes the last price publish consistent specially for galoy-pay

src/graphql/root/query/btc-price-list.ts
dolcalmi: Fixed

vindard (dismissed):

src/services/price/index.ts
```diff
@@ -1,19 +1,98 @@
...
+import { PriceHistory } from "./schema"
+import { PriceProtoDescriptor } from "./grpc"
+import { SATS_PER_BTC } from "@config/app"
+
+const priceUrl = process.env.PRICE_ADDRESS ?? "galoy-price"
+const pricePort = process.env.PRICE_PORT ?? "50051"
+const fullUrl = `${priceUrl}:${pricePort}`
+const priceClient = new PriceProtoDescriptor.PriceFeed(
+  fullUrl,
+  credentials.createInsecure(),
```
nicolasburtey, line 19: I don't remember what `createInsecure()` in term of security but I guess something we might want to look into at some point

src/services/price/update-price-history.ts
```diff
@@ -0,0 +1,108 @@
...
+    timeout: 5000,
+  })
+
+  try {
+    const ohlcv = await client.fetchOHLCV(pair, "1h", since, limit)
+    baseLogger.info("data fetched from exchange")
+    return ohlcv
+  } catch (e) {
+    if (e instanceof ccxt.NetworkError) {
+      throw new Error(`fetchTicker failed due to a network error: ${e.message}`)
```
nicolasburtey, line 99: I guess we might want to refactor this (as another PR eventually) because IIUC we don't want to throw in services, is that correct?

test/integration/services/price.spec.ts
```diff
@@ -0,0 +1,35 @@
+import { PriceService } from "@services/price"
```
nicolasburtey, line 1: could this be a unit test?

test/integration/services/price.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+  describe("listHistory", () => {
+    const rangesIntervalsToTest = [
+      { range: PriceRange.OneDay, interval: PriceInterval.OneHour, ticks: 23 },
+      { range: PriceRange.OneWeek, interval: PriceInterval.FourHours, ticks: 42 },
+      { range: PriceRange.OneMonth, interval: PriceInterval.OneDay, ticks: 30 },
+      { range: PriceRange.OneYear, interval: PriceInterval.OneWeek, ticks: 52 },
+    ]
+    rangesIntervalsToTest.forEach(({ range, interval, ticks }) => {
+      it(`returns ${range} history with ${interval} interval`, async () => {
+        const prices = await PriceService().listHistory({ range, interval })
```
nicolasburtey, line 14: is this mocked?

nicolasburtey (approved):
couple of question but mostly for me to better understand the new architecture rather than real comment on the PR. so approving it.

src/services/price/update-price-history.ts
dolcalmi: this runs only from trigger and will be moved as requested here https://github.com/GaloyMoney/galoy/issues/683

test/integration/services/price.spec.ts
dolcalmi: if we mock db yes, but in this case is to validate if the query to mongodb is returning the correct interval

test/integration/services/price.spec.ts
dolcalmi: only in unit test history is mocked (but that is only for test cache behavior) but here is query directly from db (data is generated in `01-price-history.spec.ts`)

nicolasburtey (approved):

bodymindarts (approved):

PR #687, opened 2021-11-01
Add initial doc to go with postman collection

sebastienverreault:

PR #728, opened 2021-11-08
Add tutorial to spectaql config

krtk6160:


nicolasburtey (changes requested):
the comment in https://github.com/GaloyMoney/galoy/pull/687 has not been addressed. also we might want to copy it programmatically instead of a copy the content manually (which I assume is what you have done?) otherwise it will be difficult to work on updating this

PR #729, opened 2021-11-08
Add target confirmations to onChainPay

dolcalmi:
The main purpose of this PR is to add targetConfirmations for OnChainPaymentSend in V2

this is not a use case refactor but "fix" consistency between OnChainPaymentSend mutation and OnChainTxFee query


samerbuna (approved):
LGTM

vindard (approved):
Having different configurable non-zero target confs still falls under a single use-case I believe, so this look ok to me

PR #730, opened 2021-11-09
refactoring login+(captcha)requestPhoneCode from core to apps

nicolasburtey:
doing some small improvment:
- instead of using exit(errorCode), using throw when there is a missing config. I had issue some month ago where there was a config missing, and I could not see anything in the log because nodejs was exiting early as it encounter the `exit()` method. The way to uncover the issue was to detect that the error code was 99 or 98 which is a custom error we had defined. instead, going to a throw would add something in the log, that would make it more apparent what is the issue, as looking at the exit code is typically not the first thing one will be looking at when there is a crash in a pod.
- importing subfunction in lodash instead of the whole lodash package. there is sometimes vulnerabilities discovered in lodash. importing only the few packages we are using will limit the noise we get around that. also will make our Docker image (marginally) smaller.


nicolasburtey:
integration tests are still failing but would appreciate some early feedback on the PR @bodymindarts @vindard @dolcalmi

src/app/users/login.ts
```diff
@@ -0,0 +1,131 @@
...
+  return limitOk
+}
+
+// TODO is this a candidate for domain
+const isTestAccount = ({ code, phone }: { code: PhoneCode; phone: PhoneNumber }) =>
+  yamlConfig.test_accounts.findIndex((item) => item.phone === phone) !== -1 &&
+  yamlConfig.test_accounts.filter((item) => item.phone === phone)[0].code.toString() ===
+    code.toString()
+
+export const login = async ({
```
bodymindarts, line 63: Can we put `export`ed functions at the top of the file?

bodymindarts (commented):
Its looking okay to me - but its a very big PR.
Would it be possible to extract / break it up a little and introduce some of the things incrementally.
I see at least 3 concerns you are tackling here:
- new rate limiter capabilities
- new structure for a lot of config stuff
- bring the `login` use case into `app` layer

src/app/users/login.ts
nicolasburtey: yes. is it a convention we are using?

src/app/users/login.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  const userRepo = UsersRepository()
+  let user: RepositoryError | User
+
+  user = await userRepo.findByPhone(phone)
+  if (user instanceof Error) return user
+
+  if (user) {
+    subLogger.info({ phone }, "user login")
+  } else {
```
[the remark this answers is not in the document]
nicolasburtey: good catch. is there some tool that could automatically catch this?

src/domain/users/index.types.d.ts
```diff
@@ -39,14 +54,23 @@ type User = {
   readonly defaultAccountId: AccountId
   readonly deviceTokens: DeviceToken[]
   readonly createdAt: Date
-  phone: PhoneNumber
+  readonly phone: PhoneNumber
+  twilio?: PhoneMetadata
```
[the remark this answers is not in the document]
nicolasburtey: > AFAICT it is only ever undefined during initial construction.

the info could be missing if twilio is offline, or if the info doesn't not exist (fake phone number for test account) or if the info is not available from twilio for some reasons.

agree that is could be readonly

src/graphql/error-map.ts
```diff
@@ -13,7 +13,7 @@ import {
 } from "@graphql/error"
 import { baseLogger } from "@services/logger"
 
-const assertUnreachable = (x: never): never => {
+const assertUnreachable = (x: string): never => {
```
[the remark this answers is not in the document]
nicolasburtey: at least there is an issue with main, and casting to string is silencing the issue. 
not sure to understand how to fix the issue (didn't really looked into it)

src/services/lnd/auth.types.d.ts
```diff
@@ -7,10 +7,12 @@ type LndParams = {
   port: string | number
   type: nodeType[]
   pubkey: string
+  priority: number
```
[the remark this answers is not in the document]
nicolasburtey: this field is defined in the yaml to determine which node should have the priority when several are running simulatenously

src/services/mongoose/users.ts
```diff
@@ -46,6 +47,38 @@ export const UsersRepository = (): IUsersRepository => {
     }
   }
 
+  const findByPhone = async (phone: PhoneNumber): Promise<User | RepositoryError> => {
+    try {
+      const result = await User.findOne({ phone })
+      if (!result) {
+        return new CouldNotFindUserFromPhoneError(phone)
```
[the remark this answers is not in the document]
nicolasburtey: just following the pattern that we already have here: src/domain/errors.ts

src/app/users/login.ts
bodymindarts: I always put the publicly exported parts of a file / module / class at the top no matter what language I'm working in. I think the others have also been doing this - at least for the new code. Though we haven't made it an explicit rule I consider it good practice.

bodymindarts:
Can you clean up the history? Its hard to read:
```
$ gl
*   123215ae (HEAD -> improve_config, origin/improve_config) Merge branch 'improve_config' of https://github.com/GaloyMoney/galoy into improve_config
|\
| *   60bbdc37 Merge branch 'improve_config' of https://github.com/GaloyMoney/galoy into improve_config
| |\
* | \   6ae86b4b Merge branch 'improve_config' of https://github.com/GaloyMoney/galoy into improve_config
|\ \ \
| |/ /
|/| /
| |/
| * 1d20d285 work on fixing initialization issue
* | 2eb21fce work on fixing initialization issue
```

nicolasburtey:
> Can you clean up the history? Its hard to read:
> 
> ```
> $ gl
> *   123215ae (HEAD -> improve_config, origin/improve_config) Merge branch 'improve_config' of https://github.com/GaloyMoney/galoy into improve_config
> |\
> | *   60bbdc37 Merge branch 'improve_config' of https://github.com/GaloyMoney/galoy into improve_config
> | |\
> * | \   6ae86b4b Merge branch 'improve_config' of https://github.com/GaloyMoney/galoy into improve_config
> |\ \ \
> | |/ /
> |/| /
> | |/
> | * 1d20d285 work on fixing initialization issue
> * | 2eb21fce work on fixing initialization issue
> ```

what the right way to clean it up? trying `rebase -i` and this doesn't seem to work

nicolasburtey:
> > Can you clean up the history? Its hard to read:
> > ```
> > $ gl
> > *   123215ae (HEAD -> improve_config, origin/improve_config) Merge branch 'improve_config' of https://github.com/GaloyMoney/galoy into improve_config
> > |\
> > | *   60bbdc37 Merge branch 'improve_config' of https://github.com/GaloyMoney/galoy into improve_config
> > | |\
> > * | \   6ae86b4b Merge branch 'improve_config' of https://github.com/GaloyMoney/galoy into improve_config
> > |\ \ \
> > | |/ /
> > |/| /
> > | |/
> > | * 1d20d285 work on fixing initialization issue
> > * | 2eb21fce work on fixing initialization issue
> > ```
> 
> what the right way to clean it up? trying `rebase -i` and this doesn't seem to work

ok followed this tutorial: https://blog.oddbit.com/post/2019-06-17-avoid-rebase-hell-squashing-wi/

package.json
```diff
@@ -77,6 +77,7 @@
     "lightning": "^4.10.6",
     "ln-service": "^52.8.0",
     "lodash": "^4.17.21",
+    "lodash.merge": "^4.6.2",
```
dolcalmi, line 80: why add an additional package if we are not removing the main lodash package?

nicolasburtey: because we need to switch also other lodash subpackage. for now this PR removed the `{merge} from lodash` but there are other to be removed in later PR.

src/graphql/error-map.ts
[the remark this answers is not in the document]
vindard: Ok that's exactly how that `never` type is supposed to work. The error is a signal that there are Error types we've defined that are missing from the switch (see image for which ones from linter).

To fix, simply add case lines for the new errors, and where applicable can add human readable messages and return in specific graphql errors to be picked up at outer layers.

Screenshot from 2021-11-22 11-00-27

src/app/users/login.ts
vindard: Taking a note of this as well on my end

vindard: Most other files seem to have imports after exports and in some cases an `export default <thing>` right at the end as the last line. Not sure what the usual JS convention is here or if there are any gotchas with putting `export` at the top? cc @samerbuna @dolcalmi

src/services/mongoose/users.ts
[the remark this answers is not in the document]
vindard: Needed for the right message/graphql-error in the exhaustiveness swich

src/app/users/login.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  rewardFailedLoginAttemptPerIpLimits(ip)
+
+  const userRepo = UsersRepository()
+  let user: RepositoryError | User = await userRepo.findByPhone(phone)
+
+  if (user instanceof CouldNotFindUserFromPhoneError) {
+    subLogger.info({ phone }, "new user signup")
+
+    const userRaw = { phone } as NewUserInfo
```
vindard, line 60: Not sure if we'd prefer `const userRaw: NewUserInfo = { phone }` here instead to more tightly catch potential inconsistencies

src/services/phone-provider.ts
```diff
@@ -1,38 +1,36 @@
...
-  const apiSecret = process.env.TWILIO_API_SECRET
-
-  const client = twilio(apiKey, apiSecret, { accountSid })
-  return client
-}
+const client = twilio(getTwilioConfig().apiKey, getTwilioConfig().apiSecret, {
+  accountSid: getTwilioConfig().accountSid,
+})
 
 export const sendTwilioText = async ({ body, to, logger }) => {
```
vindard, line 11: These set of abstractions may be a little off here.

The idea here so far has been to define generic interfaces (e.g. `ILightningService`) and generic mehods within then (e.g. `lookupInvoice`) and then to implement these interfaces for specific services (e.g. `lndService` implements the `ILightningService` interface).

If we followed that pattern here, we'd probably more end up with something like an `IPhoneProviderService` interface with generic `sendText` and `getCarrier` methods and defined types, and a `twilioService` that implements this interface.

vindard (changes requested):
Adding a few preliminary comments here

src/config/app.ts
```diff
@@ -182,6 +210,31 @@ export const getSpecterWalletConfig = (): SpecterWalletConfig => {
   }
 }
 
+type TwilioConfig = {
```
vindard, line 213: Can maybe place this in `src/core/types.d.ts` to be consistent with all types being in declaration files and injected into global environment

vindard (changes requested):

test/integration/graphql/admin/queries.spec.ts
```diff
@@ -3,6 +3,8 @@ import { graphql } from "graphql"
 import { User } from "@services/mongoose/schema"
 import { gqlAdminSchema } from "@graphql/admin"
 
+jest.mock("@services/phone-provider", () => require("test/mocks/phone-provider"))
```
dolcalmi, line 6: why do we need to add this mock here?

test/integration/graphql/admin/mutations.spec.ts
```diff
@@ -3,6 +3,8 @@ import { graphql } from "graphql"
 import { User } from "@services/mongoose/schema"
 import { gqlAdminSchema } from "@graphql/admin"
 
+jest.mock("@services/phone-provider", () => require("test/mocks/phone-provider"))
```
dolcalmi, line 6: why do we need mock here?

test/integration/01-setup/00-create-mandatory-users.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { getAndCreateUserWallet } from "test/helpers"
+
+jest.mock("@services/phone-provider", () => require("test/mocks/phone-provider"))
```
dolcalmi, line 3: why do we need this mock here?

src/services/phone-provider.ts
```diff
@@ -1,38 +1,36 @@
...
   return true
 }
 
 export const getCarrier = async (phone: string) => {
-  const result = await getTwilioClient()
-    .lookups.phoneNumbers(phone)
-    .fetch({ type: ["carrier"] })
-  baseLogger.info({ result }, "result carrier info")
-  return result
+  try {
```
dolcalmi, line 29: this service/methods should be refactored as Arvin suggested. We can even go further and make it part of notification service (open to discussion)

src/services/mongoose/phone-code.ts
```diff
@@ -0,0 +1,38 @@
...
+  }: {
+    phone: PhoneNumber
+    code: PhoneCode
+  }): Promise<true | RepositoryError> => {
+    try {
+      const phoneCode = await PhoneCode.findOne({
+        phone,
+        code,
+        created_at: {
+          $gte: moment().subtract(VALIDITY_TIME_CODE, "seconds"),
```
dolcalmi, line 23: please use only date (as far as I remember we are trying to remove moment package)

src/services/mongodb/index.ts
```diff
@@ -45,8 +43,7 @@ export const setupMongoConnection = async () => {
     })
   } catch (err) {
     baseLogger.fatal({ err, user, address, db }, `error connecting to mongodb`)
-    await sleep(100)
-    exit(99)
+    throw err
```
dolcalmi, line 46: @bodymindarts @krtk6160 dont we need the exit code to handle CI errors??

src/services/jwt.ts
```diff
@@ -21,5 +18,5 @@ export const createToken = ({ uid, network }) => {
     //
     // see: https://www.theregister.com/2018/04/24/myetherwallet_dns_hijack/
     algorithm: "HS256",
-  })
+  }) as JWTToken
```
dolcalmi, line 21: this is fine but we are using typescript implicit type to discover the return for createToken. Please add it to the createToken signature. 

Also not sure if this should be a service (is not querying external services, db, etc). Maybe would be authentication service but only when we add the new auth (in a different pod) 

@vindard @bodymindarts what do you think?

src/graphql/root/mutation/user-login.ts
```diff
@@ -35,10 +35,9 @@ const UserLoginMutation = GT.Field({
       }
     }
 
-    // TODO: Make this through a new app use-case
     const authToken = await login({ phone, code, logger, ip })
 
-    if (!authToken) {
+    if (authToken instanceof Error) {
       return { errors: [{ message: "Invalid request" }] }
     }
```
dolcalmi, line 42:
```suggestion
    if (authToken instanceof Error) {
      const appErr = mapError(authToken)
      return { errors: [{ message: appErr.message }] }
    }
```

src/domain/users/index.types.d.ts
```diff
@@ -39,14 +54,22 @@ type User = {
...
+type NewUserInfo = {
+  phone: PhoneNumber
+  twilio?: PhoneMetadata
+}
+
 interface IUsersRepository {
   findById(userId: UserId): Promise<User | RepositoryError>
   findByUsername(username: Username): Promise<User | RepositoryError>
+  findByPhone(phone: PhoneNumber): Promise<User | RepositoryError>
+  persistNew({ phone, twilio }: NewUserInfo): Promise<User | RepositoryError>
```
dolcalmi, line 72: please rename twilio param, repo implementation should take care of the legacy name

src/domain/users/index.types.d.ts
```diff
@@ -30,6 +33,18 @@ type UserQuizQuestion = {
   completed: boolean
 }
 
+type PhoneMetadata = {
+  // from twilio
+  carrier: {
+    error_code: string // check this is the right syntax
+    mobile_country_code: string
+    mobile_network_code: string
```
dolcalmi, line 41: in domain layer we should keep our variable name standard, please use camel case and do the parse in repo implementation

src/core/types.d.ts
```diff
@@ -54,6 +54,14 @@ type TwoFAConfig = {
   threshold: number
 }
 
+type TestAccounts = {
+  phone: PhoneNumber
+  code: PhoneCode
+  username: Username | undefined
+  role: string | undefined // FIXME
+  currencies // FIXME
```
dolcalmi, line 62: what need to be fixed?

src/core/text.ts
```diff
@@ -1,25 +1,11 @@
+import { getGaloyInstanceName, getGaloySMSProvider, yamlConfig } from "@config/app"
```
dolcalmi, line 1: isn't requestPhoneCode part of the login? shouldn't be part of this PR?

src/config/app.ts
dolcalmi: we should not add more code to the core folder. Maybe a new type file in config folder?

src/config/app.ts
```diff
@@ -1,23 +1,39 @@
...
+}
 
 import { baseLogger } from "@services/logger"
 
 const defaultContent = fs.readFileSync("./default.yaml", "utf8")
 export const defaultConfig = yaml.load(defaultContent)
 
-export const JWT_SECRET = process.env.JWT_SECRET
-export const BTC_NETWORK = process.env.NETWORK as BtcNetwork
+const jwt_secret = process.env.JWT_SECRET
```
dolcalmi, line 14: if this is only for local validation please use camel case

src/config/app.ts
```diff
@@ -1,23 +1,39 @@
...
+if (!jwt_secret) {
+  throw new ConfigError("missing JWT_SECRET")
+}
+
+export const JWT_SECRET = jwt_secret
+
+const btc_network = process.env.NETWORK
```
dolcalmi, line 21: please use camel case

src/app/users/login.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  // TODO:
+  // add fibonachi on failed login
+  // https://github.com/animir/node-rate-limiter-flexible/wiki/Overall-example#dynamic-block-duration
+
+  const validCode = await isCodeValid({ phone, code })
+  if (validCode instanceof Error) return validCode
+
+  // reseting the limiter for this phone
+  // limiterLoginAttempt.delete(phone) // no need to await the promise
```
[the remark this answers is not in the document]
dolcalmi: if it is not necessary please remove the comments

dolcalmi (changes requested):

src/app/users/login.ts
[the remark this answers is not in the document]
nicolasburtey: it will put a limit on how frequently someone can log in. previously it was possible to login infinitely on a single phone number. but this reset is actually harmful after thinking more about it. so I'll delete.

src/services/phone-provider.ts
nicolasburtey: I have refactored the file to make it a class instead of function at the root of the file

Is this not an somewhat non necessary overhead to do an interface here?

Unless I'm mistaken, Typescript will deduce the type from the file anyway. and the interface is really only useful if we plan to have several services implementing the same interface.

test/integration/servers/graphql-old-server.spec.ts
```diff
@@ -60,7 +65,7 @@ describe("graphql", () => {
 
     // exhaust the limiter
     const requestPhoneCodeLimits = getRequestPhoneCodeLimits()
-    for (let i = 0; i < requestPhoneCodeLimits.points; i++) {
+    for (let i = 0; i < requestPhoneCodeLimits.points - 1; i++) {
```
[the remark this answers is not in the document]
nicolasburtey: before `requestPhoneCodeLimits` was checked after the main associated query. now it's being done before.

test/integration/graphql/admin/queries.spec.ts
nicolasburtey: no longer necessary after moving text.ts to a class model

test/integration/graphql/admin/mutations.spec.ts
nicolasburtey: no longer necessary after moving text.ts to a class model

src/core/text.ts
nicolasburtey: I planned to do this for the next PR

src/services/mongoose/phone-code.ts
nicolasburtey: just a copy paste here. I recommend doing this in a new PR.

src/services/mongodb/index.ts
nicolasburtey: I don't think so. I was the one introducing this logic. but it's not very practical. node would still exit with 1 if an error is being thrown AFAIK

src/services/phone-provider.ts
bodymindarts: Its just about separating declaration (which is part of domain) from implementation.
It is useful for testing, clean boundaries, and potentially interchangeability of implementation. We have after all had 2 implementations in the past and now moved back to 1. Its not a lot of overhead and it keeps the structure we have introduced consistent.

nicolasburtey: > Its just about separating declaration (which is part of domain) from implementation.

why should declaration part of `domain` but implementation part of `app`?

src/core/types.d.ts
nicolasburtey: the type is an array of currencies, should be `any`. but because this is likely to be removed with the USD integration I didn't bother defining the type properly.

test/integration/servers/graphql-old-server.spec.ts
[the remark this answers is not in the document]
nicolasburtey: actually it's still ok to have +1. I'll revert this.

src/domain/users/index.types.d.ts
nicolasburtey: what would you rename it to?

dolcalmi: uhmm the type is `PhoneMetadata` so I think is fine to use `metadata` or `phoneMetadata`

src/services/phone-provider.ts
bodymindarts: implementation is part of `services` not `app`. 
Domain is unit-testable and app could be to if we did dependency injection (haven't so far).
> It is useful for testing, clean boundaries, and potentially interchangeability of implementation. We have after all had 2 implementations in the past and now moved back to 1. Its not a lot of overhead and it keeps the structure we have introduced consistent.

src/domain/users/index.types.d.ts
nicolasburtey: this require a migration to do properly. so will do in a later PR.

src/app/users/login.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    const carrierInfo = await TwilioClient().getCarrier(phone)
+    if (carrierInfo instanceof Error) {
+      // non fatal error
+      subLogger.warn({ phone }, "impossible to fetch carrier")
+    } else {
+      userRaw.phoneMetadata = carrierInfo
+    }
+
+    user = await userRepo.persistNew(userRaw)
```
vindard, line 67: Small nit, this could probably be `const persistedUser = ...` to reflect the difference between this variable and the `user` we fetch from the `findOne` higher up to help with keeping these concepts separate.

src/services/geetest.types.d.ts
```diff
@@ -1,3 +1,5 @@
+import { GeetestError } from "@domain/geetest/error"
```
vindard, line 1: Imports should be done like this so that we don't lose the implicit imports from this file.

We should also probably case this as `GeeTest..` to be consistent with other types.

```suggestion
type GeeTestError = import("@domain/geetest/error").GeeTestError
```

src/services/geetest.ts
```diff
@@ -3,8 +3,14 @@
...
 // doing this: "If the storage space is not sufficient: Send request to check bypass status before starting the verfication process."
 
+import {
+  GeetestError,
+  GeetestUserFailToPassError,
+  UnknownGeetestError,
+} from "@domain/geetest/error"
 import axios from "axios"
 import GeetestLib from "gt3-server-node-express-sdk/sdk/geetest_lib" // galoy fork
+import { GeeTestType } from "./geetest.types"
```
vindard, line 13: Can be deleted after this
```suggestion
```

src/app/users/request-phone-code.ts
```diff
@@ -0,0 +1,138 @@
...
+import { UnknownPhoneProviderServiceError } from "@domain/errors"
+import { GeetestError } from "@domain/geetest/error"
+import { RateLimitPrefix } from "@domain/rate-limit"
+import {
+  RateLimiterExceededError,
+  UserPhoneCodeAttemptIpRateLimiterExceededError,
+  UserPhoneCodeAttemptMinIntervalLimiterExceededError,
+  UserPhoneCodeAttemptPhoneRateLimiterExceededError,
+} from "@domain/rate-limit/errors"
+import { GeeTestType } from "@services/geetest.types"
```
vindard, line 17: Can be deleted after this
```suggestion
```

src/core/captcha-challenge-create.ts
```diff
@@ -1,3 +1,5 @@
+import { GeeTestType } from "@services/geetest.types"
```
vindard, line 1: Can be deleted after this
```suggestion
```

src/services/geetest.ts
```diff
@@ -46,17 +52,28 @@ const GeeTest = (config): GeeTestType => {
...
+        result = await gtLib.successValidate(challenge, validate, seccode, params)
+      } else {
+        result = gtLib.failValidate(challenge, validate, seccode)
+      }
+      if (result.status !== 1) {
+        return new GeetestUserFailToPassError()
+      }
+      return true
+    } catch (err) {
+      return new UnknownGeetestError()
```
vindard, line 75: Small nit, we try to do this where possible to get visibility on the unknown error later on.
```suggestion
      return new UnknownGeetestError(err)
```

src/domain/errors.ts
```diff
@@ -11,13 +11,17 @@ export class UnknownRepositoryError extends RepositoryError {}
 export class PersistError extends RepositoryError {}
 export class DuplicateError extends RepositoryError {}
 
+export class UnknownPhoneProviderServiceError extends RepositoryError {}
```
vindard, line 14: This should be in `src/domain/phone-provider/errors.ts` and extend `PhoneProviderServiceError`

src/domain/phone-provider/index.types.d.ts
```diff
@@ -0,0 +1,31 @@
+import { UnknownPhoneProviderServiceError } from "@domain/errors"
+import { PhoneNumberInstance } from "twilio/lib/rest/lookups/v1/phoneNumber"
```
vindard, line 2: We shouldn't have imports like this in declaration files (see here)
```suggestion
type PhoneProviderServiceError = import("./errors").PhoneProviderServiceError
type UnknownPhoneProviderServiceError = import("@domain/errors").UnknownPhoneProviderServiceError
```

src/domain/phone-provider/index.types.d.ts
```diff
@@ -0,0 +1,31 @@
...
+
+type IPInfo = {
+  provider: string
+  country: string
+  region: string
+  city: string
+  type: string
+  status: string
+}
+
```
vindard, line 17: These were already declared in `src/domain/ipfetcher/index.types.d.ts`
```suggestion
```

src/services/twilio.ts
```diff
@@ -0,0 +1,54 @@
+import { getTwilioConfig } from "@config/app"
+import { UnknownPhoneProviderServiceError } from "@domain/errors"
+import { IPhoneProviderService } from "@domain/phone-provider/index.types"
```
vindard, line 3: Can be removed after this
```suggestion
```

src/graphql/root/mutation/captcha-request-auth-code.ts
```diff
@@ -33,27 +33,27 @@ const CaptchaRequestAuthCodeMutation = GT.Field({
...
+      geetest,
+      geetestChallenge,
+      geetestValidate,
+      geetestSeccode,
+      logger,
+      ip,
+    })
+
+    if (result instanceof Error) {
       return {
```
vindard, line 47: This is a little trick we've been doing in resolvers to convert domain errors to graphql errors

```suggestion
      const appErr = mapError(result)
      return {
```

src/graphql/root/mutation/captcha-request-auth-code.ts
```diff
@@ -33,27 +33,27 @@ const CaptchaRequestAuthCodeMutation = GT.Field({
...
-        errors: [{ message: err.message }],
+        errors: [{ message: result.message }],
```
vindard, line 48:
```suggestion
        errors: [{ message: appErr.message }],
```

src/graphql/root/mutation/captcha-request-auth-code.ts
```diff
@@ -1,8 +1,8 @@
 import { GT } from "@graphql/index"
-import verifyCaptchaAndReturnOTP from "@core/auth-code-request"
 
 import Phone from "@graphql/types/scalar/phone"
 import SuccessPayload from "@graphql/types/payload/success-payload"
+import { requestPhoneCodeWithCaptcha } from "@app/users/request-phone-code"
```
vindard, line 5:
```suggestion
import { requestPhoneCodeWithCaptcha } from "@app/users/request-phone-code"
import { mapError } from "@graphql/error-map"
```

vindard (changes requested):
The new `IPhoneServiceProvider` abstraction looks really good! I found a few other remaining inconsistencies though that I tried to flag here.

src/app/users/login.ts
nicolasburtey: they're the same concept? first one is the case the user exist, second is the case the user didn't exist because, but they both return `persistedUser`

src/services/geetest.types.d.ts
nicolasburtey: > Imports should be done like this so that we don't lose the implicit imports from this file.

I don't really understand this. where can I learn more about it?

nicolasburtey: I think I get some of it. if it's in a `d.ts` then type is being automatically added, but only if this is a `type` not an `import`?

src/app/users/login.ts
dolcalmi: imports should be at the beginning and after that we can have non-exported/exported functions, I prefer exported functions at the end but maybe is because I have done "some" js projects/libraries. We should define the rule in the next code review.

src/domain/users/index.types.d.ts
```diff
@@ -39,14 +66,22 @@ type User = {
   readonly defaultAccountId: AccountId
   readonly deviceTokens: DeviceToken[]
   readonly createdAt: Date
-  phone: PhoneNumber
+  readonly phone: PhoneNumber
+  readonly twilio: PhoneMetadata | null
```
dolcalmi, line 70: same here, if you do the parsing in the service you can use here `phoneMetadata` as property name

src/graphql/root/mutation/user-request-auth-code.ts
```diff
@@ -29,10 +29,9 @@ const UserRequestAuthCodeMutation = GT.Field({
       return { errors: [{ message: phone.message }] }
     }
 
-    // TODO: Make this through a new app use-case
     const status = await requestPhoneCode({ phone, logger, ip })
 
-    if (!status) {
+    if (status instanceof Error) {
```
dolcalmi, line 34: now we can use map error to show the proper error message `"Could not complete the operation successfully"` should be a message from error mapping

src/services/mongoose/phone-code.types.d.ts
```diff
@@ -0,0 +1,11 @@
+interface IPhoneCodesRepository {
```
dolcalmi, line 1: this is used in app layer, should be defined in domain layer not in service layer (service layer only should have the implementation)

src/services/rate-limit/index.ts
```diff
@@ -33,5 +33,14 @@ export const RedisRateLimitService = ({
     }
   }
 
-  return { consume, reset }
+  const reward = async (key) => {
```
dolcalmi, line 36: although is not necessary, we should try to keep the same signature defined in domain layer (readability)

src/services/twilio.ts
```diff
@@ -0,0 +1,53 @@
...
+import { UnknownPhoneProviderServiceError } from "@domain/phone-provider"
+import { baseLogger } from "@services/logger"
+import twilio from "twilio"
+
+export const TwilioClient = (): IPhoneProviderService => {
+  const client = twilio(getTwilioConfig().apiKey, getTwilioConfig().apiSecret, {
+    accountSid: getTwilioConfig().accountSid,
+  })
+
+  const sendText = async ({ body, to, logger }) => {
```
dolcalmi, line 11: same comment about signature, we should try to keep the same signature

src/services/twilio.ts
```diff
@@ -0,0 +1,53 @@
...
+    } catch (err) {
+      logger.error({ err }, "impossible to send text")
+      return new UnknownPhoneProviderServiceError(err)
+    }
+
+    logger.info({ to }, "sent text successfully")
+    return true
+  }
+
+  const getCarrier = async (phone: PhoneNumber) => {
```
dolcalmi, line 28: same here (signature)

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
```diff
@@ -8,24 +8,22 @@ import {
...
 let bitcoindOutside
 
 beforeAll(async () => {
   // load funder wallet before use it
-  await getUserWallet(4)
+  await getAndCreateUserWallet(4)
 
   // "bankowner" user
-  await getUserWallet(14)
+  await getAndCreateUserWallet(14)
```
dolcalmi, line 26: now that we have `00-create-mandatory-users.spec.ts ` is this necessary?

dolcalmi (dismissed):

test/integration/01-setup/01-funds-bitcoind-lnds.spec.ts
nicolasburtey: seems `00-create-mandatory-users.spec.ts` is no longer needed. was probably a transient issue while I was working on this PR. deleting

src/app/users/login.ts
```diff
@@ -0,0 +1,126 @@
...
+
+const rewardFailedLoginAttemptPerIpLimits = async (
+  ip: IpAddress,
+): Promise<true | RateLimiterExceededError> => {
+  const limiter = RedisRateLimitService({
+    keyPrefix: RateLimitPrefix.failedAttemptLoginIp,
+    limitOptions: getFailedLoginAttemptPerIpLimits(),
+  })
+  const limitOk = await limiter.reward(ip)
+  if (limitOk instanceof Error) return new UnknownRepositoryError()
```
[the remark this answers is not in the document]
nicolasburtey: this is a reward here, so this should not fail. the idea is to reward login from succesful login such that if many people try to login from the same network (shared wifi), the rate limiter should not kick in too early

src/config/app.ts
```diff
@@ -61,28 +77,37 @@ export const USER_ACTIVENESS_MONTHLY_VOLUME_THRESHOLD =
...
+        throw new ConfigError(`lnd params missing for: ${input.name}${key}`)
+      }
+    })
+  })
+
   return config.map((input) => ({
-    cert: process.env[`${input.name}_TLS`] || exit(98),
-    macaroon: process.env[`${input.name}_MACAROON`] || exit(98),
-    node: process.env[`${input.name}_DNS`] || exit(98),
+    cert: process.env[`${input.name}_TLS`],
```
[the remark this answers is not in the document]
nicolasburtey: `make trigger` locally works at least? did you have something else in mind?

src/domain/geetest/error.ts
```diff
@@ -0,0 +1,6 @@
+export class GeetestError extends Error {
+  name = this.constructor.name
+}
+
+export class GeetestUserFailToPassError extends GeetestError {}
```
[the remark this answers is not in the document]
nicolasburtey: is this what you have in mind as a fix: https://github.com/GaloyMoney/galoy/pull/730/commits/bb8d498d2c6d27159a5071c0abc7e673d199c8a5 ?

nicolasburtey:
> no, this does not require migration, this is at domain definition. You should do the parsing in service implementation (persistNew and userFromRaw)

yes but this means we need both a mapping from twilio and mongodb. a better fix would be to just map the value from twilio to a camelCase such that we don't have a map the value to/from mongodb?

src/domain/users/index.types.d.ts
nicolasburtey: fixed

src/services/mongoose/phone-code.types.d.ts
nicolasburtey: is this what should be done https://github.com/GaloyMoney/galoy/pull/730/commits/921a61184dff4705b376c4c280a6d103713c6675?

src/services/rate-limit/index.ts
nicolasburtey: is that the fix you're looking for? https://github.com/GaloyMoney/galoy/pull/730/commits/5c6ffbf8df01bf50483bedf4b802b2d07a5fdb9c or is it something else?

src/services/twilio.ts
nicolasburtey: the signature is the same?

src/services/twilio.ts
nicolasburtey: is this the fix you're looking for? https://github.com/GaloyMoney/galoy/pull/730/commits/bfce0ba30bcf717d31ef4744c9fe858b16d1a493

src/app/users/login.ts
```diff
@@ -0,0 +1,126 @@
...
+    subLogger.info({ phone }, "new user signup")
+
+    const userRaw = { phone } as NewUserInfo
+
+    const carrierInfo = await TwilioClient().getCarrier(phone)
+    if (carrierInfo instanceof Error) {
+      // non fatal error
+      subLogger.warn({ phone }, "impossible to fetch carrier")
+    } else {
+      userRaw.phoneMetadata = carrierInfo
```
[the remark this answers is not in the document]
nicolasburtey: is this the fix you're looking for? https://github.com/GaloyMoney/galoy/pull/730/commits/a5b07e1a778e0c438ec77fe5751f9e43c2ec7440

src/domain/phone-code/index.types.d.ts
```diff
@@ -0,0 +1,11 @@
+interface IPhoneCodesRepository {
+  findRecent: (arg: {
```
bodymindarts, line 2: `find` prefix implied it returns an entity (that is the pattern we've been using so far). Since its returning a boolean perhaps a better name would be `checkExistance` or something like that.

src/domain/users/index.types.d.ts
bodymindarts: This is the whole reason to have the hexagonal architecture... keep the domain / app layers clean regardless of services / repo implementations. Thats what the translation layers (aka adaptors) are for.

src/domain/users/index.types.d.ts
```diff
@@ -30,6 +33,30 @@ type UserQuizQuestion = {
...
+// }
+
+type PhoneMetadata = {
+  // from twilio
+  carrier: {
+    error_code: string // check this is the right syntax
+    mobile_country_code: string
+    mobile_network_code: string
+    name: string
+    type: "landline" | "voip" | "mobile"
```
bodymindarts, line 55: For this kind of type that has only certain allowed values we've been defining a `const Object` and making a type out of it. Look at the `LedgerTransactionType` for example:
[Type](https://github.com/GaloyMoney/galoy/blob/main/src/domain/ledger/index.types.d.ts#L16-L17) [object](https://github.com/GaloyMoney/galoy/blob/main/src/domain/ledger/index.ts#L10-L26). With this method you can then reference the values like `PhoneType.Landline`.

src/domain/users/index.types.d.ts
```diff
@@ -39,14 +66,22 @@ type User = {
...
-  phone: PhoneNumber
+  readonly phone: PhoneNumber
+  readonly phoneMetadata: PhoneMetadata | null
   language: UserLanguage
   twoFA: TwoFAForUser
 }
 
+type NewUserInfo = {
+  phone: PhoneNumber
+  phoneMetadata?: PhoneMetadata
```
bodymindarts, line 77: Should be consistently `| null`. In general we are avoiding `?` arguments as they are typically an indication of multiple use-cases being merged into one. Legitimate reasons for `absence` should be codified via `null`.

bodymindarts (commented):
This PR seems to be getting bigger and bigger. Is there no way to bring in the changes incrementally?

nicolasburtey:
> This PR seems to be getting bigger and bigger. Is there no way to bring in the changes incrementally?

not straightforward. had to squash the commits to remove the multiple parent commits issue.

src/app/users/login.ts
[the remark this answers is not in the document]
dolcalmi: uhmm what I mean is that we shouldn't use `UnknownRepositoryError` at app layer, shoud be an specific method (like you implemented it in `checkFailedLoginAttemptPerPhoneLimits`)

src/config/app.ts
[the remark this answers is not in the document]
dolcalmi: uhmm I thought that if env variables were not setup the process exited but it keeps running so there is no difference (at least for trigger).

src/domain/geetest/error.ts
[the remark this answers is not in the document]
dolcalmi: yes

src/services/rate-limit/index.ts
dolcalmi: what I mean is: if you have this in domain
```js
  consume(key: string): Promise<true | RateLimitServiceError>
  reset(key: string): Promise<true | RateLimitServiceError>
  reward(key: string): Promise<true | RateLimitServiceError>
```

then the service implementation should be something like: 
```js
  const consume = async (key: string): Promise<true | RateLimitServiceError> => {...}
  const reset = async (key: string): Promise<true | RateLimitServiceError> => {...}
  const reward = async (key: string): Promise<true | RateLimitServiceError> => {...}
```

src/services/twilio.ts
dolcalmi: please check this comment https://github.com/GaloyMoney/galoy/pull/730#discussion_r757807499

src/services/twilio.ts
dolcalmi: please check this comment https://github.com/GaloyMoney/galoy/pull/730#discussion_r757807499

src/services/rate-limit/index.ts
nicolasburtey: I don't get it. this is what the code is currently doing?

image

are you suggesting it's needed to have explicit declaration of type versus implicit?

nicolasburtey: assuming your comment is related to this, typescript is inferring the type automatically so I'm not sure why we would want to specify it manually

src/domain/geetest/error.ts
```diff
@@ -0,0 +1,6 @@
+export class GeetestError extends Error {
```
bodymindarts: `Geetest` is an implementation that domain should now know about. I think the component is a `CaptchaProvider`.

src/domain/rate-limit/errors.ts
```diff
@@ -8,3 +8,8 @@ export class UnknownRateLimitServiceError extends RateLimitServiceError {}
 export class RateLimiterExceededError extends RateLimitServiceError {}
 export class InvoiceCreateRateLimiterExceededError extends RateLimiterExceededError {}
 export class OnChainAddressCreateRateLimiterExceededError extends RateLimiterExceededError {}
+export class UserPhoneCodeAttemptMinIntervalLimiterExceededError extends RateLimitError {}
```
bodymindarts: Are all of these errors treated differently by the client?

src/graphql/error-map.ts
```diff
@@ -185,6 +188,13 @@ export const mapError = (error: ApplicationError): CustomApolloError => {
     case "LocalCacheServiceError":
     case "LocalCacheUndefinedError":
     case "UnknownLocalCacheServiceError":
+    case "UserLoginIpRateLimiterExceededError":
```
bodymindarts, line 191: As all these errors are collapsed I guess we don't need the granularity or? Wouldn't a simple `RateLimitExceeded` be enough?

src/services/geetest.types.d.ts
```diff
@@ -7,5 +9,9 @@ type GeeTestRegister = {
 
 type GeeTestType = {
```
bodymindarts: If the captcha logic is being handled in the domain layer this declaration should be in domain (ie `CaptchaProvider`) and `GeeTestService` should implement it. This is the same pattern as with `Twillio` and `IPhoneProviderService`.

src/services/jwt.ts
```diff
@@ -1,14 +1,11 @@
...
 // TODO: replace network by uri of the server
 // the uri will embed the network, ie: graphql.mainnet.server.io
 // and provide more information than just the network
-export const createToken = ({ uid, network }) => {
-  if (!JWT_SECRET) {
-    throw new Error("JWT_SECRET env variable is missing")
-  }
-
+export const createToken = ({ uid }: { uid: UserId }): JwtToken => {
+  const network = BTC_NETWORK
```
bodymindarts, line 8: Config should be injected via constructor.

src/services/mongoose/phone-code.ts
```diff
@@ -0,0 +1,54 @@
...
+import {
+  CouldNotFindPhoneCodeError,
+  RepositoryError,
+  UnknownRepositoryError,
+} from "@domain/errors"
+import moment from "moment"
+import { PhoneCode } from "./schema"
+
+export const PhoneCodesRepository = (): IPhoneCodesRepository => {
+  const findRecent = async ({
```
bodymindarts: The definition of `recent` is business logic and should be hard coded in the repo - it should be part of the domain.

perhaps something like:
```
existsNewerThan({phone: PhoneNumber, code: PhoneCode, age: Seconds})
```

src/services/mongoose/phone-code.ts
```diff
@@ -0,0 +1,54 @@
+import { VALIDITY_TIME_CODE } from "@config/app"
```
bodymindarts, line 1: Values like this are critical business logic and should either be defined in domain - or if they are configurable then injected from the app layer - see below.

src/services/mongoose/phone-code.types.d.ts
bodymindarts: Did the fixing commit get dropped? This file is back in the latest diff.

src/services/mongoose/schema.types.d.ts
```diff
@@ -19,8 +19,14 @@ type OnChainObjectForUser = {
   address: string
 }
 
-type TwillioObjectForUser = {
-  carrier: string
+type TwilioObjectForUser = {
+  carrier: {
+    error_code: string // check this is the right syntax
+    mobile_country_code: string
```
bodymindarts, line 25: snake case.

src/services/mongoose/schema.types.d.ts
```diff
@@ -19,8 +19,14 @@ type OnChainObjectForUser = {
...
+    mobile_network_code: string
+    name: string
+    type: "landline" | "voip" | "mobile"
```
bodymindarts, line 28: type should be const object

src/app/users/request-phone-code.ts
```diff
@@ -0,0 +1,138 @@
...
+import { PhoneCodesRepository } from "@services/mongoose/phone-code"
+import { RedisRateLimitService } from "@services/rate-limit"
+import { TwilioClient } from "@services/twilio"
+import { isTestAccountPhone } from "."
+
+export const requestPhoneCodeWithCaptcha = async ({
+  phone,
+  geetest,
+  geetestChallenge,
```
bodymindarts, line 22: I see `geetest` as a service implementation - not sure these fields should be coupled to it.

bodymindarts (commented):
I think the biggest open issues are:
- there is an `as` cast (simple to fix).
- the whole `Geetest` stuff is not correctly separated into `interface` / `ServiceImplementation` - but it can be fixed in a later PR. Especially since we will be reworking authentication anyhow in the next months and this will probably change a lot then anyhow.
- there are some boundary issues regarding the rate limiter errors.

I think the only thing that is really concerning is the cast. Other than that - while we can certainly keep improving things - I think we're getting close to the point of diminishing returns and should probably clean things up in follow up PRs when we need to touch the code again.

Also some issues were already fixed but showed up again in the diff - perhaps there was a history re-write that dropped some commits.

src/app/users/login.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  await rewardFailedLoginAttemptPerIpLimits(ip)
+
+  const userRepo = UsersRepository()
+  let user: RepositoryError | User = await userRepo.findByPhone(phone)
+
+  if (user instanceof CouldNotFindUserFromPhoneError) {
+    subLogger.info({ phone }, "new user signup")
+
+    const userRaw = { phone } as NewUserInfo
```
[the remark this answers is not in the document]
nicolasburtey: can you explain an example on how this could be dangerous?

src/app/users/login.ts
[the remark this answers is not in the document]
nicolasburtey: @bodymindarts const says the reference to the object can't change. the object itself it mutable.

src/app/users/login.ts
```diff
@@ -0,0 +1,126 @@
...
+const checkFailedLoginAttemptPerIpLimits = async (
+  ip: IpAddress,
+): Promise<true | RateLimiterExceededError> => {
+  const limiter = RedisRateLimitService({
+    keyPrefix: RateLimitPrefix.failedAttemptLoginIp,
+    limitOptions: getFailedLoginAttemptPerIpLimits(),
+  })
+  const limitOk = await limiter.consume(ip)
+  if (limitOk instanceof RateLimiterExceededError)
+    return new UserLoginIpRateLimiterExceededError()
```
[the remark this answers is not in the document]
nicolasburtey: > is the additional granularity needed? Are we treating UserLoginIpRateLimiterExceededError any differently than RateLimiterExceededError in the calling code?

no opinion. just applying existing pattern here.

> As this is inherited from RateLimitError why does it not originate from the service or domain code under rate-limit?

I don't really understand the point. a diff would be useful probably.

src/app/users/request-phone-code.ts
nicolasburtey: I have not idea how to interpret those fields / give them a generic name. I don't think we should assume captcha would easily be interchangeable, they're meant to be a black box and I'm sure they would have different input/outflow if we choose to add another one

src/graphql/error-map.ts
nicolasburtey: I see a point to know which rate limiter are most frequently applied, specially following potential DDOs, therefore it make sense to differentiate them.

src/services/mongoose/schema.types.d.ts
nicolasburtey: will do a migration later

src/app/users/login.ts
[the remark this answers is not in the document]
bodymindarts: It is dangerous because it exposes the possibility for runtime errors where they could be caught at compile time if you don't use the cast.

EG: say the `NewUserInfo` type evolves and we want to add a fiel:
```
type NewUserInfo
{
  phone: string
  anotherField: number
}
```

now you pass the new info to the repository:
```
const newUser = { phone } as NewUserInfo
repo.persistNew(newUser)
```
And the repo function is defined as:
```
persistNew = ({phone, anotherField} :NewUserInfo) => {
 anotherField += 5
}
```

This is just a simple example but its highlighting that because of the cast this will compile - and not only that it will ALWAYS compile even if the definition of `NewUserInfo` changes.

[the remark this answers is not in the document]
nicolasburtey: ok makes sense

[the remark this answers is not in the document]
nicolasburtey: fixed

bodymindarts (approved):

PR #742, opened 2021-11-12
Introduce a Casbin as authorization implementation

bodymindarts:
This introduces the Casbin library for a maintainable way to add role based permissions to resources. Needed for the business account use case but already useful now as we already have a need for some permissions checks.

IMO authorization must be part of the domain and not handled via middleware. This PR is a POC of how I think it should be introduced. Tests are failing but would appreciate some feedback.


src/app/wallets/create-on-chain-address.ts
```diff
@@ -31,9 +33,23 @@ export const createOnChainAddress = async (
...
+export const createOnChainAddressByWalletPublicId = async ({
+  authorizationService,
+  userId,
+  walletPublicId,
+}: {
+  authorizationService: IAuthorizationService
+  userId: UserId
+  walletPublicId: WalletPublicId
+}): Promise<OnChainAddress | ApplicationError> => {
+  const authResult = await authorizationService.checkPermission({
```
nicolasburtey, line 45: shound't this be done in a middleware?

src/config/app.ts
```diff
@@ -200,6 +200,19 @@ export const getIpConfig = (config = yamlConfig): IpConfig => ({
...
+  const user = process.env.MONGODB_USER ?? "testGaloy"
+  const password = process.env.MONGODB_PASSWORD
+  const address = process.env.MONGODB_ADDRESS ?? "mongodb"
+  const db = process.env.MONGODB_DATABASE ?? "galoy"
+
+  return {
+    user,
+    address,
+    db,
+    path: `mongodb://${user}:${password}@${address}/${db}`,
```
nicolasburtey, line 213: this code is already present in src/services/mongodb/index.ts

src/domain/authorization/primitives.ts
```diff
@@ -0,0 +1,15 @@
+export const Role = {
+  AccountOwner: "account.owner",
+} as const
+
+export const Permission = {
+  WalletView: "wallet:view",
+  WalletOnChainAddressCreate: "wallet:onChainAddressCreate",
```
nicolasburtey, line 7: in my mind I would have use the pattern `onchain:create` or `onchain:send`

src/services/casbin/casbin.conf
```diff
@@ -0,0 +1,14 @@
+[request_definition]
```
nicolasburtey, line 1: what is the difference between what settings will be stored here versus in mongodb?

nicolasburtey:
with casbin we are not adding any pods to the cluster?

nicolasburtey:
tests are failing also

src/app/wallets/create-on-chain-address.ts
bodymindarts: Authorization can be middleware but when it becomes very granular / complex its better to have it as part of the domain layer.
With the use cases we want to target (merchant / business accounts with RBAC). I think we should just introduce it into the right layer from the beginning - even though right now it could also be a middleware.

src/domain/authorization/primitives.ts
bodymindarts: Should definitely preface it with the resource but good idea for improvement.
Perhaps:
```
wallet:onchain:createAddress
wallet:onchain:send
```

src/services/casbin/casbin.conf
bodymindarts: MongoDB stores the individual policies - this file defines the logic of how the policies are interpreted. Its like a template for all authorization logic.

Casbin is very flexible and supports up to 6 fields per policy that can be used to determine authorization. Also you can define multiple rules of how policies apply so you can represent things like inheritance etc.

bodymindarts:
> with casbin we are not adding any pods to the cluster?

No casbin only requires a database - decisions are made locally. Only requirement is that all pods load the same `casbin.conf` (to interpret the policies in the same way).

Could also use this as an opportunity to introduce Postgres and store casbin policies there as the data is completely separate from the rest of the persistence stuff.

> tests are failing also

Correct as I mentioned `Tests are failing but would appreciate some feedback.`.
Thanks for commenting - starting this discussion is exactly why I opened the PR with failing tests. Want to gain general consensus before getting everything merge ready.

src/services/casbin/casbin.conf
nicolasburtey: is it an out of the box model or a modified one? if the first, maybe we could use the filename by default on https://casbin.org/docs/en/supported-models

nicolasburtey: sounds to be this model: https://github.com/casbin/casbin/blob/master/examples/rbac_with_domains_model.conf ?

src/app/wallets/create-on-chain-address.ts
nicolasburtey: From Casbin documentation, the recommend way of doing it is with a middleware: https://casbin.org/docs/en/graphql-middlewares

bodymindarts: Thats funny because if you follow the referenced link https://graphql.org/learn/authorization/ you will read that the official graphql docs consider it best to delegate authorization to the business logic layer - which is the pattern I'm introducing.

It seems like casbin and graphql docs are not suggesting the same thing - or they are but semantics don't match.

In both cases AFAICT the authorization code happens in the resolver via delegating to a business logic function NOT as a GQL middleware (like it is currently)

src/services/casbin/casbin.conf
bodymindarts: Its not exactly that model.
The one you are referencing requires 4 arguments:
```
[request_definition]
r = sub, dom, obj, act
```

Meaning to test authorization you need to also pass in the specific domain.
In the modified version that I am proposing casbin would figure out by itself wether a permission exists via any domain. That way we only need to pass in 3 arguments which makes things a lot simpler.

src/servers/graphql-server.ts
```diff
@@ -150,6 +151,7 @@ export const startApolloServer = async ({
             geetest,
             account,
             ip,
+            authorizationService: await CasbinService(),
```
vindard, line 154: Curious why we chose to initialize here instead of within use-case methods like we do with other services

src/core/text.ts
```diff
@@ -179,6 +181,15 @@ export const login = async ({
       subLogger.info({ phone }, "user logged in")
     } else {
       user = await User.findOneAndUpdate({ phone }, {}, { upsert: true, new: true })
+      const res = await Authorization.initAccountPermissions({
```
vindard, line 184: Couldn't find the implementation for this, is it uncommitted?

vindard (commented):
As a general pattern were we pass `userId` in to domain functions as well to check authentication against an authentication service, this looks ok to me (especially with this context: https://graphql.org/learn/authorization/).

An additional idea I had was, I wonder if we can reduce some of the boilerplate somehow by using typescript decorators on domain methods (see [here](https://www.typescriptlang.org/docs/handbook/decorators.html)). I'm familiar with them from other languages, can look into whether they can be applicable here too.

_**Edit:** It looks like decorators can only be used inside classes_

src/servers/graphql-server.ts
bodymindarts: To re-use the database connection. This now creates 1 connection per request. Better would be to use a connection pool but I haven't seen a simple way to pass a pre-existing connection to the casbin libraries. Might investigate this when switching to postgres.

Either way 1 instance (and connection) per domain function would be too much overhead. (Note the repositories re-use the connection that is created when mongoose is initialized)

src/core/text.ts
bodymindarts: You're right I had forgotten to commit it - look here https://github.com/GaloyMoney/galoy/blob/3cbd7760485d285b493e0f42ad1fd4ba5e1a6a49/src/app/authorization/index.ts#L10

dolcalmi:
@bodymindarts can we close this?

PR #743, opened 2021-11-13
Add on chain payment mutations/query to walletId middleware

dolcalmi:
- Does not include on chain pay use case migration for `onChainPaymentSend` and `onChainPaymentSendAll`
- Fix `onChainAddressCurrent` mutation and `onChainTxFee` query logic
- Includes `onChainAddressCurrent`, `onChainPaymentSend` and `onChainPaymentSendAll` in walletId middleware
- Fix empty error message for old app errors

### DIFF
--- a/src/graphql/error-map.ts
+++ b/src/graphql/error-map.ts
@@ -18,8 +18,8 @@ const assertUnreachable = (x: never): never => {
 }
 
 export const mapError = (error: ApplicationError): CustomApolloError => {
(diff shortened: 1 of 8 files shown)


samerbuna (approved):

bodymindarts (approved):
I’m approving but also working on a refactor to have a more future proof way of doing authentication. Can talk about it in person in ES or at code review.

PR #744, opened 2021-11-14
Add intraLedgerPaymentSend mutation to walletId middleware

dolcalmi:
- Add walletId to intraLedgerPaymentSend
- Fix error mapping 
- The use case was not changed to use wallet public id because this will need to be updated when we implement the new authorization system.

### DIFF
--- a/src/app/wallets/index.ts
+++ b/src/app/wallets/index.ts
@@ -21,3 +21,10 @@ export const getWallet = async (walletId: WalletId) => {
   const wallets = WalletsRepository()
   return wallets.findById(walletId)
 }
+
+export const getWalletByPublicId = async (
+  walletPublicId: WalletPublicId,
+): Promise<Wallet | ApplicationError> => {
+  const wallets = WalletsRepository()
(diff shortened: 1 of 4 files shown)


bodymindarts (approved):

src/app/wallets/index.ts
```diff
@@ -21,3 +21,10 @@ export const getWallet = async (walletId: WalletId) => {
   const wallets = WalletsRepository()
   return wallets.findById(walletId)
 }
+
+export const getWalletByPublicId = async (
+  walletPublicId: WalletPublicId,
+): Promise<Wallet | ApplicationError> => {
+  const wallets = WalletsRepository()
+  return wallets.findByPublicId(walletPublicId)
+}
```
samerbuna, line 30: Why this wrapper around `findByPublicId`?

dolcalmi: Servers should only access app layer (not service layer) also, as mentioned in main comment, was not refactored in main use case because will be updated with the new authorization schema

samerbuna (approved):

PR #753, opened 2021-11-23
Expose the account id in the API

samerbuna:


dolcalmi (dismissed):

vindard (approved):

bodymindarts:
Why are we exposing the database ID directly here when we had so much discussion around introducing a WalletPublicID to avoid this?

vindard:
> Why are we exposing the database ID directly here when we had so much discussion around introducing a WalletPublicID to avoid this?

If I understood correctly, the purpose of this is to have an id to cache against at the graphql layer for Account-related requests, and it's not meant to be used anywhere else. From my end at least, wasn't sure if this would require us doing a new `AccountPublicId` property + db migration if it won't be used outside of caching.

dolcalmi (approved):

dolcalmi:
> > Why are we exposing the database ID directly here when we had so much discussion around introducing a WalletPublicID to avoid this?
> 
> If I understood correctly, the purpose of this is to have an id to cache against at the graphql layer for Account-related requests, and it's not meant to be used anywhere else. From my end at least, wasn't sure if this would require us doing a new `AccountPublicId` property + db migration if it won't be used outside of caching.

yes, adding AccountPublicId could be a lot of work just for client side graphql cache. I think the "deterministic" uuid approach (temporal solution) is fine because after account collection migration we only need to remove the resolvers.

vindard (approved):

PR #759, opened 2021-11-27
Transactions API data

samerbuna:
- Refactor WalletTransaction types
- Abstract common props into 'baseTransaction' in 'fromLedger'
- Make new 'ExtendedLedgerTransactionType' type
- Refactor 'fromLedger' to reflect new WalletTransaction types
- Refactor 'filterPendingIncoming to return new LedgerTransaction types
- Refactor new LedgerTransaction types into old graphql server
- Refactor initiationVia/settlementVia in graphql objects
- Refactor new LedgerTransaction types into tests
- Temporary: Set PaymentSecret to null
- Implement GraphQL types for the new InitiationVia/SettlementVia unions


src/domain/wallets/index.types.d.ts
```diff
@@ -12,11 +10,47 @@ type Deprecated = {
...
+  readonly paymentHash: PaymentHash
+  readonly pubkey: Pubkey
+}
+
+type InitiationViaOnChain = {
+  readonly type: PaymentInitiationMethod["OnChain"]
+  readonly address: OnChainAddress
+}
+
+type InitiationViaOnChainLegacy = {
```
nicolasburtey, line 30: why is that needed?

src/domain/wallets/index.types.d.ts
```diff
@@ -12,11 +10,47 @@ type Deprecated = {
...
+
+type SettlementViaIntraledger = {
+  readonly type: SettlementMethod["IntraLedger"]
+  readonly walletId: WalletId
+  readonly counterPartyUsername: Username | null
+}
+
+type SettlementViaLn = {
+  readonly type: SettlementMethod["Lightning"]
+  paymentSecret: PaymentSecret | null
```
nicolasburtey, line 43: this shouldn't be `null`? we should always know the `paymentSecret`  once this has been settled? maybe the reason this may not be `null` is if the payment is in flight?

src/domain/wallets/tx-history.ts
```diff
@@ -17,23 +17,26 @@ const filterPendingIncoming = (
...
-          otherPartyUsername: null,
-          settlementFee: toSats(0),
-          transactionHash: rawTx.txHash,
-          status: TxStatus.Pending,
-          memo: null,
-          createdAt: createdAt,
-          settlementAmount: sats,
-          settlementUsdPerSat: usdPerSat,
-          address,
+          initiationVia: {
```
nicolasburtey, line 32: if this is `initiationVia` without an appendix telling which protocol we are using, I would assume this would cover all the initiation option, yet we have only have `OnChain` in the object below?

src/domain/wallets/tx-history.ts
```diff
@@ -65,92 +68,143 @@ export const fromLedger = (
...
+          walletTransaction = {
+            ...baseTransaction,
+            initiationVia: {
+              type: PaymentInitiationMethod.Lightning,
+              paymentHash: paymentHash as PaymentHash,
+              pubkey: pubkey as Pubkey,
+            },
+            settlementVia: {
+              type: SettlementMethod.Lightning,
+              paymentSecret: null,
```
nicolasburtey, line 185: should not be `null`?

nicolasburtey: `paymentSecret` should probably be treated differently whether it's a `Payment` of an `Invoice`:

- in the case of an invoice, "we" (ie: our lnd node) actually know the `secret` right from the beginning because we created it. but we should not disclosed it before the invoice is being paid, because we don't want to let a end user know/share it until, this would be an attack vector. ie: an intermediary node could know the secret before it reaches our node, and the payment would never arrived to us so our user would not be credited, but the payer would have sent the money. maybe not a high risk scenario but something to consider.

- in the case of a payment, when we know the secret, it means the invoice has been settled, so there is no risk here

src/domain/wallets/tx-history.ts
```diff
@@ -65,92 +68,143 @@ export const fromLedger = (
...
+            },
+          }
+          return walletTransaction
+      }
+
+      walletTransaction = {
```
nicolasburtey, line 191: shouldn't this also be part of the `case:` scenario?

nicolasburtey: that would be cleaner that assuming a default?

nicolasburtey (commented):
couple of comments/questions

src/domain/wallets/index.types.d.ts
vindard: This was for onchain transaction that have `undefined` address in the db for older version of the codebase that didn't persist this

src/domain/wallets/index.types.d.ts
vindard: This was a temporary thing because we don't persist this yet for successful payments. I'm planning to add this in after the persist-payments PR (#665) is merged.

src/domain/wallets/tx-history.ts
vindard: True, I can make another type to reflect this (`WalletLnSettledSentTransaction` & `WalletLnSettledReceivedTransaction`) with different names for that property (`paymentSecretGenerated` vs. `paymentSecretReceived` for e.g.)

src/domain/wallets/tx-history.ts
vindard: We differentiate this via the `type` sub-property inside `initiationVia`, not the actual property name. In this case though, I could actually make this more accurate by changing the return type of the `filterPendingIncoming` to be `WalletOnChainTransaction`.

src/domain/wallets/tx-history.ts
vindard: This part of the code was a little messy and needs to be cleaned up a bit. It's tricky because we have send/receive types mixed in with things like `"routing_fee"` and `"exchange_rebalance"` under the same `LedgerTransactionType` definition. I tried a few things but ended up in some dead-ends, so opted to maintain the current behaviour which defaults to Intraledger initiated and settled and then to clean this up in another later PR (to not block the rest of Samer's integration work).

src/domain/wallets/tx-history.ts
vindard: > should not be `null`?

See this [comment](https://github.com/GaloyMoney/galoy/pull/759#discussion_r757837347)

src/domain/wallets/index.types.d.ts
```diff
@@ -12,11 +10,47 @@ type Deprecated = {
...
+  readonly type: PaymentInitiationMethod["Lightning"]
+  readonly paymentHash: PaymentHash
+  readonly pubkey: Pubkey
+}
+
+type InitiationViaOnChain = {
+  readonly type: PaymentInitiationMethod["OnChain"]
+  readonly address: OnChainAddress
+}
+
```
nicolasburtey, line 29:
```suggestion
FIXME: add OnChainAddress associated with old transaction to remove this legacy type
```

src/graphql/types/abstract/settlement-via.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const SettlementViaIntraLedger = new GT.Object({
+  name: "SettlementViaIntraLedger",
+  fields: () => ({
+    walletId: {
+      type: GT.NonNull(WalletId),
+    },
+    counterPartyUsername: {
+      type: Username,
+      description: dedent`Settlement destination: Could be null the payee does not have a username`,
```
nicolasburtey, line 20:
```suggestion
      description: dedent`Settlement destination: Could be null if the payee does not have a username`,
```

src/graphql/types/object/transaction.ts
```diff
@@ -0,0 +1,75 @@
...
+      type: GT.NonNull(InitiationVia),
+      description: "From which protocol the payment has been initiated.",
+    },
+    settlementVia: {
+      type: GT.NonNull(SettlementVia),
+      description: "To which protocol the payment has settled on.",
+    },
+    settlementAmount: {
+      type: GT.NonNull(SatAmount),
+      description: "Amount of sats paid by the user.",
```
nicolasburtey, line 36: this is inclusive fees? maybe should be added to the description?

src/graphql/types/abstract/initiation-via.ts
```diff
@@ -0,0 +1,43 @@
...
+      type: Username,
+    },
+  }),
+})
+
+const InitiationViaLn = new GT.Object({
+  name: "InitiationViaIntraLn",
+  fields: () => ({
+    paymentRequest: {
+      type: GT.NonNull(LnPaymentRequest),
```
nicolasburtey: sometime the amount is mandatory (amountless invoice), sometimes it's explicitly forbidden. how are we dealing with this here?

because amount is somehow related to the initiation type

src/domain/wallets/tx-history.ts
nicolasburtey: no strong opinion on this except that `"routing_fee"` and `"exchange_rebalance"` are not meant to be returned to any normal user.

it's a transaction that is related to the admin/"bank owner", so this won't be expose to the public API in my mind.

vindard: Yup exactly, we should probably differentiate those into another domain type

src/domain/wallets/index.types.d.ts
vindard: I added this commit, but thinking about it now I'm not sure if we'd have this information anywhere else to backfill for transactions that were initiated onchain but settled intraledger 🤔

src/domain/wallets/tx-history.ts
```diff
@@ -65,92 +68,143 @@ export const fromLedger = (
...
+      }
+
+      let walletTransaction: WalletTransaction
+      switch (txType) {
+        case ExtendedLedgerTransactionType.IntraLedger:
+          walletTransaction = {
+            ...baseTransaction,
+            initiationVia: {
+              type: PaymentInitiationMethod.IntraLedger,
+              walletId: walletId as WalletId,
```
dolcalmi, line 117: why do we need walletId here? is the same base walletId

src/graphql/types/object/transaction.ts
dolcalmi: yes, this amount includes fee

image

dolcalmi (approved):
as reference, this issues need to be solved after merge

- Add counterPartyWalletId (publicid) to medici transactions https://github.com/GaloyMoney/galoy/issues/765
- Add payment secret for invoices and payments https://github.com/GaloyMoney/galoy/issues/764

src/domain/ledger/index.ts
```diff
@@ -25,6 +26,11 @@ export const LedgerTransactionType = {
   ToHotWallet: "to_hot_wallet",
 } as const
 
+export const ExtendedLedgerTransactionType = {
```
bodymindarts, line 29: Why was this added as an additional type.

bodymindarts (approved):

src/graphql/types/abstract/initiation-via.ts
```diff
@@ -0,0 +1,48 @@
...
+import OnChainAddress from "../scalar/on-chain-address"
+import { PaymentInitiationMethod } from "@domain/wallets"
+import PaymentHash from "../scalar/payment-hash"
+
+const InitiationViaIntraLedger = new GT.Object({
+  name: "InitiationViaIntraLedger",
+  isTypeOf: (source) => source.type === PaymentInitiationMethod.IntraLedger,
+  fields: () => ({
+    counterPartyWalletId: {
+      // type: GT.NonNull(WalletId),
```
vindard, line 14: Based on https://github.com/GaloyMoney/galoy/pull/771#issuecomment-983855316 should I change this back as well?

src/domain/ledger/index.ts
vindard: It was to help us distinguish ln-intraledger from normal intraledger transactions since we currently lump both under `on_us` in the db whereas onchain-intraledger has its own `onchain_on_us` type.

It got used in the `tx-history` switch when categorizing different transactions into their respective `initiationVia` and `settlementVia` types.

PR #760, opened 2021-11-27
migrating rest of user from core to app

nicolasburtey:


src/app/users/update-user-level.ts
```diff
@@ -0,0 +1,5 @@
+import { update } from "@services/mongoose/users-admin"
+
+export const updateUserLevel = async ({ uid, level }): Promise<UserAdmin | Error> => {
+  return update({ id: uid, level })
```
nicolasburtey, line 4: @bodymindarts what is the right way/level of abstraction to update a single field ?

`update` for `users-admin` can currently update 2 fields, either `level` or `status`. 

I'm just not sure how to design the repository.

I'm confused by https://github.com/GaloyMoney/galoy/blob/204da8da0fc4704b393e6942222d3cdc19c7e713/src/services/mongoose/users.ts#L67, because it takes `{phone,
    language,
    contacts,
    deviceTokens,
    twoFA}` as argument, but there is always only 1 argument that will be updated by a migration, yet all the argument are mandatory.

nicolasburtey: ok looking at it we systematically load the object before. it's a weird pattern to me as this is not atomic and therefore can introduce race condition. but I'll use this pattern for now.

bodymindarts: Correct the basic pattern is:
```
entity = repo.findById(id)
entity.field = newValue
repo.update(entity)
```

While the current approach is executed atomically inside the database it doesn't prevent race conditions as there is no pre-condition check within the transactional boundaries.
For updates that need consistency checks the way to go is to use [optimistic concurrency](https://en.wikipedia.org/wiki/Optimistic_concurrency_control). For the most critical parts of the domain I would even opt for [event sourcing](https://martinfowler.com/eaaDev/EventSourcing.html) as it gets you the highest level of guarantees when it comes to consistency and auditing.

As to designing the repository - entity boundaries should align with consistency requirements derived from business imperatives. Things that must be consistent / updated atomically must be on the same entity.

In general most values in our system aren't at risk of causing breakage due to concurrent updates - so I don't consider the possibility for race conditions (which existed previously as well) as a big deal here.

PR #762, opened 2021-11-28
Organizing scripts

samerbuna:
This is a proposal for how we can organize npm scripts under a scripts folder and cleanup package.json.

Note, I changed the syntax to use the `:` separator as it's a more common way to namespace scripts.

If you all think this is good, I'll convert the rest. @nicolasburtey @bodymindarts @vindard @dolcalmi


dolcalmi (dismissed):

dolcalmi (commented):
looks good, my only concern is about the use of `:` in filenames (in general is better to avoid it). if someone has windows would be good to test it

scripts/eslint:check
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env sh
+
+eslint "{src,test}/**/*.ts"
```
bodymindarts, line 3: What if `eslint` isn't installed locally? The `yarn eslint` prefix ensures that the correct version from `package.json` is used right?

bodymindarts:
I like having a consistent script organisation - I would prefer to use `make` as the primary task driver to be consistent with all our other repositories, but its not a big deal.

Only concern is that the scripts are calling the dev tools directly instead of via `yarn`. This creates a dependency on the local dev environment which we need to avoid!

package.json
```diff
@@ -1,29 +1,30 @@
...
-    "update-postman": "curl --location --request GET \"https://api.getpostman.com/collections/4e3fee88-7267-44e1-a316-b81e90b0bd2b\" --header \"X-Api-Key: $POSTMANAPIKEY\" | jq . > test/postman/graphql.postman-collection.json",
-    "test": "export JEST_JUNIT_OUTPUT_DIR=\"./artifacts\" && jest --ci --reporters=default --reporters=jest-junit",
-    "test:unit": ". ./.envrc && LOGLEVEL=warn jest --config ./test/jest-unit.config.js --bail --verbose $TEST | yarn pino-pretty -c -l",
-    "test:integration": ". ./.envrc && LOGLEVEL=warn jest --config ./test/jest-integration.config.js --bail --runInBand --verbose $TEST | yarn pino-pretty -c -l",
-    "build-docs": "npx spectaql spectaql-config.yml -1"
+    "build:docs": "./scripts/build-docs",
+    "build": "./scripts/build-tsc",
+    "code:check": "./scripts/code-check",
+    "deps:clean": "./scripts/deps-clean",
+    "deps:reset": "./scripts/deps-reset",
```
bodymindarts, line 8: Strongly disagree with using yarn to drive tasks that do not depend on nodes.

scripts/deps-clean
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env sh
+
+docker-compose down
```
bodymindarts, line 3: Don't see the value in the indirection for 1 liner commands. It just makes figuring out what is actually going less obvious

scripts/code-check
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
+
+yarn tsc:check
```
bodymindarts, line 3: Circular dependency. `yarn` calls `sh` calls `yarn`.

scripts/start
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env sh
+
+./scripts/start-deps
```
bodymindarts, line 3: This script serves no purpose other than delegating to other scripts. The more indirection the less obvious it is to see what is going on.

scripts/eslint:check
samerbuna: It's still the same with this method. You can only use yarn to run these scripts and yarn would use the correct version.

package.json
samerbuna: Why?

I removed the deps tasks and included their actions directly in other tasks.

scripts/code-check
samerbuna: It's not really a "circular" dependency. Yarn run can call other yarn runs. I did however make this call the other scripts directly

scripts/start
samerbuna: The start script? It serves the purpose of running multiple servers. We can put all the commands here but I think it's helpful to have them separate so that they can be run individually if needed

samerbuna:
@dolcalmi, we had a discussion about this while you were not there. @bodymindarts and I have a conflict here:

- I strongly vote that npm scripts should be the only task runner in this Node repo (and all JavaScript repos really). They are optimized for node_modules binaries and it's the common and expected method out there. Having `make` in the mix for some tasks is a bit confusing (and unnecessary), IMO.

- Justin strongly votes that we keep `make`. One argument there is that it's good to have the same task runner in all of galoy repos, and we have non JS repos under galoy.

I disagree with that argument, it's kinda like saying: in this house, we only eat food with a fork.

I guess you are the deciding vote on this? Either way we need to pick one tool and remove the other, if we're going with `yarn`, merge this PR. If we're going with `make`, close this PR and I'll work on another PR to remove yarn scripts and keep only make scripts.

bodymindarts:
To lay out my position I think we should be using npm scripts (ie. 1 liners) for simple commands where we need the environment wrapping that it provides.
Make should be the default especially when things are more complex, there are no nodejs dependencies involved, tasks depend on other tasks etc. This is a common language-agnostic approach.

To see an example take a look at the way `Makefile` and `package.json` are used here: https://github.com/ory/kratos-selfservice-ui-node/tree/hydra-integration

dolcalmi:
:thinking: I personally prefer to use make files but agree with @samerbuna because:

- Is common and "well known" by js developers (backend, frontend and library projects).
- One liners or passing params to internal cmds some times are common problems but I think is solved using yarn and script files (this also fix editor support)
- In regards to this "it's good to have the same task runner in all of galoy repos": I like consistency but in this moment galoy-pay, galoy-mobile and admin don't use make files so not sure if is enough to decide against npm/yarn scripts

PR #763, opened 2021-11-28
refactor: migrate admin requests to apps

nicolasburtey:
working on removing core/


dolcalmi:
please create PR per case, also just for your information arvin is migrating fee probe and me onChainPay

src/app/admin/index.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  const username_ = username as Username
+
+  const accounts = AccountsRepository()
+  return accounts.findByUsername(username_)
+}
+
+export const getUserByPhone = async function (phone: string) {
+  const accounts = AccountsRepository()
+  return accounts.findByPhone(phone as PhoneNumber)
```
dolcalmi, line 16: phone must not be part of account, also avoid this type of direct cast (you can implement the same strategy used in `getUserByUsername` using `checkedToUsername` to validate input)

src/app/users/update-business-map-info.ts
```diff
@@ -0,0 +1,20 @@
...
+  coordinate: { latitude: number; longitude: number }
+  title: string
+}): Promise<Account | Error> => {
+  const accountsRepo = AccountsRepository()
+
+  const account = await accountsRepo.findByUsername(username as Username)
+  if (account instanceof Error) return account
+
+  account.coordinate = coordinate as Coordinates
+  account.title = title as BusinessMapTitle
```
dolcalmi, line 18: if we will have business info associated to username then it shoud be part of user repo.

src/app/users/update-business-map-info.ts
```diff
@@ -0,0 +1,20 @@
...
+  return accountsRepo.update(account)
```
dolcalmi, line 19: this kind of update is not good at account level... please use a specific repo for businesses (this will facilitate the subsequent migration)

src/app/admin/index.ts
```diff
@@ -0,0 +1,17 @@
+import { checkedToUsername } from "@domain/users"
```
dolcalmi, line 1: not sure about file location, if this is `getUserByUsername` use case then should be in user folder

src/app/users/update-business-map-info.ts
```diff
@@ -0,0 +1,20 @@
+import { AccountsRepository } from "@services/mongoose/accounts"
```
dolcalmi, line 1: this is contradictory, this use case is in users folder but we are using account repo (for all logic)

dolcalmi (changes requested):
partial review ... maybe we should discuss users / accounts / wallets again before continuing

src/app/admin/index.ts
```diff
@@ -0,0 +1,17 @@
...
+import { AccountsRepository } from "@services/mongoose"
+
+export const getUserByUsername = async function (username: string) {
```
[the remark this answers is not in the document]
nicolasburtey: this is the inferred return type: `const getUserByUsername: (username: string) => Promise<RepositoryError | Account>` 

What is the benefits of having to manually add the return type? it creates more code boilerplate (so code is harder to read)

reading on typescript best practice, what I'm reading is: make sure typescript can infer the type, and best to not have to type everything as long as typescript can understand what this is

src/app/admin/index.ts
nicolasburtey: I've been thinking about it, but the only use case is from the admin panel?

src/app/admin/index.ts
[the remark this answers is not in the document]
nicolasburtey: please use arrow functions: done

src/app/admin/index.ts
```diff
@@ -0,0 +1,17 @@
...
+  const usernameValid = checkedToUsername(username)
+  if (usernameValid instanceof Error) throw usernameValid
+
+  const username_ = username as Username
+
+  const accounts = AccountsRepository()
```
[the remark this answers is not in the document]
nicolasburtey: agree for the phone. but we may not want to have user have username (only phone or email) to avoid the confusion on the use of username: it's to receive a payment, not to login.

src/app/users/update-business-map-info.ts
nicolasburtey: is that not good enough for now? we want to migrate all the use cases to app shortly, then we can improve those.

src/app/users/request-phone-code.ts
```diff
@@ -22,16 +23,19 @@ export const requestPhoneCodeWithCaptcha = async ({
...
-  phone: PhoneNumber
+  phone: string
   geetest: GeetestType
   geetestChallenge: string
   geetestValidate: string
   geetestSeccode: string
   logger: Logger
   ip: IpAddress
-}): Promise<true | CaptchaError | UnknownPhoneProviderServiceError> => {
+}) => {
```
nicolasburtey, line 33: type is automatically being detected by Typescript

src/domain/accounts/index.ts
```diff
@@ -11,3 +13,24 @@ export const AccountStatus = {
...
+  }
+  const coordinates: Coordinates = { latitude, longitude }
+  return coordinates
+}
+
+export const checkedMapTitle = (title: string) => {
+  if (title.length < 3 || title.length > 100) {
+    return new InvalidTitleLengthError()
+  }
+  return title as BusinessMapTitle
```
nicolasburtey, line 35: @bodymindarts is that how it's supposed to be cast in this instance?

src/graphql/admin/index.ts
```diff
@@ -1,16 +1,8 @@
...
-import { ALL_INTERFACE_TYPES } from "@graphql/types"
-
 import QueryType from "./queries"
 import MutationType from "./mutations"
-import { isDev } from "@core/utils"
-
-export const gqlAdminSchema = new GraphQLSchema({
-  query: QueryType,
-  mutation: MutationType,
-  types: ALL_INTERFACE_TYPES,
```
nicolasburtey, line 12: @samerbuna can you let me know if this is ok?

nicolasburtey: needed to remove type to avoid collision between Account in admin and root

src/graphql/admin/lexicographic-sort-schema.graphql
```diff
@@ -1,8 +1,17 @@
-interface Account {
-  csvTransactions(walletIds: [WalletId!]!): String!
-  defaultWalletId: WalletId!
+type Account {
```
nicolasburtey, line 1: oh could it have anything to do with a mix between `interface` and `type` that I may have done?

nicolasburtey: @samerbuna

src/graphql/admin/root/mutation/account-update-level.ts
```diff
@@ -0,0 +1,46 @@
...
+const AccountUpdateLevelMutation = GT.Field({
+  type: GT.NonNull(AccountDetailPayload),
+  args: {
+    input: { type: GT.NonNull(AccountUpdateLevelInput) },
+  },
+  resolve: async (_, args) => {
+    const { uid, level } = args.input
+
+    // FIXME: nicolas: I don't understand the purpose of this below
+    for (const input of [uid, level]) {
```
nicolasburtey, line 28: @samerbuna is this that uid and level are being checked by graphql with `parseLiteral` and/or `parseValue`?

src/graphql/admin/types/object/account.ts
```diff
@@ -0,0 +1,46 @@
...
+    owner: {
+      // should be used for individual account only,
+      // ie: when there are no multiple users
+      // probably separating AccountDetail to DetailConsumerAccount
+      // with DetailCorporateAccount is a way to have owner only in DetailConsumerAccount
+      // and users: [Users] in DetailCorporateAccount
+
+      type: GT.NonNull(User),
+      resolve: async (source) => {
+        const user = await UsersRepository().findById(source.id) // FIXME/TODO should it be source.ownerId?
```
nicolasburtey, line 27: @bodymindarts @samerbuna I don't yet fully understand how the recursion work with graphql. I presume `source` here will be resolve to what the Account is. but how to get the `UserId` from the `Account` object? 

currently they're the same but it's an implementation detail.

src/graphql/main/lexicographic-sort-schema.graphql
```diff
@@ -46,6 +46,33 @@ type AuthTokenPayload {
   errors: [Error!]!
 }
 
+type BTCWallet implements Wallet {
```
nicolasburtey, line 49: not clear why this has been added to the schema

bodymindarts: Its a potential future proofing as we will very likely have a different type for `FiatWallet` or `USDWallet`

src/app/accounts/index.ts
```diff
@@ -10,11 +10,21 @@ export * from "./add-api-key-for-account"
...
+
 export const getAccount = async (accountId: AccountId) => {
-  const accounts = AccountsRepository()
   return accounts.findById(accountId)
 }
 
+export const getWalletPublicIdFromUsername = async (
+  username: Username,
+): Promise<WalletPublicId | Error> => {
+  const account = await accounts.findByUsername(username)
```
bodymindarts, line 22: `username` will be renamed here in a followup PR right?

src/app/accounts/index.ts
```diff
@@ -10,11 +10,21 @@ export * from "./add-api-key-for-account"
...
+  if (account instanceof Error) return account
+
+  return account.walletPublicId
```
bodymindarts, line 25: `account` doesn't have a `walletPublicId` a `Wallet` has it. We have to go from account to:
```
wallets.findById(account.defaultWalletId)
```

or?

src/app/accounts/update-account-status.ts
```diff
@@ -0,0 +1,17 @@
...
+}: {
+  id: string
+  status: string
+}): Promise<Account | Error> => {
+  const accountsRepo = AccountsRepository()
+
+  const account = await accountsRepo.findById(id as AccountId)
+  if (account instanceof Error) return account
+
+  account.status = status as AccountStatus
```
bodymindarts, line 15: account status needs a `checkedToXxx` method. It is not safe to cast a `string` as an `AccountStatus`.

src/app/accounts/update-business-map-info.ts
```diff
@@ -0,0 +1,27 @@
...
+  coordinates: { latitude, longitude },
+  title,
+}: {
+  username: string
+  coordinates: { latitude: number; longitude: number }
+  title: string
+}): Promise<Account | Error> => {
+  const accountsRepo = AccountsRepository()
+
+  const account = await accountsRepo.findByUsername(username as Username)
```
bodymindarts, line 15: Casting `as Username` is unsafe - need to use `checkedToUsername`

src/app/users/request-phone-code.ts
bodymindarts: But if we don't set it explicitly (ie. communicate to the client what the return type should be) and we change the implementation to return something else (perhaps by accident) then this can cause unintended errors right?

In other words if the return is inferred and the implementation changes to return a different type then we have broken the contract inadvertently.

bodymindarts (commented):
Need to revisit all `as` casting you're doing and prefer using `checkedToXxx` methods.

Haven't gotten through the whole PR - will continue review later.

src/app/accounts/index.ts
nicolasburtey: yes

nicolasburtey: assuming you mean renaming `username` to `handle`

src/app/accounts/index.ts
nicolasburtey: you're right.

src/app/accounts/update-account-status.ts
nicolasburtey: the check is already done at the [graphql level](https://github.com/GaloyMoney/galoy/blob/521310ac85202eb0158a6fca72d33b3b000418d5/src/graphql/admin/root/mutation/user-update-status.ts#L14). maybe casting should happen at the graphql layer?

otherwise doing the check here would mean we're looking at the validity of status enum twice

bodymindarts: that is a different type. GraphQL layer has its own AccountStatus:
```
import AccountStatus from "@graphql/admin/types/scalar/account-status"
```

bodymindarts: The graphql layer check gives an early return to clients (at query parsing time) - the domain layer check is authoritative.

nicolasburtey:
> But if we don't set it explicitly (ie. communicate to the client what the return type should be) and we change the implementation to return something else (perhaps by accident) then this can cause unintended errors right?
In other words if the return is inferred and the implementation changes to return a different type then we have broken the contract inadvertently.

that is correct, but if new type if been added inadvertently, Typescript would not let the app compile because the function that depend on it would have a broken contract

here is the advantage of not making the type explicit when it's detected by Typescript. in many functions, the type is overly broad. instead of including errors manually, like `Promise<ApplicationError1, ApplicationError2>`, the error returned is `Promise<Error>`, which is valid, but we're loosing the granularity of the possible errors. 

This is especially true when the function embedded other function that can return `Error`. It become almost impossible to properly cast the function because at some point the number of `Error`s possibly returned compound, and we don't want to update all the functions that depend on this function to properly cast the error while Typescript can figure out it.

it will also help reduce the number of codeline, making the code more compact.

src/graphql/main/lexicographic-sort-schema.graphql
nicolasburtey: I know but I don't know why this appear here in this PR (this is the auto generated file), as I have not changed anything regarding the BTCWallet interface on the graphql layer AFAIK

src/config/app.ts
```diff
@@ -249,6 +249,12 @@ export const getBuildVersions = (): {
 
 export const PROXY_CHECK_APIKEY = yamlConfig?.PROXY_CHECK_APIKEY
 
+// FIXME: we have process.env.NODE_ENV === "production" | "development" | "test"
+// "test" might not be needed
```
bodymindarts, line 253: Why is this comment necessary? Where are we setting or depending on `NODE_ENV="test"`?

src/domain/accounts/index.ts
bodymindarts: Yes - inside the `checkedToXxx` function is the only place where casting like this is allowed.

src/domain/users/index.types.d.ts
```diff
@@ -80,6 +79,25 @@ type NewUserInfo = {
   phoneMetadata: PhoneMetadata | null
 }
 
+type TestAccounts = {
```
bodymindarts, line 82: if these are `Accounts` they probably shouldn't be under `domain/users`... or are they more accurately `TestUser`?

Also the type shouldn't be pluralised:
```
testAccount: TestAccount = {..}
testAccounts = [testAccount]
```

src/graphql/admin/index.ts
```diff
@@ -1,19 +1,20 @@
...
-})
+import { isDev } from "@config/app"
 
 if (isDev) {
   import("@services/fs").then(({ writeSDLFile }) => {
     writeSDLFile(__dirname + "/schema.graphql", printSchema(gqlAdminSchema))
+    writeSDLFile(
+      __dirname + "/lexicographic-sort-schema.graphql",
```
bodymindarts: why is this in the diff? Its in another PR

src/graphql/admin/lexicographic-sort-schema.graphql
bodymindarts: Don't understand this comment.

src/graphql/admin/root/mutation/account-update-level.ts
samerbuna: They are not, but we're generically checking all inputs here in case in the future graphql parsing checks are added.

src/domain/accounts/index.types.d.ts
```diff
@@ -62,10 +68,14 @@ type LimitsChecker = {
     walletVolume: TxVolume
   }): true | LimitsExceededError
 }
+
 interface IAccountsRepository {
   findById(accountId: AccountId): Promise<Account | RepositoryError>
   listByUserId(userId: UserId): Promise<Account[] | RepositoryError>
   findByWalletId(walletId: WalletId): Promise<Account | RepositoryError>
   findByWalletPublicId(walletPublicId: WalletPublicId): Promise<Account | RepositoryError>
+  findByPhone(phone: PhoneNumber): Promise<Account | RepositoryError>
```
bodymindarts, line 77: I think this can be deleted as you've already implemented the use case correctly:
```
user = userRepo.findByPhone(phone)
account = accounts.findById(user.defaultAccountId)
```

src/graphql/admin/index.ts
```diff
@@ -18,5 +16,4 @@ if (isDev) {
 export const gqlAdminSchema = new GraphQLSchema({
   query: QueryType,
   mutation: MutationType,
-  types: ALL_INTERFACE_TYPES,
```
samerbuna, line 21: This is needed to include interface types in the schema

nicolasburtey: even for admin when those types are not used AFAIK?

src/graphql/admin/types/object/account.ts
```diff
@@ -0,0 +1,46 @@
...
+        if (user instanceof Error) {
+          throw user
+        }
+
+        return user
+      },
+    },
+    coordinates: {
+      type: Coordinates,
+      resolve: (source) => source.coordinates,
```
samerbuna, line 37: The resolve function here can be dropped (this is the default behavior when a resolve function is not defined)

src/app/accounts/update-account-level.ts
```diff
@@ -0,0 +1,17 @@
+import { AccountsRepository } from "@services/mongoose"
+
+export const updateAccountLevel = async ({
+  id,
+  level,
+}: {
+  id: string
```
vindard, line 7: We've been typing these at this point as `id: AccountId` to avoid the cast a few lines down as well

src/app/accounts/update-account-status.ts
```diff
@@ -0,0 +1,17 @@
+import { AccountsRepository } from "@services/mongoose"
+
+export const updateAccountStatus = async ({
+  id,
+  status,
+}: {
+  id: string
```
vindard, line 7: Same here

src/app/users/request-phone-code.ts
bodymindarts: should be `| ApplicationError`

bodymindarts:
Definition of `ApplicationError` https://github.com/GaloyMoney/galoy/blob/main/src/app/index.types.d.ts#L10

src/app/accounts/update-account-status.ts
bodymindarts: The GQL layer definition should reference the domain definition 
here is the source of truth definition: https://github.com/GaloyMoney/galoy/blob/main/src/domain/accounts/index.ts#L10-L13
Look how we are doing it here:
https://github.com/GaloyMoney/galoy/blob/main/src/graphql/types/scalar/tx-status.ts#L7-L9

nicolasburtey: I guess i don't understand why we want to make a distinction between the `graphql` and `domain` layer. seems we are making twice the work here: ie checking twice for the same thing, with a lot of code duplication.

src/config/app.ts
nicolasburtey: don't know. the comment with `FIXME` is that we look into it in future PR.

src/domain/accounts/index.types.d.ts
nicolasburtey: you're right

src/domain/users/index.types.d.ts
nicolasburtey: they're both account + user given that they're very coupled right now and creating one will automatically create the other. I'll rename to User to be consistent.

nicolasburtey: I'll actually keep testAccount :)

src/graphql/admin/root/mutation/account-update-level.ts
nicolasburtey: why having this code if this is a no-op for the time being? isn't better to add the check when we are actually using the graphql check? otherwise it may make developer thinks/mislead the inputs are actually already checked.

samerbuna: Fair point. We should really leverage TS types here to figure out which inputs are checked. I'll look into the best way to do that.

src/app/accounts/update-account-status.ts
bodymindarts: Graphql = coupled to clients shouldn't break
Domain = coupled to business rules - will evolve -> breaking changes

Their lifecycles, reasons to change is different. If we don't decouple them we cannot evolve our domain logic without causing breaking changes to clients (which we don't want for API v2).

Of course keeping them in sync as much as possible is useful to minimize cognitive overhead - but coupling them will significantly slow us down in the mid term unless we are also willing to push breaking changes to GQL.

nicolasburtey: > Decoupled domain layer is a great asset to have in my experience.
not my experience so far. I have lost 1d on trying to get my head [around this issue](https://github.com/GaloyMoney/galoy/issues/781). 

Per this [book](https://book.productionreadygraphql.com/), Facebook (creator of GraphQL) is coupling their graphql layer with their domain. There is a [popular library](https://typegraphql.com/) to do it. not saying we should move to a 1-1 mapping, but the further we get away from a tight coupling, the more the developer overhead there is. 

our v2 api already need some deprecation (some User stuff needs to be renamed to Account), so while we want to keep a API fairly stable, migration is something we have to plan anyway.

bodymindarts: He warns that it can be dangerous (p90):
> An annotation-based schema implementation usually uses your domain entities, objects, classes, etc. and annotates them to generate a GraphQL schema out of them. The big advantage here is that the interface layer is tiny. You don’t need hundreds of objects defining a GraphQL schema. Instead, you may simply reuse the domain entities you already have.
The danger with this approach is that it depends on your “entities” making a good API in the first place. If all you have are ActiveRecord objects closely coupled to your database schema or implementation details, an annotation-based approach risks making your resulting API not so great or stable. However, if you do have well-defined entities that represent your domain without being coupled to implementation details like the database, this might be a really great approach.


Also from the same book (p93):
> No matter how you’re building a schema, you will have to write resolvers at some point. The resolver pattern of GraphQL’s execution engine is great. It helps teams add new use cases by working on their isolated resolver and makes it simple to think about the logic of a single field. However, it can be overused. At its core, GraphQL is an API interface. It’s an interface to your domain and business logic, and in most cases, should not be the source of truth for your business.

Its an API *interface* (...) should *NOT* be source of truth!

I consider a library like [this](https://typegraphql.com/) to be poison similar to ORMs. They provide initial fast iteration at the cost of long term sustainable progress and maintainability.

If you want to revisit this we should probably bring it up @ code review to get more opinions.

src/graphql/admin/root/mutation/account-update-status.ts
```diff
@@ -29,11 +29,11 @@ const UserUpdateStatusMutation = GT.Field({
       }
     }
 
-    const user = await updateUserAccountStatus({ uid, status })
-    if (user instanceof Error) {
-      return { errors: [{ message: user.message }] }
+    const account = await updateAccountStatus({ id: uid, status })
+    if (account instanceof Error) {
+      return { errors: [{ message: account.message }] }
```
vindard, line 34:
```suggestion
      return { errors: [{ message: mapError(account).message }] }
```

src/graphql/admin/root/mutation/business-update-map-info.ts
```diff
@@ -37,23 +35,27 @@ const BusinessUpdateMapInfoMutation = GT.Field({
...
+      latitude,
+      longitude,
+    }
+
+    const account = await updateBusinessMapInfo({ username, title, coordinates })
 
-    if (user instanceof Error) {
-      return { errors: [{ message: user.message }] }
+    if (account instanceof Error) {
+      return { errors: [{ message: account.message }] }
```
vindard, line 46:
```suggestion
      return { errors: [{ message: mapError(account).message }] }
```

src/graphql/admin/root/query/account-details-by-phone.ts
```diff
@@ -0,0 +1,27 @@
...
+import { getAccountByUserPhone } from "@app/admin"
+
+const AccountDetailsByUserPhoneQuery = GT.Field({
+  type: GT.NonNull(GraphQLUser),
+  args: {
+    phone: { type: GT.NonNull(Phone) },
+  },
+  resolve: async (parent, { phone }) => {
+    if (phone instanceof Error) {
+      throw phone
```
vindard, line 15: I saw this was from a simple switchover, but should this be `return` instead of `throw` here?

src/graphql/admin/root/query/account-details-by-username.ts
```diff
@@ -0,0 +1,26 @@
...
+import { getAccountByUsername } from "@app/admin"
+
+const AccountDetailsByUsernameQuery = GT.Field({
+  type: GT.NonNull(GraphQLAccount),
+  args: {
+    username: { type: GT.NonNull(Username) },
+  },
+  resolve: async (parent, { username }) => {
+    if (username instanceof Error) {
+      throw username
```
vindard, line 14: Same comment `return` vs. `throw`?

src/graphql/root/query/user-default-wallet-id.ts
```diff
@@ -1,37 +1,28 @@
...
-
-    if (walletPublicId instanceof CouldNotFindError) {
-      throw new NotFoundError("User not found", { forwardToClient: true, logger })
-    }
-
-    if (walletPublicId instanceof Error) {
-      throw new UnknownClientError("Something went wrong")
-    }
+    const account = await AccountsRepository().findByUsername(username)
+    if (account instanceof Error) return account
```
vindard, line 21:
```suggestion
    if (account instanceof Error) return mapError(account)
```

src/servers/graphql-server.ts
```diff
@@ -84,6 +85,13 @@ const sessionContext = ({ token, ips, body, apiKey, apiSecret }) => {
             reason: loggedInUser,
           })
         domainUser = loggedInUser
+
+        const loggedInDomainAccount = await AccountsRepository().findById(
+          domainUser.defaultAccountId,
+        )
+        if (loggedInDomainAccount instanceof Error) throw Error
```
vindard, line 92: Can probably throw some sort of `ApolloError` here with the `reason: loggedInDomainAccount` property

vindard (changes requested):
This is getting very close now I think. Just left a few small cleanup suggestions.

src/graphql/admin/root/query/account-details-by-phone.ts
nicolasburtey: every `error` in graphql should throw no?

@samerbuna @bodymindarts

src/domain/users/index.ts
```diff
@@ -14,3 +14,25 @@ export const checkedToUsername = (username: string): Username | ValidationError
...
+export const checkedToPhoneNumber = (
+  phoneNumber: string,
+): PhoneNumber | ValidationError => {
+  if (!phoneNumber.match(PhoneNumberRegex)) {
+    return new InvalidPhoneNumber(phoneNumber)
+  }
+  return phoneNumber as PhoneNumber
+}
+
+export const TestAccounts: TestAccounts = (testAccounts: TestAccount[]) => ({
```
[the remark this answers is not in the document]
nicolasburtey: done. also move to `accounts/` instead of `/users`.

src/graphql/admin/root/mutation/account-update-status.ts
```diff
@@ -17,7 +18,7 @@ const UserUpdateStatusInput = new GT.Input({
 })
 
 const UserUpdateStatusMutation = GT.Field({
-  type: GT.NonNull(UserDetailPayload),
+  type: GT.NonNull(AccountDetailPayload),
```
dolcalmi, line 21: please update UserUpdateStatusMutation to AccountUpdateStatusMutation and the same for UserUpdateStatusInput

src/graphql/admin/root/mutation/business-update-map-info.ts
```diff
@@ -37,23 +36,27 @@ const BusinessUpdateMapInfoMutation = GT.Field({
...
-        coordinate: user.coordinate,
-        created_at: user.created_at,
+      accountDetails: {
+        id: account.id,
+        username: account.username,
+        level: account.level,
+        status: account.status,
+        title: account.title,
+        coordinates: account.coordinates,
+        createdAt: account.createdAt,
```
dolcalmi, line 59: if the attributes are the same you dont need to do this `accountDetails: account` is enough

src/graphql/admin/root/query/account-details-by-phone.ts
dolcalmi: that was true until the new error map was introduced (this applies for V2, for v1 is not necessary)... now should be something like:

for graphql input validations (line 14 repace input by phone) 
```
if (input instanceof Error) {
}
```
and for returns from app layer (line 19 replace result with account)
```
  return { errors: [{ message: appErr.message }] }
}
```

src/graphql/main/queries.ts
```diff
@@ -17,7 +17,7 @@ const QueryType = new GT.Object({
     globals: GlobalsQuery,
     me: MeQuery,
     usernameAvailable: UsernameAvailableQuery,
-    userDefaultWalletId: UserDefaultWalletIdQuery,
+    userDefaultWalletId: AccountDefaultWalletIdQuery, // FIXME: migrate to AccountDefaultWalletId
```
dolcalmi, line 20: uhmm this is a bit confusing, imported from `"@graphql/root/query/user-default-wallet-id"` `userDefaultWalletId` as attribute but with `AccountDefaultWalletIdQuery` type name

src/graphql/root/query/user-default-wallet-id.ts
```diff
@@ -1,37 +1,29 @@
+import { mapError } from "@graphql/error-map"
```
dolcalmi, line 1: file needs to be renamed

src/servers/graphql-server.ts
```diff
@@ -101,13 +109,15 @@ const sessionContext = ({ token, ips, body, apiKey, apiSecret }) => {
...
       }
 
-      addAttributesToCurrentSpan({ [ENDUSER_ALIAS]: domainUser?.username })
+      addAttributesToCurrentSpan({ [ENDUSER_ALIAS]: domainAccount?.username })
 
       return {
         logger,
         uid: userId,
         wallet,
+        // FIXME: we should not return this for the admin graphql endpoint
```
bodymindarts, line 118: What should we return? What is the fix? Is _this_ really the right place to add this information. Wouldn't it make more sense to add a comment to the admin/user resolver (ie. where it might potentially be (mis-)used) rather than here where it is being injected?

src/migrations/20211202002129-coordinate-to-coordinates.ts
```diff
@@ -0,0 +1,19 @@
...
+            coordinates: "$coordinate",
+          },
+        },
+      ])
+
+    console.log({ result }, "migration completed")
+  },
+
+  // async down(db) {
+  // },
```
dolcalmi, line 18: please remove

dolcalmi (dismissed):

bodymindarts (approved):

src/servers/graphql-server.ts
nicolasburtey: every resolver get the `context` object, so this is not practical. this should only be used for the main graphql server. but the refactor can wait until api v1 is being deprecated/server removed.

src/graphql/admin/root/mutation/business-update-map-info.ts
nicolasburtey: good point

nicolasburtey:
> that was true until the new error map was introduced (this applies for V2, for v1 is not necessary)... now should be something like:
for graphql input validations (line 14 repace input by phone)
if (input instanceof Error) {
}
and for returns from app layer (line 19 replace result with account)
  return { errors: [{ message: appErr.message }] }
}

ok I'll work on rewriting all those errors @dolcalmi

src/graphql/main/queries.ts
nicolasburtey: file renamed.

`userDefaultWalletId -> accountDefaultWalletId` will need a migration from the mobile app AFAIK

so my intent is to do it later in another PR because we will need both fields for some time during the transition

src/graphql/admin/root/query/account-details-by-phone.ts
dolcalmi: since there are many mutations/queries missing mapError I think is better to do it in another PR

src/graphql/admin/root/query/account-details-by-username.ts
dolcalmi: in general queries return object doesn't include `errors` so, at least in queries, throw is necessary (although for app returned errors is necessary to do the mapping)

src/graphql/admin/root/mutation/account-update-level.ts
```diff
@@ -0,0 +1,45 @@
...
+  resolve: async (_, args) => {
+    const { uid, level } = args.input
+
+    for (const input of [uid, level]) {
+      if (input instanceof Error) {
+        return { errors: [{ message: input.message }] }
+      }
+    }
+
+    const account = await updateAccountLevel({ id: uid, level } as {
```
bodymindarts, line 33: Casting should happen in domain layer or via `checkedToXxx` function. Don't like this TODO - but not a blocker.

src/services/mongoose/accounts.ts
```diff
@@ -76,25 +104,60 @@ export const AccountsRepository = (): IAccountsRepository => {
...
+          coordinates: account.coordinates as Coordinates,
         },
       }))
     } catch (err) {
       return new UnknownRepositoryError(err)
     }
   }
+
+  // currently only used by Admin
+  const update = async ({
```
bodymindarts, line 116: The comment could be encoded in the function name:
```
updateFromAdminUseCase
```

Not a blocker - just a thought.

bodymindarts (approved):

src/graphql/admin/root/query/account-details-by-username.ts
nicolasburtey: but the same object are being used whether it's a `query` of a `mutation`? like in both case we can return an `Account` object for instance? for a query because we fetch account details, for a mutation because we update the account status.

vindard (approved):

PR #766, opened 2021-11-29
Implement new Ln Fee Probe use-case

vindard:
## Description

This PR implements the new feeProbe use-case methods and removes the old `getLightningFee` method from `core`.


src/app/lightning/get-lightning-fee.ts
```diff
@@ -0,0 +1,80 @@
...
+  LnFeeCalculator,
+  RouteNotFoundError,
+} from "@domain/bitcoin/lightning"
+import { LnPaymentRequestZeroAmountRequiredError } from "@domain/errors"
+import { CachedRouteLookupKeyFactory } from "@domain/routes/key-factory"
+import { LndService } from "@services/lnd"
+import { RoutesCache } from "@services/redis/routes"
+
+export const lnInvoiceFeeProbe = async ({
+  paymentRequest,
```
dolcalmi, line 13: this use case needs to add amount validation so we will require here wallePublicId and move this use case to wallets folder.

@nicolasburtey mentions this validation as a possible attack vector and graphql already has the walletId https://github.com/GaloyMoney/galoy/pull/752

src/app/lightning/get-lightning-fee.ts
```diff
@@ -0,0 +1,80 @@
...
+  decodeInvoice,
+  LnFeeCalculator,
+  RouteNotFoundError,
+} from "@domain/bitcoin/lightning"
+import { LnPaymentRequestZeroAmountRequiredError } from "@domain/errors"
+import { CachedRouteLookupKeyFactory } from "@domain/routes/key-factory"
+import { LndService } from "@services/lnd"
+import { RoutesCache } from "@services/redis/routes"
+
+export const lnInvoiceFeeProbe = async ({
```
dolcalmi, line 12: please dont use graphql naming conventions in app, domain or service layer

src/app/lightning/get-lightning-fee.ts
```diff
@@ -0,0 +1,80 @@
...
+}): Promise<Satoshis | ApplicationError> => {
+  const decodedInvoice = await decodeInvoice(paymentRequest)
+  if (decodedInvoice instanceof Error) return decodedInvoice
+  const paymentAmount = checkedToSats(decodedInvoice.amount || 0)
+  if (paymentAmount instanceof Error) return paymentAmount
+
+  return feeProbe({ decodedInvoice, paymentAmount })
+}
+
+export const lnNoAmountInvoiceFeeProbe = async ({
```
dolcalmi, line 25: same here (verb should be first)

src/app/lightning/get-lightning-fee.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  const lndService = LndService()
+  if (lndService instanceof Error) throw lndService
+  if (lndService.isLocal(destination)) {
+    return toSats(0)
+  }
+
+  const key = CachedRouteLookupKeyFactory().create({
+    paymentHash,
+    milliSats: milliSatsAmount,
```
dolcalmi, line 63: what happen if is no amount invoice? we need to use paymentAmount

src/app/lightning/get-lightning-fee.ts
```diff
@@ -0,0 +1,80 @@
...
+  })
+  const routeFromCache = await RoutesCache().findByKey(key)
+  const validCachedRoute = !(routeFromCache instanceof Error)
+  if (validCachedRoute) return toSats(routeFromCache.route.fee)
+
+  const maxFee = LnFeeCalculator().max(paymentAmount)
+
+  const rawRoute = await lndService.invoiceProbeForRoute({ decodedInvoice, maxFee })
+  if (rawRoute instanceof RouteNotFoundError) return rawRoute
+  if (rawRoute instanceof Error) return rawRoute
```
dolcalmi, line 73: you are not doing error parsing here so having both is unnecessary

src/domain/bitcoin/lightning/index.types.d.ts
```diff
@@ -104,6 +105,24 @@ interface ILightningService {
...
+  noAmountInvoiceProbeForRoute({
+    decodedInvoice,
+    maxFee,
+    amount,
+  }: {
+    decodedInvoice: LnInvoice
+    maxFee: Satoshis
+    amount: Satoshis
+  }): Promise<RawRoute | LightningServiceError>
+
```
dolcalmi, line 125: same issue with naming convention

src/graphql/root/mutation/ln-invoice-fee-probe.ts
```diff
@@ -25,17 +27,18 @@ const LnInvoiceFeeProbeMutation = GT.Field({
...
-        amount: feeSatAmount,
-      }
-    } catch (err) {
-      return { errors: [{ message: err.message }] }
+    const feeSatAmount = await lnInvoiceFeeProbe({
+      paymentRequest,
+    })
+    if (feeSatAmount instanceof Error) {
+      const appErr = mapError(feeSatAmount)
+      return { errors: [{ message: appErr.message || appErr.name }] } // TODO: refine error
```
dolcalmi, line 35: `|| appErr.name` is unnecessary (was added to `mapError` function)

src/graphql/root/mutation/ln-noamount-invoice-fee-probe.ts
```diff
@@ -27,18 +29,16 @@ const LnNoAmountInvoiceFeeProbeMutation = GT.Field({
...
-      return {
-        errors: [],
-        amount: feeSatAmount,
-      }
-    } catch (err) {
-      return { errors: [{ message: err.message }] }
+    const feeSatAmount = await lnNoAmountInvoiceFeeProbe({ amount, paymentRequest })
+    if (feeSatAmount instanceof Error) {
+      const appErr = mapError(feeSatAmount)
+      return { errors: [{ message: appErr.message || appErr.name }] } // TODO: refine error
```
dolcalmi, line 35: `|| appErr.name` is unnecessary (was added to `mapError` function)

src/servers/graphql-old-server.ts
```diff
@@ -374,7 +379,24 @@ const resolvers = {
...
+      getFee: async ({ amount, invoice }) => {
+        let feeSatAmount: Satoshis | ApplicationError
+        if (amount && amount > 0) {
+          feeSatAmount = await lnNoAmountInvoiceFeeProbe({
+            amount,
+            paymentRequest: invoice,
+          })
+          if (!(feeSatAmount instanceof Error)) return feeSatAmount
+          if (!(feeSatAmount instanceof LnPaymentRequestZeroAmountRequiredError))
+            throw mapError(feeSatAmount)
```
dolcalmi, line 391: not sure if I am missing something but maybe is cleaner to have something similar to `lnInvoiceFeeProbe` 
```
if (feeSatAmount instanceof Error) throw mapError(feeSatAmount)
return feeSatAmount
```

src/services/lnd/index.ts
```diff
@@ -51,6 +53,101 @@ export const LndService = (): ILightningService | LightningServiceError => {
...
+        "Invalid amount passed to method for invoice with no amount. Expected a valid amount to be passed.",
+      )
+    return _probeForRoute({
+      decodedInvoice,
+      maxFee,
+      amount,
+    })
+  }
+
+  const _probeForRoute = async ({
```
dolcalmi, line 94: this is a private method but here is not necessary to `mark` with `_` the method because is not part of a calls and we are not exporting it. please just use `probeForRoute`

src/services/lnd/index.ts
```diff
@@ -51,6 +53,101 @@ export const LndService = (): ILightningService | LightningServiceError => {
...
+              base_fee_mtokens: hop.baseFeeMTokens
+                ? parseFloat(hop.baseFeeMTokens)
+                : undefined,
+              channel: hop.channel,
+              cltv_delta: hop.cltvDelta,
+              fee_rate: hop.feeRate,
+              public_key: hop.nodePubkey,
+            }),
+          )
+          routes.push(rawRoute)
```
dolcalmi, line 119: if we dont have additional logic please use map

dolcalmi (changes requested):

src/app/lightning/get-lightning-fee.ts
vindard: Fixed https://github.com/GaloyMoney/galoy/pull/766/commits/1f3051187030ee5597d048cf5ea12c9b1c7100a4

src/servers/graphql-old-server.ts
vindard: I believe with the current code we want the function to continue on if it's an error that isn't a `LnPaymentRequestZeroAmountRequiredError` error.

With the suggested change the function would stop here on all errors.

src/app/lightning/get-lightning-fee.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  const routeToCache = { pubkey: lndService.defaultPubkey(), route: rawRoute }
+  const cachedRoute = await RoutesCache().store({ key, routeToCache })
+  if (cachedRoute instanceof Error) return cachedRoute
+
+  return toSats(rawRoute.fee)
+}
+
+const noAmountProbeForFee = async ({
```
nicolasburtey, line 81: shouldn't this be dedupe from `feeProbe`?

src/app/lightning/get-lightning-fee.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  const lndService = LndService()
+  if (lndService instanceof Error) throw lndService
+  if (lndService.isLocal(destination)) {
+    return toSats(0)
+  }
+
+  const key = CachedRouteLookupKeyFactory().create({
+    paymentHash,
+    milliSats: toMilliSatsFromNumber(paymentAmount),
```
nicolasburtey, line 98: is this the only difference from `lnNoAmountInvoiceProbeForFee`?

src/graphql/error-map.ts
```diff
@@ -105,7 +106,14 @@ export const mapError = (error: ApplicationError): CustomApolloError => {
         "Too many onchain addresses creation, please wait for a while and try again."
       return new TooManyRequestError({ message, logger: baseLogger })
 
+    case "UnknownRouteNotFoundError":
     case "UnknownLnInvoiceDecodeError":
```
nicolasburtey, line 110: I don't think the message for UnknownLnInvoiceDecodeError should be `""Unable to find a route for payment"`

src/app/lightning/get-lightning-fee.ts
vindard: This and the call to `lndService.noAmountInvoiceProbeForRoute`. I started de-duping but these two differences are probably enough of a difference to signal separate use-cases.

The heuristic we've been using is to not try to de-dupe separate use-cases, and to not de-dupe until the 3rd duplication of code (guided by @bodymindarts)

src/app/lightning/get-lightning-fee.ts
vindard: See https://github.com/GaloyMoney/galoy/pull/766#discussion_r759399286

src/servers/graphql-old-server.ts
```diff
@@ -374,7 +379,24 @@ const resolvers = {
...
       },
-      getFee: async ({ amount, invoice }) => wallet.getLightningFee({ amount, invoice }),
+      getFee: async ({ amount, invoice }) => {
+        let feeSatAmount: Satoshis | ApplicationError
+        if (amount && amount > 0) {
+          feeSatAmount = await lnNoAmountInvoiceProbeForFee({
+            amount,
+            paymentRequest: invoice,
+          })
+          if (!(feeSatAmount instanceof Error)) return feeSatAmount
```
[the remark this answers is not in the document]
vindard: Yup, but it's old-server code that we'll likely be deprecating soon.

The intention is to try noAmountProbe first and if it fails with a specific error to then do amountProbe. Open to suggestions on how this can be cleaned up.

src/app/lightning/get-lightning-fee.ts
vindard: De-dupe discussion resolved in dev call yesterday

src/domain/bitcoin/lightning/index.types.d.ts
vindard: How do you mean? `invoiceProbeForRoute` seems to be a verb that already doesn't exist at graphql layer

src/app/lightning/get-lightning-fee.ts
vindard: Added the user balance check and moved to `wallets` dir

src/graphql/error-map.ts
```diff
@@ -105,7 +106,21 @@ export const mapError = (error: ApplicationError): CustomApolloError => {
...
+    case "InsufficientBalanceForRoutingError":
+      message = "Insufficient balance for finding a route."
+      return new RouteFindingError({ message, logger: baseLogger })
+
+    case "UnknownRouteNotFoundError":
+      message = "Unknown error occurred when trying to find a route for payment."
+      return new RouteFindingError({ message, logger: baseLogger })
+
     case "UnknownLnInvoiceDecodeError":
+      // Consider using `error.message` somehow since lib returns semi-sensible details
```
nicolasburtey, line 122:
```suggestion
      // TODO: Consider using `error.message` somehow since lib returns semi-sensible details
```

nicolasburtey: TODO and FIXME are 2 keywords that would be automatically highlighted in many IDE. so when there are things left to do in later PR it's always good to add the comment in code directly (typically better than adding a related issues that get lost eventually)

src/graphql/root/mutation/ln-invoice-fee-probe.ts
```diff
@@ -25,17 +27,19 @@ const LnInvoiceFeeProbeMutation = GT.Field({
...
-      })
-      // TODO: validate feeSatAmount
-      return {
-        errors: [],
-        amount: feeSatAmount,
-      }
-    } catch (err) {
-      return { errors: [{ message: err.message }] }
+    const feeSatAmount = await lnInvoiceProbeForFee({
+      walletPublicId: walletId,
```
nicolasburtey, line 31: doesn't sound right to attach walletPublicId to a walletId because as type they're 2 different things?

src/graphql/root/mutation/ln-noamount-invoice-fee-probe.ts
```diff
@@ -27,18 +29,20 @@ const LnNoAmountInvoiceFeeProbeMutation = GT.Field({
...
-      })
-      // TODO: validate feeSatAmount
-      return {
-        errors: [],
-        amount: feeSatAmount,
-      }
-    } catch (err) {
-      return { errors: [{ message: err.message }] }
+    const feeSatAmount = await lnNoAmountInvoiceProbeForFee({
+      walletPublicId: walletId,
```
nicolasburtey, line 33: same comment as above

src/services/lnd/index.ts
```diff
@@ -316,3 +421,10 @@ const lookupPaymentByPubkeyAndHash = async ({
     return new UnknownLightningServiceError(err)
   }
 }
+
+const parseLndErrorDetails = (err) =>
```
nicolasburtey, line 425: note: this was probably a hacky script I came along with? maybe this could be improve. or at least add a FIXME to it.

src/graphql/root/mutation/ln-invoice-fee-probe.ts
vindard: This is a little strange part of our codebase. We have:

**Domain**
`WalletId`: internal db id
`WalletPublicId`: uuid4 generated wallet id

**GraphQL**
`WalletId`: uuid4 generated wallet id that maps to domain `WalletPublicId`

src/graphql/root/mutation/ln-noamount-invoice-fee-probe.ts
vindard: https://github.com/GaloyMoney/galoy/pull/766#discussion_r760603642

src/services/lnd/index.ts
vindard: This is how I would've done it too 🙈

src/graphql/root/mutation/ln-invoice-fee-probe.ts
nicolasburtey: We should probably “fix” it by aligning the same from graphql to domain. Otherwise this will confuse everyone working on our codebase

vindard: cc @samerbuna @bodymindarts @dolcalmi

src/app/wallets/ln-probe-for-fee.ts
```diff
@@ -0,0 +1,166 @@
...
+export const lnInvoiceProbeForFee = async ({
+  walletPublicId,
+  paymentRequest,
+  logger,
+}: {
+  walletPublicId: WalletPublicId
+  paymentRequest: EncodedPaymentRequest
+  logger: Logger
+}): Promise<Satoshis | ApplicationError> => {
+  const decodedInvoice = await decodeInvoice(paymentRequest)
```
dolcalmi, line 26: await is not needed, `decodeInvoice` is not an async function

src/app/wallets/ln-probe-for-fee.ts
```diff
@@ -0,0 +1,166 @@
...
+  if (decodedInvoice instanceof Error) return decodedInvoice
+
+  const { amount: lnInvoiceAmount } = decodedInvoice
+  if (lnInvoiceAmount && lnInvoiceAmount > 0) {
+    return new LnPaymentRequestZeroAmountRequiredError()
+  }
+
+  if (!(amount && amount > 0)) {
+    return new SatoshiAmountRequiredError()
+  }
```
dolcalmi, line 55: checkedToSats in line 57 does this validation

src/domain/bitcoin/lightning/index.types.d.ts
dolcalmi: the methods are returning routes so could be something like `findRouteForInvoice` and `findRouteForNoAmountInvoice`, having invoice and no amount invoice at the beginning does not match the other methods (used naming convention) in the service

dolcalmi (changes requested):

dolcalmi (dismissed):

src/graphql/root/mutation/ln-invoice-fee-probe.ts
nicolasburtey: https://github.com/GaloyMoney/galoy/issues/781

bodymindarts: Agree with aligning naming as much as possible. WalletPublicId or just `publicId` in domain makes sense.

src/app/wallets/get-lightning-fee.ts
```diff
@@ -0,0 +1,161 @@
...
+  const balance = await getBalanceForWallet({ walletId, logger })
+  if (balance instanceof Error) return balance
+  if (balance < paymentAmount) {
+    return new InsufficientBalanceError(
+      `Payment amount '${paymentAmount}' exceeds balance '${balance}'`,
+    )
+  }
+
+  const lndService = LndService()
+  if (lndService instanceof Error) throw lndService
```
nicolasburtey, line 83: should we only throw from the graphql layer @dolcalmi ?

src/app/wallets/get-lightning-fee.ts
```diff
@@ -0,0 +1,161 @@
...
```
nicolasburtey, line 133: should we only throw from the graphql layer?

src/app/wallets/get-lightning-fee.ts
```diff
@@ -0,0 +1,161 @@
...
+  if (lndService.isLocal(destination)) {
+    return toSats(0)
+  }
+
+  const key = CachedRouteLookupKeyFactory().create({
+    paymentHash,
+    milliSats: toMilliSatsFromNumber(paymentAmount),
```
nicolasburtey, line 140: what is the associated path for noInvoicePayment that is using the cached route?

a quick search only lead me to https://github.com/GaloyMoney/galoy/blob/b98adfcb4cdbb607cac0bc913ba6274bbe0c0660/src/app/wallets/ln-send-payment.ts#L464, but this assume the invoice is not amountLess.

Could this be the issue I have seen regarding the cache/probe that is not been used when I try paying an invoice? (and maybe you only try reproducing with invoiceFull amount?)

src/app/wallets/get-lightning-fee.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  const maxFee = LnFeeCalculator().max(paymentAmount)
+
+  const rawRoute = await lndService.findRouteForNoAmountInvoice({
+    decodedInvoice,
+    maxFee,
+    amount: paymentAmount,
+  })
+  if (rawRoute instanceof RouteNotFoundError) return rawRoute
+  if (rawRoute instanceof Error) return rawRoute
```
nicolasburtey, line 154: is there any benefits to keeping those both lines, versus just using the latter?

src/app/wallets/get-lightning-fee.ts
vindard: Oh this is a good catch, we should ideally be using `amount` there instead of `decodedInvoice.milliSatsAmount` to also have unique caches for each different amount that may be attempted on a noAmount invoice.

vindard: Done https://github.com/GaloyMoney/galoy/pull/766/commits/fe518b1a8570a7a455fa1c1c71b436b85f3e2fc3

src/app/wallets/get-lightning-fee.ts
vindard: Oversight here from earlier version of the PR, fixed

dolcalmi (approved):

nicolasburtey (approved):

PR #767, opened 2021-11-29
fix msBeforeNext

nicolasburtey:
msBeforeNext was from yaml was not passed correctly
also improve label for default.yaml

### DIFF
--- a/src/config/app.ts
+++ b/src/config/app.ts
@@ -143,24 +143,30 @@ const getRateLimits = (config): RateLimitOptions => {
    * Returns a subset of the required parameters for the
    * 'rate-limiter-flexible.RateLimiterRedis' object.
    */
-  return {
+  const rateLimitOptions: RateLimitOptions = {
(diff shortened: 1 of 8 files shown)


src/config/app.ts
```diff
@@ -143,24 +143,30 @@ const getRateLimits = (config): RateLimitOptions => {
    * Returns a subset of the required parameters for the
    * 'rate-limiter-flexible.RateLimiterRedis' object.
    */
-  return {
+  const obj: RateLimitOptions = {
```
dolcalmi, line 146: I know is a tmp variable but maybe `rateLimitOptions` or just `opts` is better than `obj`

dolcalmi (dismissed):
just minor comment but approved anyway

dolcalmi (approved):

bodymindarts (approved):
This will also need updating in charts and deployments repos

PR #771, opened 2021-12-01
Convert 'address' in graphql to allow null

vindard:
### Description

We have a "legacy" type for onchain transaction where the 'address' property wasn't previously recorded in the database. This means that at the graphql level this property can't be non-null until we either figure out a way to backfill this data or place a non-null placeholder value in there instead.

_**Note:** We have similar issues with `counterPartyWalletId` and `paymentSecret` properties where these have also been made non-null for now until we migrate the db, see #759._

### DIFF
--- a/src/domain/wallets/tx-history.ts
+++ b/src/domain/wallets/tx-history.ts
@@ -108,6 +108,8 @@ export const fromLedger = (
         txType = ExtendedLedgerTransactionType.LnIntraLedger
       }
 
+      const defaultOnChainAddress = "<no-address>" as OnChainAddress
+
       let walletTransaction: WalletTransaction
       switch (txType) {
         case ExtendedLedgerTransactionType.IntraLedger:
@@ -131,7 +133,7 @@ export const fromLedger = (
             ...baseTransaction,
             initiationVia: {
               type: PaymentInitiationMethod.OnChain,
-              address,
+              address: address || defaultOnChainAddress,
             },
             settlementVia: {
               type: SettlementMethod.IntraLedger,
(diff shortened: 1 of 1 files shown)


nicolasburtey (dismissed):

src/graphql/types/abstract/initiation-via.ts
```diff
@@ -35,7 +35,8 @@ const InitiationViaOnChain = new GT.Object({
   isTypeOf: (source) => source.type === PaymentInitiationMethod.OnChain,
   fields: () => ({
     address: {
-      type: GT.NonNull(OnChainAddress),
+      // type: GT.NonNull(OnChainAddress),
```
nicolasburtey, line 38:
```suggestion
      // FIXME(after migration to fill the gap): type: GT.NonNull(OnChainAddress),
```

nicolasburtey (approved):

bodymindarts:
We don’t need to bankroll the database. We can just return a default in the graphql layer if it’s null. That would preserve the type we want at the GeaphQL layer

vindard:
Changed the code to return a default address at the domain layer instead of changing the gql interface based on @bodymindarts comments

samerbuna (approved):

nicolasburtey:
> It should definitely be a real address AND one we control. Just incase someone sends us money again based on seeing the address.

yeah maybe a better idea than a sink valid bitcoin address. but at the same time, we don't want to start monitoring an address that is not attached to a user. so maybe the not valid address is a idea after all

nicolasburtey (approved):

PR #772, opened 2021-12-01
Run `yarn install` before tests

jireva:
Ensures that tests will work when switching branches.


bodymindarts:
IMO yarn install adds a lot to the runtime of testing. Should be a manually 1-off rather than part of the test command.

PR #777, opened 2021-12-02
chore: run github workflows on main branch

bodymindarts:
This is somewhat redundant as all commits that land on main were already tested with the merged result of their respective PR - but this way ppl can see the current result of the tests on main GitHub page.

### DIFF
--- a/.github/workflows/check-code.yml
+++ b/.github/workflows/check-code.yml
@@ -1,5 +1,7 @@
 name: "Check code"
 on:
+  push:
+    branches: [ main ]
   pull_request:
     branches: [ main ]
 jobs:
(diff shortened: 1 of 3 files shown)


nicolasburtey:
are those tests not run on the concourse simultaneously? can't we just have the link for those tests instead? 

specially with the false positive we have on the integration test, I was more thinking as a link to the tests run on concourse than running those tests both in github actions + concourse

nicolasburtey:
> I am not aware of a way to make the result of concourse jobs show up in GitHub UI.

this is a pretty standard features (ie: CircleCI, GitGuardian does it out of the box)

bodymindarts:
> > I am not aware of a way to make the result of concourse jobs show up in GitHub UI.
> 
> this is a pretty standard features (ie: CircleCI, GitGuardian does it out of the box)

Those are hosted services. Concourse is self-hosted.

PR #778, opened 2021-12-03
Docker compose v2

JimenezWeerden:
- Make local setup work with docker-compose v2


JimenezWeerden:
@bodymindarts Here's the PR

bodymindarts:
Thanks - this branch is too noisy. Will curate another one.

PR #782, opened 2021-12-04
chore: make it easier to understand the diff of a .graphql

nicolasburtey:
by computing another generated file which is lexicographic based
to remove the non deterministic aspect of the .graphql file creation

still keeping the standard schema.graphql as this is the one
that make most sense for an API user

### DIFF
--- a/src/graphql/admin/index.ts
+++ b/src/graphql/admin/index.ts
@@ -1,4 +1,4 @@
-import { GraphQLSchema, printSchema } from "graphql"
+import { GraphQLSchema, lexicographicSortSchema, printSchema } from "graphql"
 
 import { ALL_INTERFACE_TYPES } from "@graphql/types"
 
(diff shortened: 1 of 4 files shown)


bodymindarts:
Haven’t looked at the diff but just a small comment. This does not qualify as a feature on the sense of semantic versioning. IMO it’s a chore or improvement.

nicolasburtey:
> Haven’t looked at the diff but just a small comment. This does not qualify as a feature on the sense of semantic versioning. IMO it’s a chore or improvement.
> […](#)
> Sent from my iPhone
> On 4. Dec 2021, at 15:26, nicolasburtey ***@***.***> wrote: ﻿ @nicolasburtey requested your review on: #782 feat: make it easier to understand the diff of a .graphql. — You are receiving this because your review was requested. Reply to this email directly, view it on GitHub, or unsubscribe. Triage notifications on the go with GitHub Mobile for iOS or Android.

fixed

samerbuna:
This SDL looks much better. I'd say let's just use it for the main schema files and not generate a different file.

bodymindarts (approved):
Just as a reminder we need to add a failing condition to the GitHub action integration test that fails if there is a diff post run.

dolcalmi (approved):
I am approving but I agree with @samerbuna.

nicolasburtey:
updated with your reco @samerbuna @dolcalmi

samerbuna (approved):

Issue #783, opened 2021-12-04
Push notifications in multiple devices


dolcalmi:
when a user log in in a new device the notifications are sent to the old one.

We should guarantee that only 1 device id/token is related to 1 user 

Possible solutions:
- Remove previous device id/tokens when a user log in to its account
- Remove all device id/tokens related to a user when logouts

nicolasburtey:
problem still present

PR #784, opened 2021-12-04
fix: send push notification only to the last registered device token

dolcalmi:
- Only send push notifications to the last registered device
- Partial fix to https://github.com/GaloyMoney/galoy/issues/783
- Needs to be tested in staging (important before prod deployment)


samerbuna:
I don't think we should do this. if a user is logged-in to the app from multiple devices, we should notify all the devices.

PR #785, opened 2021-12-05
refactor: replace getOnChainTransactions dependency

dolcalmi:
this is the 1st split of the on-chain pay use case refactor

Includes:
- Removes getOnchainReceipt
- Replace getOnChainTransactions
- Renames getIncomingTransactions to match repos standard (list prefix)

### DIFF
--- a/src/config/app.ts
+++ b/src/config/app.ts
@@ -3,6 +3,7 @@ import yaml from "js-yaml"
 import merge from "lodash.merge"
 
 import { baseLogger } from "@services/logger"
+import { checkedToScanDepth } from "@domain/bitcoin/onchain"
 
 const defaultContent = fs.readFileSync("./default.yaml", "utf8")
 export const defaultConfig = yaml.load(defaultContent)
@@ -67,10 +68,18 @@ export const MEMO_SHARING_SATS_THRESHOLD = yamlConfig.limits.memoSharingSatsThre
 
 export const ONCHAIN_MIN_CONFIRMATIONS = yamlConfig.onChainWallet.minConfirmations
 // how many block are we looking back for getChainTransactions
-export const ONCHAIN_LOOK_BACK = yamlConfig.onChainWallet.lookBack
-export const ONCHAIN_LOOK_BACK_OUTGOING = yamlConfig.onChainWallet.lookBackOutgoing
-export const ONCHAIN_LOOK_BACK_CHANNEL_UPDATE =
-  yamlConfig.onChainWallet.lookBackChannelUpdate
+const getOnChainScanDepth = (val: number): ScanDepth => {
+  const scanDepth = checkedToScanDepth(val)
+  if (scanDepth instanceof Error) throw scanDepth
+  return scanDepth
+}
+export const ONCHAIN_SCAN_DEPTH = getOnChainScanDepth(yamlConfig.onChainWallet.scanDepth)
+export const ONCHAIN_SCAN_DEPTH_OUTGOING = getOnChainScanDepth(
+  yamlConfig.onChainWallet.scanDepthOutgoing,
+)
+export const ONCHAIN_SCAN_DEPTH_CHANNEL_UPDATE = getOnChainScanDepth(
+  yamlConfig.onChainWallet.scanDepthChannelUpdate,
+)
 
 export const USER_ACTIVENESS_MONTHLY_VOLUME_THRESHOLD =
   yamlConfig.userActivenessMonthlyVolumeThreshold
--- a/src/core/specter-wallet.ts
+++ b/src/core/specter-wallet.ts
@@ -8,7 +8,7 @@ import { ledger } from "@services/mongodb"
 
 import { UserWallet } from "./user-wallet"
 import { btc2sat, sat2btc } from "./utils"
-import { BTC_NETWORK, ONCHAIN_LOOK_BACK_OUTGOING } from "@config/app"
+import { BTC_NETWORK, ONCHAIN_SCAN_DEPTH_OUTGOING } from "@config/app"
 import { toSats } from "@domain/bitcoin"
 import { OnChainService } from "@services/lnd/onchain-service"
 import { TxDecoder } from "@domain/bitcoin/onchain"
@@ -208,9 +208,9 @@ export class SpecterWallet {
       const onChainService = OnChainService(TxDecoder(BTC_NETWORK))
       if (onChainService instanceof Error) return toSats(0)
 
-      const onChainTxFee = await onChainService.findOnChainFee({
+      const onChainTxFee = await onChainService.lookupOnChainFee({
         txHash,
-        scanDepth: ONCHAIN_LOOK_BACK_OUTGOING,
+        scanDepth: ONCHAIN_SCAN_DEPTH_OUTGOING,
       })
       if (onChainTxFee instanceof Error) return toSats(0)
 
--- a/src/core/specter-wallet.ts
+++ b/src/core/specter-wallet.ts
@@ -6,9 +6,12 @@ import { bitcoindDefaultClient, BitcoindWalletClient } from "@services/bitcoind"
 import { getActiveOnchainLnd, lndsBalances } from "@services/lnd/utils"
 import { ledger } from "@services/mongodb"
 
-import { getOnChainTransactions } from "./on-chain"
 import { UserWallet } from "./user-wallet"
 import { btc2sat, sat2btc } from "./utils"
+import { BTC_NETWORK, ONCHAIN_LOOK_BACK_OUTGOING } from "@config/app"
+import { toSats } from "@domain/bitcoin"
+import { OnChainService } from "@services/lnd/onchain-service"
+import { TxDecoder } from "@domain/bitcoin/onchain"
 
 // TODO no longer used in tests, removing the creation of the default wallet didn't break anything
 const staticClient = ""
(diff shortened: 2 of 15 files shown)


src/services/lnd/onchain-service.ts
```diff
@@ -27,26 +28,42 @@ export const OnChainService = (
...
+    scanDepth: number,
+  ): Promise<SubmittedTransaction[] | OnChainServiceError> => {
+    const txs = await listTransactions(scanDepth)
+    if (txs instanceof Error) return txs
+
+    return extractIncomingTransactions({ decoder, txs })
+  }
+
+  const listOutgoingTransactions = async (
+    scanDepth: number,
```
nicolasburtey, line 59: this could be its own type

src/services/lnd/onchain-service.ts
```diff
@@ -27,26 +28,42 @@ export const OnChainService = (
...
-
-      return extractIncomingTransactions(decoder, result)
     } catch (err) {
       return new UnknownOnChainServiceError(err)
     }
   }
 
+  const listIncomingTransactions = async (
+    scanDepth: number,
+  ): Promise<SubmittedTransaction[] | OnChainServiceError> => {
```
nicolasburtey, line 51: in my mind `SubmittedTransaction` suggested more an intent of our node to send/submit transaction (ie: outgoing transaction) versus transaction coming to us.

src/services/lnd/onchain-service.ts
```diff
@@ -82,17 +110,21 @@ export const OnChainService = (
...
-  decoder: TxDecoder,
-  { transactions }: GetChainTransactionsResult,
-): SubmittedTransaction[] => {
-  return transactions
+export const extractIncomingTransactions = ({
+  decoder,
+  txs,
+}: {
+  decoder: TxDecoder
+  txs: GetChainTransactionsResult
```
nicolasburtey, line 125: does the `Get` in the `GetChainTransactionsResult` make sense in that context, in term of type name?

src/services/lnd/onchain-service.ts
```diff
@@ -104,3 +136,23 @@ export const extractIncomingTransactions = (
         }),
     )
 }
+
+export const extractOutgoingTransactions = ({
```
nicolasburtey, line 140: shoundn't this belong to domain?

src/services/lnd/onchain-service.ts
```diff
@@ -62,6 +79,17 @@ export const OnChainService = (
     }
   }
 
+  const findOnChainFee = async ({
```
nicolasburtey, line 82: shouldn't this belong to domain?

nicolasburtey (commented):
couple of questions, mainly around naming

src/services/lnd/onchain-service.ts
dolcalmi: `GetChainTransactionsResult` is a type from `lightning` library

src/services/lnd/onchain-service.ts
dolcalmi: :thinking: yep although this is just a refactor (this type was not introduced in this PR). what name do you suggest? (we already have `OnChainTransaction` and is part of this type https://github.com/GaloyMoney/galoy/blob/main/src/domain/bitcoin/onchain/index.types.d.ts#L25)

src/services/lnd/onchain-service.ts
dolcalmi: this methods are like `userFromRaw`, `translateToLedgerTx` or `userIPsFromRaw` (internal implementation of service) the difference here is that are exported  to do unit testing.

src/services/lnd/onchain-service.ts
dolcalmi: this is from service, I encapsulated the use of internal lightning library using listOutgoingTransactions but this is an specific implementation (after on-chain migration probably will be queried directly to bitcoind)

src/services/lnd/onchain-service.ts
nicolasburtey: `IncomingTransaction[]`?

src/services/lnd/onchain-service.ts
nicolasburtey: note: we got the input that bitcoind may not scale well in term of how many onchain address it can monitor (it's made as a individual wallet, not an entreprise that could monitor millions of addresses), so not sure we'll migrate anytime soon

src/services/lnd/onchain-service.ts
nicolasburtey: there are some logic in here with the `filter`, so in my mind, this is not just a mapping

dolcalmi: that could be the case if we add a generic listTransactions in domain interface but we only have `listIncomingTransactions` src/domain/bitcoin/onchain/index.types.d.ts (how that is implemented is responsability of the service).

nicolasburtey: probably not really needed as of now

src/domain/bitcoin/onchain/index.types.d.ts
```diff
@@ -44,11 +44,21 @@ type TxFilter = {
...
+  scanDepth: number
+}
+
 interface IOnChainService {
-  getIncomingTransactions(
+  listIncomingTransactions(
     scanDepth: number,
   ): Promise<SubmittedTransaction[] | OnChainServiceError>
 
+  findOnChainFee({
```
bodymindarts, line 57: `find` seems weird as a prefix here.
Find implies it could be there - or not (like `repo.findById`)... here its more a `lookup` or something like that.

src/services/lnd/onchain-service.ts
bodymindarts: Hmm I think the `SubmittedTransaction` name isn't great. Its intended to signify a tx that can be in the mempool or confirmed. From looking through the code I think `IncomingTransaction` (or `IncomingOnChainTransaction`) works everywhere its being used.

src/services/lnd/onchain-service.ts
bodymindarts: True - when there is a primitive type that naturally describes the value I don't mind using the primitive type. Here scan depth really is a number. But I guess we could add additional constraints (`x > 0 && < maxDepth`) which would merit the type.

bodymindarts (approved):

src/services/lnd/onchain-service.ts
dolcalmi: change done here https://github.com/GaloyMoney/galoy/pull/787 to avoid a bigger PR

src/services/lnd/onchain-service.ts
dolcalmi: done https://github.com/GaloyMoney/galoy/commit/f308efa7545cfab98b612b82ecedb06c0ada099f

src/domain/bitcoin/onchain/index.types.d.ts
dolcalmi: changed

vindard (dismissed):
Looks ok to me now, tested with the `OnChainPaymentSend` as well

vindard (approved):

src/config/app.ts
```diff
@@ -67,10 +68,18 @@ export const MEMO_SHARING_SATS_THRESHOLD = yamlConfig.limits.memoSharingSatsThre
 
 export const ONCHAIN_MIN_CONFIRMATIONS = yamlConfig.onChainWallet.minConfirmations
 // how many block are we looking back for getChainTransactions
-export const ONCHAIN_LOOK_BACK = yamlConfig.onChainWallet.lookBack
-export const ONCHAIN_LOOK_BACK_OUTGOING = yamlConfig.onChainWallet.lookBackOutgoing
-export const ONCHAIN_LOOK_BACK_CHANNEL_UPDATE =
-  yamlConfig.onChainWallet.lookBackChannelUpdate
+const getOnChainScanDepth = (val: number): ScanDepth => {
+  const scanDepth = checkedToScanDepth(val)
```
bodymindarts, line 72: Just a comment - not a blocker. I wish we could do a complete config check at startup to fail early and not when these values get read the first time at runtime.

src/domain/bitcoin/onchain/index.types.d.ts
```diff
@@ -22,14 +25,17 @@ type OnChainTransaction = {
...
-type IncomingOnChainTransaction = {
+type BaseOnChainTransaction = {
   confirmations: number
   rawTx: OnChainTransaction
   fee: Satoshis
   createdAt: Date
   uniqueAddresses: () => OnChainAddress[]
 }
 
+type IncomingOnChainTransaction = BaseOnChainTransaction
```
bodymindarts, line 36: This alias will not be type safe as it is not actually changing the signature.

What we want from these types is that you cannot pass an `incomingTx` to a function expecting an `outgoingTx` but I think this doesn't prevent it.

src/config/app.ts
dolcalmi: if you like I can update this in the next PR

bodymindarts (approved):
I'm approving with 2 small comments.

The config thing is another issue but the type thing should probably be fixed soon. Up to you to fix it pre or post merge.

src/core/specter-wallet.ts
```diff
@@ -201,8 +204,20 @@ export class SpecterWallet {
 
     const memo = `deposit of ${sats} sats to the cold storage wallet`
 
-    const outgoingOnchainTxns = await getOnChainTransactions({ lnd, incoming: false })
-    const [{ fee }] = outgoingOnchainTxns.filter((tx) => tx.id === id)
+    const getOnChainFee = async (txHash: OnChainTxHash): Promise<Satoshis> => {
+      const onChainService = OnChainService(TxDecoder(BTC_NETWORK))
+      if (onChainService instanceof Error) return toSats(0)
```
nicolasburtey, line 209: not sure we should return 0 in case of an error?

nicolasburtey: it actually depends if this is for a settleTransaction or for a request for fee from the user. we should probably not return 0 for the latter.

nicolasburtey: I see this is in specter. nevermind.

src/config/app.ts
```diff
@@ -67,10 +68,18 @@ export const MEMO_SHARING_SATS_THRESHOLD = yamlConfig.limits.memoSharingSatsThre
...
+  if (scanDepth instanceof Error) throw scanDepth
+  return scanDepth
```
nicolasburtey, line 74: shouldn't it be `  return scanDepth as ScanDepth`? or Typescript does it automatically?

src/domain/bitcoin/onchain/index.ts
```diff
@@ -23,3 +31,39 @@ export const IncomingOnChainTransaction = ({
...
     }, []),
 })
+
+export const IncomingOnChainTransaction = ({
+  confirmations,
+  rawTx,
+  fee,
+  createdAt,
+}: {
+  confirmations: number
```
nicolasburtey, line 41: could be its own type also. don't think that's critical / could be done in another PR.

PR #786, opened 2021-12-06
Add ln service error switch

vindard:
## Description

This PR is to:
- fix a small leak where domain errors were getting through to the app layer for unhandled errors in our error map (see **Screenshot**)
- to introduce a pattern for parsing and handling known errors from the lnd service library

---

### Screenshot
image

### DIFF
--- a/src/graphql/error-map.ts
+++ b/src/graphql/error-map.ts
@@ -210,6 +210,7 @@ export const mapError = (error: ApplicationError): CustomApolloError => {
     case "CouldNotFindPhoneCodeError":
     case "PhoneProviderServiceError":
     case "UnknownPhoneProviderServiceError":
+      message = `Unknown error occurred (code: ${error.name})`
       return new UnknownClientError({ message, logger: baseLogger })
 
     default:
--- a/src/graphql/error-map.ts
+++ b/src/graphql/error-map.ts
@@ -11,6 +11,7 @@ import {
   ValidationInternalError,
   TooManyRequestError,
   RouteFindingError,
+  PhoneCodeError,
 } from "@graphql/error"
 import { baseLogger } from "@services/logger"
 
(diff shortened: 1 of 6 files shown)


src/graphql/error-map.ts
```diff
@@ -106,6 +107,10 @@ export const mapError = (error: ApplicationError): CustomApolloError => {
         "Too many onchain addresses creation, please wait for a while and try again."
       return new TooManyRequestError({ message, logger: baseLogger })
 
+    case "CouldNotFindPhoneCodeError":
+      message = "Invalid or incorrect phone code entered."
```
nicolasburtey, line 111: do we really need to override the message? maybe we could just set a default message of `PhoneCodeError` as this one?

nicolasburtey (approved):

src/services/lnd/index.ts
```diff
@@ -418,13 +454,20 @@ const lookupPaymentByPubkeyAndHash = async ({
...
-    return new UnknownLightningServiceError(err)
+    const errDetails = parseLndErrorDetails(err)
+    switch (errDetails) {
+      default:
+        return new UnknownLightningServiceError(err)
+    }
   }
 }
 
 const parseLndErrorDetails = (err) =>
```
nicolasburtey, line 465: some documentation could be helpful here (maybe on the way `ln-service`/`lightning` manages error in a non conventional way.

dolcalmi (approved):
looks good, although this 'pattern' could be a generic parse function
```
    const errDetails = parseLndErrorDetails(err)
    switch (errDetails) {
      default:
        return new UnknownLightningServiceError(err)
    }
```

vindard:
> looks good, although this 'pattern' could be a generic parse function

I was thinking that too, but was also thinking there could be cases where the same lnd error could lead to different domain errors e.g. `""insufficient local balance"` from lnd resolving to either `InsufficientBalanceForRoutingError` or `InsufficientBalanceForPaymentError` depending on the service method it is thrown from

src/services/lnd/index.ts
vindard: Added: https://github.com/GaloyMoney/galoy/pull/804

PR #788, opened 2021-12-06
fix: update on-chain receipt to use lock by hash

dolcalmi:
After docker v2 update the integration test  lock issues have been increasing, this fix change lock from wallet to hash for on chain receipts


vindard (approved):
Looks ok to me, corresponds to how we lock by payment hash when updating ln txns. Approving because I can't see any immediate gotchas from switching away from locking with `walletId`

bodymindarts (approved):

PR #791, opened 2021-12-08
refactor: add payToAddress and getBalance to on-chain service

dolcalmi:
this is the 2nd split of the on-chain pay use case refactor

- Move on-chain address validator to domain
- Add payToAddress and getBalance to onchain service
- Update error parsing in on-chain service
- Update getOnChainFeeEstimate with 1 param policy

### DIFF
--- a/src/core/on-chain/index.ts
+++ b/src/core/on-chain/index.ts
@@ -7,12 +7,12 @@ import {
   checkWithdrawalLimits,
 } from "@app/wallets/check-limit-helpers"
 import { BTC_NETWORK, ONCHAIN_SCAN_DEPTH_OUTGOING } from "@config/app"
-import { toSats } from "@domain/bitcoin"
-import { TxDecoder } from "@domain/bitcoin/onchain"
+import { checkedToTargetConfs, toSats, toTargetConfs } from "@domain/bitcoin"
+import { checkedToOnChainAddress, TxDecoder } from "@domain/bitcoin/onchain"
 import { TwoFANewCodeNeededError } from "@domain/twoFA"
+import { WithdrawalFeeCalculator } from "@domain/wallets"
 import { LedgerService } from "@services/ledger"
 import { OnChainService } from "@services/lnd/onchain-service"
-import { getActiveOnchainLnd } from "@services/lnd/utils"
 import { LockService } from "@services/lock"
 import { ledger } from "@services/mongodb"
 import { User } from "@services/mongoose/schema"
@@ -261,7 +261,7 @@ export const OnChainMixin = (superclass) =>
         const targetConfs = getTargetConfirmations()
 
         const estimatedFee = await Wallets.getOnChainFeeByWalletId({
-          walletId: this.user.id,
+          walletId: this.user.walletId,
           amount: checksAmount,
           address: checkedAddress,
           targetConfirmations: targetConfs,
--- a/src/core/on-chain/index.ts
+++ b/src/core/on-chain/index.ts
@@ -17,7 +17,6 @@ import { LockService } from "@services/lock"
 import { ledger } from "@services/mongodb"
 import { User } from "@services/mongoose/schema"
 import { NotificationsService } from "@services/notifications"
-import { getChainBalance, getChainFeeEstimate, sendToChainAddress } from "lightning"
 
 import {
   DustAmountError,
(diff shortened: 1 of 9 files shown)


src/core/on-chain/index.ts
```diff
@@ -306,17 +329,21 @@ export const OnChainMixin = (superclass) =>
...
-          try {
-            ;({ id } = await sendToChainAddress({
-              address,
-              lnd,
-              tokens: amountToSend,
-              utxo_confirmations: 0,
-              target_confirmations: targetConfs,
-            }))
-          } catch (err) {
+          const payToAddress = () => {
```
bodymindarts, line 332: I don't understand the purpose of this variable. Its only being referenced in 1 place and may as well be inline, or?

src/domain/bitcoin/onchain/index.ts
```diff
@@ -1,16 +1,32 @@
-import { InvalidScanDepthAmount } from "@domain/errors"
+import { InvalidOnChainAddress, InvalidScanDepthAmount } from "@domain/errors"
 
 export * from "./errors"
 export * from "./tx-filter"
 export * from "./tx-decoder"
 
+export const checkedToOnChainAddress = (
```
bodymindarts, line 7: Should take an additional `network` arg and only return the addresses that are valid for the specific network.

src/domain/bitcoin/onchain/index.ts
```diff
@@ -42,28 +58,14 @@ export const IncomingOnChainTransaction = ({
...
-  fee: Satoshis
-  createdAt: Date
-}): OutgoingOnChainTransaction =>
-  BaseOnChainTransaction({
-    confirmations,
-    rawTx,
-    fee,
-    createdAt,
-  })
+  uniqueAddresses: () =>
```
nicolasburtey, line 66: a comment to explain what we're doing could be useful. I guess we're trying to get a list of onchain addresses but why do we need a TxOut is is not obvious

src/domain/bitcoin/onchain/index.ts
```diff
@@ -32,7 +48,7 @@ export const BaseOnChainTransaction = ({
     }, []),
 })
 
-export const IncomingOnChainTransaction = ({
+export const OutgoingOnChainTransaction = ({
   confirmations,
   rawTx,
```
nicolasburtey, line 53: why should we pass rawTx to an `OutgoingOnChainTransaction` is not obvious to me. I would expect a function named `OutgoingOnChainTransaction` to return to me a list of transactions.

nicolasburtey:
you want to update this as part of another PR @dolcalmi ? 

https://github.com/GaloyMoney/galoy/blob/e4ab3c0790fe3ceeffa48a5fd2dcdd5e4dfab786/src/graphql/root/mutation/onchain-payment-send.ts#L35

nicolasburtey:
> you want to update this as part of another PR @dolcalmi ?
> 
> https://github.com/GaloyMoney/galoy/blob/e4ab3c0790fe3ceeffa48a5fd2dcdd5e4dfab786/src/graphql/root/mutation/onchain-payment-send.ts#L35

not sure if this is ideal to have api v2 still referring to v1 code

src/core/on-chain/index.ts
dolcalmi: is to avoid to have `onchainLogger.error` twice. This willl be remove in the final PR

src/domain/bitcoin/onchain/index.ts
dolcalmi: I didnt change the original implementation, just matched the previous. If you want we can discuss it in code review and updated in the next PR

dolcalmi:
> you want to update this as part of another PR @dolcalmi ?
> 
> https://github.com/GaloyMoney/galoy/blob/e4ab3c0790fe3ceeffa48a5fd2dcdd5e4dfab786/src/graphql/root/mutation/onchain-payment-send.ts#L35

yes, this is part of a 3 steps migration

src/domain/bitcoin/onchain/index.ts
dolcalmi: this is part of a previous implementation. I guess was done to fix a bug with a deposit but really not sure. @bodymindarts do you remember?

bodymindarts: A transaction has multiple UTXOs - this function returns a list of unique addresses from all the UTXOs in the transaction. Note not every UTXO must send to an address (due to `OP_RETURN`) and some UTXOs could send to the same address (though it would be bad).
Its used to get all wallets that need updating when a tx comes in [here](https://github.com/GaloyMoney/galoy/blob/main/src/app/wallets/update-on-chain-receipt.ts#L40-L41)

bodymindarts (dismissed):

bodymindarts (approved):

nicolasburtey (approved):

PR #792, opened 2021-12-09
feat: ledger to use walletId instead of userId

nicolasburtey:
- remove the hard dependency between userId and walletId for transaction. this is necessary for having multiple wallets/account
- remove the use of walletPublicId

### DIFF
--- a/src/app/wallets/intraledger-send-payment.ts
+++ b/src/app/wallets/intraledger-send-payment.ts
@@ -165,6 +165,7 @@ const executePaymentViaIntraledger = async ({
   if (recipientUser instanceof Error) return recipientUser
   if (recipientUser.id === userId) return new SelfPaymentError()
   // FIXME: selfPayment should be at the walletId level, no longer at the UserId
+  // but this is not a bloquer until we have multiple wallets per account
 
   const recipientAccount = await AccountsRepository().findById(
     recipientUser.defaultAccountId,
@@ -206,7 +207,7 @@ const executePaymentViaIntraledger = async ({
         fee: lnFee,
         usd,
         usdFee,
-        recipientLiabilitiesAccountId: toLiabilitiesWalletId(recipientWalletId),
+        recipientLiabilitiesWalletId: toLiabilitiesWalletId(recipientWalletId),
         payerUsername: username,
         recipientUsername,
         memoPayer,
--- a/src/app/wallets/intraledger-send-payment.ts
+++ b/src/app/wallets/intraledger-send-payment.ts
@@ -13,7 +13,7 @@ import {
   SatoshiAmountRequiredError,
   SelfPaymentError,
 } from "@domain/errors"
-import { toLiabilitiesAccountId } from "@domain/ledger"
+import { toLiabilitiesWalletId } from "@domain/ledger"
 import { LedgerService } from "@services/ledger"
 import { LockService } from "@services/lock"
 import {
@@ -115,6 +115,7 @@ const lnSendPayment = async ({
     username,
     logger,
   })
+  if (paymentSendStatus instanceof Error) return paymentSendStatus
 
   const addContactToPayerResult = await addNewContact({
     userId,
@@ -157,15 +158,18 @@ const executePaymentViaIntraledger = async ({
     amount,
     walletId,
   })
+
   if (intraledgerLimitCheck instanceof Error) return intraledgerLimitCheck
 
   const recipientUser = await UsersRepository().findByUsername(recipientUsername)
   if (recipientUser instanceof Error) return recipientUser
   if (recipientUser.id === userId) return new SelfPaymentError()
+  // FIXME: selfPayment should be at the walletId level, no longer at the UserId
 
   const recipientAccount = await AccountsRepository().findById(
     recipientUser.defaultAccountId,
   )
+
   if (recipientAccount instanceof Error) return recipientAccount
   if (!(recipientAccount.walletIds && recipientAccount.walletIds.length > 0)) {
     return new NoWalletExistsForUserError(recipientUsername)
@@ -193,16 +197,16 @@ const executePaymentViaIntraledger = async ({
       )
     }
 
-    const liabilitiesAccountId = toLiabilitiesAccountId(walletId)
+    const liabilitiesWalletId = toLiabilitiesWalletId(walletId)
     const journal = await LockService().extendLock({ logger, lock }, async () =>
       LedgerService().addUsernameIntraledgerTxSend({
-        liabilitiesAccountId,
+        liabilitiesWalletId,
         description: "",
         sats,
         fee: lnFee,
         usd,
         usdFee,
-        recipientLiabilitiesAccountId: toLiabilitiesAccountId(recipientWalletId),
+        recipientLiabilitiesAccountId: toLiabilitiesWalletId(recipientWalletId),
         payerUsername: username,
         recipientUsername,
         memoPayer,
--- a/src/app/wallets/intraledger-send-payment.ts
+++ b/src/app/wallets/intraledger-send-payment.ts
@@ -13,7 +13,6 @@ import {
   SatoshiAmountRequiredError,
   SelfPaymentError,
 } from "@domain/errors"
-import { toLiabilitiesWalletId } from "@domain/ledger"
 import { LedgerService } from "@services/ledger"
 import { LockService } from "@services/lock"
 import {
@@ -198,16 +197,15 @@ const executePaymentViaIntraledger = async ({
       )
     }
 
-    const liabilitiesWalletId = toLiabilitiesWalletId(walletId)
     const journal = await LockService().extendLock({ logger, lock }, async () =>
       LedgerService().addUsernameIntraledgerTxSend({
-        liabilitiesWalletId,
+        walletId,
         description: "",
         sats,
         fee: lnFee,
         usd,
         usdFee,
-        recipientLiabilitiesWalletId: toLiabilitiesWalletId(recipientWalletId),
+        recipientWalletId,
         payerUsername: username,
         recipientUsername,
         memoPayer,
--- a/src/core/on-chain/index.ts
+++ b/src/core/on-chain/index.ts
@@ -1,10 +1,25 @@
+import { getCurrentPrice } from "@app/prices"
+import * as Wallets from "@app/wallets"
+import {
+  checkAndVerifyTwoFA,
+  checkIntraledgerLimits,
+  checkWithdrawalLimits,
+} from "@app/wallets/check-limit-helpers"
+import { BTC_NETWORK, ONCHAIN_SCAN_DEPTH_OUTGOING } from "@config/app"
+import { toSats } from "@domain/bitcoin"
+import { TxDecoder } from "@domain/bitcoin/onchain"
+import { toLiabilitiesWalletId } from "@domain/ledger"
+import { TwoFANewCodeNeededError } from "@domain/twoFA"
+import { LedgerService } from "@services/ledger"
+import { OnChainService } from "@services/lnd/onchain-service"
+import { getActiveOnchainLnd } from "@services/lnd/utils"
+import { LockService } from "@services/lock"
+import { ledger } from "@services/mongodb"
+import { UsersRepository, WalletsRepository } from "@services/mongoose"
+import { User } from "@services/mongoose/schema"
+import { NotificationsService } from "@services/notifications"
 import assert from "assert"
 import { getChainBalance, getChainFeeEstimate, sendToChainAddress } from "lightning"
-
-import { getActiveOnchainLnd } from "@services/lnd/utils"
-import { ledger } from "@services/mongodb"
-import { User } from "@services/mongoose/schema"
-
 import {
   DustAmountError,
   InsufficientBalanceError,
@@ -18,24 +33,6 @@ import {
 import { redlock } from "../lock"
 import { UserWallet } from "../user-wallet"
 import { LoggedError } from "../utils"
-import { BTC_NETWORK, ONCHAIN_SCAN_DEPTH_OUTGOING } from "@config/app"
-import * as Wallets from "@app/wallets"
-import { toSats } from "@domain/bitcoin"
-import { UsersRepository, WalletsRepository } from "@services/mongoose"
-import { CouldNotFindError } from "@domain/errors"
-import { LedgerService } from "@services/ledger"
-import { toLiabilitiesWalletId } from "@domain/ledger"
-import { LockService } from "@services/lock"
-import {
-  checkAndVerifyTwoFA,
-  checkIntraledgerLimits,
-  checkWithdrawalLimits,
-} from "@app/wallets/check-limit-helpers"
-import { TwoFANewCodeNeededError } from "@domain/twoFA"
-import { getCurrentPrice } from "@app/prices"
-import { NotificationsService } from "@services/notifications"
-import { OnChainService } from "@services/lnd/onchain-service"
-import { TxDecoder } from "@domain/bitcoin/onchain"
 
 export const OnChainMixin = (superclass) =>
   class extends superclass {
@@ -80,7 +77,8 @@ export const OnChainMixin = (superclass) =>
         /// TODO: unable to check balanceSats vs this.dustThreshold at this point...
       }
 
-      return redlock({ path: this.user._id, logger: onchainLogger }, async (lock) => {
+      const walletId_ = this.user.walletId // FIXME: just set this variable for easier code review. long variable would trigger adding a tab and much bigger diff
+      return redlock({ path: walletId_, logger: onchainLogger }, async (lock) => {
         const balanceSats = await Wallets.getBalanceForWallet({
           walletId: this.user.walletId,
           logger: onchainLogger,
@@ -155,10 +153,8 @@ export const OnChainMixin = (superclass) =>
           const usdFee = onChainFee * price
 
           const payerWallet = await WalletsRepository().findById(this.user.walletId)
-          if (payerWallet instanceof CouldNotFindError) throw payerWallet
           if (payerWallet instanceof Error) throw payerWallet
           const recipientWallet = await WalletsRepository().findById(payeeUser.walletId)
-          if (recipientWallet instanceof CouldNotFindError) throw recipientWallet
           if (recipientWallet instanceof Error) throw recipientWallet
 
           const journal = await LockService().extendLock(
@@ -363,7 +359,7 @@ export const OnChainMixin = (superclass) =>
               description: memo,
               sats,
               fee: this.user.withdrawFee,
-              account: this.user.accountPath,
+              walletPath: this.user.walletPath,
               metadata,
             })
 
--- a/src/core/on-chain/index.ts
+++ b/src/core/on-chain/index.ts
@@ -169,7 +169,7 @@ export const OnChainMixin = (superclass) =>
                 usdFee,
                 payeeAddresses: [address as OnChainAddress],
                 sendAll,
-                recipientLiabilitiesAccountId: toLiabilitiesWalletId(payeeUser.walletId),
+                recipientLiabilitiesWalletId: toLiabilitiesWalletId(payeeUser.walletId),
                 payerUsername: this.user.username,
                 recipientUsername: payeeUser.username,
                 memoPayer: memo || null,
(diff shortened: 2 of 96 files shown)


src/core/on-chain/index.ts
```diff
@@ -80,9 +77,10 @@ export const OnChainMixin = (superclass) =>
         /// TODO: unable to check balanceSats vs this.dustThreshold at this point...
       }
 
-      return redlock({ path: this.user._id, logger: onchainLogger }, async (lock) => {
+      const walletId_ = this.user.walletId // FIXME: just set this variable for easier code review. long variable would trigger adding a tab and much bigger diff
+      return redlock({ path: walletId_, logger: onchainLogger }, async (lock) => {
         const balanceSats = await Wallets.getBalanceForWallet({
-          walletId: this.user.id,
+          walletId: this.user.walletId,
```
bodymindarts, line 83: The old user doesn't have a `walletId` does it?

src/domain/ledger/index.ts
```diff
@@ -4,8 +4,8 @@ export * from "./errors"
 
 export const liabilitiesMainAccount = "Liabilities"
 
-export const toLiabilitiesAccountId = (walletId: WalletId): LiabilitiesAccountId =>
-  `${liabilitiesMainAccount}:${walletId}` as LiabilitiesAccountId
+export const toLiabilitiesWalletId = (walletId: WalletId): LiabilitiesWalletId =>
```
bodymindarts, line 7: I'm confused about this change. The ledger doesn't have a wallet does it? The ledger has accounts or journals or journal entries or something like that but not a wallet or? I don't think wallet is the correct naming for a grouping identifier in the ledger subdomain.
There is clearly some kind of mapping between wallets and the scope of journals in the ledger but I wouldn't use the same naming.

src/domain/ledger/index.types.d.ts
```diff
@@ -28,7 +28,7 @@ type LedgerJournal = {
 // Differentiate fields depending on what 'type' we have (see domain/wallets/index.types.d.ts)
 type LedgerTransaction = {
   readonly id: LedgerTransactionId
-  readonly walletId: WalletId | null
+  readonly walletId: WalletId | null // FIXME create a subclass so that this field is always set for liabilities wallets
```
bodymindarts, line 31: Can we document todo's in other places than in the code? If it is important for the functioning of the code it should be done in the same or directly following PR. If it is not important it probably won't get done and this comment will age and create confusion further down the road. Better to document it in a GitHub issue than in code for that case IMO.

Also I'm again confused regarding `liabilities wallets` term. I think there is an overloading of the term `wallet`. My idea of a wallet is `the thing / entity that can send / receive money` (ie. is connected to unchain addresses and lightning invoices etc). The concept of liabilities belongs to the `ledger` domain which shouldn't really know about wallets or?

src/graphql/admin/root/mutation/account-update-level.ts
```diff
@@ -7,6 +7,7 @@ import { GT } from "@graphql/index"
 const AccountUpdateLevelInput = new GT.Input({
   name: "AccountUpdateLevelInput",
   fields: () => ({
+    // FIXME: should no longer be uid
```
bodymindarts, line 10: What should it be then?

src/graphql/admin/root/mutation/account-update-level.ts
```diff
@@ -22,6 +23,7 @@ const AccountUpdateLevelMutation = GT.Field({
     input: { type: GT.NonNull(AccountUpdateLevelInput) },
   },
   resolve: async (_, args) => {
+    // FIXME: should no longer be uid
```
bodymindarts, line 26: What should it be then?
I don't see the value in this comment - saying what should _not_ be the case doesn't really help in indicating what _should_ be done. As a reader I don't get any information from this other than being confused. The questions in my head on reading this is:
- so the current code is wrong?
- does that means its broken?
- what is broken about it?
- if its broken why don't we fix it?
- why do tests pass?
- what about clients using this field?
... etc.

Maybe I'm lacking context here - I'm just giving my train of thought when I read this comment.

src/graphql/types/object/business-account.ts
```diff
@@ -7,12 +7,11 @@ import Transaction from "./transaction"
 import WalletId from "../scalar/wallet-id"
 
 import * as Wallets from "@app/wallets"
-import * as Accounts from "@app/accounts"
 
 const BusinessAccount = new GT.Object({
   name: "BusinessAccount",
   interfaces: () => [IAccount],
-  isTypeOf: (source) => source.title || source.coordinates, // TODO: improve
+  // isTypeOf: (source) => source.title || source.coordinates, // TODO: improve
```
bodymindarts, line 14: Can we delete this and document a TODO as GitHub issue instead of keeping commented code?

src/services/ledger/accounts.ts
```diff
@@ -32,21 +32,22 @@ export const resolveAccountId = (accountPath: string | string[]) => {
     id = path[1]
   }
 
-  return id
+  // TODO: add check for WalletId syntax validity
```
bodymindarts, line 35: What does this mean `syntax validity`?

bodymindarts (commented):
I've gotten through about half the PR - its quite long and seems to be doing more than the title says.

I think the code around removing `walletPublicId` looks okay. But then it also changes `liabilitiesAccountId` which is a significant change in semantics which I do not understand and seems unrelated.
Also it adds a lot of comments that are more confusing than valuable when I read it.

src/core/on-chain/index.ts
nicolasburtey: it does (with this PR)

nicolasburtey:
> I think the code around removing `walletPublicId` looks okay. But then it also changes `liabilitiesAccountId` which is a significant change in semantics which I do not understand and seems unrelated.

It's related. after this PR, `account.id == user.id`, and now the books are recorded with the `wallet.id`, so saying `liabilitiesAccountId` is no longer relevant. it's really referring to a `walletId`.

src/services/ledger/accounts.ts
nicolasburtey: should be `uuid-v4`. but not sure this is needed.

src/graphql/types/object/business-account.ts
nicolasburtey: IMO it's a better practice to document in code as opposed to issue. it's contextualized. 

but actually I should take care of this one before merging. I thought BusinessAccount were not used (I had businessAccount as corporate account in my mind, but they are consumer account with maps metadata), but they actually are already.

bodymindarts:
> It's related. after this PR, account.id == user.id, and now the books are recorded with the wallet.id, so saying liabilitiesAccountId is no longer relevant. it's really referring to a walletId.

In my mind a LiabilitiesAccount within a Ledger is semantically distinct from both a galoy `Account` and a `Wallet` - which is why I'm confused. We should probably have a sync with a few engineers to get the semantics right here.

From [wikipedia article on double-entry bookkeeping](https://en.wikipedia.org/wiki/Double-entry_bookkeeping):
> Double-entry bookkeeping, in accounting, is a system of bookkeeping where every entry to an account requires a corresponding and opposite entry to a different account. The double-entry system has two equal and corresponding sides known as debit and credit. A transaction in double-entry bookkeeping always affects at least two accounts, always includes at least one debit and one credit, and always has total debits and total credits that are equal. For example, if a business takes out a bank loan for $10,000, recording the transaction would require a debit of $10,000 to an asset account called "Cash", as well as a credit of $10,000 to a liability account called "Notes Payable".

There is no mention of `wallet` in that field. From my reading `account` is the correct word for 'a grouping of related entries. Hence `LiabilitiesAccount` regardless of its reference to a wallet or account or user in other parts of the galoy system.

src/services/ledger/accounts.ts
bodymindarts: For types that originate from our domain we should assume validity. The only place walletId's can be generated is internally so for this not to be valid here would be very strange.

src/graphql/types/object/business-account.ts
bodymindarts: > IMO it's a better practice to document in code as opposed to issue. it's contextualized.
Something to gain team consensus on I guess

src/domain/ledger/index.ts
nicolasburtey: so we have, at the domain level:
User
Account
Wallet

That is map at the mongodb level:
user.id
user.id
wallet.id (and it used to be used.id before this PR)

Ledger tx are being identify by wallet.id now.
Ledger still use an internal id, but the domain is not making use of that any more.

src/domain/ledger/index.types.d.ts
nicolasburtey: > Better to document it in a GitHub issue than in code for that case IMO.

I disagree on this. we have 100 issues on github. when was the last time you look at all of them to see if something needs improvement?

nicolasburtey: >  The concept of liabilities belongs to the ledger domain which shouldn't really know about wallets or? 

This is correct IMO and could be improved in another PR. doing this with this PR would make it even bigger. This PR doesn't change things on this.

I'm just renaming `LiabitilitiesAccountId` to `LiabilitiesWalletId` because ledger has nothing to do with account.id any more.

src/graphql/admin/root/mutation/account-update-level.ts
nicolasburtey: `accountId`

src/graphql/admin/root/mutation/account-update-level.ts
nicolasburtey: it should not be `uid`/`user id` but `aid`/`account id`. comment updated

nicolasburtey: it's part of the refactoring to dissociated user/account/wallet to 3 different entities.

src/graphql/types/object/business-account.ts
nicolasburtey: we could discuss it during code review, my assumption is that is already part of the consensus, the code has ~200 FIXME and TODO.

src/services/ledger/accounts.ts
nicolasburtey: yup, removing the comment/type

nicolasburtey:
> > It's related. after this PR, account.id == user.id, and now the books are recorded with the wallet.id, so saying liabilitiesAccountId is no longer relevant. it's really referring to a walletId.
> 
> In my mind a `LiabilitiesAccount` within a Ledger is semantically distinct from both a galoy `Account` and a `Wallet` - which is why I'm confused. We should probably have a sync with a few engineers to get the semantics right here.

This is what we talked about last Code review and we agreed upon (maybe without full understanding of what we agreed upon for everyone).

The goal with this PR is to have:

User <-> Account <-> Wallet <-> Ledger

The idea is that 1-n user have an account, and each account can have 1-n wallet. so the wallet is what needs to be register to the ledger. account and ledger have no direct relationship here because we have the level of abstraction necessary for having multiple currency/wallet per account.

> From [wikipedia article on double-entry bookkeeping](https://en.wikipedia.org/wiki/Double-entry_bookkeeping):
> 
> > Double-entry bookkeeping, in accounting, is a system of bookkeeping where every entry to an account requires a corresponding and opposite entry to a different account. The double-entry system has two equal and corresponding sides known as debit and credit. A transaction in double-entry bookkeeping always affects at least two accounts, always includes at least one debit and one credit, and always has total debits and total credits that are equal. For example, if a business takes out a bank loan for $10,000, recording the transaction would require a debit of $10,000 to an asset account called "Cash", as well as a credit of $10,000 to a liability account called "Notes Payable".
> 
> There is no mention of `wallet` in that field. From my reading `account` is the correct word for 'a grouping of related entries. Hence `LiabilitiesAccount` regardless of its reference to a wallet or account or user in other parts of the galoy system.

That is correct, but if you remember the discussion we had (this one some months ago), the use of wallet was intentional and reflective of the use in the bitcoin/crypto space. you say you have different "wallets", not different "accounts". 

 > Think we need a broader discussion here.

src/domain/ledger/index.types.d.ts
```diff
@@ -44,7 +44,6 @@ type LedgerTransaction = {
   readonly feeUsd: number
 
   // for IntraLedger
-  readonly walletPublicId?: WalletPublicId
```
vindard, line 47: This was originally meant to be the recipient's `walletPublicId`. Should we instead rename this to something like `payeeWalletId` or `recipientWalletId`?

vindard, line 47: This was originally meant to be the recipient's `walletPublicId`. Should we instead rename this to something like `recipientWalletId` instead of removing?

src/domain/wallets/tx-history.ts
```diff
@@ -117,12 +116,12 @@ export const fromLedger = (
             ...baseTransaction,
             initiationVia: {
               type: PaymentInitiationMethod.IntraLedger,
-              counterPartyWalletPublicId: walletPublicId as WalletPublicId,
+              counterPartyWalletId: walletId as WalletId,
```
vindard, line 119: This one's wrong, because we originally named `walletPublicId` badly in the `LedgerTransaction` type (see related comment). It should actually be the `recipientWalletId` property.

src/domain/wallets/tx-history.ts
```diff
@@ -117,12 +116,12 @@ export const fromLedger = (
...
               counterPartyUsername: username as Username,
             },
             settlementVia: {
               type: SettlementMethod.IntraLedger,
-              counterPartyWalletPublicId: walletPublicId as WalletPublicId,
+              counterPartyWalletId: walletId as WalletId,
```
vindard, line 124: Same here

src/domain/wallets/tx-history.ts
```diff
@@ -137,7 +136,7 @@ export const fromLedger = (
...
```
vindard, line 139: Same here

src/domain/wallets/tx-history.ts
```diff
@@ -168,7 +167,7 @@ export const fromLedger = (
...
```
vindard, line 170: Same here

src/domain/wallets/tx-history.ts
```diff
@@ -195,12 +194,12 @@ export const fromLedger = (
         ...baseTransaction,
         initiationVia: {
           type: PaymentInitiationMethod.IntraLedger,
-          counterPartyWalletPublicId: walletPublicId as WalletPublicId,
+          counterPartyWalletId: walletId as WalletId,
```
vindard, line 197: Same here

src/domain/wallets/tx-history.ts
```diff
@@ -195,12 +194,12 @@ export const fromLedger = (
...
           counterPartyUsername: username as Username,
         },
         settlementVia: {
           type: SettlementMethod.IntraLedger,
-          counterPartyWalletPublicId: walletPublicId as WalletPublicId,
+          counterPartyWalletId: walletId as WalletId,
```
vindard, line 202: Same here

vindard (changes requested):
I'm still working my way through it. I've looked at a lot of the `WalletPublicId` to `WalletId` code, but am still working my way through the ledger account(_1_) change from account(_2_) ids to wallet ids which on the surface makes sense to me.

There's one obvious little mix-up with the removal of `walletPublicId` from `LedgerTransaction` type which was named badly and actually refers to the recipient's `walletPublicId`. Should be renamed instead of removed as it doesn't refer to the sender's new `walletId`.

---

_**1** account as in a collection of records in an accounting ledger_
_**2** account as in our logical grouping of multiple wallets and the entity against which payments are made/received_

nicolasburtey:
> I'm still working my way through it. I've looked at a lot of the `WalletPublicId` to `WalletId` code, but am still working my way through the ledger account(_**1**_) change from account(_**2**_) ids to wallet ids which on the surface makes sense to me if we can keep the account(_**2**_) link somehow maybe e.g. `Liabilities:<account-id>:<wallet-id>`.

I think that's a great idea. I'll bring a lot of optimization down the road by having the account Id readily available. I'll work on that.

src/domain/ledger/index.types.d.ts
nicolasburtey: does this resolves your concern: https://github.com/GaloyMoney/galoy/pull/792/commits/82d3cc1ecb878e975215fd3ea6326a6d566da18b ?

src/services/ledger/index.ts
```diff
@@ -72,23 +77,25 @@ export const LedgerService = (): ILedgerService => {
   }
 
   const getLiabilityTransactions = async (
-    liabilitiesAccountId: LiabilitiesAccountId,
+    walletId: WalletId,
   ): Promise<LedgerTransaction[] | LedgerError> => {
+    const liabilitiesWalletId = toLiabilitiesWalletId(walletId)
```
vindard, line 82: I believe we deliberately didn't include these inside service methods previously to keep boundaries clear. @bodymindarts would have to confirm.

This would also be related to the function signature changes in `src/domain/ledger/index.types.d.ts`

src/app/wallets/intraledger-send-payment.ts
```diff
@@ -115,6 +114,7 @@ const lnSendPayment = async ({
     username,
     logger,
   })
+  if (paymentSendStatus instanceof Error) return paymentSendStatus
```
vindard, line 117: Good catch!

src/core/user-wallet.ts
```diff
@@ -35,13 +35,14 @@ export abstract class UserWallet {
   }
 
   async getBalances(lock?): Promise<Balances> {
-    Wallets.updatePendingInvoices({
-      walletId: this.user.id as WalletId,
+    // was the await omit on purpose?
```
vindard, line 38: Yes, they both were originally left without the `await` and then we added it back to the `Wallets.updatePendingPayments` line to be able to see if there were any errors to return from. We didn't need the error feedback from `Wallets.updatePendingInvoices` so we left it without the await, though I don't remember why we made this distinction at the time.

src/services/mongoose/accounts.ts
```diff
@@ -34,8 +34,8 @@ export const AccountsRepository = (): IAccountsRepository => {
     walletId: WalletId,
   ): Promise<Account | RepositoryError> => {
     try {
-      const result: UserType = await User.findOne({ _id: walletId }, projection)
-      if (!result) return new CouldNotFindError()
+      const result: UserType = await User.findOne({ walletId }, projection)
+      if (!result) return new CouldNotFindError("Invalid wallet")
```
vindard, line 38: We don't usually parse the arguments from these sorts of errors, so if we'd like to add additional context for the final error-map we'd usually make a new distinct error type and add it to the error-map switch.

src/core/user-wallet.ts
nicolasburtey: I feel a comment would be nice to have in future PR. will leave the comment for now on if that's ok.

src/services/ledger/index.ts
nicolasburtey: > I believe we deliberately didn't include these inside service methods previously to keep boundaries clear.

hmm. I think not including those function within ledger made the boundary not clear in my mind. because now we have added to to the conversion of `walletId` to `liabilitiesWalletId` everywhere in the codebase, where `liabilitiesWalletId` is a notion that should only concern the ledger IMO.

vindard: Marking this as resolved based on our discussion about making the Ledger sub-domain non-generic (i.e. aware of core domain concepts)

bodymindarts (dismissed):

vindard (approved):

vindard (approved):

src/graphql/types/object/btc-wallet.ts
```diff
@@ -15,7 +15,7 @@ const BTCWallet = new GT.Object({
   fields: () => ({
     id: {
       type: GT.NonNullID,
-      resolve: (source) => source.publicId,
+      resolve: (source) => source.walletId,
```
vindard, line 18: On local testing I got source as type `Wallet`. Is that right or does something else have to change as well?
```suggestion
      resolve: (source: Wallet) => source.id,
```

nicolasburtey: fix in last commit

vindard (approved):

vindard:
I did a comparison against #814 and found just one inconsistency (fixed: https://github.com/GaloyMoney/galoy/pull/792/commits/7c0b98a7ae19f7f34f2a9e67d46362f29f1e515e). This should be good to go now.

dolcalmi (approved):

PR #795, opened 2021-12-10
fix: trigger health check

dolcalmi:
Add mongo, redis and lnds status validation to trigger healthz endpoint

How to test:

- Run `TEST="01|02" make reset-integration`
- Run `make trigger`
- GET `http://localhost:8888/healthz` (should receive 200 after trigger initialization)
- Run `docker ps` to get container ids 
- Run `docker stop <galoy-backend-lnd2-1 container id>`
- GET `http://localhost:8888/healthz` (should receive 503)
- Run `docker start <galoy-backend-lnd2-1 container id>` 
- GET `http://localhost:8888/healthz` (should receive 200)

ps: you can use the same procedure with redis and mongo

### DIFF
--- a/src/servers/exporter.ts
+++ b/src/servers/exporter.ts
@@ -154,7 +154,7 @@ const main = async () => {
     "/healthz",
     healthzHandler({
       checkDbConnectionStatus: true,
-      checkRedisStatus: true, // TODO: set to false when invoices/payments updates are removed from get balance
+      checkRedisStatus: false,
       checkLndsStatus: true,
     }),
   )
--- a/src/servers/exporter.ts
+++ b/src/servers/exporter.ts
@@ -14,6 +14,8 @@ import { User } from "@services/mongoose/schema"
 import express from "express"
 import client, { register } from "prom-client"
 
+import healthzHandler from "./healthz-handler"
+
 const logger = baseLogger.child({ module: "exporter" })
 
 const server = express()
@@ -148,9 +150,14 @@ const main = async () => {
     res.end(await register.metrics())
   })
 
-  server.get("/healthz", async (req, res) => {
-    res.send("OK")
-  })
+  server.get(
+    "/healthz",
+    healthzHandler({
+      checkDbConnectionStatus: true,
+      checkRedisStatus: true, // TODO: set to false when invoices/payments updates are removed from get balance
+      checkLndsStatus: true,
+    }),
+  )
 
   const port = process.env.PORT || 3000
   logger.info(`Server listening to ${port}, metrics exposed on /metrics endpoint`)
--- a/src/servers/healthz-handler.ts
+++ b/src/servers/healthz-handler.ts
@@ -1,0 +1,44 @@
+import express from "express"
+import mongoose from "mongoose"
+import { redis } from "@services/redis"
+import { lndStatusEvent } from "@services/lnd/health"
+
+type HealthzArgs = {
+  checkDbConnectionStatus: boolean
+  checkRedisStatus: boolean
+  checkLndsStatus: boolean
+}
+
+export default function ({
+  checkDbConnectionStatus,
+  checkRedisStatus,
+  checkLndsStatus,
+}: HealthzArgs) {
+  const lndStatus: { [key: string]: boolean } = {}
+  if (checkLndsStatus) {
+    lndStatusEvent.on("started", ({ pubkey, active }) => {
+      lndStatus[pubkey] = active
+    })
+
+    lndStatusEvent.on("stopped", ({ pubkey, active }) => {
+      lndStatus[pubkey] = active
+    })
+  }
+
+  return async (_req: express.Request, res: express.Response) => {
+    const isMongoAlive = !checkDbConnectionStatus || mongoose.connection.readyState === 1
+    const isRedisAlive = !checkRedisStatus || (await isRedisAvailable())
+    const statuses = Object.values(lndStatus)
+    const areLndsAlive =
+      !checkLndsStatus || (statuses.length > 0 && statuses.some((s) => s))
+    res.status(isMongoAlive && isRedisAlive && areLndsAlive ? 200 : 503).send()
+  }
+}
+
+const isRedisAvailable = async (): Promise<boolean> => {
+  try {
+    return (await redis.ping()) === "PONG"
+  } catch {
+    return false
+  }
+}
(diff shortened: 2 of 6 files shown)


src/servers/trigger.ts
```diff
@@ -255,7 +261,13 @@ const main = () => {
 const healthCheck = () => {
   const app = express()
   const port = 8888
-  app.get("/healthz", (req, res) => res.sendStatus(200))
+  app.get("/healthz", async (_req, res) => {
+    const isMongoAlive = mongoose.connection.readyState === 1
+    const isRedisAlive = (await redis.ping()) === "PONG"
+    const statuses = Object.values(lndStatus)
+    const areLndsAlive = statuses.length > 0 && statuses.every((s) => s)
```
nicolasburtey, line 268: I'm not sure to understand the `statuses.every((s) => s)`. It's ok for 1 lnd to be offline. this is the idea behind the high availability that we're working on. I think the code here assumes that all lnd needs to be online?

src/servers/trigger.ts
```diff
@@ -255,7 +261,13 @@ const main = () => {
 const healthCheck = () => {
   const app = express()
   const port = 8888
-  app.get("/healthz", (req, res) => res.sendStatus(200))
+  app.get("/healthz", async (_req, res) => {
+    const isMongoAlive = mongoose.connection.readyState === 1
+    const isRedisAlive = (await redis.ping()) === "PONG"
+    const statuses = Object.values(lndStatus)
```
nicolasburtey, line 267: maybe this should be refactored so that the other servers has the code/use of lndStatus.

dolcalmi: but should we check only main lnd? or both should be active?

nicolasburtey: at least 1

nicolasburtey: one thing I'm not sure how this would work: if one lnd were to restart, but not trigger, would trigger be able to reconnect to it?

dolcalmi: yep, it works, you can test locally with:

- Run docker stop <galoy-backend-lnd2-1 container id>
- GET http://localhost:8888/healthz (should receive 503)
- Run docker start <galoy-backend-lnd2-1 container id>
- GET http://localhost:8888/healthz (should receive 200)

nicolasburtey: what works exactly? trigger would restart?

nicolasburtey: we don't want trigger to continuously restart if one lnd is continuously failing

dolcalmi: I changed the validation to some, so trigger will not restart until both are offline (or if is not able to connect)

dolcalmi: Refactor done for exporter, graphql and trigger but I couldn't add `is_ready` (Wallet Is Ready For RPC Calls) because is not supported on LND 0.13.4 and below

src/servers/exporter.ts
```diff
@@ -143,9 +144,14 @@ const main = async () => {
...
   })
 
-  server.get("/healthz", async (req, res) => {
-    res.send("OK")
-  })
+  server.get(
+    "/healthz",
+    healthzHandler({
+      checkDbConnectionStatus: true,
+      checkRedisStatus: true, // TODO: set to false when invoices/payment updates are removed from get balance
```
vindard, line 151: What's this comment referring to?

vindard (dismissed):
Looks ok to me. We've abstracted the checks into a single function that we can toggle and apply in different places. The "how to test" steps were appreciated 🙌🏾

src/servers/exporter.ts
dolcalmi: exporter only needs db and lnds to get the balances but since we use getBalanceForWallet the update of `pending invoices and payments` are triggered and that functions need redis

vindard (dismissed):

nicolasburtey:
note to self: it seems that using the merge tool from github is not great because it does a merge commit instead of rebase

src/servers/exporter.ts
```diff
@@ -148,9 +150,14 @@ const main = async () => {
...
   })
 
-  server.get("/healthz", async (req, res) => {
-    res.send("OK")
-  })
+  server.get(
+    "/healthz",
+    healthzHandler({
+      checkDbConnectionStatus: true,
+      checkRedisStatus: true, // TODO: set to false when invoices/payments updates are removed from get balance
```
nicolasburtey, line 157: could you explain more in depth the comment?

src/servers/healthz-handler.ts
```diff
@@ -0,0 +1,44 @@
+import express from "express"
```
nicolasburtey, line 1: should this be in `src/servers/`? it's not meant to be run independantly? I think this should be somewhere else (not sure where)

nicolasburtey (commented):
two questions. 

otherwise I didn't really check the logic but this sounds to be an improvement over what we currently have so I think we should definitively merge it

src/servers/healthz-handler.ts
dolcalmi: I can rename `src/servers/graphql-middlewares` to `src/servers/middlewares` and move it to that folder. This is another middleware implementation. what do you think?

src/servers/exporter.ts
dolcalmi: Previously we were using `getBalanceForWallet` (runs updatePendingInvoices and updatePendingPayments that need redis to lock wallets) but last week we updated to `getBalanceForWalletId` so now we no longer need redis check. Updated here https://github.com/GaloyMoney/galoy/pull/795/commits/cf8b2324ae2a1b56207a9d84a9b21eae8b71055a

src/servers/exporter.ts
dolcalmi: Last week we updated to `getBalanceForWalletId` so now we no longer need redis check. Updated here https://github.com/GaloyMoney/galoy/pull/795/commits/cf8b2324ae2a1b56207a9d84a9b21eae8b71055a

bodymindarts (approved):

src/servers/healthz-handler.ts
nicolasburtey: does this matters if this files is included in this pattern? https://github.com/GaloyMoney/galoy/blob/a9cdea18a037720d749657c879012b8dc4c3b756/tsconfig-build.json#L3

and is it included in this pattern if we add it in a middlewares folder?

dolcalmi: `**` should take care of inner folders, it is safe to move it. let me know if is necessary

vindard (approved):

Issue #793, opened 2021-12-09
Latency in calls to MainQuery


daviroo:
@nicolasburtey experienced a period of latency between making an onchain payment and seeing his balance update in the app.  I checked a trace of the call to the MainQuery from @nicolasburtey user [here](https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/vdLxTcjzHZy/trace/EXCRs7tuv2c?span=86bd7f1592fa09b2).  We can see that the trace contains a lot of http calls to `https://fcm.googleapis.com/fcm/send`, but as these seem to be asynchronous we don't think this is the cause of the latency. Instead it seems to be coming from the part of the code which fetches the current balance which loops through all the unpaid invoices / unprocessed payments.  More investigation is needed to confirm but we need to trim down the time it takes to make this call as it's the main query we use in the app to display all of the data to the user so it should never take 3.5 seconds to respond.

daviroo:
@nicolasburtey thinks it might be related to the issue discussed in #604 but it's probably worth looking into other aspects as the delay might be coming from multiple places.

nicolasburtey:
the problem has been amplified by the fact `trigger` was offline since yesterday migration. let's look at https://github.com/GaloyMoney/galoy/issues/604 as a potential solution?

dolcalmi:
this https://github.com/GaloyMoney/galoy/commit/c5c64c8acea9075525e0d363293c6bd81205ee61 could help too. We are querying 360 blocks for pending incoming on-chain txs (in theory we need only 1 day max, if cron is running 😅 )

vindard:
@dolcalmi could we maybe do something like the `runInParallel` function you introduced a while back too?

I noticed from honeycomb that each `updatePendingInvoice` seems to happen sequentially, and I couldn't quite figure how to collect a bunch of promises from the async generator in that loop and fire them all off in a `Promise.all` or something similar 🤔

nicolasburtey:
most/(all?) of the significant latency issues has been addressed over the last 2 weeks. closing this for now. we can reopen if needed.

PR #796, opened 2021-12-10
chore(tracing): wrap 'getBalanceForWallet' use-case in span

vindard:
## Description

Wrap `getBalanceForWallet` in a tracing span, related to debugging done in #793.

### DIFF
--- a/src/app/wallets/get-balance-for-wallet.ts
+++ b/src/app/wallets/get-balance-for-wallet.ts
@@ -1,6 +1,11 @@
 import { toLiabilitiesAccountId } from "@domain/ledger"
 import { LedgerService } from "@services/ledger"
 import * as Wallets from "@app/wallets"
+import {
+  asyncRunInSpan,
+  ENDUSER_ACCOUNT_WALLETID,
+  SemanticAttributes,
+} from "@services/tracing"
 
 export const getBalanceForWallet = async ({
   walletId,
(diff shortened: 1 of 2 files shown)


dolcalmi (approved):

src/app/wallets/get-balance-for-wallet.ts
```diff
@@ -10,23 +11,31 @@ export const getBalanceForWallet = async ({
...
-      logger,
-    }),
-  ])
-  if (updatePaymentsResult instanceof Error) throw updatePaymentsResult
+}): Promise<Satoshis | ApplicationError> =>
+  asyncRunInSpan(
+    "app.getBalanceForWallet",
+    {
+      [SemanticAttributes.CODE_FUNCTION]: "getBalanceForWallet",
+      "enduser.walletId": walletId,
```
bodymindarts, line 19: This should be defined globally like [`ENDUSER_ALIAS`](https://github.com/GaloyMoney/galoy/blob/main/src/services/tracing.ts#L189).
Also it is semantically wrong as the enduser doesn't have a wallet - an account has wallets. Not a blocker as this adds value.

bodymindarts (approved):

nicolasburtey (approved):
we would benefits to override some methods (the resolver?) to avoid doing this for every use case

vindard:
Waiting to see what comes from #813 for this PR

nicolasburtey:
should we close this one now @vindard ?

vindard:
Work done in #837

PR #797, opened 2021-12-10
chore: add cron run to github integration job

dolcalmi:
- Add an additional task to integration test to run cron and try to detect configuration issues
- Is necessary to run cron after integration test because DbMetadata needs to be initialized
- This is not a final solution because config can be updated in staging or prod (CI tasks need to be updated) but will help developers to avoid issues with cron job

Example
image

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -15,5 +15,3 @@ jobs:
     - run: yarn install
     - name: Run integration tests
       run: docker compose up -d && make integration-in-ci
-    - name: Run cron
-      run: make cron-in-ci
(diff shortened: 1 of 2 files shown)


krtk6160 (approved):

bodymindarts:
If we run it here we also need to run it in concourse - and I think the best place to test is part of the [galoy smokiest](https://github.com/GaloyMoney/charts/blob/main/ci/tasks/galoy-smoketest.sh) that gets executed on the real cluster.

bodymindarts:
This is now being run as part of the smoketest @ staging rollout.

bodymindarts (dismissed):

vindard (dismissed):

nicolasburtey:
a rebase is needed @dolcalmi

bodymindarts (approved):

vindard (approved):

PR #800, opened 2021-12-11
refactor: remove sendBalance from core

nicolasburtey:
it's not perfect, there are still use of UserType in apps, but this can be refactor again once core is being eliminated.

because this is a fairly isolate piece of code for now, I think this is a good tradeoff.


src/app/users/active-users.ts
```diff
@@ -0,0 +1,42 @@
+import { MS_PER_30_DAYs, USER_ACTIVENESS_MONTHLY_VOLUME_THRESHOLD } from "@config/app"
+import { baseLogger } from "@services/logger"
+import { User } from "@services/mongoose/schema"
+
+// user is considered active if there has been one transaction of more than 1000 sats in the last 30 days
+const isUserActive = async (user: UserType /* FIXME */) => {
```
bodymindarts, line 6: IMO moving this out of core into app is misleading.
All the code under app should be using the domain types. If we are not ready to refactor this use-case to using domain types then it should not go under app imo but stay under core.

I don't mind the cleanup - it is an improvement to remove these functions from the mongoose object but still think this file should stay under core for the time being.

src/app/users/active-users.ts
```diff
@@ -0,0 +1,42 @@
+import { MS_PER_30_DAYs, USER_ACTIVENESS_MONTHLY_VOLUME_THRESHOLD } from "@config/app"
+import { baseLogger } from "@services/logger"
+import { User } from "@services/mongoose/schema"
```
bodymindarts, line 3: App should not depend on mongoose that is what the repository is for - an adapter between domain and persistence.

src/app/accounts/active-accounts.ts
```diff
@@ -0,0 +1,38 @@
+import { MS_PER_30_DAYS, USER_ACTIVENESS_MONTHLY_VOLUME_THRESHOLD } from "@config/app"
+import { LedgerService } from "@services/ledger"
+import { AccountsRepository } from "@services/mongoose"
+
+// user is considered active if there has been one transaction of more than USER_ACTIVENESS_MONTHLY_VOLUME_THRESHOLD sats in the last 30 days
+const isAccountRecentlyActive = async (walletId: WalletId) => {
```
bodymindarts, line 6: there is a naming inconsistency here `isAccount...` but argument `walletId` - why?
In the desired future state are we measuring activity of the account or the user or the wallet?

src/app/accounts/active-accounts.ts
```diff
@@ -0,0 +1,38 @@
+import { MS_PER_30_DAYS, USER_ACTIVENESS_MONTHLY_VOLUME_THRESHOLD } from "@config/app"
+import { LedgerService } from "@services/ledger"
+import { AccountsRepository } from "@services/mongoose"
+
+// user is considered active if there has been one transaction of more than USER_ACTIVENESS_MONTHLY_VOLUME_THRESHOLD sats in the last 30 days
```
bodymindarts, line 5: Is this explanation documented in form of a test?

src/app/accounts/active-accounts.ts
```diff
@@ -0,0 +1,38 @@
...
+    walletId,
+  })
+  if (volume instanceof Error) return volume
+
+  return (
+    volume.outgoingSats > activenessThreshold || volume.incomingSats > activenessThreshold
+  )
+}
+
+export const getRecentlyActiveAccount = async (): Promise<
```
bodymindarts, line 22: Can we have the public (ie `export`ed functions) above the private ones in the file?
IMO its easier to understand quickly what the most important entry points of a module are if they are at the top. I think we discussed this before - but if not we can bring it up at code review this evening to gain consensus.

src/services/mongoose/accounts.ts
```diff
@@ -19,9 +19,20 @@ const projection = {
 }
 
 export const AccountsRepository = (): IAccountsRepository => {
+  const listUnlockedAccounts = async (): Promise<Account[] | RepositoryError> => {
+    try {
+      const result: UserType[] /* UserType actually not correct with {projection} */ =
+        await User.find({ status: AccountStatus.Active }, projection)
+      if (result.length === 0) return new CouldNotFindError()
```
bodymindarts, line 26: This isn't really an error or? When calling a `listXxxCriteria` function then an empty array is a valid return. It would signify there were no records that match the criteria.
```
if (result.length === 0) return []
```

src/services/mongoose/accounts.ts
```diff
@@ -19,9 +19,20 @@ const projection = {
...
+      return result.map((a) => translateToAccount(a))
+    } catch (err) {
+      return new UnknownRepositoryError(err)
+    }
+  }
+
   const findById = async (accountId: AccountId): Promise<Account | RepositoryError> => {
     try {
-      const result: UserType /* UserType actually not correct with {projections} */ =
+      const result: UserType /* UserType actually not correct with {projection} */ =
```
bodymindarts, line 35: what does `projection` mean in this context?

src/services/notifications/index.ts
```diff
@@ -191,13 +194,59 @@ export const NotificationsService = (logger: Logger): INotificationsService => {
...
   }
 
+  const sendBalance = async (account: Account): Promise<void> => {
+    const balanceSats = await getBalanceForWallet({
+      walletId: account.defaultWalletId,
+      logger,
+    })
+    if (balanceSats instanceof Error) throw balanceSats
+
+    // Add commas to balancesats
```
bodymindarts, line 204: Are both these comments above and below 205 necessary? Can we encode the comment in the variable name:
```
const balanceSatsAsFormattedString = ...
```

src/services/notifications/index.ts
```diff
@@ -191,13 +194,59 @@ export const NotificationsService = (logger: Logger): INotificationsService => {
...
+      // TODO: i18n
+      title = `Your balance is $${balanceUsd} (${balanceSatsPrettified} sats)`
+    }
+
+    logger.info(
+      { balanceSatsPrettified, title, ownerId: account.ownerId },
+      `sending balance notification to user`,
+    )
+
+    // FIXME:
```
bodymindarts, line 230: What needs to be fixed here?

src/app/accounts/active-accounts.ts
nicolasburtey: we're measuring the activity of the wallets attached to an account. I've done a refactoring to make it future proof when we'll have multiple wallet per account (also this assume account are denominated BTC currently which is still la shortcut)

src/app/accounts/active-accounts.ts
nicolasburtey: there are integration tests related to this, but I guess you're thinking of a unit test?

src/app/accounts/active-accounts.ts
nicolasburtey: will do. I'm just used to do the opposite

src/services/mongoose/accounts.ts
nicolasburtey: which element fo we want from mongodb: https://www.geeksforgeeks.org/mongodb-projection/

src/services/notifications/index.ts
nicolasburtey: the line below. we should not pass a `UserType` to `sendNotification`. but this requires more refactoring of `sendNotification` (not part of what this PR do)

src/services/notifications/index.ts
nicolasburtey: at least the comment is useful to me. I would need to know otherwise that `toLocalString("en")` will add the comms the `balanceSats` to understand this line, and this is not really something we use a lot in the codebase

src/services/notifications/index.ts
```diff
@@ -191,13 +194,58 @@ export const NotificationsService = (logger: Logger): INotificationsService => {
     }
   }
 
+  const sendBalance = async (account: Account): Promise<void> => {
```
vindard, line 197: In the other `sendNotification` methods in this file we pass in the pre-processed values to be included in the notification, whereas in this new method we're determining the values internally using `Wallet` and `User` concepts.

I'm wondering if this breaks one of the boundaries we've setup and whether we should be determining these outside this function and passing them in as arguments? (or is this something we already plan to do in follow-up PRs?)

src/app/accounts/active-accounts.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (volume instanceof LedgerServiceError) return volume
+    volumeCum.incomingSats = toSats(volume.incomingSats + volumeCum.incomingSats)
+    volumeCum.outgoingSats = toSats(volume.outgoingSats + volumeCum.outgoingSats)
+  }
+
+  return volumeCum
+}
+
+// user is considered active if there has been one transaction of more than USER_ACTIVENESS_MONTHLY_VOLUME_THRESHOLD sats in the last 30 days
+const isAccountRecentlyActive = async (walletIds: WalletId[]) => {
```
vindard, line 53:
```suggestion
const isAccountRecentlyActive = async (
  walletIds: WalletId[],
): Promise<boolean | ApplicationError> => {

```

src/app/accounts/active-accounts.ts
```diff
@@ -0,0 +1,69 @@
...
+
+const sumActivityWalletIds = async ({
+  walletIds,
+  timestamp,
+  getVolumeFn,
+}: {
+  walletIds: WalletId[]
+  timestamp: Date
+  getVolumeFn: (args: IGetVolumeArgs) => VolumeResult
+}) => {
```
vindard, line 33:
```suggestion
}): Promise<TxVolume | ApplicationError> => {

```

vindard (commented):
Some small nits and one question, but otherwise looks ok to me

src/app/accounts/active-accounts.ts
```diff
@@ -0,0 +1,69 @@
...
+  const activenessThreshold = USER_ACTIVENESS_MONTHLY_VOLUME_THRESHOLD
+
+  const ledger = LedgerService()
+
+  const volume = await sumActivityWalletIds({
+    timestamp: timestamp30DaysAgo,
+    walletIds,
+    getVolumeFn: ledger.allTxVolumeSince,
+  })
+  if (volume instanceof LedgerServiceError) return volume
```
vindard, line 64: `  if (volume instanceof Error)...` case needs to be handled before moving on to the final return below

vindard (changes requested):

src/services/notifications/index.ts
nicolasburtey: good point, I can at least compute the balance outside of this function

src/services/notifications/index.ts
```diff
@@ -191,13 +193,58 @@ export const NotificationsService = (logger: Logger): INotificationsService => {
...
+    ownerId,
+  }: {
+    balance: Satoshis
+    ownerId: UserId
+  }): Promise<void> => {
+    // Add commas to balancesats
+    const balanceSatsAsFormattedString = balance.toLocaleString("en")
+
+    let balanceUsdAsFormattedString: string, title: string
+    const price = await getCurrentPrice()
```
vindard, line 207: We usually call this function at the app layer as well (not service layer)

vindard (changes requested):
Just one more suggested change and I think this looks ok to me to approve

vindard (approved):

src/servers/daily-balance-notification.ts
```diff
@@ -1,8 +1,9 @@
-import { setupMongoConnection } from "@services/mongodb"
-import { User } from "@services/mongoose/schema"
+import { getRecentlyActiveAccounts } from "@app/accounts/active-accounts"
+import { getCurrentPrice } from "@app/prices"
+import { getBalanceForWallet } from "@app/wallets"
```
dolcalmi, line 3: this should be consumed from `@app` it is required if we want to have tracing from this

src/servers/daily-balance-notification.ts
```diff
@@ -16,11 +17,23 @@ const main = async () => {
...
+    if (balance instanceof Error) throw balance
 
-  for (const user of users) {
-    const userWallet = await WalletFactory({ user, logger })
-    await userWallet.sendBalance()
+    await NotificationsService(logger).sendBalance({
+      balance,
+      ownerId: account.ownerId,
+      price,
+    })
```
dolcalmi, line 36: this should be part of app layer

dolcalmi (changes requested):

src/servers/daily-balance-notification.ts
nicolasburtey: how should this be called?

```
import Wallet from "@app"
Wallet.getBalanceForWallet()
```
?

src/servers/daily-balance-notification.ts
nicolasburtey: ok will do as subsequent PR

PR #802, opened 2021-12-11
fix: real fix for business account?

nicolasburtey:


dolcalmi:
we need to enable `BusinessAccount` in `src/graphql/types/index.ts`

nicolasburtey:
yes. I think the definition of a `businessAccount` here is a bit misleading. we can probably live with the fix we have for now until we have real business account (account with many wallets)

Issue #757, opened 2021-11-25
prepopulate the playground with some query


nicolasburtey:
currently if we open the playgound, we see an empty query field.

It would help developers to be able to have a query to go with when they first open the page.

I don't know if/how this could be done, but it would be useful to have also all the queries that are [here](https://galoymoney.github.io/galoy/#intro-item-0)

nicolasburtey:
looking at github: https://docs.github.com/en/graphql/overview/explorer

there is an additional tab called "Explorer" which make it extra easy to add/create a query or a mutation.

![image](https://user-images.githubusercontent.com/11156370/143501939-1813b2bb-c4b1-460e-aa3d-2ca1a4e3ce87.png)

the Explorer tab is not part of our playground.

moksha256:
@nicolasburtey assuming [this](https://api.staging.galoy.io/graphql) is the playground we're talking about, and [this](https://github.com/GaloyMoney/galoy/blob/1ed5feb03b2870b81e5352db2ae8ba522055a45d/src/servers/graphql-server.ts#L149) where it's configured, it looks like we should be able to add a `tabs` property with a `query` specifying default text ([reference](https://www.apollographql.com/docs/apollo-server/v2/testing/graphql-playground/#configuring-playground)).

Is that correct? If it is, I can make a pull request.

nicolasburtey:
this issue has been done

PR #804, opened 2021-12-13
docs: add placeholder query to playground

moksha256:
To address #757.

Tested locally and worked fine.

Placeholder query is currently `btcPriceList` since it doesn't require authentication. This query also uses a variable, which may be useful for people to see. But let me know if you would prefer a different query.

### DIFF
--- a/src/graphql/playground.ts
+++ b/src/graphql/playground.ts
@@ -17,4 +17,3 @@ export const playgroundTabs = {
     variables: JSON.stringify({ range: "ONE_DAY" }),
   },
 }
-
--- a/src/graphql/playground.ts
+++ b/src/graphql/playground.ts
@@ -1,0 +1,17 @@
+const defaultQuery = `query btcPriceList($range: PriceGraphRange!) {
+    btcPriceList(range: $range) {
+        timestamp
+        price {
+        base
+        offset
+        currencyUnit
+        formattedAmount
+      }
+    }
+}`
+
+const defaultQueryName = "btcPriceList (Sample Query)"
+
+const defaultQueryVars = '{"range": "ONE_DAY"}'
+
+export { defaultQuery, defaultQueryName, defaultQueryVars }
--- a/src/graphql/playground.ts
+++ b/src/graphql/playground.ts
@@ -1,17 +1,20 @@
-const defaultQuery = `query btcPriceList($range: PriceGraphRange!) {
-    btcPriceList(range: $range) {
-        timestamp
-        price {
-        base
-        offset
-        currencyUnit
-        formattedAmount
-      }
+export const BTC_PRICE_QUERY = `query btcPriceList($range: PriceGraphRange!) {
+  btcPriceList(range: $range) {
+      timestamp
+      price {
+      base
+      offset
+      currencyUnit
+      formattedAmount
     }
+  }
 }`
 
-const defaultQueryName = "btcPriceList (Sample Query)"
+export const playgroundTabs = {
+  default: {
+    query: BTC_PRICE_QUERY,
+    name: "btcPriceList (Sample Query)",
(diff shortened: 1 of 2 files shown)


nicolasburtey:
tests are not passing @moksha256 . you'll need to run prettier. `yarn prettier-write` should do it.

moksha256:
> `yarn prettier-write` should do it.

Thanks, all checks now passed.

samerbuna:
Let's put the query text in a variable using a template string (using backticks and dropping the `\n`).  Define it in `src/graphql/playground.ts` and import it in `graphql-server.ts`

samerbuna (changes requested):
Let's put the query text in a variable using a template string (using backticks and dropping the `\n`).  Define it in `src/graphql/playground.ts` and import it in `graphql-server.ts`

vindard (dismissed):
Looks good. Only suggestion I'd add is to potentially also have a `name` property on the tab, maybe something like `btcPriceList Sample Query`. Not a blocker though.

moksha256:
@samerbuna @vindard thanks for the feedback. I added a tab name and defined all 3 query-related strings in `src/graphql/playground.ts` since it seemed like they should be together.

src/graphql/playground.ts
```diff
@@ -0,0 +1,17 @@
...
+        formattedAmount
+      }
+    }
+}`
+
+const defaultQueryName = "btcPriceList (Sample Query)"
+
+const defaultQueryVars = '{"range": "ONE_DAY"}'
+
+export { defaultQuery, defaultQueryName, defaultQueryVars }
```
vindard, line 17:
```suggestion
export const BTC_PRICE_QUERY = `query btcPriceList($range: PriceGraphRange!) {
  btcPriceList(range: $range) {
      timestamp
      price {
      base
      offset
      currencyUnit
      formattedAmount
    }
  }
}`

export const playgroundTabs = {
  default: {
    query: BTC_PRICE_QUERY,
    name: "btcPriceList (Sample Query)",
    variables: JSON.stringify({ range: "ONE_DAY" }),
  },
}

```

src/servers/graphql-server.ts
```diff
@@ -147,7 +148,17 @@ export const startApolloServer = async ({
...
     playground: apolloConfig.playground
-      ? { settings: { "schema.polling.enable": false } }
+      ? {
+          settings: { "schema.polling.enable": false },
+          tabs: [
+            {
+              endpoint: "https://api.staging.galoy.io/graphql",
+              query: defaultQuery,
+              name: defaultQueryName,
+              variables: defaultQueryVars,
```
vindard, line 158:
```suggestion
              ...playgroundTabs.default,
```

src/servers/graphql-server.ts
```diff
@@ -31,6 +31,7 @@ import {
   ENDUSER_ALIAS,
 } from "@services/tracing"
 import { AccountsRepository } from "@services/mongoose"
+import { defaultQuery, defaultQueryName, defaultQueryVars } from "../graphql/playground"
```
vindard, line 34:
```suggestion
import { playgroundTabs } from "../graphql/playground"
```

vindard (commented):
Oh that makes sense collecting all the values in one file. I'm suggesting an alternative layout here that we can optionally use as well.

moksha256:
Not sure if anyone else has feedback for vindard's changes so I just committed them and tested...looks good from my end.

vindard (approved):

nicolasburtey (approved):

vindard:
> Let's put the query text in a variable using a template string (using backticks and dropping the `\n`).

@samerbuna how would you drop the `\n` from the query string here?

PR #807, opened 2021-12-14
docs(lightning): add a rough description of the lib's error type

vindard:
### Description

An attempt at describing the object that gets thrown from the `lightning` library we use, based on the request in [this comment](https://github.com/GaloyMoney/galoy/pull/786#discussion_r762862086).

It wasn't possible (afaik) to define as a usual Typescript type since the object was an array with specific values expected at specific indices. I instead included a rough description as a multi-line comment above the method we've implemented to parse these errors.

### DIFF
--- a/src/services/lnd/index.ts
+++ b/src/services/lnd/index.ts
@@ -485,7 +485,7 @@ const lookupPaymentByPubkeyAndHash = async ({
 // where '<Error Code Details Object>' is an Error object with
 // the usual 'message', 'stack' etc. properties and additional
 // properties: 'code', 'details', 'metadata'.
-const parseLndErrorDetails = (err) =>
+export const parseLndErrorDetails = (err) =>
   err[2]?.err?.details || err[2]?.failures?.[0]?.[2]?.err?.details || err[1]
 
 const KnownLndErrorDetails = {
--- a/src/services/lnd/index.ts
+++ b/src/services/lnd/index.ts
@@ -462,6 +462,29 @@ const lookupPaymentByPubkeyAndHash = async ({
   }
 }
 
+// A rough description of the error type we get back from the
+// 'lightning' library can be described as:
+//
+// [
+//   0: <Error Classification Code Number>
+//   1: <Error Type String>
+//   2: {
+//     err?: <Error Code Details Object>
+//     failures?: [
+//       [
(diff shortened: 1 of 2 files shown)


src/services/lnd/index.ts
```diff
@@ -462,6 +462,27 @@ const lookupPaymentByPubkeyAndHash = async ({
...
+//         }
+//     ][]
+//   }
+// ]
+//
+// where '<Error Code Details Object>' is an Error object with
+// the usual 'message', 'stack' etc. properties and additional
+// properties: 'code', 'details', 'metadata'.
 const parseLndErrorDetails = (err) =>
   err[2]?.err?.details || err[2]?.failures?.[0]?.[2]?.err?.details || err[1]
```
nicolasburtey, line 487: shouldn't it be `err[2]?.failures?.[2]?.err?.details` then? (dropping a `[0]`)

vindard: `failures` is an array of failure objects (where was object has the 0/1/2 indices). I tried to capture this with the `[]` at the very end of that failures object.

nicolasburtey: yup ok

vindard: [Refactored](https://github.com/GaloyMoney/galoy/compare/8b5e80ebeeec87b4f9b1d42e5c547739a78b1c8a..2c07f214fb63bf88559d4b9b3145fc596449365e) to hopefully make that clearer

dolcalmi (approved):

src/services/lnd/index.ts
```diff
@@ -462,7 +462,30 @@ const lookupPaymentByPubkeyAndHash = async ({
   }
 }
 
-const parseLndErrorDetails = (err) =>
+// A rough description of the error type we get back from the
```
bodymindarts, line 465: Is this comment still needed with the test? Can we just say something like:
```
// to see our assumption of how the error is constructed see the <xxx> test
```

bodymindarts (approved):
Approving with comment

src/services/lnd/index.ts
vindard: @nicolasburtey had originally requested it, so once he's ok with removing I don't see why not

PR #808, opened 2021-12-15
refactor: removing walletpublicid inside our app

nicolasburtey:


src/app/wallets/get-csv-for-wallets.ts
```diff
@@ -1,12 +1,12 @@
...
-import { accountPath } from "@services/ledger/accounts"
+import { walletPath } from "@services/ledger/accounts"
 
 export const getCSVForWallets = async (
   walletIds: WalletId[],
 ): Promise<string | ApplicationError> => {
   const csv = new CSVAccountExport()
   for (const walletId of walletIds) {
-    await csv.addAccount({ account: accountPath(walletId) })
+    await csv.addAccount({ account: walletPath(walletId) })
```
bodymindarts, line 9: This seems to still be on the wrong side of the core / ledger boundary - is that right?
Not saying it needs to be done in this PR just asking to check if we have a shared understanding.

src/app/wallets/intraledger-send-payment.ts
```diff
@@ -157,15 +157,19 @@ const executePaymentViaIntraledger = async ({
...
     walletId,
   })
+
   if (intraledgerLimitCheck instanceof Error) return intraledgerLimitCheck
 
   const recipientUser = await UsersRepository().findByUsername(recipientUsername)
   if (recipientUser instanceof Error) return recipientUser
   if (recipientUser.id === userId) return new SelfPaymentError()
+  // FIXME: selfPayment should be at the walletId level, no longer at the UserId
+  // but this is not a bloquer until we have multiple wallets per account
```
bodymindarts, line 167: Typo in `blocker` and I prefer to have comments above the line that they refer to....

no big deal.

src/core/lightning/wallet.ts
```diff
@@ -35,7 +35,7 @@ export class LightningUserWallet extends OnChainMixin(UserWallet) {
       logger: this.logger,
     })
 
-    return redlock({ path: this.user._id, logger: this.logger }, async () => {
+    return redlock({ path: this.user.id, logger: this.logger }, async () => {
```
bodymindarts, line 38: Will this change be an issue if there is a lock inflight during upgrade?

src/domain/ledger/index.ts
```diff
@@ -31,8 +31,8 @@ export const ExtendedLedgerTransactionType = {
   LnIntraLedger: "ln_on_us",
 } as const
 
-export const toWalletId = (accountId: LiabilitiesAccountId): WalletId | null => {
-  const path = accountId.split(":")
+export const toWalletId = (walletIdPath: LiabilitiesWalletId): WalletId | null => {
```
bodymindarts, line 34: I thought we agreed to go with `CoreAccount` and `LiabilitiesAccount` but here its still `LiabilitiesWallet`? Is this pending a follow up PR?

src/domain/wallets/index.ts
```diff
@@ -1,19 +1,21 @@
-import { InvalidPublicWalletId } from "@domain/errors"
+// import { InvalidWalletId } from "@domain/errors"
```
bodymindarts, line 1: Delete dead code?

src/domain/wallets/index.ts
```diff
@@ -1,19 +1,21 @@
...
-  if (!walletPublicId.match(WalletPublicIdRegex)) {
-    return new InvalidPublicWalletId(walletPublicId)
-  }
-  return walletPublicId as WalletPublicId
+export const checkedToWalletId = (walletId: string): WalletId | ValidationError => {
+  return walletId as WalletId
+
+  // TODO: uncomment after walletId is using uuid v4
+
+  // if (!walletId.match(WalletIdRegex)) {
```
bodymindarts, line 17: Committed dead code?

src/migrations/20211209211641-use-walletid-for-ledger.ts
```diff
@@ -0,0 +1,18 @@
+module.exports = {
+  async up(db) {
+    {
+      const result = await db.collection("users").updateMany({}, [
```
bodymindarts, line 4: I thought this PR should be non-risk ie. without a migration?

src/app/wallets/get-csv-for-wallets.ts
nicolasburtey: that is correct. `walletPath` should not be used outside ledger (eventually)

src/app/wallets/intraledger-send-payment.ts
nicolasburtey: oh yeah I had used the french version lol

src/core/lightning/wallet.ts
nicolasburtey: `_id` and `.id` are the same value, so no.

src/domain/ledger/index.ts
nicolasburtey: yes

src/app/wallets/get-csv-for-wallets.ts
bodymindarts: Right so probably this use case is not on `app/wallets` but `app/ledger`

src/core/lightning/wallet.ts
bodymindarts: afaik one of them is a string serizliazation and the other is an object

src/migrations/20211209211641-use-walletid-for-ledger.ts
nicolasburtey: this is a migration just to copy `walletPublicId` to `walletId` in the user collection. this is not the risky part. 

the "risky" part is related to the update of the Ledger

bodymindarts (dismissed):
I'm approving to move forward. Don't see any critical issues.

src/servers/graphql-old-server.ts
```diff
@@ -384,7 +361,7 @@ const resolvers = {
         let feeSatAmount: Satoshis | ApplicationError
         if (amount && amount > 0) {
           feeSatAmount = await getNoAmountLightningFee({
-            walletPublicId: wallet.user.walletPublicId,
+            walletId: wallet.user.id,
```
vindard, line 364: Should this be `wallet.user.walletId` after the migration?

vindard (commented):
Just one question about the name of the `walletId` property on the `UserType` object we get back from mongo. Everything else looks ok to approve after that question.

src/servers/graphql-old-server.ts
nicolasburtey: I think this should be `user.id` here. where are the other place where we still access `user.walletId`? 

there are some placed that I have replaced here: https://github.com/GaloyMoney/galoy/pull/808/commits/e6d1022f74ca432b0711aae63bfe24b85c98af21

bodymindarts (approved):

vindard (approved):

src/servers/graphql-old-server.ts
vindard: Sorry I was looking at the wrong branch locally 🙈  ... approved ✅

PR #810, opened 2021-12-15
refactoring: remove some dependancy on non typed function in ledger f…

nicolasburtey:
…rom core/test

ledger/query.ts is code copied/paste from core without type function. remove the one that have no longer any reasons to be here


src/services/ledger/csv-wallet-export.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    await csvWriter.writeRecords(this.entries)
+    console.log("saving complete")
+  }
+
+  async addWallet(wallet: WalletId): Promise<void | Error> {
+    // TODO: interface could be improved by returning self, so that it's
+    // possible to run csv.addWallet(wallet).getBase64()
+    const txs = await LedgerService().getLiabilityTransactions(wallet)
+    if (txs instanceof Error) return txs
```
vindard, line 66:
```suggestion
  async addWallet(walletId: WalletId): Promise<void | ApplicationError> {
    // TODO: interface could be improved by returning self, so that it's
    // possible to run csv.addWallet(wallet).getBase64()
    const txs = await LedgerService().getLiabilityTransactions(walletId)
    if (txs instanceof Error) return txs
```

src/services/ledger/index.ts
```diff
@@ -617,6 +617,26 @@ export const LedgerService = (): ILedgerService => {
     }
   }
 
+  const getWalletIdByTransactionHash = async (
+    hash,
+  ): Promise<WalletId | CouldNotFindTransactionError | UnknownLedgerError> => {
```
vindard, line 622: It looks like we've been keeping this more generic so far

```suggestion
  ): Promise<WalletId | LedgerError> => {
```

test/integration/02-user-wallet/02-index.spec.ts
```diff
@@ -134,9 +135,11 @@ describe("UserWallet", () => {
 
   describe("getStringCsv", () => {
     const csvHeader =
-      "voided,approved,_id,accounts,credit,debit,_journal,book,unix,date,datetime,currency,username,type,hash,txid,fee,feeUsd,sats,usd,memo,memoPayer,meta,pending"
+      "id,walletId,type,credit,debit,fee,currency,timestamp,pendingConfirmation,journalId,lnMemo,usd,feeUsd,recipientWalletId,username,memoFromPayer,paymentHash,pubkey,feeKnownInAdvance,address,txHash"
     it("exports to csv", async () => {
-      const base64Data = await userWallet0.getStringCsv()
```
vindard, line 139: Would we be able to make this switch in `graphql-old-server.ts` as well and delete this `getStringCsv` method entirely?

vindard (changes requested):
Some small suggestions, but otherwise looks good

src/services/ledger/csv-wallet-export.ts
nicolasburtey: but `LedgerServiceError` is not a child of `ApplicationError`?

test/integration/02-user-wallet/02-index.spec.ts
nicolasburtey: I'm not actively looking at removing code from `core` until there is dependency outside of core. the idea is to come to a stage where we can just delete the core folder entirely.

src/services/ledger/csv-wallet-export.ts
vindard: It should be, included at this point: https://github.com/GaloyMoney/galoy/blob/709a3e1727bf101a70a3514f28a043612159e470/src/app/errors.ts#L17

nicolasburtey: ok we use composition for `ApplicationError`. is it the only place we do it so?

vindard: I think the easiest way to test that is to see if it's included in the `error-map.ts` switch (https://github.com/GaloyMoney/galoy/blob/709a3e1727bf101a70a3514f28a043612159e470/src/graphql/error-map.ts#L169)

That switch is designed to not compile if:
- something that isn't an `ApplicationError` gets included
- something that is an `ApplicationError` gets left out

vindard: > ok we use composition for `ApplicationError`. is it the only place we do it so?

Yup only place where we don't use the `class ... extends .. ` pattern to create the error so far.

vindard (approved):

PR #811, opened 2021-12-15
fix: regression publicWalletId -> walletId PR

nicolasburtey:


bodymindarts (approved):
Why isn’t this caught statically? @dolcalmi

PR #813, opened 2021-12-15
feat: add telemetry wrappers

dolcalmi:
Findings:

- Is not possible to do dynamic exports (so we cant do automatic wrapping without modify TS pipeline)  
- The alternative to add span at function level is to wrap at module level (index files)
- This implementation does not break Typescript types and/or IDE integration

The code in this PR is only to validate if we prefer it against the current implementation.
- Add wrappers in lightning `sub-module` ( 'dynamic' way). Please keep in mind that we can use this code at `app/index.ts` to support all sub-modules.
- Update current `getUserForLogin` span (manual way)
- In wrapAsyncToRunInSpan we can use args to add common attributes to the span (walletId, userId, ...)

### DIFF
--- a/src/app/lightning/index.ts
+++ b/src/app/lightning/index.ts
@@ -1,3 +1,18 @@
-export * from "./payment-status-checker"
-export * from "./lookup-invoice-by-hash"
-export * from "./lookup-payment-by-hash"
+import { wrapAsyncToRunInSpan, wrapToRunInSpan } from "@services/tracing"
+import * as LookupInvoiceByHash from "./lookup-invoice-by-hash"
+import * as LookupPaymentByHash from "./lookup-payment-by-hash"
+import { PaymentStatusChecker as PaymentStatusCheckerFn } from "./payment-status-checker"
+
+// creates an object with all the functions of imported modules
+const dynamicWrapper = { ...LookupInvoiceByHash, ...LookupPaymentByHash }
+
+for (const fn in dynamicWrapper) {
+  // wrap with async span, we need to be careful with functions that don't return a promise like PaymentStatusChecker
+  dynamicWrapper[fn] = wrapAsyncToRunInSpan({ fn: dynamicWrapper[fn] })
+}
+
+// Still we need to do a static export
+export const { PaymentStatusChecker, lookupInvoiceByHash, lookupPaymentByHash } = {
+  PaymentStatusChecker: wrapToRunInSpan({ fn: PaymentStatusCheckerFn }),
+  ...dynamicWrapper,
+}
--- a/src/services/tracing.ts
+++ b/src/services/tracing.ts
@@ -171,6 +171,64 @@ export const asyncRunInSpan = <F extends () => ReturnType<F>>(
   return ret
 }
 
+// eslint-disable-next-line  @typescript-eslint/no-explicit-any
+export const wrapToRunInSpan = <A extends Array<any>, R>({
+  fn,
+  fnName,
+  namespace = "app",
+  spanAttributes = {},
+}: {
+  fn: (...args: A) => R
+  fnName?: string
+  namespace?: string
+  spanAttributes?: SpanAttributes
+}) => {
+  return (...args: A): R => {
+    const name = fnName || fn.name
+    const spanName = `${namespace}.${name}`
+    const attributes = { [SemanticAttributes.CODE_FUNCTION]: name, ...spanAttributes }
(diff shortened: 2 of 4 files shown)


bodymindarts:
Is there a way to add the function arguments automatically as well? Otherwise we don't really win much as we must still add the attributes we want the 'old' way.

dolcalmi:
> Is there a way to add the function arguments automatically as well? Otherwise we don't really win much as we must still add the attributes we want the 'old' way.

yes, please check https://github.com/GaloyMoney/galoy/pull/813/commits/bef3ca7fc2d47b86e2f8cf691422cedf1dd30b9d

src/app/lightning/index.ts
```diff
@@ -1,3 +1,18 @@
-export * from "./payment-status-checker"
-export * from "./lookup-invoice-by-hash"
-export * from "./lookup-payment-by-hash"
+import { wrapAsyncToRunInSpan, wrapToRunInSpan } from "@services/tracing"
+import * as LookupInvoiceByHash from "./lookup-invoice-by-hash"
+import * as LookupPaymentByHash from "./lookup-payment-by-hash"
+import { PaymentStatusChecker as PaymentStatusCheckerFn } from "./payment-status-checker"
+
+// creates an object with all the functions of imported modules
+const dynamicWrapper = { ...LookupInvoiceByHash, ...LookupPaymentByHash }
```
nicolasburtey, line 7: maybe the name below make more sense?

```suggestion
const allAsyncFunctions = { ...LookupInvoiceByHash, ...LookupPaymentByHash }
```

src/app/lightning/index.ts
```diff
@@ -1,3 +1,18 @@
...
+
+for (const fn in dynamicWrapper) {
+  // wrap with async span, we need to be careful with functions that don't return a promise like PaymentStatusChecker
+  dynamicWrapper[fn] = wrapAsyncToRunInSpan({ fn: dynamicWrapper[fn] })
+}
+
+// Still we need to do a static export
+export const { PaymentStatusChecker, lookupInvoiceByHash, lookupPaymentByHash } = {
+  PaymentStatusChecker: wrapToRunInSpan({ fn: PaymentStatusCheckerFn }),
```
nicolasburtey, line 16: I presume we're wrapping `PaymentStatusChecker` differently because it's running in sync? we could add the wrapper before the `export const` to normalize this, in a `allSyncFunctions` loop.

I guess this is just an example here and when agreed upon we would replicate for other apps?

src/services/tracing.ts
```diff
@@ -171,6 +171,78 @@ export const asyncRunInSpan = <F extends () => ReturnType<F>>(
   return ret
 }
 
+// eslint-disable-next-line  @typescript-eslint/no-explicit-any
+export const wrapToRunInSpan = <A extends Array<any>, R>({
```
nicolasburtey, line 175: `unknown` doesn't work instead of `any`?

src/services/tracing.ts
```diff
@@ -171,6 +171,78 @@ export const asyncRunInSpan = <F extends () => ReturnType<F>>(
...
+  namespace = "app",
+  spanAttributes = {},
+}: {
+  fn: (...args: A) => R
+  fnName?: string
+  namespace?: string
+  spanAttributes?: SpanAttributes
+}) => {
+  return (...args: A): R => {
+    const name = fnName || fn.name
```
nicolasburtey, line 187: could we not always use `fn.name`? is there some use case where we specifically need to change the function name?

src/services/tracing.ts
dolcalmi: is giving me type errors :thinking: do you have some suggestion to implement it? cc @samerbuna

src/app/lightning/index.ts
dolcalmi: I did a different implementation, please check https://github.com/GaloyMoney/galoy/blob/13d967d7b89118d4648cc6c423e4084fca54e66a/src/app/index.ts

src/app/lightning/index.ts
dolcalmi: Updated in the new PR

src/services/tracing.ts
dolcalmi: removed

dolcalmi:
closed in favor of https://github.com/GaloyMoney/galoy/pull/837

PR #814, opened 2021-12-16
Fix all instances of our MongoDb id as our new `walletId` value

vindard:
## Description

⚠️  **_This fixes errors currently deployed to staging_**

I started by fixing a `"CastError: Cast to ObjectId failed for value \"ae2f7b11-1ee6-41fa-9264-3867dda26831\" at path \"_id\" for model \"User\""` when passing the `walletId` we persisted into requests which then broke our tests and snowballed into changing all the places I could find where were passing the mongoDb `id` in places where we should be passing our `uuid4`-formatted `walletId`.

### DIFF
--- a/src/services/mongoose/schema.ts
+++ b/src/services/mongoose/schema.ts
@@ -285,7 +285,7 @@ UserSchema.virtual("ratioBtc").get(function (this: typeof UserSchema) {
 
 // this is the accounting path in medici for this user
 UserSchema.virtual("walletPath").get(function (this: typeof UserSchema) {
-  return walletPath(this.id)
+  return walletPath(this.walletId)
 })
 
 UserSchema.virtual("oldEnoughForWithdrawal").get(function (this: typeof UserSchema) {
--- a/src/core/lightning/wallet.ts
+++ b/src/core/lightning/wallet.ts
@@ -50,7 +50,7 @@ export class LightningUserWallet extends OnChainMixin(UserWallet) {
         if (userPastState.earn.findIndex((item) => item === id) === -1) {
           // FIXME: use pay by username instead
           const lnInvoice = await addInvoice({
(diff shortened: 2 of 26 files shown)


dolcalmi (dismissed):
I did some random tests locally and looks good

bodymindarts (dismissed):

src/services/mongoose/schema.ts
```diff
@@ -285,7 +285,7 @@ UserSchema.virtual("ratioBtc").get(function (this: typeof UserSchema) {
 
 // this is the accounting path in medici for this user
 UserSchema.virtual("walletPath").get(function (this: typeof UserSchema) {
-  return walletPath(this.id)
+  return walletPath(this.walletId)
```
nicolasburtey, line 288: Need to dig deeper but IMO this will not work. we still want to use `User.id` because this is used in accounting, and currently `User.id` is what needs to be stored.

nicolasburtey (changes requested):
I think some changes here will break the accounting. digging deeper to get the bottom of it.

nicolasburtey:
just running `TEST="01" make reset-integration`

and this is what is being stored: 

image

`WalletId` is being stored in `Liabilities`, but it should be `UserId`. 

I'm wondering if we should just scratched this and get back my older PR? https://github.com/GaloyMoney/galoy/pull/792
I felt trying to separate those under 2 PR was not a great ideas but couldn't have a good argument, now I have a good argument is that we may loose a lot of time splitting this PR in 2. 

thoughts @vindard @bodymindarts ?

vindard:
> I'm wondering if we should just scratched this and get back my older PR? #792

Yup I think this ends up being a PR to show how trying to make the types consistent ends up in that `accounts` change and that the two issues are probably coupled with our current design 🙈

PR #816, opened 2021-12-16
some early test to simulate galoy-pay client request

nicolasburtey:


nicolasburtey:
taking over by https://github.com/GaloyMoney/galoy/pull/859

PR #821, opened 2021-12-17
fix: the error check was not checking anything

nicolasburtey:
the error types didn't had any overlap

note: I'm a bit loss still about the way we're created the error. sometime the error extends directly from Error, sometime from RepositoryError.

### DIFF
--- a/src/app/lightning/payment-status-checker.ts
+++ b/src/app/lightning/payment-status-checker.ts
@@ -1,6 +1,7 @@
 import { RepositoryError } from "@domain/errors"
 import { decodeInvoice } from "@domain/bitcoin/lightning"
 import { LedgerService } from "@services/ledger"
+import { LedgerServiceError } from "@domain/ledger"
 
 export const PaymentStatusChecker = ({ paymentRequest }) => {
   const decodedInvoice = decodeInvoice(paymentRequest)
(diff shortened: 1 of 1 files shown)


src/app/lightning/payment-status-checker.ts
```diff
@@ -14,7 +15,7 @@ export const PaymentStatusChecker = ({ paymentRequest }) => {
     invoiceIsPaid: async (): Promise<boolean | RepositoryError> => {
       const ledger = LedgerService()
       const recorded = await ledger.isLnTxRecorded(paymentHash)
-      if (recorded instanceof RepositoryError) return recorded
+      if (recorded instanceof LedgerServiceError) return recorded
```
dolcalmi, line 18: this is one of the first implementations (is not consistent with actual pattern). If we are not doing error wrapping then you should use only `Error`

```suggestion
      if (recorded instanceof Error) return recorded
```

dolcalmi (changes requested):

vindard (approved):
Adding approval, but make-check won't pass until we remove that unused import

nicolasburtey:
> Adding approval, but make-check won't pass until we remove that unused import

fixed

dolcalmi (approved):

vindard (approved):

PR #822, opened 2021-12-17
fix: my updates subscription regression

dolcalmi:


src/graphql/root/subscription/my-updates.ts
```diff
@@ -108,7 +108,10 @@ const MeSubscription = {
       })
     }
 
-    return pubsub.asyncIterator([USER_PRICE_UPDATE_EVENT, walletUpdateEvent(ctx.uid)])
+    return pubsub.asyncIterator([
+      USER_PRICE_UPDATE_EVENT,
+      walletUpdateEvent(ctx.wallet.user.walletId),
```
nicolasburtey, line 113: not sure what ctx here is, but couldn't it be just ctx.walletId? (assuming ctx is an InvoiceUser object?)

vindard: If I understand correctly `ctx` is [this object](https://github.com/GaloyMoney/galoy/blob/398d149f02f53d4d70c79244b57ff9d0331974f3/src/servers/graphql-server.ts#L115) with the following properties:
```
{
        logger,
        uid: userId,
        wallet,
        // FIXME: we should not return this for the admin graphql endpoint
        domainUser,
        domainAccount,
        user,
        geetest,
        account,
        ip,
}
```

In that object `wallet` still comes from `WalletFactory` and is a `LightningUserWallet` and I also don't see any other place we could access the `walletId` from as yet.

nicolasburtey: domainAccount.defaultWalletId?

vindard: Or that 🙈

dolcalmi: updated and tested

vindard (approved):

samerbuna (approved):

PR #825, opened 2021-12-17
Add concurrency group to docs gen

krtk6160:


nicolasburtey:
what is the purpose of this concurrency?

you can retrigger the build by doing a `git push --force`, this avoid the need of a debug commits/ "bad" github history

krtk6160:
> what is the purpose of this concurrency?

You had highlighted twice an issue with our doc generation github action. If there are two quick merges to main, the second one triggers a job that fails. Adding concurrency to this particular github action will fix the problem, by ensuring that new doc building jobs are queued.

> you can retrigger the build by doing a `git push --force`, this avoid the need of a debug commits/ "bad" github history

We have squashing now so I don't think that will be a problem. Unless you meant something else?

nicolasburtey:
> We have only squashing allowed now so I don't think that will be a problem. Unless you meant something else?

yes that's correct, but when reviewing the PR then you would only have the relevant commit. not a big deal with squash indeed.

nicolasburtey:
before squash it used to be that all the commits could be kept. now it means that the PR is just "harder" to review.

nicolasburtey (approved):

PR #827, opened 2021-12-18
Add earn refactor

nicolasburtey:


default.yaml
```diff
@@ -38,6 +38,7 @@ test_accounts:
   code: 321321
 - phone: "+16505554322" # user1
   code: 321432
+  phoneMetadata_carrier_type: "mobile" # TODO: verify mobile is a valid value
```
dolcalmi, line 41: Please change to camel case. 

in other configs we use camel case. Here we are using both (camel and snake case)

src/app/accounts/add-earn.ts
```diff
@@ -0,0 +1,68 @@
...
+import { onboardingEarn } from "@config/app"
+import {
+  RewardInsufficientBalanceError,
+  RewardMissingMetadataError,
+  RewardNonValidTypeError,
+  ValidationError,
+} from "@domain/errors"
+import { getFunderWalletId } from "@services/ledger/accounts"
+import { RewardsRepository } from "@services/mongoose/rewards"
+import { getAccount } from "."
```
dolcalmi, line 12: we should avoid this. importing index could lead to future circular references. I think is better move getAccount to its own file.

src/app/accounts/add-earn.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export const addEarn = async ({
+  id,
+  aid,
```
dolcalmi, line 16: please change to `accountId`

src/app/accounts/add-earn.ts
```diff
@@ -0,0 +1,68 @@
...
+  RewardMissingMetadataError,
+  RewardNonValidTypeError,
+  ValidationError,
+} from "@domain/errors"
+import { getFunderWalletId } from "@services/ledger/accounts"
+import { RewardsRepository } from "@services/mongoose/rewards"
+import { getAccount } from "."
+
+export const addEarn = async ({
+  id,
```
dolcalmi, line 15: please change to `quizQuestionId`

src/services/mongoose/rewards.ts
```diff
@@ -0,0 +1,31 @@
+import { RewardAlreadyPresentError, UnknownRepositoryError } from "@domain/errors"
+import { User } from "./schema"
+
+// FIXME: improve boundary
+export const RewardsRepository = (accountId: AccountId) => {
+  const tentativelyAddNew = async (quizQuestionId: QuizQuestionId) => {
```
dolcalmi, line 6: we are not doing a rollback so maybe tentatively is not necessary

src/services/mongoose/rewards.ts
```diff
@@ -0,0 +1,31 @@
...
+      )
+
+      const rewardNotFound =
+        oldState.earn.findIndex((item) => item === quizQuestionId) === -1
+
+      if (rewardNotFound) {
+        return true
+      } else {
+        return new RewardAlreadyPresentError()
+      }
```
dolcalmi, line 22:
```suggestion
      return rewardNotFound || new RewardAlreadyPresentError()
```

dolcalmi (changes requested):

src/services/mongoose/rewards.ts
nicolasburtey: the `tentative` is here to suggest that, if the rewards already exist, we will not credit it to the user. this may be misleading?

src/app/accounts/add-earn.ts
nicolasburtey: make sense, just reusing a pattern that was already here. how would you name that file?

nicolasburtey: I'll do that on another PR

dolcalmi: just `get-account.ts`

dolcalmi (approved):

PR #828, opened 2021-12-18
chore: first part of add-earn refactoring

nicolasburtey:
this is an extract from this PR, to merge all the non core part first
https://github.com/GaloyMoney/galoy/pull/827


src/app/users/get-user.ts
```diff
@@ -33,8 +32,9 @@ export const getUserForLogin = async ({
         return user
       }
       addAttributesToCurrentSpan({
-        [ENDUSER_ALIAS]: user.username,
+        [SemanticAttributes.ENDUSER_ID]: user.id,
```
bodymindarts, line 35: The `ENDUSER_ID` is already added https://github.com/GaloyMoney/galoy/blob/main/src/servers/graphql-server.ts#L78 `ENDUSER_ALIAS` is not.

src/app/wallets/get-balance-for-wallet.ts
```diff
@@ -30,7 +30,7 @@ export const getBalanceForWallet = async ({
 export const getBalanceForWalletId = async (
   walletId: WalletId,
 ): Promise<Satoshis | ApplicationError> => {
-  const balance = await LedgerService().getAccountBalance(walletId)
+  const balance = await LedgerService().getWalletBalance(walletId)
```
bodymindarts, line 33: As far as we discussed we agreed to go with `LedgerAccount` as the naming - so that would imply this function should be `getLedgerAccountBalanceForWalletId` or something like that.

src/servers/exporter.ts
```diff
@@ -69,6 +70,7 @@ const business_g = new client.Gauge({
 })
 
 const roles = ["dealer", "funder", "bankowner"]
+const account_roles = [getDealerWalletId(), getFunderWalletId(), getBankOwnerWalletId()]
```
bodymindarts, line 73: camlCase

bodymindarts (commented):
You are executing the `Account` -> `Wallet` renaming here for the `ledger` sub-domain but we agreed to go with `LedgerAccount` naming.

src/servers/exporter.ts
nicolasburtey: ?

src/app/users/get-user.ts
nicolasburtey: there is no longer a concept of `ENDUSER_ALIAS`, should be account related now. should I just delete this for now?

src/graphql/types/object/wallet-contact.ts
```diff
@@ -43,7 +41,7 @@ const UserContact = new GT.Object({
           throw account
         }
 
-        const transactions = await Wallets.getAccountTransactionsForContact({
+        const transactions = await getAccountTransactionsForContact({
```
vindard, line 44: Small nit, why not have this imported as `Accounts.getAccountTransactionsForContact` instead?

src/servers/exporter.ts
```diff
@@ -108,19 +110,14 @@ const main = async () => {
...
-      try {
-        const wallet = await getWalletFromRole({ role, logger })
-        const balanceSats = await Wallets.getBalanceForWallet({
-          walletId: wallet.user.walletId as WalletId,
-          logger,
-        })
-        if (balanceSats instanceof Error) throw balanceSats
+    for (const index in roles) {
+      const role = roles[index]
+      const account = await account_roles[index]
```
vindard, line 115: Could this loose coupling be tricky to maintain in the future vs. something like an object with `role` as key and `account_role` as value?

src/services/ledger/index.ts
```diff
@@ -24,19 +24,22 @@ import {
...
 }
 
 export const loadLedger = ({
   bankOwnerWalletResolver,
   dealerWalletResolver,
+  funderWalletResolver,
 }: LoadLedgerParams) => {
   accounts.setbankOwnerWalletResolver(bankOwnerWalletResolver)
   accounts.setdealerWalletResolver(dealerWalletResolver)
+  accounts.setFunderWalletResolver(funderWalletResolver)
```
vindard, line 42: Can probably make the camel-casing for these consistent as we're touching this code already here
```suggestion
  accounts.setBankOwnerWalletResolver(bankOwnerWalletResolver)
  accounts.setDealerWalletResolver(dealerWalletResolver)
  accounts.setFunderWalletResolver(funderWalletResolver)
```

vindard (commented):
Added some small suggestions

src/app/wallets/intraledger-send-payment.ts
```diff
@@ -37,10 +37,15 @@ export const intraledgerPaymentSend = async ({
   const user = await UsersRepository().findById(userId)
```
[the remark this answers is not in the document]
nicolasburtey: using this to also get the `defaultAccountId` from the user. 

Note: there are more refactor on this on the second part of the PR for refactoring for add_earn, as we can move away from "usernameSend" and use "PublicWalletIdSend" instead. the refactor allow to send money from the funder account for the add earn without using the username.

src/graphql/types/object/wallet-contact.ts
nicolasburtey: no particular reasons

src/servers/exporter.ts
nicolasburtey: the idea I have is to stop manually calling "role" outside of the ledger as I have been using it for now, but instead of load it at startup and keep in memory here. the loading happens here: [src/services/mongodb/index.ts](https://github.com/GaloyMoney/galoy/blob/main/src/services/mongodb/index.ts)

nicolasburtey: this will improve efficiency because we were fetching previously the _id associated to account for every transaction

src/services/ledger/index.ts
nicolasburtey: done

src/app/users/get-user.ts
bodymindarts: Sure delete it then

src/servers/exporter.ts
bodymindarts: `account_roles ` => `accountRoles` - or is it `snake_case` because a dependency expects it?

nicolasburtey: oh yes ok. was looking at the array itself: getDealerWalletId and co.

bodymindarts (approved):

vindard (approved):

PR #829, opened 2021-12-19
chore: one user should only connect to one account

nicolasburtey:
managing user that could have several accounts is a very complex use case, that we should not try to cover for the time being (and probably for some time). most financial product from fintech that I have used in the recent past didn't offer this capability for this same reason


bodymindarts:
This has quite some implications - not saying we shouldn't do it, but probably a good topic for code review

nicolasburtey:
> This has quite some implications - not saying we shouldn't do it, but probably a good topic for code review

why does it has some implications? I don't think there is any other places in the code with the assumption that user can be access to many accounts

PR #835, opened 2021-12-20
refactor: bitcoind/specter types

nicolasburtey:


src/domain/bitcoin/onchain/index.types.d.ts
```diff
@@ -65,3 +71,63 @@ interface IOnChainService {
     targetConfirmations: TargetConfirmations,
   ): Promise<Satoshis | OnChainServiceError>
 }
+
+interface IBitcoindService {
```
bodymindarts, line 75: I need to think about the boundaries of these interfaces some more. I don't like that `Bitcoind` is in the name.
I think there can be a lot smaller more granular interfaces like `IBlockExplorer` and `IOnChainWalletRepository` or something like that

src/domain/bitcoin/onchain/index.types.d.ts
```diff
@@ -65,3 +71,63 @@ interface IOnChainService {
...
+    | { name: BitcoindWalletName; warning: string }
+    | UnknownRepositoryError
+    | WalletAlreadyExistError
+  >
+  listWallets(): Promise<BitcoindWalletName[]>
+  loadWallet(walletName: BitcoindWalletName): Promise<{ name: string; warning: string }>
+  unloadWallet(walletName: BitcoindWalletName): Promise<{ warning: string }>
+}
+
+interface IBitcoindWalletService {
```
bodymindarts, line 90: I don't see the reason for having `bitcoind` in the name. This could also be deconstructed into an `AddressRepository` a `PsbtProcessor` and a `TransactionBroadcaster`.

Having smaller interfaces is valuable because it reduces the coupling between use cases.

src/domain/bitcoin/index.types.d.ts
```diff
@@ -1,5 +1,7 @@
 type Satoshis = number & { readonly brand: unique symbol }
 type TargetConfirmations = number & { readonly brand: unique symbol }
 type MilliSatoshis = number & { readonly brand: unique symbol }
+type WholeBitcoin = number & { readonly brand: unique symbol }
```
bodymindarts, line 4: Don't quite understand this `WholeBitcoin` implies an `int` for me but I think its actually a float isn't it? I would prefer not to add this type and stick with a consistent `Sat` type and if translation is needed to talk to an external dependency do the translation in the service / repository implementation.
Or are there any other restrictions on this type other than where the decimal point is compared to `Satoshis`?

bodymindarts (commented):
I understand your motivation for moving this over - but I think the types here need more design work to get good boundaries.

src/domain/bitcoin/index.types.d.ts
nicolasburtey: yes this is a float. the idea is to convey that this number is denominated in bitcoin, not satoshis. 

the lightning wallet either takes millisats or sats as a unit of account. 
the bitcoind wallet takes bitcoin which is why I'm introducing this new type to make it more clear what number is being passed around. 

happy to have better suggestion for the name instead of whole bitcoin.

I think what you're saying also is that wholeBitcoin should be only present in the bitcoind internal service implementation instead of the the domain?

bodymindarts: Yes that is correct - in the domain we should be consistent and only have 1 type to represent bitcoin denominated value. I think `Satoshis` is the most natural here. That bitcoind wants a float is an implementation detail that should be taken care of in the service layer that talks to bitcoind.

PR #837, opened 2021-12-21
feat: add telemetry to app layer

dolcalmi:
- Update access to app layer in a consistent way to support telemetry from servers and tests
- Update code according to PR https://github.com/GaloyMoney/galoy/pull/813 comments 
- Update security sensitive attributes in otel config


src/app/index.ts
```diff
@@ -1,2 +1,29 @@
...
+  Users: { ...UsersMod },
+  Wallets: { ...WalletsMod },
+}
+const syncFunctions = ["PaymentStatusChecker"]
+
+for (const subModule in allFunctions) {
+  for (const fn in allFunctions[subModule]) {
+    const wrapper = syncFunctions.includes(fn) ? wrapToRunInSpan : wrapAsyncToRunInSpan
+    allFunctions[subModule][fn] = wrapper({
+      namespace: "app",
```
bodymindarts, line 23: should we encode the namespace? `app.accounts`, `app.admin`... etc? Would be easy to add here. As keys we should be using the names from [SemanticConvetions](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-semantic-conventions/src/trace/SemanticAttributes.ts#L340-L345)

src/app/wallets/ln-send-payment.ts
```diff
@@ -115,37 +105,30 @@ export const lnInvoicePaymentSend = async ({
...
   userId,
   logger,
-}: LnInvoicePaymentSendArgs): Promise<PaymentSendStatus | ApplicationError> =>
-  asyncRunInSpan(
-    "app.lnInvoicePaymentSend",
-    {
-      [SemanticAttributes.CODE_FUNCTION]: "lnInvoicePaymentSend",
-      "payment.initiation_method": PaymentInitiationMethod.Lightning,
-      "payment.wallet_id": walletId,
-      "payment.request": paymentRequest,
```
bodymindarts, line 125: `payment.request` is removed but not added back anywhere... or is it part of the arguments?

src/servers/trigger.ts
```diff
@@ -166,7 +165,7 @@ export const onInvoiceUpdate = async (invoice) => {
 }
 
 const publishCurrentPrice = async () => {
-  const usdPerSat = await getCurrentPrice()
+  const usdPerSat = await Prices.getCurrentPrice()
```
bodymindarts, line 168: Or these renamings required for the wrappers to work or are you just moving to a consistent style?

src/services/tracing.ts
```diff
@@ -171,6 +172,74 @@ export const asyncRunInSpan = <F extends () => ReturnType<F>>(
...
+  fn,
+  namespace,
+  spanAttributes = {},
+}: {
+  fn: (...args: A) => Promise<PromiseReturnType<R>>
+  namespace: string
+  spanAttributes?: SpanAttributes
+}) => {
+  return (...args: A): Promise<PromiseReturnType<R>> => {
+    const name = fn.name
```
bodymindarts, line 221: This section that sets up the span attributes is duplicate with `wrapToRunInSpan` - could probably extract into a helper function or?

src/app/wallets/ln-send-payment.ts
dolcalmi: yes, `paymentRequest` is an attribute

src/servers/trigger.ts
dolcalmi: in servers we need to use `Accounts, Admin, Lightning, Prices, Users, Wallets` from `@app` if we want to use telemetry wrapped functions.

src/services/tracing.ts
dolcalmi: :) in general I prefer a helper but in a previous PR a rule was defined against this if the code is not used more than twice

dolcalmi: done https://github.com/GaloyMoney/galoy/pull/837/commits/fea23b3ea8b3294e33d4d6cbb202b35aac28de88

src/app/index.ts
dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/837/commits/fea23b3ea8b3294e33d4d6cbb202b35aac28de88

src/app/index.ts
```diff
@@ -1,2 +1,29 @@
...
+  Lightning: { ...LightningMod },
+  Prices: { ...PricesMod },
+  Users: { ...UsersMod },
+  Wallets: { ...WalletsMod },
+}
+const syncFunctions = ["PaymentStatusChecker"]
+
+for (const subModule in allFunctions) {
+  for (const fn in allFunctions[subModule]) {
+    const wrapper = syncFunctions.includes(fn) ? wrapToRunInSpan : wrapAsyncToRunInSpan
```
nicolasburtey, line 21: only `PaymentStatusChecker` is sync? why not making this function async in that case so we don't have to go through this condition?

src/app/wallets/get-lightning-fee.ts
```diff
@@ -9,6 +8,7 @@ import { CachedRouteLookupKeyFactory } from "@domain/routes/key-factory"
 import { checkedToWalletId } from "@domain/wallets"
 import { LndService } from "@services/lnd"
 import { RoutesCache } from "@services/redis/routes"
+import { getBalanceForWallet } from "./get-balance-for-wallet"
```
nicolasburtey, line 11: I guess we should no longer import with `from "@app/wallets"` anymore after this refactoring, so that we always go through the index.ts at the root of apps?

src/services/tracing.ts
```diff
@@ -171,6 +173,94 @@ export const asyncRunInSpan = <F extends () => ReturnType<F>>(
...
+    return ret
+  }
+}
+
+type PromiseReturnType<T> = T extends Promise<infer Return> ? Return : T
+
+export const wrapAsyncToRunInSpan = <A extends Array<unknown>, R>({
+  fn,
+  namespace,
+  spanAttributes = {},
```
nicolasburtey, line 237: this is never set in the auto instrumentation?

src/services/tracing.ts
```diff
@@ -171,6 +173,94 @@ export const asyncRunInSpan = <F extends () => ReturnType<F>>(
...
+}: {
+  fn: (...args: A) => Promise<PromiseReturnType<R>>
+  namespace: string
+  spanAttributes?: SpanAttributes
```
nicolasburtey, line 241: we don't need a `?` if we set a default (`{}`) when there are no argument given

nicolasburtey (commented):
PR looks good on my end. just have some questions about the way the span attributes  are working

src/services/tracing.ts
dolcalmi: for now I removed the param.. we can add it later if we need it https://github.com/GaloyMoney/galoy/pull/837/commits/6c2545bd9bf949c6db30ea9355816b37cd355b8a

src/services/tracing.ts
dolcalmi: I removed the param https://github.com/GaloyMoney/galoy/pull/837/commits/6c2545bd9bf949c6db30ea9355816b37cd355b8a

src/app/index.ts
dolcalmi: I didnt want to change the implementations but code is more cleaner. Updated here https://github.com/GaloyMoney/galoy/pull/837/commits/6c2545bd9bf949c6db30ea9355816b37cd355b8a

src/app/wallets/get-lightning-fee.ts
dolcalmi: I tried to make it consistent:
- If the method is required outside the `app` (like in servers or tests) then `from "@app"` should be used
- If the method is required in a different submodule (inside `app`) then you can use  `from "@app/wallets"`
- If the method is required in the same submodule then we should use relative imports (this is to avoid multiple loads of the same resources, circular references, ...)

src/services/tracing.ts
```diff
@@ -171,6 +173,90 @@ export const asyncRunInSpan = <F extends () => ReturnType<F>>(
...
+    const params =
+      typeof functionArgs[0] === "object" ? functionArgs[0] : { "0": functionArgs[0] }
+    for (const key in params) {
+      attributes[`${SemanticAttributes.CODE_FUNCTION}.params.${key}`] = params[key]
+    }
+  }
+  return { attributes }
+}
+
+export const wrapToRunInSpan = <A extends Array<unknown>, R>({
```
nicolasburtey, line 202: still used?

src/services/tracing.ts
```diff
@@ -171,6 +173,90 @@ export const asyncRunInSpan = <F extends () => ReturnType<F>>(
...
+  const attributes = {
+    [SemanticAttributes.CODE_FUNCTION]: functionName,
+    [SemanticAttributes.CODE_NAMESPACE]: namespace,
+    ...spanAttributes,
+  }
+  if (functionArgs && functionArgs.length > 0) {
+    const params =
+      typeof functionArgs[0] === "object" ? functionArgs[0] : { "0": functionArgs[0] }
+    for (const key in params) {
+      attributes[`${SemanticAttributes.CODE_FUNCTION}.params.${key}`] = params[key]
```
nicolasburtey, line 196: we'll have to make sure that code and 2fa are redacted also on this

nicolasburtey (dismissed):

nicolasburtey (approved):

bodymindarts (approved):

PR #838, opened 2021-12-22
refactor: move ip lookup to domain

nicolasburtey:


src/domain/users-ips/index.ts
```diff
@@ -0,0 +1,17 @@
+export const parseIps = (
+  headers: null | NodeJS.Dict<string | string[]>,
+): IpAddress | null => {
+  if (!headers) return null
```
vindard, line 4: Using `undefined` now to represent absence

vindard: Setting as `null` here also breaks the checks at these points: https://github.com/GaloyMoney/galoy/search?q=%22ip+%3D%3D%3D+undefined%22&type=code

Sidenote: are there any ideas for having those checks not break when testing locally?

src/servers/graphql-server.ts
```diff
@@ -66,21 +68,8 @@ export const isEditor = rule({ cache: "contextual" })((parent, args, ctx) => {
 const geeTestConfig = getGeetestConfig()
 const geetest = Geetest(geeTestConfig)
 
-const sessionContext = ({
-  token,
-  ips,
-  body,
-  apiKey,
-  apiSecret,
-}): Promise<GraphQLContext> => {
```
vindard, line 75: Why remove the `Promise<GraphQLContext>` type here?

src/services/mongoose/users-ips.ts
```diff
@@ -57,6 +57,6 @@ export const UsersIpRepository = (): IUsersIPsRepository => {
 const userIPsFromRaw = (result: UserIPsType): UserIPs => {
   return {
     id: result.id as UserId,
-    lastIPs: (result.lastIPs || []) as IPType[],
+    lastIPs: result.lastIPs as IPType[],
```
vindard, line 60: This cast breaks the `lastIps: IPType` contract defined in `UserIPs` type since `result.lastIPs` can also be `undefined`

vindard (dismissed):

src/servers/graphql-server.ts
daviroo: @nicolasburtey was there a reason you removed this?

src/services/mongoose/users-ips.ts
daviroo: I've added undefined to the `UserIPS` type.

src/domain/users-ips/index.ts
daviroo: Changed to `undefined` to represent absence.

src/domain/users-ips/index.ts
```diff
@@ -0,0 +1,20 @@
+import { ExpressContext } from "apollo-server-express"
+
+export const parseIps = (context: undefined | ExpressContext): IpAddress | undefined => {
+  if (!context) return undefined
+
+  const ips =
```
daviroo, line 6: This is the fix for allowing it to work in local dev.

src/servers/graphql-server.ts
nicolasburtey: probably no good reasons

src/services/mongoose/users-ips.ts
nicolasburtey: I don't think it should every be undefined? https://github.com/GaloyMoney/galoy/blob/5ee38ade1ab2c7b87cb6b3c66e78a03f6abc0b07/src/services/mongoose/schema.ts#L152

src/domain/users-ips/index.ts
```diff
@@ -1,11 +1,14 @@
-export const parseIps = (
-  headers: null | NodeJS.Dict<string | string[]>,
-): IpAddress | null => {
-  if (!headers) return null
+import { ExpressContext } from "apollo-server-express"
```
nicolasburtey, line 1: the domain should not depend on apollo-server-express. the logic around `    process.env.NODE_ENV === "development"
      ? context.req?.ip
      : context.req?.headers["x-real-ip"]` should be in the server I think.
      
      @bodymindarts ?

daviroo: @bodymindarts

src/services/mongoose/users-ips.ts
daviroo: You're right it's the `ip` in the `IpType` which could be undefined - not the `IpType` itself.

src/domain/users-ips/index.ts
bodymindarts: Thats right - domain is purely functional without external dependencies

src/app/users/get-user.ts
```diff
@@ -69,20 +65,20 @@ const updateUserIPsInfo = async ({
         return
       }
 
-      const lastIP = userIP.lastIPs.find((ipObject) => ipObject.ip === ip)
+      const lastIP = userIP.lastIPs?.find((ipObject) => ipObject.ip === ip)
```
vindard, line 68: This change probably not necessary if `lastIPs` is never undefined
```suggestion
      const lastIP = userIP.lastIPs.find((ipObject) => ipObject.ip === ip)
```

src/app/users/get-user.ts
```diff
@@ -91,7 +87,7 @@ const updateUserIPsInfo = async ({
 
           ipInfo = { ...ipInfo, ...ipFetcherInfo }
         }
-        userIP.lastIPs.push(ipInfo)
+        userIP.lastIPs?.push(ipInfo)
```
vindard, line 90: Same here

daviroo: Fixed 👍

src/app/users/get-user.ts
daviroo: Fixed 👍

src/servers/graphql-server.ts
daviroo: I've added it back in.

src/domain/users-ips/index.ts
daviroo: @nicolasburtey I've moved the dependency out of the domain and back to the server.

bodymindarts (approved):

PR #839, opened 2021-12-22
chore: removing moment

nicolasburtey:


src/domain/bitcoin/lightning/invoice-expiration.ts
```diff
@@ -1,16 +1,15 @@
-import moment from "moment"
+import { toSeconds } from "@domain/primitives"
 
 const DEFAULT_EXPIRATIONS = {
-  BTC: { value: 1, unit: "days" as moment.unitOfTime.DurationConstructor },
-  USD: { value: 2, unit: "minutes" as moment.unitOfTime.DurationConstructor },
+  BTC: { delay: toSeconds(60 * 60 * 24) },
+  USD: { delay: toSeconds(2 * 60) },
```
vindard, line 5: We have a section with constants [here](https://github.com/GaloyMoney/galoy/blob/182fe1f816c738f222d8a5ae8864931a2ad9c2bc/src/config/app.ts#L35) in `app.ts` that we can probably add some of these values to
```suggestion
  BTC: { delay: toSeconds(SECS_PER_DAY) },
  USD: { delay: toSeconds(SECS_PER_2_MINS) },
```

vindard (commented):
Just one comment about constants.

Also do we need to remove `moment` from [here](https://github.com/GaloyMoney/galoy/blob/5632db972f88db8d9bbd7c74a364b582274d9671/src/services/price/update-price-history.ts#L45) as well?

nicolasburtey:
> Just one comment about constants.
> 
> Also do we need to remove `moment` from [here](https://github.com/GaloyMoney/galoy/blob/5632db972f88db8d9bbd7c74a364b582274d9671/src/services/price/update-price-history.ts#L45) as well?

oups. good catch. the ts-ignore made typescript ignoring the issue it seems.

vindard (dismissed):

src/services/price/update-price-history.ts
```diff
@@ -41,10 +40,8 @@ export const updatePriceHistory = async (
 
   // skip if it has not been an hour since last update
   try {
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore-error: TODO
-    const diff = moment().diff(moment(last(doc.pair.exchange.price)._id))
-    if (diff < 1000 * 60 * 60) {
+    const lastUpdate: number | undefined = last(doc.pair.exchange.price._id)
+    if (!!lastUpdate && Date.now() - lastUpdate < 1000 * 60 * 60) {
```
vindard, line 44: Just one small optional nit here

```suggestion
    if (!!lastUpdate && Date.now() - lastUpdate < MS_PER_HOUR) {
```

PR #844, opened 2021-12-22
chore: define GraphQLContext and use it for the GraphQL User type

samerbuna:
@nicolasburtey we can define the context for all resolvers this way

### DIFF
--- a/src/graphql/types/object/graphql-user.ts
+++ b/src/graphql/types/object/graphql-user.ts
@@ -15,7 +15,7 @@ import Username from "../scalar/username"
 import * as Users from "@app/users"
 import { UnknownClientError } from "@core/error"
 
-const GraphQLUser = new GT.Object({
+const GraphQLUser = new GT.Object<User, GraphQLContext>({
   name: "User",
   fields: () => ({
     id: {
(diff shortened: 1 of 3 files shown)


nicolasburtey (dismissed):

nicolasburtey:
> @nicolasburtey we can define the context for all resolvers this way

This will be very helpful

src/graphql/types/object/graphql-user.ts
```diff
@@ -15,7 +15,7 @@ import Username from "../scalar/username"
 import * as Users from "@app/users"
 import { UnknownClientError } from "@core/error"
 
-const GraphQLUser = new GT.Object({
+const GraphQLUser = new GT.Object<User, GraphQLContext>({
```
nicolasburtey, line 18: @samerbuna what should we do got `GT.Field`?

vindard (approved):

src/servers/index.files.d.ts
```diff
@@ -0,0 +1,11 @@
+type GraphQLContext = {
+  logger: Logger
+  uid: UserId | null
+  wallet: Wallet | null
+  domainUser: User | null
+  domainAccount: Account | null
+  user: User | null
+  geetest: GeetestType
+  account: Account | null
+  ip: string | undefined
```
vindard, line 10: One nit, we can probably make an `Ip` type for here

nicolasburtey: I think this should be `IpAddress` IIRC

PR #845, opened 2021-12-23
chore: deprecated for username on user

nicolasburtey:


dolcalmi (approved):

vindard (commented):
Do we need to also expose the alternative `Handle` field for folks to switch to now, or is it ok to deprecate without the alt for now?

nicolasburtey:
> Do we need to also expose the alternative `Handle` field for folks to switch to now, or is it ok to deprecate without the alt for now?

I think it's ok to deprecate for now: we don't have a lot of API user so at this stage this is more a self documentation for us than something we're communicating to external developers. Eventually when we deprecated API in the future it might make sense to already have an alternative.

I was also trying to understand how to expose an property in our graphql is deprecated because we never have done so in the past.

PR #848, opened 2021-12-23
feat(lnpayments): add new lnpayments collection to mongoDb

vindard:
## Description

This PR adds a new collection for persisting payment data from our `lnd` node across into our database. We will be doing this to:
- make the data available for data collection/analysis
- remove the data from `lnd` nodes to improve how they run

The initial approach taken would be to add a method to our cron runner to periodically query `lnd` for payment transactions that have not been persisted as yet and to save these. The payments would then be deleted from `lnd` after a certain fixed period.

### `LnPaymentLookup` type and schema

This new type was structured to accommodate the different confirmed/pending/failed objects that come back from the `getPayment` library method, and the objects (slightly different) that are returned from the `getPayments` and `getFailedPayments` methods.

The idea here was to keep the properties that apply to all 3 statuses at the top level of the object and to include two sub-objects:
- `confirmedDetails`: a domain-typed object with properties that are only returned/populated for non-failed payments
- `attempts`: a generic library-typed object with details failed attempts and multiple paths in the case of MPP payments

### Considerations

There are two additional elements we've added on top of simply persisting the data as-is, to help facilitate this process:

1. **Write twice to the collection**
Ideally we would only write to the collection once from the cron job for this task. There is a problem though where the `request` property for payments is often not returned in queries, and this property carries useful information about payments. To get around this we instead create partial payment objects asynchronously as part of our regular payment flow where we have access to that property, and then the rest of the object is populated from the cron job later on.

2. **Add additional `isCompleteRecord` property to objects**
Since we do not delete payments from `lnd` at the time of persistence, knowing whether an object is fully persisted isn't as simple as checking that it no longer exists in `lnd`. To get around this challenge, I've added an additional `isCompleteRecord` property to our db payment objects that would allow us to quickly query the payments that are still incomplete and to compare them against the full set of payments from `lnd` to be completed.

## Future tasks
The following related tasks are still outstanding for later PRs

### Handle historical `lnd` data

This code is designed to start persisting from our payment flow and to then complete persistence of already-started records via our cron job. 

We would need to still capture data that has already been captured in the past and this can be done via a separate migration script that will query `lnd` and populate all old payment data into our new db collection.

_This is planned for a separate PR._

### Delete persisted payments from `lnd`

This will be a task that can query `lnd` data past a certain date, check that it's already been persisted, and then delete the payment data from `lnd`.

This will most likely be a separate task from the current persistence task given the age requirement we'd like to include.

## How to test
- run `TEST="01|02" make reset-integration` to populate the new `lnpayments` collection with partial objects

- run cronjob to complete the db objects from `lnd`
  ```bash
  . ./.envrc && \
		yarn build && \
		LOGLEVEL=error node lib/servers/cron.js
  ```

**Some sample json objects are included in [this comment](https://github.com/GaloyMoney/galoy/pull/848#issuecomment-1010052974) lower down.**

## Pending for this PR
- [x] ~finish manually testing the async trigger from payment from, and running the task from the cronjob~
- [x] ~separate `PaymentSecret` type into [the invoice secret (`s`) and the preimage for the payment hash (`p`)](https://github.com/lightning/bolts/blob/master/11-payment-encoding.md#tagged-fields)~ (moved to #906)
- [x] ~manually test flow for a pending payment that eventually settles/fails~ (tested by using our hodl invoice test)
- [x] ~add appropriate integration tests~
- [x] ~add handling for querying multiple lnds~
- [x] ~after #906, rework LnInvoiceLookup to separate LnInvoice from lookup metadata, potentially apply to LnPaymentLookup too~
- [x] ~double-check token/paging functionality (see [here](https://github.com/GaloyMoney/galoy/pull/848#issuecomment-1015058426))~ done [here](https://github.com/GaloyMoney/galoy/pull/848#issuecomment-1015470981)


src/app/lightning/update-ln-payments.ts
```diff
@@ -0,0 +1,90 @@
...
+    after,
+    pubkey,
+  }: ListLnPaymentsArgs) => Promise<ListLnPaymentsResult | LightningServiceError>
+}) => {
+  const processedLnPaymentsHashes: PaymentHash[] | ApplicationError = []
+  const incompleteLnPaymentsHashes = incompleteLnPayments.map((p) => p.paymentHash)
+
+  const lndService = LndService()
+  if (lndService instanceof Error) return lndService
+  const pubkeys = lndService.listPubkeys()
```
vindard, line 39: Should we be listing only active lnds here, or is it ok to include all lnds?

vindard:
## Data layout

This is the sample data being captured in the new `lnpayments` collection. The following files are contained in [this zip file](https://github.com/GaloyMoney/galoy/files/7847572/demo-data.zip).


- `demo-success-pre.json`
This is the prelim object that gets added to the collection on payment during our payment flow. It gets added via a fire-and-forget async call so as to not impact performance of the pay flow.

- `demo-success-post.json`
This is the complete object that get populated during our cron job.

- `demo-success-fromlnd.json`
This is the raw object that we get back from lnd. This is the thing we are translating and persisting

- demo-failed-post.json`
A sample object that's persisted for failed payments.

src/app/lightning/update-ln-payments.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export const updateLnPayments = async (): Promise<true | ApplicationError> => {
+  let incompleteLnPayments = await LnPaymentsRepository().listIncomplete()
+  if (incompleteLnPayments instanceof Error) return incompleteLnPayments
+
+  const lndService = LndService()
+  if (lndService instanceof Error) return lndService
+
+  const lndListMethods = [lndService.listSettledPayments, lndService.listFailedPayments]
+  for (const listPaymentsFn of lndListMethods) {
```
nicolasburtey, line 13: shouldn't this be a `for await`? otherwise the loop will wait for the internal await?

src/app/lightning/update-ln-payments.ts
```diff
@@ -0,0 +1,93 @@
...
+    incompleteLnPayments = await fetchAndUpdatePayments({
+      incompleteLnPayments,
+      listPaymentsFn,
+    })
+    if (incompleteLnPayments instanceof Error) return incompleteLnPayments
+    if (incompleteLnPayments.length == 0) return true
```
nicolasburtey, line 19: should we return here? this is within a loop and we're going there twice, and shouldn't return if the other loop has not completed? because we may have to return an error but the first iteration of the loop may already have returned

src/app/lightning/update-ln-payments.ts
nicolasburtey: if an lnd is offline we won't be able to do much with it

src/app/lightning/update-ln-payments.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  const lndService = LndService()
+  if (lndService instanceof Error) return lndService
+  const pubkeys = lndService.listPubkeys()
+
+  let lnPayments: LnPaymentLookup[]
+  const endCursors: { [key: Pubkey]: PagingToken | false | undefined } = {}
+  for (const key of pubkeys) endCursors[key] = undefined
+
+  // Breadth-first paginated fetch of payments across multiple lightning instances
```
nicolasburtey, line 45: this is the type of comment that make me think we have some important logic here, and maybe `app` is not the right directory? may want to move this either to `domain` or `service` and maybe see if we can test it specifically?

src/domain/bitcoin/lightning/index.types.d.ts
```diff
@@ -12,9 +12,14 @@ type PaymentIdentifyingSecret = string & { readonly brand: unique symbol }
 type FeatureBit = number & { readonly brand: unique symbol }
 type FeatureType = string & { readonly brand: unique symbol }
 
+declare const pagingTokenSymbol: unique symbol
+type PagingToken = string & { [pagingTokenSymbol]: never }
```
nicolasburtey, line 16:
```suggestion
type PagingToken = string & { readonly brand: unique symbol }
```

src/domain/bitcoin/lightning/index.types.d.ts
```diff
@@ -133,7 +173,17 @@ interface ILightningService {
   }: {
     pubkey?: Pubkey
     paymentHash: PaymentHash
-  }): Promise<LnPaymentLookup | LightningServiceError>
+  }): Promise<LnPaymentLookup | LnFailedPartialPaymentLookup | LightningServiceError>
+
+  listSettledPayments({
```
nicolasburtey, line 178: I'm wondering if writing this as: ` listSettledPayments(args: ListLnPaymentsArgs): Promise<ListLnPaymentsResult | LightningServiceError>` would not be equivalent, and less verbose?

nicolasburtey: basically not sure the destructuring is very useful for the .types.d.ts. (have not tested whether it's changing something)

src/servers/cron.ts
```diff
@@ -40,6 +41,11 @@ const main = async () => {
       const updatePendingOnChainPayments = () =>
         updateUsersPendingPayment({ onchainOnly: true })
 
+      const updateLnPaymentsCollection = async () => {
+        const result = await updateLnPayments()
+        if (result instanceof Error) throw result
```
nicolasburtey, line 46: not sure we want to throw? it may stop the other promise to complete?

I think you can just pass `updateLnPayments` to the array of tasks below instead?

src/services/lnd/index.ts
```diff
@@ -50,21 +53,13 @@ export const LndService = (): ILightningService | LightningServiceError => {
...
-      switch (errDetails) {
-        default:
-          return new UnknownLightningServiceError(err)
-      }
-    }
-    return !!offchainLnds
+  const isLocal = (pubkey: Pubkey): boolean | LightningServiceError =>
+    !!getLnds({ type: "offchain" })
       .map((item) => item.pubkey as Pubkey)
       .find((item) => item == pubkey)
```
nicolasburtey, line 59:
```suggestion
      .some((item) => item == pubkey)
```

(haven't tested it)

src/services/lnd/index.ts
```diff
@@ -50,21 +53,13 @@ export const LndService = (): ILightningService | LightningServiceError => {
...
-    } catch (err) {
-      const errDetails = parseLndErrorDetails(err)
-      switch (errDetails) {
-        default:
-          return new UnknownLightningServiceError(err)
-      }
-    }
-    return !!offchainLnds
+  const isLocal = (pubkey: Pubkey): boolean | LightningServiceError =>
+    !!getLnds({ type: "offchain" })
```
nicolasburtey, line 57: we could use `.some()` instead of `!!`

src/services/lnd/index.ts
nicolasburtey, line 59: `some` would replace `find` here and return a boolean

src/services/lnd/index.ts
```diff
@@ -256,6 +256,52 @@ export const LndService = (): ILightningService | LightningServiceError => {
...
+      const { payments, next } = await getFailedPayments({ lnd, ...pagingArgs })
+
+      return {
+        lnPayments: payments
+          .map(translateLnPaymentLookup)
+          .map((p) => ({ ...p, status: PaymentStatus.Failed })),
+        endCursor: (next as PagingToken) || false,
+      }
+    } catch (err) {
+      return new UnknownLightningServiceError(err)
```
nicolasburtey, line 278: maybe the `try catch` could/should be just around `getFailedPayments`?

src/services/lnd/schema.ts
```diff
@@ -0,0 +1,59 @@
...
+  },
+  paymentHash: {
+    type: String,
+    index: true,
+    unique: true,
+    required: true,
+  },
+  paymentRequest: {
+    type: String,
+    unique: true,
```
nicolasburtey, line 46: do we need an index on this, if we already have an index in the hash?

the index on the hash will it so that it's not possible to add the same paymentRequest twice

src/app/lightning/update-ln-payments.ts
vindard: Not sure, I assumed `for await ... of` only worked for async iterables and in this case I think `lndListMethods` is a sync iterable containing async items.

Maybe @dolcalmi / @samerbuna  / @daviroo knows as well. Would be great if we could get all the iterations to trigger simultaneously, but still resolve their internal `await fetchAndUpdatePayments` so we can check for errors.

src/app/lightning/update-ln-payments.ts
vindard: The `fetchAndUpdatePayments` takes an initial `incompleteLnPayments` array and returns a pared down version of `incompleteLnPayments` so this return is for if there are no more incomplete payments from our `lnPayments` collection to be updated. In that case we don't need to go to `lnd` for anything anymore regardless of the method we're currently in.

src/servers/cron.ts
vindard: Sure, I was following the precedent from [above](https://github.com/GaloyMoney/galoy/blob/8f7c0f0bb2f39ec955a6c7e37139f9c045857509/src/servers/cron.ts#L50).

@dolcalmi was there a reason why you used `throw` on Ln 51, or can we remove both of these?

src/services/mongoose/ln-payments.ts
```diff
@@ -0,0 +1,103 @@
...
+        revealedPreImage: result.confirmedDetails.revealedPreImage,
+        roundedUpFee: result.confirmedDetails.roundedUpFee,
+        milliSatsFee: result.confirmedDetails.milliSatsFee,
+        // cast away from CoreMongooseArray type
+        hopPubkeys: result.confirmedDetails.hopPubkeys
+          ? [...result.confirmedDetails.hopPubkeys]
+          : [],
+      }
+    : undefined,
+  // cast away from CoreMongooseArray type
```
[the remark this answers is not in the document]
vindard: Needed for the `toStrictEqual` comparison in the tests [here](https://github.com/GaloyMoney/galoy/pull/848/files#diff-15e9812156b4ff1c33faaaf99ffcc368a8af2756ccb7a5e81acdcfc46782db9fR840) and [here](https://github.com/GaloyMoney/galoy/pull/848/files#diff-15e9812156b4ff1c33faaaf99ffcc368a8af2756ccb7a5e81acdcfc46782db9fR856)

src/servers/cron.ts
dolcalmi: check line 82-83, we need to throw here because we need to catch and log the error, this will also affect exit code and will trigger an alert in slack (if we dont throw then we need to add additional logic to register the error and propagate the error)

src/app/lightning/update-ln-payments.ts
vindard: I was just looking some more into how this works ([this was helpful](https://stackoverflow.com/a/59695815)).

If we wanted to not wait for the internal `await` we could wrap the entire internal `for` loop logic in a promise and then fire these off using `Promise.all`. That would break the intent of this for loop though, since we are doing a synchronous update of the `incompletePayments` array to remove processed elements.

I'm not sure if there's a way to work on the `incompletePayments` object asynchronously and whether this optimization would be worth the added complexity at this point (premature optimization?)

dolcalmi: I think `for await (const res of items.map(e => somethingAsync(e))) ` method should work having `somethingAsync` as a function that wraps line 14 to 19. 

Also 
- `update-ln-payments.ts` is under lightning folder, `-ln-` is not necessary
- Please use `import { Lightning } from "@app"` and `Lightning.updatePayments` (to use wrapped functions in the cron)

src/app/lightning/update-ln-payments.ts
vindard: I'm not sure we can move this out of the app layer because it involves the interplay between two things from our domain: `LndService` & `LnPaymentsRepository`.

A bit of background, the reason this is "breadth-first" has to do with that we will accumulate payments in lnd for some period (maybe 2 weeks) before deleting, but the cronjob should only ever need to look through about 1 day of data to get all the updates it needs. This loop searches all lnd instances in `TRANSACTION_PER_PAGE` size chunks, instead of searching the entire history of a given lnd before moving on to another. The loop automatically stops querying lnd once all incomplete payments are processed.

src/servers/cron.ts
vindard: Ok so other promises are unaffected then? (just confirming)

src/app/lightning/update-ln-payments.ts
vindard: Ok actually, I think I have an idea for how to move this into the service layer

vindard:
Converted to _**Draft**_ to work on abstracting the payments update loop into separate Lightning service 'listPaymentsPaginated' calls breadth-first across multiple lightning instances.

vindard:
### TODO: Double-check `lnd` payments pagination

On preliminary tests I would've expected back a string for the `next` property, but am getting back an object that looks like: `'{"offset":1,"limit":250}'`

src/services/mongoose/ln-payments.ts
```diff
@@ -0,0 +1,103 @@
...
+      const result = await LnPayment.find({
+        isCompleteRecord: false,
+      })
+      return result.map(lnPaymentFromRaw)
+    } catch (err) {
+      return new UnknownRepositoryError(err)
+    }
+  }
+
+  const persist = async (
```
nicolasburtey, line 36: we use `persistNew` on other repositories

src/app/lightning/update-ln-payments.ts
```diff
@@ -0,0 +1,110 @@
...
+  incompleteLnPayments: PersistedLnPaymentLookup[]
+  lastSeenCursorByPubkey: ListSettledAndFailedLnPaymentsByPubkeyArgs
+  lndService: ILightningService
+}) => {
+  // Fetch from Lightning service
+  const results = await lndService.listSettledAndFailedPaymentsByPubkey(
+    lastSeenCursorByPubkey,
+  )
+
+  // Update cursors & drop completed pubkeys
```
nicolasburtey, line 52: I feel the logic of wrapping around pubkey to be convoluted:
- if we get multiple node to fetch over, it will be just few of them, maybe 2 or 3.
- I would favor simplifying the logic instead of trying to parallelize this code (which I think this is what we're trying to achieve here but maybe I'm mistaken): 

this is a cronjob so it's not critical to get high throughput, and iterating slowly over the payment is actually beneficial to not overload the different lnds.

so maybe a more simple approach in the like of

```
for await (const pubkey of pubkeys) {
 fetchAndUpdatePayments({pubkey, ...})
}
```

might make it easier to understand what is going on.

not a strong opinion on this as I'm trying to understand the PR still, but this is my first reaction.

src/services/lnd/index.ts
```diff
@@ -256,6 +254,81 @@ export const LndService = (): ILightningService | LightningServiceError => {
...
+          .map(translateLnPaymentLookup)
+          .filter((p) => p.status === PaymentStatus.Settled),
+        endCursor: (next as PagingStartToken) || false,
+      }
+    } catch (err) {
+      return new UnknownLightningServiceError(err)
+    }
+  }
+
+  const listSettledAndFailedPaymentsByPubkey = async (
```
nicolasburtey, line 303: the other method also take a pubkey not sure we need to add this in the name of the function

src/services/lnd/index.ts
```diff
@@ -256,6 +254,81 @@ export const LndService = (): ILightningService | LightningServiceError => {
...
+    cursors: ListSettledAndFailedLnPaymentsByPubkeyArgs,
+  ): Promise<ListSettledAndFailedLnPaymentsByPubkeyResult> => {
+    const nullListLnPaymentsResult: ListLnPaymentsResult = {
+      lnPayments: [],
+      endCursor: false,
+    }
+
+    return Promise.all(
+      cursors.map(async ({ pubkey, settledAfter, failedAfter }) => ({
```
nicolasburtey, line 312: what feels a convoluted logic mention [here](https://github.com/GaloyMoney/galoy/pull/848/files#r786639296) is continuing here. there is no reason to apply the parallelism of this method here at the service level.

src/services/lnd/schema.ts
```diff
@@ -0,0 +1,58 @@
...
+  hopPubkeys: [String],
+})
+
+const paymentAttemptSchema = Schema.Types.Mixed
+
+const paymentSchema = new Schema<LnPaymentType>({
+  createdAt: Date,
+  status: {
+    type: String,
+    enum: ["settled", "failed", "pending"],
```
nicolasburtey, line 35: could be using a type instead? and like for `ledgerTransactionTypes` in ledger/schema.ts

src/services/lnd/schema.ts
```diff
@@ -0,0 +1,58 @@
...
+    type: Number,
+    required: true,
+  },
+  hopPubkeys: [String],
+})
+
+const paymentAttemptSchema = Schema.Types.Mixed
+
+const paymentSchema = new Schema<LnPaymentType>({
+  createdAt: Date,
```
nicolasburtey, line 32: needs to set a default value here. this is not been populated from the demo-success-pre.json from your example

src/servers/cron.ts
```diff
@@ -42,6 +43,11 @@ const main = async () => {
     await deleteExpiredInvoiceUser()
   }
 
+  const updateLnPaymentsCollection = async () => {
+    const result = await updateLnPayments()
+    if (result instanceof Error) throw result
```
nicolasburtey, line 48: should it be a throw or return here @dolcalmi ?

vindard:
> ### TODO: Double-check `lnd` payments pagination
> 
> On preliminary tests I would've expected back a string for the `next` property, but am getting back an object that looks like: `'{"offset":1,"limit":250}'`


This is a sample log from how the method iterates through two lnds, one lnd with 0 payments, and the other with 524 (511 settled, 10 pending, 3 failed)

```
lastSeenCursorByPubkey start: [
  {
    settledAfter: undefined,
    failedAfter: undefined,
    pubkey: '03f7611a886ae26284d0c25726163ab7a37de2694be5b1afbe456d9d9b039a1dbb'
  },
  {
    settledAfter: undefined,
    failedAfter: undefined,
    pubkey: '03defd0750d155c4eab38b9c9987e62c76b766cd9dc49ba9b5ed0045fed6661876'
  }
]

==============================
START-WHILE

Settled pubkey: 03f7611a886ae26284d0c25726163ab7a37de2694be5b1afbe456d9d9b039a1dbb
Settled request token (03f76...): undefined

Settled pubkey: 03defd0750d155c4eab38b9c9987e62c76b766cd9dc49ba9b5ed0045fed6661876
Settled request token (03def...): undefined

Settled token returned (03def...): undefined
Settled payments # (03def...): 0

Failed pubkey: 03defd0750d155c4eab38b9c9987e62c76b766cd9dc49ba9b5ed0045fed6661876
Failed request token (03def...): undefined

Failed token returned (03def...): undefined
Failed payments # (03def...): 0

Settled token returned (03f76...): {"offset":524,"limit":250}
Settled payments # (03f76...): 250

Failed pubkey: 03f7611a886ae26284d0c25726163ab7a37de2694be5b1afbe456d9d9b039a1dbb
Failed request token (03f76...): undefined

Failed token returned (03f76...): {"offset":524,"limit":250}
Failed payments # (03f76...): 0
lastSeenCursorByPubkey continue: [
  {
    settledAfter: '{"offset":524,"limit":250}',
    failedAfter: '{"offset":524,"limit":250}',
    pubkey: '03f7611a886ae26284d0c25726163ab7a37de2694be5b1afbe456d9d9b039a1dbb'
  }
]

END-WHILE
==============================
START-WHILE

Settled pubkey: 03f7611a886ae26284d0c25726163ab7a37de2694be5b1afbe456d9d9b039a1dbb
Settled request token (03f76...): {"offset":524,"limit":250}

Settled token returned (03f76...): {"offset":274,"limit":250}
Settled payments # (03f76...): 250

Failed pubkey: 03f7611a886ae26284d0c25726163ab7a37de2694be5b1afbe456d9d9b039a1dbb
Failed request token (03f76...): {"offset":524,"limit":250}

Failed token returned (03f76...): {"offset":274,"limit":250}
Failed payments # (03f76...): 0
lastSeenCursorByPubkey continue: [
  {
    settledAfter: '{"offset":274,"limit":250}',
    failedAfter: '{"offset":274,"limit":250}',
    pubkey: '03f7611a886ae26284d0c25726163ab7a37de2694be5b1afbe456d9d9b039a1dbb'
  }
]

END-WHILE
==============================
START-WHILE

Settled pubkey: 03f7611a886ae26284d0c25726163ab7a37de2694be5b1afbe456d9d9b039a1dbb
Settled request token (03f76...): {"offset":274,"limit":250}

Settled token returned (03f76...): {"offset":24,"limit":250}
Settled payments # (03f76...): 250

Failed pubkey: 03f7611a886ae26284d0c25726163ab7a37de2694be5b1afbe456d9d9b039a1dbb
Failed request token (03f76...): {"offset":274,"limit":250}

Failed token returned (03f76...): {"offset":24,"limit":250}
Failed payments # (03f76...): 0
lastSeenCursorByPubkey continue: [
  {
    settledAfter: '{"offset":24,"limit":250}',
    failedAfter: '{"offset":24,"limit":250}',
    pubkey: '03f7611a886ae26284d0c25726163ab7a37de2694be5b1afbe456d9d9b039a1dbb'
  }
]

END-WHILE
==============================
START-WHILE

Settled pubkey: 03f7611a886ae26284d0c25726163ab7a37de2694be5b1afbe456d9d9b039a1dbb
Settled request token (03f76...): {"offset":24,"limit":250}

Settled token returned (03f76...): undefined
Settled payments # (03f76...): 11

Failed pubkey: 03f7611a886ae26284d0c25726163ab7a37de2694be5b1afbe456d9d9b039a1dbb
Failed request token (03f76...): {"offset":24,"limit":250}

Failed token returned (03f76...): {"offset":1,"limit":250}
Failed payments # (03f76...): 3
lastSeenCursorByPubkey continue: [
  {
    settledAfter: false,
    failedAfter: '{"offset":1,"limit":250}',
    pubkey: '03f7611a886ae26284d0c25726163ab7a37de2694be5b1afbe456d9d9b039a1dbb'
  }
]

END-WHILE
==============================
START-WHILE

Failed pubkey: 03f7611a886ae26284d0c25726163ab7a37de2694be5b1afbe456d9d9b039a1dbb
Failed request token (03f76...): {"offset":1,"limit":250}

Failed token returned (03f76...): undefined
Failed payments # (03f76...): 0
lastSeenCursorByPubkey continue: []

END-WHILE
==============================
```

src/app/lightning/update-ln-payments.ts
vindard: The idea here isn't so much to parallelize as it is to enforce that we look across lnds before we go down each one (breadth-first search).

Say for example these are our 3 lnds and we have relevant payments in the boxes marked `x` that we need to update from. Each box represents one page in the paginated search.

| lnd1 | lnd2 | lnd3 |
|------|------|------|
|   x  |   x  |   x  |
|   -  |   -  |   -  |
|   x  |   -  |   -  |
|   -  |   -  |   -  |
|   -  |   -  |   -  |
|   -  |   -  |   -  |
|   -  |   -  |   -  |
|   -  |   -  |   -  |

If we were to do one lnd at a time, then we would have to query down the entire `lnd1` first, and then `lnd2`, and we won't complete until we get to the first page of the `lnd3`.

What I'm trying to do is query across all lnds first so that our queries would finish on the 3rd page of `lnd1` and we would only need 3 passes instead of potentially much more if we're deleting payments every 2 weeks but running the cron task each day (or multiple times per day).

The intention is to try to avoid a significant number of unnecessary queries to the lnds in an environment where we have multiple lnds.

src/services/lnd/schema.ts
vindard: I'm not sure if there's value in this here. The intention is that `created_at` gets set from the value in lnd, and before that (`isCompleteRecord: false`) these objects aren't meant to be used but are only a marker with the bare info needed to query lnd and complete when the cronjob is run.

There's other fields we can also infer from the paymentRequest that I deliberately left out as well. I really only included the `paymentHash` as an index and the `paymentRequest` because we don't always get this back from lnd.

src/servers/cron.ts
vindard: Relevant comment: https://github.com/GaloyMoney/galoy/pull/848#discussion_r784882767

src/app/lightning/update-ln-payments.ts
```diff
@@ -0,0 +1,110 @@
...
+      ...(failed instanceof Error ? [] : failed.lnPayments),
+    ]
+
+    for (const payment of lnPayments) {
+      let persistedPaymentLookup = incompleteLnPayments.find(
+        (elem) => elem.paymentHash === payment.paymentHash,
+      )
+      if (!persistedPaymentLookup) continue
+
+      persistedPaymentLookup = Object.assign(persistedPaymentLookup, {
```
nicolasburtey, line 87: check type

src/services/lnd/schema.ts
nicolasburtey: the value I see in having the createdAt set at this time, is for us to calculate how long a payment is taking to complete (time-to-payment-completion is a key metrics to follow).

maybe we're getting this from lnd?

src/app/lightning/update-ln-payments.ts
nicolasburtey: I understand the idea, I'm not sure this is the right tradeoff. 

doing it by lnd would be:

1-0, 1-1, 1-2, 2-0, 3-0

doing it with the cursors the way it's implemented is:

1-0, 2-0, 3-0, 1-1, 1-2

The reason why we would benefits from the added algorithm complexity is if this was on a critical path for latency for a user request. but because this run as a cronjob, there is no reason to optimize for latency (assumption is we want to optimize for the fastest number of payment in the shorter time and have the potential gap later in the process?).

I'm also not saying we have to rewrite it. but I just think as a result the code is not easy to understand.

nicolasburtey (approved):

src/services/lnd/schema.ts
vindard: Exactly, the property itself is definitely valuable. What I'm saying is that we always get it from lnd, so there probably isn't value in trying to pre-emptively set it ourselves to always have it overwritten.

src/app/lightning/update-ln-payments.ts
vindard: Agreed this makes the code more complex.

Maybe a good tradeoff could be to leave it like this for now until we implement the `deletePayments` logic, and then simplify at that point when we have much lighter lnd payment histories to parse.

src/app/lightning/update-ln-payments.ts
vindard: Fixed: https://github.com/GaloyMoney/galoy/pull/848/commits/8c3537c58a2ce60064a8ba6856742e9d0af1f0f5

src/services/lnd/index.ts
vindard: Can resolve in a follow-up PR: https://github.com/GaloyMoney/galoy/pull/848#discussion_r787658488

nicolasburtey (dismissed):

nicolasburtey (approved):

vindard:
I just double-checked and the grpc trace this new method produces in cron (`lookupPayment`) is different from the current problem one (`TrackPaymentV2`) so this is probably ok to merge now.

PR #850, opened 2021-12-24
fix: filter voided txs from txVolumeSince

dolcalmi:
Fix an issue reported by an user that hit the daily limit without successful transactions (verified with logs)

How to test:
- Run `TEST="01|02" make reset-integration`
- Query `walletId` for `user1`
- Run the next query:
```js
db.getCollection('medici_transactions').aggregate([{
  $match: {
    accounts: "Liabilities:<user1.walletId>",
    voided: false,
    _original_journal:{ $exists: false},
  },
},
{
  $group: {
    _id: null,
    outgoingSats: { $sum: "$debit" },
    incomingSats: { $sum: "$credit" },
  },
}])
```
- Run the same query without `voided` and `_original_journal` conditions

### DIFF
--- a/src/services/ledger/index.ts
+++ b/src/services/ledger/index.ts
@@ -227,6 +227,8 @@ export const LedgerService = (): ILedgerService => {
         {
           $match: {
             accounts: liabilitiesWalletId,
+            voided: false, // avoid to sum voided transactions to outgoingSats
+            _original_journal: { $exists: false }, // avoid to sum voided transactions to incomingSats
             $or: txnTypesObj,
             $and: [{ timestamp: { $gte: timestamp } }],
           },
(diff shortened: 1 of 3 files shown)


nicolasburtey:
side note something somewhat related, to the void transactions: 

I think we should refactor our code to remove the notion of pending transaction for sending payment in medici, and use another way to store pending payment (either redis or another mongodb collection) such that we don't have to revert/void those transactions when they failed.

the void transaction or partial refund are also a source of confusion to the mobile user. and will be a source of confusion for api user. if you want to know how much fees has been paid for a refunded transaction, then you have to do some groupby, which is not very convenient.

nicolasburtey:
@dolcalmi I understand the purpose of the PR, but at the same time, doesn't it add open a new vector where now it's easy (easier) to somewhat ddos the server with a lot of failed transaction?

nicolasburtey (changes requested):
I propose not merging this one for now (we can still keep the changes in test/helpers/wallet.ts)

PR #851, opened 2021-12-24
chore: simplifying domain type declaration

nicolasburtey:
reading a book on [typescript](https://www.amazon.com/Programming-TypeScript-Making-JavaScript-Applications/dp/1492037656) and this is what they suggest to get the domain type. feels a better option than what we're doing.

### DIFF
--- a/src/domain/bitcoin/lightning/index.types.d.ts
+++ b/src/domain/bitcoin/lightning/index.types.d.ts
@@ -3,23 +3,12 @@ type LnInvoiceDecodeError = import("./errors").LnInvoiceDecodeError
 type LightningServiceError = import("./errors").LightningServiceError
 type RouteNotFoundError = import("./errors").RouteNotFoundError
 
-declare const invoiceExpirationSymbol: unique symbol
-type InvoiceExpiration = Date & { [invoiceExpirationSymbol]: never }
-
-declare const encodedPaymentRequestSymbol: unique symbol
-type EncodedPaymentRequest = string & { [encodedPaymentRequestSymbol]: never }
-
-declare const paymentHashSymbol: unique symbol
-type PaymentHash = string & { [paymentHashSymbol]: never }
-
-declare const paymentSecretSymbol: unique symbol
-type PaymentSecret = string & { [paymentSecretSymbol]: never }
-
-declare const featureBitSymbol: unique symbol
-type FeatureBit = number & { [featureBitSymbol]: never }
-
-declare const featureTypeSymbol: unique symbol
-type FeatureType = string & { [featureTypeSymbol]: never }
+type InvoiceExpiration = Date & { readonly brand: unique symbol }
+type EncodedPaymentRequest = string & { readonly brand: unique symbol }
+type PaymentHash = string & { readonly brand: unique symbol }
+type PaymentSecret = string & { readonly brand: unique symbol }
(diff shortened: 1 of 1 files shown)


src/domain/bitcoin/lightning/index.types.d.ts
```diff
@@ -3,23 +3,12 @@ type LnInvoiceDecodeError = import("./errors").LnInvoiceDecodeError
...
-type FeatureBit = number & { [featureBitSymbol]: never }
-
-declare const featureTypeSymbol: unique symbol
-type FeatureType = string & { [featureTypeSymbol]: never }
+type InvoiceExpiration = Date & { readonly brand: unique symbol }
+type EncodedPaymentRequest = string & { readonly brand: unique symbol }
+type PaymentHash = string & { readonly brand: unique symbol }
+type PaymentSecret = string & { readonly brand: unique symbol }
+type FeatureBit = number & { readonly brand: unique symbol }
+type FeatureType = string & { readonly brand: unique symbol }
```
vindard, line 11: As far as I can tell it gets us what we want from the type-checker

image

```suggestion
type Unique = { readonly brand: unique symbol }

type InvoiceExpiration = Date & Unique
type EncodedPaymentRequest = string & Unique
type PaymentHash = string & Unique
type PaymentSecret = string & Unique
type FeatureBit = number & Unique
type FeatureType = string & Unique
```

vindard (dismissed):
One suggested optional improvement

src/domain/bitcoin/lightning/index.types.d.ts
vindard: If we like this suggestion, this find & replace should work from VSCode

image

nicolasburtey: you master better VSCode than I do :) do you want to run it then on a new PR? we can close this one afterwards

vindard: Sure can do

nicolasburtey: can you double check that with the new type `Unique` you created that 2 types of `string & Unique` are not interchangeables (as opposed to a `Date` and `Date & Unique`)

vindard: Ok I found an issue with this approach in general.

I can do `const userId: UserId = 'id' as WalletId` and this works because it's comparing `{readonly brand: unique symbol}` in both cases. This isn't a problem with our current approach because we also define unique symbol props instead of the generic `brand` prop

nicolasburtey: what would be the message if we were to do `const userId: UserId = 'id' as WalletId` with our approach?

vindard: No error, that's seen as valid code

nicolasburtey: I just did a test, the issue seems to be using `type Unique = { readonly brand: unique symbol }`. otherwise just doing:

```
type PaymentHash = string & { readonly brand: unique symbol }
type PaymentSecret = string & { readonly brand: unique symbol }
``` 

seems to work

nicolasburtey: image

image

PR #853, opened 2021-12-24
chore: improve boundary for sendBalanceToAccounts

nicolasburtey:


src/app/accounts/send-balance-to-users.ts
```diff
@@ -0,0 +1,25 @@
+import { getRecentlyActiveAccounts } from "@app/accounts/active-accounts"
+import { getCurrentPrice } from "@app/prices"
+import { Wallets } from "@app"
```
dolcalmi, line 3:
```
If the method is required outside the app (like in servers or tests) then from "@app" should be used
If the method is required in a different submodule (inside app) then you can use from "@app/wallets"
```
here you should use `"@app/wallets"`

src/servers/daily-balance-notification.ts
```diff
@@ -1,40 +1,13 @@
-import { getRecentlyActiveAccounts } from "@app/accounts/active-accounts"
-import { getCurrentPrice } from "@app/prices"
-import { getBalanceForWallet } from "@app/wallets"
+import { sendBalanceToAccounts } from "@app/accounts/send-balance-to-users"
```
dolcalmi, line 1: this should be called direclty from "@app" (this is a call in a server). you should add send-balance-to-users to account/index

dolcalmi (changes requested):

test/integration/notifications/notification.spec.ts
```diff
@@ -1,8 +1,9 @@
 import { Prices } from "@app"
 import { getRecentlyActiveAccounts } from "@app/accounts/active-accounts"
+import { sendBalanceToAccounts } from "@app/accounts/send-balance-to-accounts"
```
dolcalmi, line 3: is not mandatory but if we want to verify honeycomb with local data then this should be used through wrapped functions

src/app/accounts/send-balance-to-accounts.ts
```diff
@@ -0,0 +1,25 @@
+import { getRecentlyActiveAccounts } from "@app/accounts/active-accounts"
```
dolcalmi, line 1: not mandatory, but we should try to keep it consistent. Lets discuss tomorrow if we should use this way or relative for imports in the same sub-module

dolcalmi (approved):
looks good, just minor comments to discuss tomorrow (code review)

PR #856, opened 2021-12-26
chore: simplyfing brand type declaration

nicolasburtey:
more context can be read here: https://github.com/GaloyMoney/galoy/pull/851


dolcalmi (approved):
The only issue with this approach is that IDEs shows `brand` as a property (just to keep in mind)

PR #858, opened 2021-12-27
chore: force import order

nicolasburtey:
there is a frequent issue when doing rebase which is that the order of import changes.

this is easy to fix generally, but still require time.

this is the tutorial I have followed to get the vscode working, it is now doing both linting and prettier at the same time: https://medium.com/how-to-react/config-eslint-and-prettier-in-visual-studio-code-for-react-js-development-97bb2236b31a


dolcalmi (approved):
if we automate this we need to be careful with medici schema import

nicolasburtey:
> if we automate this we need to be careful with medici schema import

yup already took care of it: https://github.com/GaloyMoney/galoy/pull/858/commits/26e7f800c126a7bec043059fd5ab7e9ac5f6edda

nicolasburtey:
I'm just not sure how to enforce the rules correctly on every save from VSCode right now

vindard (approved):

package.json
```diff
@@ -4,8 +4,6 @@
     "tsc-check": "tsc --noEmit -p tsconfig.d.json && tsc --noEmit --skipLibCheck",
     "eslint-check": "eslint \"{src,test}/**/*.ts\"",
     "eslint-fix": "eslint \"{src,test}/**/*.ts\" --fix",
-    "prettier-check": "prettier --check .",
-    "prettier-write": "prettier --write .",
```
dolcalmi, line 8: why do we need to delete this?

src/services/ledger/index.ts
```diff
@@ -4,29 +4,34 @@
...
   LedgerTransactionType,
   liabilitiesMainAccount,
   toLiabilitiesWalletId,
   toWalletId,
 } from "@domain/ledger"
-import { lndAccountingPath, getBankOwnerWalletId } from "./accounts"
+
 import { wrapAsyncToRunInSpan } from "@services/tracing"
 
+import { Transaction } from "./schema"
```
dolcalmi, line 25: we need to disable rule here too

dolcalmi (changes requested):

package.json
nicolasburtey: because prettier is now run with `eslint`.

dolcalmi (approved):

.eslintrc.json
```diff
@@ -28,6 +29,24 @@
...
+      "quoteProps": "consistent",
+      "singleQuote": false,
+      "tabWidth": 2,
+      "useTabs": false,
+      "bracketSpacing": true,
+      "arrowParens": "always",
+      "proseWrap": "preserve",
+      "endOfLine": "lf"
+    }],
+
```
samerbuna, line 47: @nicolasburtey why did you add prettier to eslint? 

We have prettier enforced already. Having it here adds a lot of noise while code is being typed.

nicolasburtey: because I had issue with rebasing PR from my workflow. to make sure import are not duplicated or there are no import that are not used, I'm using `Shift + option + O` command from vs code. but this make the import order non really deterministic. 

to make it more deterministic and avoid a lot of rebase noise, I wanted to be able to use eslint as a formatter that can handle the import ordering. the best way I found to do it was to add prettier to eslint and then use eslint as the default formatter in VS Code. 

I tried other way to batch prettier + eslint from VS Code on save but didn't end up finding any workable solution

the new way is working a lot better for me. I'm saving files frequently to do the formatting basically. But it seems you have a different flow, and maybe you didn't had the issue I had previously

PR #860, opened 2021-12-27
feat: implement a btcPrice top-level query field

samerbuna:


dolcalmi (approved):

src/graphql/root/query/btc-price.ts
```diff
@@ -0,0 +1,25 @@
...
+    if (satUsdPrice instanceof Error) {
+      throw satUsdPrice
+    }
+
+    const price = 100 * satUsdPrice
+
+    return {
+      formattedAmount: price.toString(),
+      base: Math.round(price * 10 ** 12),
+      offset: 12,
```
dolcalmi, line 19: btcPriceList use `offset: 4` why are we using 12 here?

dolcalmi (changes requested):

src/graphql/root/query/btc-price.ts
samerbuna: I ported that from the subscription. This new field is actually the sat price, so maybe we should name `satPrice`?

dolcalmi: uhmm in theory price is a generic type so I think is ok

dolcalmi (approved):

src/graphql/root/query/btc-price.ts
nicolasburtey: I agree the 10 ** 12 is confusing. I think we can multiple by `10 ** 8` using `SATS_PER_BTC`, and then multiple back by `** 4` so that the calculation is bit more explicit.

also, there is no issue with overflow with javascript int with such big number?

nicolasburtey (changes requested):
one request to detail the magic number, also a question on potential use of big num

src/graphql/root/query/btc-price.ts
samerbuna: The magic number here is 12. That's the offset we're using to go from sat price to a number that can represent enough precision to be future-proof. As the sat price increases, we can lower that offset.

We're using the `SafeInt` type for the `base` of the price, and it's handling that level fine.

The `btcPriceList` endpoint returns prices for btc, not sats, and that's why the offset there is only 4.

Should I name this new field `satPrice` instead of `btcPrice`?

nicolasburtey: my point is not on whether the `satPrice` or `btcPrice`, but how did we came to the 10 ** 12. I guess the answer if `10 ** (8 + 4)`, 8 because it's the ratio of conversion from btc to sats, and 4 is an arbitrary number to increase precision.

this would not be obvious by reading the code so we should improve this (either with a comment or by creating variable that would explain this logic flow)

samerbuna: I named the 12 `SAT_PRICE_PRECISION_OFFSET`. It's 12 because it's for a sat price and it was the arbitrary pick to get enough precision. The other one is 4 because 12 - 8

nicolasburtey (approved):

PR #871, opened 2021-12-29
chore: add index for volume query

dolcalmi:
Transaction volume query is taking more than 6 seconds in production.

This is difficult to test but you can check index stats after run `TEST="01|02" make reset-integration` and you will notice that is the most used index (this is only a good indication that we need this index)

How to check index stats:
- Access db as a root
- Run `db.medici_transactions.aggregate([{ $indexStats: {} }])`
- Check `accesses.ops` value

### DIFF
--- a/src/services/ledger/schema.ts
+++ b/src/services/ledger/schema.ts
@@ -147,6 +147,7 @@ const transactionSchema = new Schema({
 })
 
 //indexes used by our queries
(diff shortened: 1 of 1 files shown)


nicolasburtey:
we should have a debug script to generate payments in loop on devnet - so we can have a 100k+ payment if we run the script for some hours - to assess those performance issues/optimization

nicolasburtey (approved):

samerbuna (approved):

PR #874, opened 2021-12-30
fix: use getBalanceForWalletId in btc wallet

dolcalmi:
Issue: MyUpdatesSubscription triggers every time price is updated (every 30 seconds)

This is a problem because we query the balance in each update https://github.com/GaloyMoney/galoy-mobile/blob/main/app/hooks/user-hooks.ts#L49 so we do all invoices and payments lookups for a user (lnd) every 30 seconds.

The solution is to separate price subscription from my updates but we cant change mobile app quickly so we will need to trust trigger until this is fixed.


samerbuna (approved):

dolcalmi:
closed in favor of https://github.com/GaloyMoney/galoy/pull/875

PR #875, opened 2021-12-30
fix: remove the me field when replying with a price update

samerbuna:


dolcalmi (approved):
I tested locally but only with playground

sebastienverreault (approved):

PR #877, opened 2021-12-30
fix: Fix addEarn call

daviroo:
The addEarn call currently returns a single quiz question but the graphql expects an array of quiz questions as can be seen in this line https://github.com/GaloyMoney/galoy/blob/3bd8eb7882ce382b5d169e545ca8774e9ee76946/src/graphql/old-schema.graphql#L272.

### DIFF
--- a/src/app/accounts/add-earn.ts
+++ b/src/app/accounts/add-earn.ts
@@ -54,5 +54,5 @@ export const addEarn = async ({
   })
   if (payment instanceof Error) return payment
 
-  return { id: quizQuestionId, value: amount, completed: true }
+  return [{ id: quizQuestionId, value: amount, completed: true }]
 }


vindard:
We'll have to do this fix in a different way. The old graphql server expects an array but the new one expects a single question. Fix would have to be in the `graphql-old-server.ts` somehow most likely. I can take a look if you'd like @daviroo.

vindard (commented):
We'll have to do this fix in a different way. The old graphql server expects an array but the new one expects a single question. Fix would have to be in the `graphql-old-server.ts` somehow most likely. I can take a look if you'd like @daviroo.

daviroo:
Closing since this fix will break the new endpoint.

PR #879, opened 2021-12-30
fix: addEarn method return handling in new graphql endpoint

vindard:
## Description

This PR adds types to the `addEarn` use-case method and refactors the return type from that method. It also refactors some of the error types and handling in the method.

### The Issue
Marking as **draft** until we can figure out why this doesn't break in the currently deployed app (cc @daviroo). It's also curious that we haven't been able to observe the `Cannot read properties of undefined (reading 'id')` in honeycomb as yet anywhere.

This is the response we get back when we call `userQuizQuestionUpdateCompleted` right now directly from something like Postman on both prod and dev:
```ts
{
    "data": {
        "userQuizQuestionUpdateCompleted": {
            "userQuizQuestion": null,
            "errors": [
                {
                    "message": "Cannot read properties of undefined (reading 'id')"
                }
            ]
        }
    }
}
```

### DIFF
--- a/src/app/accounts/add-earn.ts
+++ b/src/app/accounts/add-earn.ts
@@ -19,7 +19,7 @@ export const addEarn = async ({
   quizQuestionId: QuizQuestionId
   accountId: AccountId /* AccountId: aid validation */
   logger: Logger
-}) => {
+}): Promise<QuizQuestion | ApplicationError> => {
   const amount = onboardingEarn[quizQuestionId]
   if (!amount) {
     return new ValidationError("incorrect reward id")
@@ -55,5 +55,5 @@ export const addEarn = async ({
   })
   if (payment instanceof Error) return payment
 
-  return { id: quizQuestionId, value: amount, completed: true }
+  return { id: quizQuestionId, earnAmount: amount }
 }
--- a/src/app/accounts/add-earn.ts
+++ b/src/app/accounts/add-earn.ts
@@ -2,10 +2,10 @@ import { getUser } from "@app/users"
 import { intraledgerPaymentSendWalletId } from "@app/wallets"
 import { onboardingEarn } from "@config/app"
 import {
+  InvalidPhoneMetadataForRewardError,
   InvalidQuizQuestionIdError,
-  RewardMissingMetadataError,
-  RewardNonValidTypeError,
 } from "@domain/errors"
+import { PhoneMetadataValidator } from "@domain/users/phone-metadata-validator"
 import { getFunderWalletId } from "@services/ledger/accounts"
 import { RewardsRepository } from "@services/mongoose"
 
@@ -31,13 +31,9 @@ export const addEarn = async ({
   const user = await getUser(recipientAccount.ownerId)
   if (user instanceof Error) return user
 
-  if (!user.phoneMetadata?.carrier) {
-    return new RewardMissingMetadataError()
-  }
-
-  if (user.phoneMetadata.carrier.type === "voip") {
-    return new RewardNonValidTypeError()
-  }
+  const validatedPhoneMetadata = PhoneMetadataValidator().validate(user.phoneMetadata)
+  if (validatedPhoneMetadata instanceof Error)
+    return new InvalidPhoneMetadataForRewardError(validatedPhoneMetadata.name)
 
   const recipientWalletId = recipientAccount.defaultWalletId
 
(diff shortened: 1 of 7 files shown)


vindard:
_Responding to @daviroo's question about why we got the internal `Cannot read properties of undefined (reading 'id')` error exposed_

### How our error-handling works
Our error translation works in 2 ways:
- old-api: we would throw `CustomApolloError` child errors directly from the internal codebase to try to mitigate this. This has been mostly removed now and replaced by our new approach.

- new-api: we would catch and **return** all errors as `DomainError` types which would then be filtered at the graphql resolver layer and mapped (`mapError`) to a `CustomApolloError` type to be returned from the resolver. This is currently a manual process where we must:
  1. do an `instanceof Error` check on things returned from the domain, and
  2. pass any domain errors through `mapError` before attaching it to the resolvers returned object

### How it failed here
In this particular case, the internal error got through because:
- the use-case return wasn't properly typed so we didn't catch the index attempt on an object
- we used a `try-catch` at the graphql resolver layer, which could potentially bypass our `DomainError` handling which happened in this case

Avoiding this sort of thing is currently a manual process at the graphql layer that we're mostly able to avoid, but there can be cases like here where things slip through manual reviews.

### Potential mitigation
I'm not sure if there's a way to guarantee this doesn't happen via our automated checks, maybe via:
- typing returns at graphql layer so that things like the `err<any>.message` statement wouldn't pass lints, or
- having some sort of a wrapper on all resolvers to catch/translate (no idea how this might work)

vindard:
## Next Steps

We decided that we probably don't want to bubble up all instances of the errors that come back from the nested section since those can be user errors and not system errors. Ideally we'd like a way to distinguish between the two (exception vs. non-exception).

The next steps on this line of work would be to not create an error span event for all errors, but to instead:
- find a way to differentiate user vs system errors
- create span error events only for system errors
- bubble up span events to be captured by the honeycomb board (this PR)
- think about whether the dynamic mutation/query names in the attribute make sense for consistent query-ability

vindard:
@daviroo confirmed that the issue that was obscuring this on mobile end has been fixed and tested now, so this is ok to merge

daviroo:
I tested this locally and confirmed that the broken mutation is now fixed.  It would be good to make the phone validation optional in dev environment as right now this endpoint won't work in local dev unless you comment the phone validation out. I think that can be done in a separate change though as this fixes something which is currently broken.

src/graphql/root/mutation/user-quiz-question-update-completed.ts
```diff
@@ -19,29 +19,22 @@ const UserQuizQuestionUpdateCompletedMutation = GT.Field({
...
-    if (!onboardingEarn[id]) {
-      return { errors: [{ message: "Invalid input" }] }
+    const question = await Accounts.addEarn({
+      quizQuestionId: id,
+      accountId: uid,
+      logger,
+    })
+    if (question instanceof Error) {
+      const appErr = mapError(question)
+      return { errors: [{ message: appErr.message || appErr.name }] }
```
dolcalmi, line 29: map already assign appErr.name so you can remove `|| appErr.name`

src/graphql/root/mutation/user-quiz-question-update-completed.ts
```diff
@@ -19,29 +19,22 @@ const UserQuizQuestionUpdateCompletedMutation = GT.Field({
   resolve: async (_, args, { uid, logger }) => {
     const { id } = args.input
 
-    if (!onboardingEarn[id]) {
-      return { errors: [{ message: "Invalid input" }] }
+    const question = await Accounts.addEarn({
+      quizQuestionId: id,
+      accountId: uid,
```
nicolasburtey, line 24: `domainAccount.id` would be better type wise

nicolasburtey (commented):
just one comment but looks good over otherwise

nicolasburtey (approved):

src/app/accounts/add-earn.ts
```diff
@@ -19,11 +19,9 @@ export const addEarn = async ({
   quizQuestionId: QuizQuestionId
   accountId: AccountId /* AccountId: aid validation */
   logger: Logger
-}) => {
+}): Promise<QuizQuestion | ApplicationError> => {
   const amount = onboardingEarn[quizQuestionId]
-  if (!amount) {
-    return new ValidationError("incorrect reward id")
-  }
+  if (amount === undefined) return new InvalidQuizQuestionIdError()
```
dolcalmi, line 24:
```suggestion
  if (!amount) return new InvalidQuizQuestionIdError()
```

dolcalmi (changes requested):
looks good, just minor comment/suggestions

dolcalmi (approved):

PR #882, opened 2022-01-02
chore: move fees to the account level

nicolasburtey:
this PR will make segregating `Wallet` to a new collection easier, for the multi wallet architecture.


vindard:
> the setter should be at the `Account` level
> the getter should be (most frequently at least) at the `Wallet` level

I think fees at the account level makes sense. Is there any (non-performance) downside to always fetching the fee from AccountsRepo like you've done here?

nicolasburtey:
> Is there any (non-performance) downside to always fetching the fee from AccountsRepo like you've done here?

I guess a bit more complexity, as we need to move the reference of the account or at least the fees from the account around. 

but there is complexity of having the fees at the wallet level as well: it will need to be updated across all wallets for a particular account when fees are being updated for a particular user.

bodymindarts (approved):

PR #886, opened 2022-01-02
chore: adding update username use case to apps

nicolasburtey:


src/core/user-wallet.ts
```diff
@@ -137,9 +118,8 @@ export abstract class UserWallet {
     return true
   }
 
-  async updateUsername({
-    username,
-  }): Promise<{ username: string | undefined; id: string }> {
+  // TODO(remove) deprecated
+  async setUsername({ username }): Promise<{ username: string | undefined; id: string }> {
```
vindard, line 122: Would it be possible to swap this method in the old server for the new `Accounts.setUsername` method and then delete this here?

nicolasburtey: taking the approach of not really touching much api v1 code, and then bulk delete it shortly when we can remove the pod from the server.

otherwise there are things like type return that can have issues and I don't think it's relevant to spend time testing right now

vindard: Ok yea makes sense 👍🏾

src/graphql/root/mutation/user-update-username.ts
```diff
@@ -15,14 +16,16 @@ const UserUpdateUsernameMutation = GT.Field({
   args: {
     input: { type: GT.NonNull(UserUpdateUsernameInput) },
   },
-  resolve: async (_, args, { wallet }) => {
+  deprecationReason:
+    "Username will be moved to @Handle in Accounts. Also SetUsername should be used instead of UpdateUpdate to reflect the idempotenty of Handles",
```
vindard, line 20:
```suggestion
    "Username will be moved to @Handle in Accounts. Also SetUsername should be used instead of UpdateUsername to reflect the idempotency of Handles",
```

vindard (changes requested):
Some small suggestions

src/graphql/root/mutation/user-update-username.ts
```diff
@@ -1,3 +1,4 @@
+import { Accounts } from "@app"
 import { GT } from "@graphql/index"
```
vindard, line 2:
```suggestion
import { mapError } from "@graphql/error-map"
import { GT } from "@graphql/index"
```

vindard (approved):

PR #890, opened 2022-01-03
chore: add account walletId property to span attributes

vindard:
## Description

Add account `defaultWalletId` values alongside username values in spans. This should with querying traces help since the `walletId` is mandatory but the `username` can be unset.

**Sample trace:** [https://ui.honeycom...298f2cf](https://ui.honeycomb.io/galoy/datasets/arvin-dev/result/2HCZradDVY6/trace/xLHgf5CgnXN?span=50c82e5a7298f2cf)

### DIFF
--- a/src/graphql/root/mutation/ln-invoice-payment-send.ts
+++ b/src/graphql/root/mutation/ln-invoice-payment-send.ts
@@ -9,7 +9,9 @@ import {
   addAttributesToCurrentSpanAndPropagate,
   SemanticAttributes,
   ENDUSER_ALIAS,
+  ENDUSER_PUBLICID,
 } from "@services/tracing"
+import getUuidByString from "uuid-by-string"
 
 const LnInvoicePaymentInput = new GT.Input({
   name: "LnInvoicePaymentInput",
@@ -29,6 +31,9 @@ const LnInvoicePaymentSendMutation = GT.Field({
     addAttributesToCurrentSpanAndPropagate(
       {
         [SemanticAttributes.ENDUSER_ID]: domainUser?.id,
+        [ENDUSER_PUBLICID]: domainUser?.id
+          ? getUuidByString(domainUser?.id)
+          : domainUser?.id,
         [ENDUSER_ALIAS]: domainUser?.username,
         [SemanticAttributes.HTTP_CLIENT_IP]: ip,
       },
--- a/src/graphql/root/mutation/ln-invoice-payment-send.ts
+++ b/src/graphql/root/mutation/ln-invoice-payment-send.ts
@@ -9,7 +9,6 @@ import {
   addAttributesToCurrentSpanAndPropagate,
   SemanticAttributes,
   ENDUSER_ALIAS,
-  ENDACCOUNT_DEFAULTWALLETID,
 } from "@services/tracing"
 
 const LnInvoicePaymentInput = new GT.Input({
@@ -26,12 +25,11 @@ const LnInvoicePaymentSendMutation = GT.Field({
   args: {
     input: { type: GT.NonNull(LnInvoicePaymentInput) },
   },
-  resolve: async (_, args, { ip, domainUser, domainAccount, logger }) =>
+  resolve: async (_, args, { ip, domainUser, logger }) =>
     addAttributesToCurrentSpanAndPropagate(
       {
         [SemanticAttributes.ENDUSER_ID]: domainUser?.id,
         [ENDUSER_ALIAS]: domainUser?.username,
-        [ENDACCOUNT_DEFAULTWALLETID]: domainAccount?.defaultWalletId,
         [SemanticAttributes.HTTP_CLIENT_IP]: ip,
       },
       async () => {
--- a/src/graphql/root/mutation/ln-invoice-payment-send.ts
+++ b/src/graphql/root/mutation/ln-invoice-payment-send.ts
@@ -9,6 +9,7 @@ import {
   addAttributesToCurrentSpanAndPropagate,
   SemanticAttributes,
   ENDUSER_ALIAS,
+  ENDACCOUNT_DEFAULTWALLETID,
 } from "@services/tracing"
 
 const LnInvoicePaymentInput = new GT.Input({
@@ -25,11 +26,12 @@ const LnInvoicePaymentSendMutation = GT.Field({
   args: {
     input: { type: GT.NonNull(LnInvoicePaymentInput) },
   },
-  resolve: async (_, args, { ip, domainUser, logger }) =>
+  resolve: async (_, args, { ip, domainUser, domainAccount, logger }) =>
     addAttributesToCurrentSpanAndPropagate(
       {
         [SemanticAttributes.ENDUSER_ID]: domainUser?.id,
         [ENDUSER_ALIAS]: domainUser?.username,
+        [ENDACCOUNT_DEFAULTWALLETID]: domainAccount?.defaultWalletId,
         [SemanticAttributes.HTTP_CLIENT_IP]: ip,
       },
       async () => {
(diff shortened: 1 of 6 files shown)


nicolasburtey:
there has been honeycomb queries where this would be been helpful?

vindard:
> there has been honeycomb queries where this would be been helpful?

Not directly as yet, but I could see it being useful in the future. I noticed we were missing it when I was trying to query some of my own traces by walletId and saw we were missing it but had our db id and username properties already added as attributes to spans. I see this one as the always-present counterpart to sometimes-present usernames.

nicolasburtey:
> > there has been honeycomb queries where this would be been helpful?
> 
> Not directly as yet, but I could see it being useful in the future. I noticed we were missing it when I was trying to query some of my own traces by walletId and saw we were missing it but had our db id and username properties already added as attributes to spans. I see this one as the always-present counterpart to sometimes-present usernames.

can we set an array instead of a single value in opentelemetry?

defaultWalletId will quickly be not that useful is an environment where account has multiple wallets. so we could expose the `walletIds` array

src/servers/graphql-server.ts
```diff
@@ -114,7 +115,10 @@ const sessionContext = ({
         account = loggedInAccount
       }
 
-      addAttributesToCurrentSpan({ [ENDUSER_ALIAS]: domainAccount?.username })
+      addAttributesToCurrentSpan({
```
dolcalmi, line 118: if we use here `addAttributesToCurrentSpanAndPropagate` would avoid the implementation in the resolvers?

nicolasburtey: IIRC @bodymindarts told me this would not work (didn't try myself) because the `context` function doesn't launch the resolvers, it's more like a:
```
context() {}
resolver() {}
``` 

than 
```
context() {
...
...
...
... resolver() {}
}
```

bodymindarts: That is correct it won't work like @nicolasburtey says

src/services/tracing.ts
```diff
@@ -278,3 +278,4 @@ export const addAttributesToCurrentSpanAndPropagate = <F extends () => ReturnTyp
 export { SemanticAttributes, SemanticResourceAttributes }
 
 export const ENDUSER_ALIAS = "enduser.alias"
+export const ENDACCOUNT_DEFAULTWALLETID = "endaccount.defaultWalletId"
```
bodymindarts, line 281: I don't understand the prefix `end` here.
Also I would say we should be recording the `targetAccount.id` and / or `targetWallet.id` or something like that rather than `defaultWalletId` - it seems a bit indirect.

vindard: > I don't understand the prefix `end` here.

Sounded strange to me as well, but left it in to keep the `enduser` context and to see if anyone else noticed it was odd.

> Also I would say we should be recording the `targetAccount.id` and / or `targetWallet.id` or something like that rather than `defaultWalletId` - it seems a bit indirect.

The original idea here was to record something that we had access to at the graphql layer, and my 1st thought was walletId since we include that as input in a lot of requests. Thinking about it again though, the [uuid-representation of the userId](https://github.com/GaloyMoney/galoy/blob/70cba538e035acdc0af33ca84fc143208ee76127/src/graphql/types/object/graphql-user.ts#L26) could work just as well for this. I'm trying to find something that's non-internal to record.

vindard:
I changed away to using the enduser's public-facing id instead for this, under a new `enduser.publicId` attribute.

Sample trace here: [https://ui.honeycom...abfc](https://ui.honeycomb.io/galoy/datasets/arvin-dev/result/BPNySHjuPVT/trace/HSajGshR9Hy?span=e10f8449f166abfc)

src/services/tracing.ts
nicolasburtey: what is the goal of `getUuidByString(source.id)` for the user.id? not exposing the user.id? (this is what is in the token so it's exposed anyway currently)

vindard: > what is the goal of `getUuidByString(source.id)` for the user.id? not exposing the user.id? (this is what is in the token so it's exposed anyway currently)

Being able to quickly move from something like a `me` query to being able to trace a user's actions through honeycomb without having to necessarily decode the auth token or lookup the user in the database. It's a very specific use-case, so I'm ok either way if we decide to include or not, or if there is a better way to quickly find a given user in honeycomb from easily accessible info on the user.

nicolasburtey: but the `getUuidByString(source.id)` is also something that is exposed to the user directly right? do you make any use of it @daviroo ?

vindard: > that is exposed to the user directly right

I'm not sure if we present it in the UI anywhere, but it does come back on a `me` query

src/graphql/root/mutation/ln-invoice-payment-send.ts
```diff
@@ -29,6 +31,9 @@ const LnInvoicePaymentSendMutation = GT.Field({
     addAttributesToCurrentSpanAndPropagate(
       {
         [SemanticAttributes.ENDUSER_ID]: domainUser?.id,
+        [ENDUSER_PUBLICID]: domainUser?.id
```
bodymindarts, line 34: This is duplication of the line above right? `[SemanticAttributes.ENDUSER_ID]: domainUser?.id`

bodymindarts (changes requested):
I don't understand this PR - we are already reporting ` [SemanticAttributes.ENDUSER_ID]` what is the difference between that and `ENDUSER_PUBLICID`?

vindard:
> I don't understand this PR - we are already reporting ` [SemanticAttributes.ENDUSER_ID]` what is the difference between that and `ENDUSER_PUBLICID`?

The idea here is to expose something linked to the user that's more accessible, primarily for debugging purposes. It's to be able to move more easily from something we can easily see from the mobile-UI / API (e.g. `phone`) to something that's tagged in honeycomb traces and query-able.

❌ `ENDUSER_ALIAS` is both exposed and added to traces, **_but not all users have their alias set_** so we can't always use this.

❌ `ENDUSER_ID` is our database id and for a dev to fetch it, the process is to take a user's phone number/username and do a db query to retrieve the user.id that we can then use in honeycomb queries (**_problematic if no db access like in prod_**).

❓ `ENDUSER_PUBLICID` is exposed in the API (not in mobile-UI though) and can be used in place of `ENDUSER_ALIAS` to jump from API response to traces without having a dev query the db. There's the added potential benefit that if we switch to Postgres then this property would likely be migrated across but internal db ids would not.

---

**Feedback needed:**

- Does the general reasoning for adding this attribute make sense?

- Are there any tradeoffs when adding a new attribute like this that I'm missing?

- I'm leaving both `ENDUSER_ID` and `ENDUSER_PUBLICID` in right now to not break the traceability of existing users traces over time. Does this make sense, or should we only have one of these attributes?

The current need I opened this for is a niche debugging one, being able to easily see the logs for a single user without having to query mongodb manually. I'm ok if we decide to not push this change for now as well because it's niche.

src/graphql/root/mutation/ln-invoice-payment-send.ts
vindard: It's part of the multi-line ternary on the two lines just below this

nicolasburtey:
> ❌ ENDUSER_ID is our database id and for a dev to fetch it, the process is to take a user's phone number/username and do a db query to retrieve the user.id that we can then use in honeycomb queries (problematic if no db access like in prod).

I'll favor this option. I think the `ENDUSER_PUBLICID` is a bit abstract right now because it's not used in the mobile wallet. and it's also not possible to do a ENDUSER_PUBLICID --> ENDUSER_ID conversion right now because this is a hash. ID --> PUBLICID

PR #891, opened 2022-01-03
chore: upgrade graphql and enable typing for GT.Field

samerbuna:


src/graphql/admin/root/mutation/account-update-level.ts
```diff
@@ -17,7 +17,16 @@ const AccountUpdateLevelInput = new GT.Input({
   }),
 })
 
-const AccountUpdateLevelMutation = GT.Field({
+const AccountUpdateLevelMutation = GT.Field<
+  null,
+  GraphQLContext,
+  {
```
nicolasburtey, line 23: the third object is the type of input?

samerbuna: Yep. It's the type of the `args` object.

samerbuna: Ok, I made args the first argument since the others can be set default.

So the order now is TArgs, TSource, TContext with the TSource defaulting to null and TContext defaulting to GraphQLContext

We have a ton of errors already. Will work through them soon.

nicolasburtey:
I have done a rebase @samerbuna

src/graphql/types/object/wallet-contact.ts
```diff
@@ -9,7 +9,7 @@ import Username from "../scalar/username"
 
 import { TransactionConnection } from "./transaction"
 
-const AccountContact = new GT.Object({
+const AccountContact = GT.Object<UserRecord, GraphQLContextForUser>({
```
nicolasburtey, line 12: shouldn't `GraphQLContextForUser` be the third argument?

samerbuna: Nope. You don't define the args for resolvers here. We can use `GT.Field` to define args for a field, which we're doing for the mutations, but not queries (yet)

src/app/index.types.d.ts
```diff
@@ -4,6 +4,8 @@ type PartialResult<T> = {
   partialResult: true
 }
 
+type ValueOf<T> = T[keyof T]
```
nicolasburtey, line 7: what is the goal of this type?

src/graphql/root/mutation/ln-noamount-invoice-payment-send.ts
```diff
@@ -17,23 +18,42 @@ const LnNoAmountInvoicePaymentInput = new GT.Input({
...
+      return { errors: [{ message: walletId.message }] }
+    }
+    if (paymentRequest instanceof InputValidationError) {
+      return { errors: [{ message: paymentRequest.message }] }
+    }
+    if (amount instanceof InputValidationError) {
+      return { errors: [{ message: amount.message }] }
+    }
+    if (memo instanceof InputValidationError) {
+      return { errors: [{ message: memo.message }] }
```
dolcalmi, line 50: why do we need to remove the  for? we are not doing additional logic

dolcalmi (approved):
Just a small comment but not a blocker. Also I think this need another approval before merge. We can trust in graphql tests but we will need to do additional testing with the app in staging

src/graphql/root/mutation/ln-noamount-invoice-payment-send.ts
nicolasburtey: typescript doesn't seem smart enough to handle the type from the loop?

nicolasburtey: image

nicolasburtey (approved):

src/graphql/root/mutation/ln-noamount-invoice-payment-send.ts
samerbuna: Yep

src/app/index.types.d.ts
samerbuna: I used this in the language scalar.

PR #894, opened 2022-01-04
fix: disable lnd check in exporter server

dolcalmi:
Temporal fix testflight initialization


nicolasburtey (approved):

nicolasburtey:
merging to see if this is fixing testflight

PR #896, opened 2022-01-04
fix: add earn memos as valid to the memo filter

vindard:
## Description

This includes a fix to exclude onboarding earn memos from our memo filter.

I had originally tried to implement this using "approved walletIds" (like the `funder`) that would be excluded from the filter, but it turned out that it wasn't easy to get the sender's walletId for an intraledger transaction with the info we have on any given medici transaction record without going back to the journal.

### DIFF
--- a/src/domain/wallets/tx-history.ts
+++ b/src/domain/wallets/tx-history.ts
@@ -235,13 +235,8 @@ const shouldDisplayMemo = ({
   return isAuthorizedMemo(memo) || credit === 0 || credit >= MEMO_SHARING_SATS_THRESHOLD
 }
 
-const isAuthorizedMemo = (memo: string | undefined): boolean => {
-  if (memo && Object.keys(onboardingEarn).includes(memo)) {
-    return true
-  }
-
-  return false
-}
+const isAuthorizedMemo = (memo: string | undefined): boolean =>
+  !!memo && Object.keys(onboardingEarn).includes(memo)
 
 export const translateDescription = ({
   memoFromPayer,
--- a/src/domain/wallets/tx-history.ts
+++ b/src/domain/wallets/tx-history.ts
@@ -1,6 +1,6 @@
 import { toSats } from "@domain/bitcoin"
 import { ExtendedLedgerTransactionType, LedgerTransactionType } from "@domain/ledger"
-import { MEMO_SHARING_SATS_THRESHOLD } from "@config/app"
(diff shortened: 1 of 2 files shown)


src/domain/wallets/tx-history.ts
```diff
@@ -225,8 +225,18 @@ export const fromLedger = (
...
-  return credit === 0 || credit >= MEMO_SHARING_SATS_THRESHOLD
+const shouldDisplayMemo = ({
+  memo,
+  credit,
+}: {
+  memo: string | undefined
+  credit: number
+}) => {
+  return (
+    (memo && Object.keys(onboardingEarn).includes(memo)) ||
```
nicolasburtey, line 236: instead of `Object.keys(onboardingEarn).includes(memo)` maybe we could have another function `isAuthorizedMemo({memo}): boolean`. it would clarify the intent and also if we want to have other field here (like fees related that could be below threshold) we could do it easily.

src/domain/wallets/tx-history.ts
```diff
@@ -225,8 +225,22 @@ export const fromLedger = (
...
+}) => {
+  return isAuthorizedMemo(memo) || credit === 0 || credit >= MEMO_SHARING_SATS_THRESHOLD
+}
+
+const isAuthorizedMemo = (memo: string | undefined): boolean => {
+  if (memo && Object.keys(onboardingEarn).includes(memo)) {
+    return true
+  }
+
+  return false
```
dolcalmi, line 243:
```suggestion
  return memo && Object.keys(onboardingEarn).includes(memo)
```

dolcalmi (changes requested):

src/domain/wallets/tx-history.ts
```diff
@@ -225,10 +225,19 @@ export const fromLedger = (
...
+  credit,
+}: {
+  memo: string | undefined
+  credit: number
+}) => {
+  return isAuthorizedMemo(memo) || credit === 0 || credit >= MEMO_SHARING_SATS_THRESHOLD
 }
 
+const isAuthorizedMemo = (memo: string | undefined): boolean =>
+  !!memo && Object.keys(onboardingEarn).includes(memo)
```
vindard, line 239: Had to add the `!!` here because the linter didn't like it without that 🤷🏾‍♂️

dolcalmi (approved):

PR #898, opened 2022-01-05
fix: remove null values from business map query

dolcalmi:
Fix a production issue that is crashing the app in the map view.

We should investigate a bit more about why we have some users with null values in title and/or coordinates


sebastienverreault (approved):
LGTM

PR #899, opened 2022-01-05
feat: make playgroundUrl configurable

bodymindarts:
In client environments that expose the playground the url is pointing to our staging instead of the environment its deployed too.

### DIFF
--- a/default.yaml
+++ b/default.yaml
@@ -152,6 +152,7 @@ onChainWallet:
 
 apollo:
   playground: true
+  playgroundUrl: "https://api.staging.galoy.io/graphql"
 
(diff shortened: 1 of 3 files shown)


krtk6160 (approved):

default.yaml
```diff
@@ -152,6 +152,7 @@ onChainWallet:
 
 apollo:
   playground: true
+  playgroundUrl: "https://api.staging.galoy.io/graphql"
```
samerbuna, line 155: @bodymindarts can I make the default here http://localhost:4002/graphql ?

It's annoying for local dev to have the default on staging.

PR #900, opened 2022-01-05
chore: better user feedback when providing a bad phone number

bodymindarts:
Currently the response when requesting a phone code for a bad phone number is:
```
{
  "errors": [
    {
      "message": "Unknown error occurred (code: UnknownPhoneProviderServiceError)",
      "code": "UNKNOWN_CLIENT_ERROR"
    }
  ],
  "data": null
}
```
from the log output I can see that twilio is returning a useful error message:
```
"err":{"type":"RestException","message":"The 'To' number +123456 is not a valid phone number."
```

So I've added some translation to return this to the user.

### DIFF
--- a/src/domain/phone-provider/errors.ts
+++ b/src/domain/phone-provider/errors.ts
@@ -2,4 +2,5 @@ export class PhoneProviderServiceError extends Error {
   name = this.constructor.name
 }
 
+export class InvalidPhoneNumberPhoneProviderError extends PhoneProviderServiceError {}
 export class UnknownPhoneProviderServiceError extends PhoneProviderServiceError {}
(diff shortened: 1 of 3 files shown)


src/domain/phone-provider/errors.ts
```diff
@@ -2,4 +2,5 @@ export class PhoneProviderServiceError extends Error {
   name = this.constructor.name
 }
 
+export class NotAValidPhoneNumberPhoneProviderError extends PhoneProviderServiceError {}
```
krtk6160, line 5: imo 'invalid' would be a better prefix than 'notAValid'

bodymindarts: fixed

krtk6160 (approved):

PR #902, opened 2022-01-05
fix(tracing): set span error event only for top-level grapqhl errors

vindard:
## Description

This PR builds on the changes from #897 and #901.

Errors should only be marked as an `error` in honeycomb if they're system-level and at the top level of the response return on any graphql request. The errors nested within selection sets are usually considered as user errors and can be recorded in honeycomb but shouldn't be set as span error events.

This change returns us back to original state of these changes except with greater visibility into selection-set-level events at the span level only.

- **Sample trace:** [https://ui.honeycom...52f0](https://ui.honeycomb.io/galoy/datasets/arvin-dev/result/fhbVFoJTUA8/trace/HSziTJtaYQc?span=fb9f0c0b545a52f0)

### To improve in future

We currently still have system-level events that can be recorded at the user-level (selection set) level. We should filter for these and either:
1. Leave them at the selection set level, but add error span events for them
2. Throw them appropriately within resolvers so that they end up in the top level event

### DIFF
--- a/src/graphql/root/mutation/user-quiz-question-update-completed.ts
+++ b/src/graphql/root/mutation/user-quiz-question-update-completed.ts
@@ -3,6 +3,11 @@ import { mapError } from "@graphql/error-map"
 import { GT } from "@graphql/index"
 
 import UserQuizQuestionUpdateCompletedPayload from "@graphql/types/payload/user-quiz-question-update-completed"
+import {
+  addAttributesToCurrentSpanAndPropagate,
+  SemanticAttributes,
+  ENDUSER_ALIAS,
+} from "@services/tracing"
 
 const UserQuizQuestionUpdateCompletedInput = new GT.Input({
   name: "UserQuizQuestionUpdateCompletedInput",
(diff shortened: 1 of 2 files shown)


dolcalmi (dismissed):

vindard:
Do we know of any way we can get the `enduser.id` attribute attached to the `graphql.execute` span?

image

src/graphql/root/mutation/user-quiz-question-update-completed.ts
```diff
@@ -16,34 +21,41 @@ const UserQuizQuestionUpdateCompletedMutation = GT.Field({
   args: {
     input: { type: GT.NonNull(UserQuizQuestionUpdateCompletedInput) },
   },
-  resolve: async (_, args, { uid, logger }) => {
-    const { id } = args.input
+  resolve: async (_, args, { uid, domainUser, logger }) =>
+    addAttributesToCurrentSpanAndPropagate(
+      {
+        [SemanticAttributes.ENDUSER_ID]: domainUser?.id,
+        [ENDUSER_ALIAS]: domainUser?.username,
```
nicolasburtey, line 28: should be domainAccount

src/graphql/root/mutation/user-quiz-question-update-completed.ts
```diff
@@ -16,34 +21,41 @@ const UserQuizQuestionUpdateCompletedMutation = GT.Field({
   args: {
     input: { type: GT.NonNull(UserQuizQuestionUpdateCompletedInput) },
   },
-  resolve: async (_, args, { uid, logger }) => {
-    const { id } = args.input
+  resolve: async (_, args, { uid, domainUser, logger }) =>
```
nicolasburtey, line 24: you can get the correct type until @samerbuna PR https://github.com/GaloyMoney/galoy/pull/891 is being merged with: `{ uid, domainUser, logger }: {uid: UserId, domainUser: User, logger: Logger}`

src/services/tracing.ts
```diff
@@ -30,25 +30,37 @@ propagation.setGlobalPropagator(new W3CTraceContextPropagator())
...
   }
 }
 
-const recordGqlErrors = ({ errors, span, subPathName }) => {
+const recordGqlErrors = (args) => {
+  recordFirstGqlError(args)
+  recordAllGqlErrorsByIndex(args)
+}
+
+const triggerGqlErrorSpanEvent = ({ errors, span, subPathName }) => {
```
nicolasburtey, line 63:
```suggestion
const recordGqlErrors = ({
  errors,
  span,
  subPathName,
}: {
  span: Span
  errors: Record<string, unknown>
  subPathName: string
}) => {```

not sure what error type should be

nicolasburtey: `recordException` accept both a code and stack. we should probably set both?

vindard:
It looks like this branch has some stale good that we've moved away from. The only material change here is us adding a span for the `addEarns` use-case which we should probably do in a more structured way if we'd like to add spans for all resolvers.

Closing for now

PR #904, opened 2022-01-05
feat: add env port config to v2 and admin servers

dolcalmi:
Add env PORT configs to allow port update in api and admin servers (this is not an issue in our infra but may be useful in dev environments)

reported by @fiatjaf

### DIFF
--- a/src/servers/graphql-admin-server.ts
+++ b/src/servers/graphql-admin-server.ts
@@ -7,6 +7,8 @@ import { setupMongoConnection } from "@services/mongodb"
 
 import { activateLndHealthCheck } from "@services/lnd/health"
 
+import { GALOY_ADMIN_PORT } from "@config/app"
+
 import { gqlAdminSchema } from "../graphql"
 
 import { startApolloServer, isAuthenticated, isEditor } from "./graphql-server"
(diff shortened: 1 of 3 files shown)


src/servers/graphql-admin-server.ts
```diff
@@ -36,8 +36,9 @@ export async function startApolloServerForAdminSchema() {
     { allowExternalErrors: true },
   )
 
+  const port = process.env.PORT || 4001
```
nicolasburtey, line 39: I think we might want to have different environment value, someone might want to have both the graphql-admin and graphql-main running, but with this PR you can do it only with the default settings

nicolasburtey (changes requested):
environment variable should have different name

src/servers/graphql-admin-server.ts
nicolasburtey: probably the port name should be related to our software also?

dolcalmi: Fixed here https://github.com/GaloyMoney/galoy/pull/904/commits/292aeccaf515cefd092ceffcc5bab76a6e3d3162, removed v1 update because will be removed soon

nicolasburtey (approved):

PR #905, opened 2022-01-06
chore: try opentelemetry event

nicolasburtey:


vindard:
Do you have any sample honeycomb traces (before/after) for this change?

nicolasburtey:
> Do you have any sample honeycomb traces (before/after) for this change?

nope have not tried it. haven't tried to send data to honeycomb from a local setup up to now. do you think I should?

I intent to just look at it in staging or production afterwards.

dolcalmi:
> > Do you have any sample honeycomb traces (before/after) for this change?
> 
> nope have not tried it. haven't tried to send data to honeycomb from a local setup up to now. do you think I should?
> 
> I intent to just look at it in staging or production afterwards.

just need to add these envs and run the tests
```
export HONEYCOMB_DATASET="nicolas-dev"
export HONEYCOMB_API_KEY="api key from https://ui.honeycomb.io/teams/galoy/dataset_instructions" 
```

nicolasburtey:
image

This is how the span event are been added on honeycomb @vindard @dolcalmi 

when the span is being selected, when clicking on span event (2), we can see the list of events. it is automatically being shown how long after the even has been trigger after the span started. this is a useful way to get basic knowledge about  algorithms. 

and it's not bloating the UX at all, so it's really information that can be checked as needed.

src/services/ledger/index.ts
```diff
@@ -230,26 +230,24 @@ export const LedgerService = (): ILedgerService => {
...
-            {
-              $group: {
-                _id: null,
-                outgoingSats: { $sum: "$debit" },
-                incomingSats: { $sum: "$credit" },
-              },
-            },
-          ]),
-      })()
+      addEventToCurrentSpan("volume aggregation stats")
```
bodymindarts, line 233: having a start/end event seem redundant - if you want to record a start and end then that is what a span is for. Events are intended for 1-off bits of information.

nicolasburtey: fair. I'll still suggest to merge this one so we have an example on how to do this.

bodymindarts (approved):
I disagree with this usage of events and don't think it should be merged but will approve to move forward.

PR #906, opened 2022-01-06
refactor: separate PaymentSecret type into 3 types

vindard:
## Description

This PR is to fix two wrongly conflated concepts into the single `PaymentSecret` type.

### Background
There are two types of secrets associated with payment requests:
- payment identifying secret (included in encoded request)
- preimage (stored locally in lnd db)

We need to decide how to name these within our domain, and how to handle some considerations around the preimage.

#### Preimage
In the simplest use-case, a preimage is either generated if we are the recipient or received against payment if we are the sender.

- for preimages that we generate, this should never be shared/revealed in our domain code before the given invoice it is associated with is paid

- for preimages that we receive, this can be revealed since it acts as a proof-of-payment for the payment sent

## ~Feedback required~

1. ~[Two new types](https://github.com/GaloyMoney/galoy/pull/906/files#diff-da98a1543d9a0ee6bbbc2bd61e7c2b5800683f2daf2c6d6044ee77297d7a8fc3R9) have been created to represent preimages that are safe to share and ones that are not. Does this approach of having two preimage types make sense?~ ✅ 

2. ~We've had to clean up this conflation at the GraphQL level too which would mean [one small change](https://github.com/GaloyMoney/galoy/pull/906/files#diff-c4f4521cdf4877b970c28bba5e1258f07252e75f47ba97f5515cc4ae303bbc40R31) to our API. Are the changes I made here ok (breaking changes) or should they be done in some other way?~ **(see [comment](https://github.com/GaloyMoney/galoy/pull/906#issuecomment-1007999011))**

## Open `TODO`s
- [x]  deprecate `paymentSecret` field in gql instead of deleting it

### DIFF
--- a/src/domain/bitcoin/lightning/index.types.d.ts
+++ b/src/domain/bitcoin/lightning/index.types.d.ts
@@ -6,7 +6,9 @@ type RouteNotFoundError = import("./errors").RouteNotFoundError
 type InvoiceExpiration = Date & { readonly brand: unique symbol }
 type EncodedPaymentRequest = string & { readonly brand: unique symbol }
 type PaymentHash = string & { readonly brand: unique symbol }
-type PaymentSecret = string & { readonly brand: unique symbol }
+type SecretPreImage = string & { readonly brand: unique symbol }
+type RevealedPreImage = string & { readonly brand: unique symbol }
+type PaymentIdentifyingSecret = string & { readonly brand: unique symbol }
 type FeatureBit = number & { readonly brand: unique symbol }
 type FeatureType = string & { readonly brand: unique symbol }
 
@@ -41,7 +43,7 @@ type LnInvoiceLookup = {
   readonly isSettled: boolean
   readonly received: Satoshis
   readonly request: string | undefined
-  readonly secret: PaymentSecret
+  readonly secretPreImage: SecretPreImage
 }
 
 type LnPaymentLookup = {
@@ -51,7 +53,7 @@ type LnPaymentLookup = {
   readonly createdAt: Date
   readonly confirmedAt: Date | undefined
   readonly amount: Satoshis
-  readonly secret: PaymentSecret
+  readonly revealedPreImage: RevealedPreImage
   readonly request: string | undefined
   readonly destination: Pubkey
 }
@@ -65,7 +67,7 @@ type LnInvoice = {
   readonly cltvDelta: number | null
   readonly amount: Satoshis | null
   readonly routeHints: Hop[][]
-  readonly paymentSecret: PaymentSecret | null
+  readonly paymentSecret: PaymentIdentifyingSecret | null
   readonly features: LnInvoiceFeature[]
 }
 
--- a/src/graphql/admin/schema.graphql
+++ b/src/graphql/admin/schema.graphql
@@ -108,6 +108,8 @@ type LightningPayment {
   status: LnPaymentStatus
 }
 
+scalar LnPaymentPreImage
+
 """
 BOLT11 lightning invoice payment request with the amount included
 """
@@ -197,6 +199,10 @@ type SettlementViaIntraLedger {
 
 type SettlementViaLn {
   paymentSecret: LnPaymentSecret
+    @deprecated(
+      reason: "Shifting property to 'preImage' to improve granularity of the LnPaymentSecret type"
+    )
+  preImage: LnPaymentPreImage
 }
 
 type SettlementViaOnChain {
--- a/src/graphql/types/scalar/ln-payment-preimage.ts
+++ b/src/graphql/types/scalar/ln-payment-preimage.ts
@@ -1,0 +1,25 @@
+import { GT } from "@graphql/index"
+import { UserInputError } from "apollo-server-errors"
+
+const LnPaymentPreImage = new GT.Scalar({
+  name: "LnPaymentPreImage",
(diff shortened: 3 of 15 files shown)


src/domain/bitcoin/lightning/index.types.d.ts
```diff
@@ -6,7 +6,9 @@ type RouteNotFoundError = import("./errors").RouteNotFoundError
 type InvoiceExpiration = Date & { readonly brand: unique symbol }
 type EncodedPaymentRequest = string & { readonly brand: unique symbol }
 type PaymentHash = string & { readonly brand: unique symbol }
-type PaymentSecret = string & { readonly brand: unique symbol }
+type SecretPreImage = string & { readonly brand: unique symbol }
+type RevealedPreImage = string & { readonly brand: unique symbol }
+type PaymentIdentifyingSecret = string & { readonly brand: unique symbol }
```
nicolasburtey, line 11: here is another option for naming :

- PaymentPreimage
- PaymentSecret
- PaymentSecretRevealed

nicolasburtey:
> We've had to clean up this conflation at the GraphQL level too which would mean one small change to our API. Are the changes I made here ok (breaking changes) or should they be done in some other way?

not completely sure to understand if you want to know if your changes are breaking changes, or if you know they are and want to know if that's ok.

I think they are great tool for graphql to know when we do breaking change automatically. we should dive into this. I think apollo studio can help but I'm sure there are other tools that could work as well

vindard:
> not completely sure to understand if you want to know if your changes are breaking changes, or if you know they are and want to know if that's ok.

I'm assuming this _is_ a breaking change since anyone trying to access the `settlementVia.paymentSecret` property now can't because it's shifted to `settlementVia.preImage`. My question is, is it ok to just rip out `paymentSecret` and switch it to `preImage` since we're still pretty early and haven't officially announced the API as yet, or should I do something more like adding an `@deprecated` flag and leaving `paymentSecret` in while also adding the new `preImage` selection that would carry the same value?

src/domain/bitcoin/lightning/index.types.d.ts
vindard: Do we want to attach this to the `Payment` concept given that we have explicit `LnPayment` types that are different from `LnInvoice` types and both types an carry a preimage?

nicolasburtey:
> I'm assuming this _is_ a breaking change since anyone trying to access the `settlementVia.paymentSecret` property now can't because it's shifted to `settlementVia.preImage`. My question is, is it ok to just rip out `paymentSecret` and switch it to `preImage` since we're still pretty early and haven't officially announced the API as yet, or should I do something more like adding an `@deprecated` flag and leaving `paymentSecret` in while also adding the new `preImage` selection that would carry the same value?

My assumption is that if this breaks mobile, of course we should do a transition, otherwise I think we can move forward and go for the breaking change without intermediary steps because we don't have API users yet on this (this later assumption could be confirmed by looking at the logs/traces, but I'm confident this is the case)

src/domain/bitcoin/lightning/index.types.d.ts
nicolasburtey: oh you're right. a reason to separate those types further? 😬

vindard: > oh you're right. a reason to separate those types further? grimacing

I was thinking that as well, but that probably introduces unnecessary complexity.

For example, if in the future we wanted to do hodl invoice tricks and we want to for e.g. generate the preimage+hash for an LnPayment (see [these scenarios](https://bolt.fun/guide/invoices/hodl-invoice#scenarios-source--more-examples)), we could still fulfill this use-case with just `secret`/`revealed` types without having to figure out whether it's attached to an invoice we generate or a payment we make, or some in-between concept of those two.

vindard:
Based on discussion in Friday Dev Call West we've decided to do the process of deprecating `paymentSecret` since the `settlementVia.paymentSecret` is currently being used in the mobile query.

src/graphql/types/scalar/ln-payment-preimage.ts
```diff
@@ -0,0 +1,25 @@
...
+    if (ast.kind === GT.Kind.STRING) {
+      return validLnPaymentPreImage(ast.value)
+    }
+    return new UserInputError("Invalid type for LnPaymentPreImage")
+  },
+})
+
+function validLnPaymentPreImage(value) {
+  // TODO: verify/improve
+  if (value.match(/^[A-Fa-f0-9]{64}$/)) {
```
nicolasburtey, line 19: we should probably extract some `isSha256hash()` function from this

src/graphql/admin/schema.graphql
```diff
@@ -197,6 +199,10 @@ type SettlementViaIntraLedger {
 
 type SettlementViaLn {
   paymentSecret: LnPaymentSecret
+    @deprecated(
+      reason: "This value/type is mismatched with our domain and will be shifted across to the 'preImage' property."
```
bodymindarts, line 203: I don't know if we should be talking about our domain here as from the POV of an API user it is an implementation detail.
A better way to say it might be: `reason: "Improving granularity of the paymentSecret type"` or something like that.

bodymindarts (approved):

bodymindarts (approved):

PR #911, opened 2022-01-07
refactor: onchain pay use case

dolcalmi:
This is the 3rd (and final) PR to refactor onchain pay use case.

Includes:
- Use case migration to app layer
- Graphql v1 and v2 updates
- Tests updates
- Core clean up (Lightning and Onchain folder). This required to remove abstract from UserWallet to be backward compatible while we remove/migrate the pending functionality
- [Remove oldEnoughForWithdrawal ](https://galoymoney-workspace.slack.com/archives/C025EMR24KB/p1641509333004800?thread_ts=1641508887.004100&cid=C025EMR24KB)


src/app/wallets/pay-on-chain.ts
```diff
@@ -0,0 +1,348 @@
...
+    ? await getBalanceForWalletId(senderWalletId)
+    : checkedToSats(amount)
+  if (checkedAmount instanceof Error) return checkedAmount
+
+  const user = await getUser(payerUserId)
+  if (user instanceof Error) return user
+
+  const { twoFA } = user
+
+  const twoFACheck = twoFA?.secret
```
nicolasburtey, line 56: I think the call should fail if 2fa is not set (as opposed to continue as if the code was correct). wdyt?

src/app/wallets/pay-on-chain.ts
```diff
@@ -0,0 +1,348 @@
...
+
+  const usdPerSat = await getCurrentPrice()
+  if (usdPerSat instanceof Error) return usdPerSat
+
+  const fee = toSats(0)
+  const sats = toSats(amount + fee)
+  const usd = sats * usdPerSat
+  const usdFee = fee * usdPerSat
+
+  const getUserName = async (walletId: WalletId) => {
```
nicolasburtey, line 165: so far we have been spelling username as `Username` (no capital on N)

test/integration/02-user-wallet/02-send-onchain.spec.ts
```diff
@@ -46,30 +58,41 @@ const date = Date.now() + 1000 * 60 * 60 * 24 * 8
...
+// using walletId11 and walletId12 to sendAll
+let walletId11: WalletId
+let walletId12: WalletId
+let userId0: UserId
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const { sendNotification } = require("@services/notifications/notification")
 
 beforeAll(async () => {
   userWallet0 = await getAndCreateUserWallet(0)
```
nicolasburtey, line 75: I would try to delete `userWallet0`. this would allow the deletion of `getAndCreateUserWallet` method.

nicolasburtey: if you need the user object associated with user0, there is a new method for it `getUserTypeByTestUserIndex`. it'll give you the `userType` object, equivalent to `userWallet0.user`.

nicolasburtey (commented):
few comments, but otherwise LGTM

src/app/wallets/pay-on-chain.ts
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/911/commits/7a9d01b554c4b98e50409140ed6c3014f859763d

test/integration/02-user-wallet/02-send-onchain.spec.ts
dolcalmi: Updated and `getAndCreateUserWallet` was removed https://github.com/GaloyMoney/galoy/pull/911/commits/7a9d01b554c4b98e50409140ed6c3014f859763d

src/app/wallets/pay-on-chain.ts
dolcalmi: I think is better to do it in another PR because we will need to update intraledger and lightning too

nicolasburtey (approved):

PR #912, opened 2022-01-08
chore: remove some code related to multi currency

nicolasburtey:
this code will be reimplemented in a subsequent PR

### DIFF
--- a/test/integration/services/ledger/transaction.__spec__.ts
+++ b/test/integration/services/ledger/transaction.__spec__.ts
@@ -501,30 +501,30 @@ describe("rebalancePortfolio", () => {
     })
     await expectBalance({ account: lndAccountingPath, currency: "BTC", balance: -1000 })
 
-    wallet.user.currencies = _5050meta.currencies
-    const error = wallet.user.validateSync()
-    expect(error).toBeFalsy()
+    // wallet.user.currencies = _5050meta.currencies
+    // const error = wallet.user.validateSync()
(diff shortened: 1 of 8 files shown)


test/integration/services/ledger/transaction.__spec__.ts
```diff
@@ -501,30 +501,30 @@ describe("rebalancePortfolio", () => {
...
+    // wallet.user.currencies = _5050meta.currencies
+    // const error = wallet.user.validateSync()
+    // expect(error).toBeFalsy()
 
-    await ledger.rebalancePortfolio({
-      description: "rebalancePortfolio",
-      metadata: { type: "user_rebalance", pending: false },
-      wallet,
-    })
+    // await ledger.rebalancePortfolio({
```
bodymindarts, line 508: Why are you commenting this code? If you ever need to look it up again its already in the git history and you can just check out this file before it was deleted. No need to keep it or?

bodymindarts (approved):
Approving but think we should delete all the commented code - its still in the git history.

test/integration/services/ledger/transaction.__spec__.ts
nicolasburtey: the tests will be re-implemented shortly with the new code infra. so keep this code around is saving time, versus trying to find back where it's from the git history.

nicolasburtey:
merging following rebase

PR #915, opened 2022-01-08
refactor: Separate ios and android build versions

krtk6160:


bodymindarts (approved):

nicolasburtey (approved):

src/servers/graphql-old-server.ts
```diff
@@ -142,16 +142,20 @@ const resolvers = {
...
     buildParameters: async () => {
-      const { minBuildNumber, lastBuildNumber } = getBuildVersions()
+      const {
+        minBuildNumberAndroid,
+        lastBuildNumberAndroid,
+        minBuildNumberIos,
+        lastBuildNumberIos,
+      } = getBuildVersions()
       return {
-        id: lastBuildNumber,
```
dolcalmi, line 147: @daviroo @samerbuna could you please confirm if this id is required for graphql client?

daviroo: For V1 api calls we did indeed use this field https://github.com/GaloyMoney/galoy-mobile/blob/04bca8d0ecd6b42fab02907aa012856477bbca46/app/graphql/query.ts#L208

PR #917, opened 2022-01-09
fix: username is null when not present, not undefined

nicolasburtey:


test/integration/02-user-wallet/02-index.spec.ts
```diff
@@ -97,24 +96,22 @@ describe("UserWallet", () => {
...
+      expect(result).not.toBeInstanceOf(Error)
     })
 
     it("does not allow set username if already taken", async () => {
       const username = "user0"
-
-      await createUserWallet(2)
       await expect(setUsername({ username, id: accountId2 })).resolves.toBeInstanceOf(
-        UsernameNotAvailableError,
+        AccountError,
```
dolcalmi, line 107: please use the specific error. The same applies for other changes in this file

src/app/accounts/set-username.ts
```diff
@@ -16,8 +16,7 @@ export const setUsername = async ({
   const accountsRepo = AccountsRepository()
   const account = await accountsRepo.findById(id as AccountId)
   if (account instanceof Error) return account
-
-  if (account.username !== undefined) return new UsernameIsImmutableError()
+  if (account.username != null) return new UsernameIsImmutableError()
```
dolcalmi, line 19: did you test this with current data? maybe is better to validate undefined, null and empty

```suggestion
  if (account.username) return new UsernameIsImmutableError()
```

dolcalmi (changes requested):

test/integration/02-user-wallet/02-index.spec.ts
nicolasburtey: the specific error is path dependant, (depending on why tests have been run, one error (UsernameNotAvailableError ) or another one (UsernameIsImmutableError) might be returned which is why I return the parent common class.

src/app/accounts/set-username.ts
nicolasburtey: it's been write as null in mongodb. maybe there are some scenario where it is undefined. it can't be empty string IMU because there is a non empty index on it.

test/integration/02-user-wallet/02-index.spec.ts
dolcalmi: this is a common mistake in tests, if we are testing a case for specific error then we need to use the proper instance of the error to guarantee that the code is validating the correct path

src/app/accounts/set-username.ts
dolcalmi: if it is a possibility then we should use `if (account.username)`

test/integration/02-user-wallet/02-index.spec.ts
bodymindarts: If this test is non-deterministic then there is something wrong with the setup.
Tests should be idempotent and deterministic.
What is preventing you from making this test order independent?

nicolasburtey: because you had to test one condition first, and then the next one: https://github.com/GaloyMoney/galoy/blob/1807822e46e5b34722c0868c80bace6a6ba53ea5/src/app/accounts/set-username.ts#L20

nicolasburtey: so if username is already set you won't go further.

bodymindarts: I'm not sure I understand what is preventing you from weighting this pseudocode:

test1:
```
user = getNewUser();
setUsername(user.id, "name");
expect(setUsername(user.id, "name")).toThrow(UsernameIsImmutableError()
```
test2:
```
user=getNewUser();
setUsername(user.id, ("name"));
user2=getNewUser();
expect(setUsername(userd.id, "name")).toThrow(UsernameNotAvailable)()
```

bodymindarts (approved):

PR #918, opened 2022-01-09
refactor: replace User with WalletsRepo

nicolasburtey:


src/servers/trigger.ts
```diff
@@ -107,13 +107,18 @@ export async function onchainTransactionEventHandler(tx) {
...
+      const walletsRepo = WalletsRepository()
+
+      // TODO: tx.output_addresses pass an array of address
+      // in the unlikely event multiple destination addresses are belong to the Galoy wallet
+      // only the first one will be notified
+      const wallet_ = await walletsRepo.findByAddress(tx.output_addresses)
+      if (wallet_ instanceof Error) return
+      wallet = wallet_
+
+      if (!wallet) {
```
dolcalmi, line 121: this validation is not necessary, if address is not assigned to an user wallet will be an instance of Error

src/servers/trigger.ts
```diff
@@ -107,13 +107,18 @@ export async function onchainTransactionEventHandler(tx) {
...
-      )
-      if (!user) {
+      const walletsRepo = WalletsRepository()
+
+      // TODO: tx.output_addresses pass an array of address
+      // in the unlikely event multiple destination addresses are belong to the Galoy wallet
+      // only the first one will be notified
+      const wallet_ = await walletsRepo.findByAddress(tx.output_addresses)
+      if (wallet_ instanceof Error) return
+      wallet = wallet_
```
dolcalmi, line 119: we dont need to have this inside the try/catch so you dont need the additional variable `wallet_`

dolcalmi (changes requested):

src/servers/trigger.ts
```diff
@@ -107,24 +107,14 @@ export async function onchainTransactionEventHandler(tx) {
...
-      throw error
-    }
+    const walletsRepo = WalletsRepository()
+
+    // TODO: tx.output_addresses pass an array of address
+    // in the unlikely event multiple destination addresses are belong to the Galoy wallet
+    // only the first one will be notified
+    const wallet_ = await walletsRepo.findByAddress(tx.output_addresses)
+    if (wallet_ instanceof Error) return
+    const wallet = wallet_
```
dolcalmi, line 117: after instanceof validation wallet can be used as wallet type.. you dont have to re-assign it

dolcalmi (changes requested):

src/servers/trigger.ts
```diff
@@ -107,24 +107,13 @@ export async function onchainTransactionEventHandler(tx) {
...
-        "issue in onchainTransactionEventHandler to get User id attached to output_addresses",
-      )
-      throw error
-    }
+    const walletsRepo = WalletsRepository()
+
+    // TODO: tx.output_addresses pass an array of address
+    // in the unlikely event multiple destination addresses are belong to the Galoy wallet
+    // only the first one will be notified
+    const wallet = await walletsRepo.findByAddress(tx.output_addresses)
```
bodymindarts, line 115: I'm confused that this compiles - I think we are missing types.
The name `tx.output_addresses` implies an array type `[]` but `walletsRepo.findByAddress` expects a single `OnChainAddress`

nicolasburtey: `tx` is of type `any`. Will add the correct type

src/servers/trigger.ts
```diff
@@ -135,12 +118,15 @@ export async function onchainTransactionEventHandler(tx) {
...
-        walletId: user.walletId,
-        amount: toSats(Number(tx.tokens)),
-        txHash: tx.id,
-        usdPerSat,
-      })
+
+      wallets.forEach((wallet) =>
+        NotificationsService(onchainLogger).onChainTransactionReceivedPending({
+          walletId: wallet.id,
+          amount: toSats(Number(tx.tokens)),
```
vindard, line 125: I was going to say that if we're separating by wallets we might also want to sum up the relevant utxo amounts for a given wallet here instead of reporting the entire tx amount across all utxos

... but then I noticed from the type that the `lightning` library doesn't give us that level of granularity

vindard (dismissed):
Looks ok to me now after latest changes. Just left one comment about tx amounts and multiple wallets.

src/servers/trigger.ts
nicolasburtey: oh right. let me add a TODO at least.

vindard (approved):
Just one small nit, but approving anyway

PR #919, opened 2022-01-09
refactor: move contact to account

nicolasburtey:


dolcalmi:
as far as I remember contact was part of the user not accounts... did this change?

nicolasburtey:
> as far as I remember contact was part of the user not accounts... did this change?

we're moving using username to `handle`, and those will be attached either to account or wallets. this is just a first PR is that direction.

src/app/accounts/update-contact-alias.ts
```diff
@@ -0,0 +1,31 @@
...
+}): Promise<AccountContact | ApplicationError> => {
+  const repo = AccountsRepository()
+  const account = await repo.findById(accountId)
+  if (account instanceof Error) {
+    return account
+  }
+
+  const contact = account.contacts.find((contact) => contact.username === username)
+  if (!contact) {
+    return new ValidationError(`User doesn't have contact ${username}`)
```
bodymindarts, line 21: `ValidationError` is just a base class and shouldn't be instantiated directly - please create a new error type.

src/app/accounts/update-contact-alias.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+  contact.alias = alias as ContactAlias
```
bodymindarts, line 23: does the alias names have any constraints (length / charset)? do we need a `checkedToContactAlias`?

src/app/users/add-new-contact.ts
```diff
@@ -1,35 +1,35 @@
-import { UsersRepository } from "@services/mongoose"
+import { AccountsRepository } from "@services/mongoose"
 
 export const addNewContact = async ({
```
bodymindarts, line 3: file should be moved under `app/accounts` or?

src/app/accounts/update-contact-alias.ts
nicolasburtey: yup. good point. actually as of now this could be a source of ddos.

bodymindarts (approved):

PR #920, opened 2022-01-09
refactor: using User instead of UserType for notifications

nicolasburtey:


src/services/notifications/index.ts
```diff
@@ -166,19 +173,26 @@ export const NotificationsService = (logger: Logger): INotificationsService => {
...
         type: NotificationType.IntraLedgerPayment,
-        user: payerUser,
+        user: senderUser,
         logger,
         amount,
         usdPerSat,
       })
 
-      const recipientUser = await User.findOne({ walletId: recipientWalletId })
+      const recipientAccount = await AccountsRepository().findByWalletId(senderWalletId)
```
dolcalmi, line 191: please use `recipientWalletId` instead of `senderWalletId`

dolcalmi (changes requested):

src/services/notifications/index.ts
nicolasburtey: good catch.

dolcalmi (approved):

PR #921, opened 2022-01-09
feat: separating Wallets collection from Users

nicolasburtey:


src/app/accounts/add-wallet.ts
```diff
@@ -0,0 +1,28 @@
+import { AccountsRepository, WalletsRepository } from "@services/mongoose"
+
+export const addWallet = async ({
+  accountId,
+  type,
+}: {
+  accountId: AccountId
+  type: WalletType
+}) => {
```
dolcalmi, line 9:
```suggestion
}): Promise<Wallet | ApplicationError> => {
```

src/app/accounts/set-default-walletid.ts
```diff
@@ -0,0 +1,22 @@
+import { InvalidWalletId } from "@domain/errors"
+import { AccountsRepository } from "@services/mongoose"
+
+export const setDefaultWalletId = async ({
+  accountId,
+  walletId,
+}: {
+  accountId: AccountId
+  walletId: WalletId
+}) => {
```
dolcalmi, line 10:
```suggestion
}): Promise<Account | ApplicationError> => {
```

src/app/accounts/set-default-walletid.ts
```diff
@@ -0,0 +1,22 @@
...
+  const account = await AccountsRepository().findById(accountId)
+  if (account instanceof Error) return account
+
+  if (!account.walletIds.some((wid) => wid === walletId)) return InvalidWalletId
```
dolcalmi, line 14: should this be `return new InvalidWalletId()`? (another reason to add return type to the method)

src/app/users/create-user.ts
```diff
@@ -0,0 +1,41 @@
...
+      userRaw.phoneMetadata = carrierInfo
+    }
+  }
+
+  const user = await UsersRepository().persistNew(userRaw)
+  if (user instanceof Error) return user
+
+  const account = await AccountsRepository().findByUserId(user.id)
+  if (account instanceof Error) return account
+
```
dolcalmi, line 36: not sure if I am understanding this logic, if we create an user from scratch this will always return and error (after collection separation)

src/app/users/login.ts
```diff
@@ -57,18 +58,8 @@ export const login = async ({
...
-    const carrierInfo = await TwilioClient().getCarrier(phoneNumberValid)
-    if (carrierInfo instanceof Error) {
-      // non fatal error
-      subLogger.warn({ phone }, "impossible to fetch carrier")
-    } else {
-      userRaw.phoneMetadata = carrierInfo
-    }
-
-    user = await userRepo.persistNew(userRaw)
+    user = await createUser(userRaw)
```
dolcalmi, line 62: we need to add account creation in `createUser` or `createAccount` and pass accountId to `createUser`

src/domain/wallets/index.types.d.ts
```diff
@@ -124,14 +124,29 @@ type WalletTransactionHistoryWithPending = {
...
+  readonly type: WalletType
   readonly onChainAddressIdentifiers: OnChainAddressIdentifier[]
   onChainAddresses(): OnChainAddress[]
 }
 
 interface IWalletsRepository {
-  findById(walletId: WalletId): Promise<Wallet | RepositoryError>
+  persistNew({ accountId, type }: NewWalletInfo): Promise<Wallet | RepositoryError>
+  findById(id: WalletId): Promise<Wallet | RepositoryError>
+  findByAccountId(accountId: AccountId): Promise<Wallet | RepositoryError>
```
dolcalmi, line 149: the structure  is `accounts 1 -> * wallets` so in theory we should not be able to get a wallet by account Id (this method should not be in this repo)

src/domain/wallets/index.types.d.ts
```diff
@@ -124,14 +124,29 @@ type WalletTransactionHistoryWithPending = {
   readonly transactions: WalletTransaction[]
 }
 
+type WalletName = string & { readonly brand: unique symbol }
```
dolcalmi, line 127: is this used?

src/services/mongoose/accounts.ts
```diff
@@ -36,7 +36,7 @@ export const AccountsRepository = (): IAccountsRepository => {
     walletId: WalletId,
   ): Promise<Account | RepositoryError> => {
     try {
-      const result: UserType = await User.findOne({ walletId }, projection)
+      const result: UserType = await User.findOne({ walletIds: walletId }, projection)
```
dolcalmi, line 39: please rename to `walletId` param or rename the method to use list prefix and pass an array

src/services/mongoose/wallet-on-chain-addresses.ts
```diff
@@ -1,21 +1,21 @@
...
   RepositoryError,
   UnknownRepositoryError,
 } from "@domain/errors"
 import { baseLogger } from "@services/logger"
+import { Wallet } from "@services/mongoose/schema"
 
 export const WalletOnChainAddressesRepository = (): IWalletOnChainAddressesRepository => {
   const persistNew = async (
-    walletId: WalletId,
+    id: WalletId,
```
dolcalmi, line 12: we need to update this to one param rule

dolcalmi (changes requested):
needs a rebase too

src/domain/wallets/index.types.d.ts
```diff
@@ -124,14 +124,29 @@ type WalletTransactionHistoryWithPending = {
...
+  type: WalletType
+}
+
 type Wallet = {
   readonly id: WalletId
+  readonly accountId: AccountId
+  readonly type: WalletType
   readonly onChainAddressIdentifiers: OnChainAddressIdentifier[]
   onChainAddresses(): OnChainAddress[]
 }
```
vindard, line 144:
```suggestion
type Wallet = NewWalletInfo & {
  readonly id: WalletId
  readonly onChainAddressIdentifiers: OnChainAddressIdentifier[]
  onChainAddresses(): OnChainAddress[]
}
```

vindard (commented):
Preliminary thought about how we will do migrations for this.

We are coupling `accountId` in the wallet to `User.id` on the User object in this PR. This means that when we eventually split out the `Account` collection, in the migration we will have to:
- create an Account for each user
- create a new `defaultAccountId` field on each User
- change the `accountId` on each wallet by:
  - querying the current `accountId` value on a wallet
  - using that value to query the `User` collection and fetching its `defaultAccountId` value
  - re-assign `Wallet.accountId` to this new value

I'm wondering if it would be simpler to separate `Account` collection _before_ `Wallet` collection for a less risky migration when we get to that point, since `Wallet.accountId` will simply point to the `Account.id` property in the already created `Account` collection.

src/domain/wallets/index.ts
```diff
@@ -15,3 +15,8 @@ export const checkedToWalletId = (walletId: string): WalletId | ValidationError
   }
   return walletId as WalletId
 }
+
+export const WalletType = {
+  CheckingBTC: "checkingbtc",
+  CheckingUSD: "checkingusd",
+} as const
```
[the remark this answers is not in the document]
nicolasburtey: I'm thinking that we'll add more local currency as the need arise. not sure having a single FiatCurrency to be the way to go just yet (maybe it is... but some countries has 2 fiat currencies, like a local currency + USD)

src/app/users/login.ts
nicolasburtey: Account and User are still the same entity currently. This PR is about separating User/Account from Wallet.

nicolasburtey: I think I understand what you mean. Splitting this further so that the Wallet is only being created from create-account?

src/domain/wallets/index.types.d.ts
nicolasburtey: we can get back to account from wallet from this PR. `Wallet` has a pointer back to Account. This is not strictly necessary but a welcome optimization because many queries will comes from a wallet (like a transaction being confirmed) and getting the accountId handy will avoid fetching back the Account (currently User) collection.

src/domain/wallets/index.types.d.ts
nicolasburtey: probably comes from this PR: https://github.com/GaloyMoney/galoy/pull/835. will remove it from now

src/services/mongoose/accounts.ts
nicolasburtey: I don't understand this comment:

> please rename to walletId param

we want to get one `walletId` out of the arrays of `walletIds`

> rename the method to use list prefix and pass an array

can you provide an example if you think my syntax is not correct? (it seems to be working at least)

src/services/mongoose/wallet-on-chain-addresses.ts
nicolasburtey: I don't understand this comment?

test/integration/app/multiple-wallet-account.spec.ts
```diff
@@ -0,0 +1,55 @@
+import { addWallet, getAccount, setDefaultWalletId } from "@app/accounts"
```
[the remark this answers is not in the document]
nicolasburtey: are we using honeycomb with the tests?

nicolasburtey:
> I'm wondering if it would be simpler to separate `Account` collection _before_ `Wallet` collection for a less risky migration when we get to that point, since `Wallet.accountId` will simply point to the `Account.id` property in the already created `Account` collection.

Don't think that's a good idea. doing the migration for this will not be a big issue (just the same type of migration I have been running for last migration/this one).

But more importantly, we don;t know how the User/Account relationship will look like using Ory. and until we do, I'm taking the "postpone to later" approach on this decision so until we have the full context (but still continue to work on the multi currency)

src/domain/wallets/index.types.d.ts
vindard: What happens if we have 2 wallets with the same `accountId` value for e.g.? Which `Wallet` gets returned?

src/services/mongoose/wallet-on-chain-addresses.ts
vindard: He means single object arg

```ts
const persistNew = async ({
  id,
  onChainAddress,
}: {
  id: WalletId
  onChainAddress: OnChainAddressIdentifier
}): Promise<OnChainAddressIdentifier | RepositoryError> => {
```

src/app/accounts/index.ts
```diff
@@ -15,6 +15,8 @@ export * from "./username-available"
 export * from "./get-contact-by-username"
 export * from "./update-contact-alias"
 export * from "./add-new-contact"
+export * from "./add-wallet"
```
vindard, line 18: How are we planning to use the `src/app/wallets` dir? (I'm wondering if this file should be there or here)

vindard (commented):
Just one question aside from Juan's question about the `findByAccountId` question.

Only thing I haven't tested as yet is the migration.

src/services/mongoose/accounts.ts
vindard: Worked ok for me on testing. Looks like Mongoose is able to use `walletIds: walletId` to search for the `walletId` inside an array and return the first object it finds that has an array with the target `walletId`.

src/domain/wallets/index.types.d.ts
nicolasburtey: ohhh. should be `listByAccountId` and return a list of Wallets. you're right. thanks for pointing this out!

src/services/mongoose/accounts.ts
nicolasburtey: yup that is also my understanding @vindard

src/domain/wallets/index.types.d.ts
nicolasburtey: actually the function was not used. I have removed it.

src/app/accounts/index.ts
nicolasburtey: we could argue this could go to wallet as well. not sure if there is a good way to think about it.

nicolasburtey: maybe a good counter argument is that because this method return wallet it should be under wallet.

bodymindarts: The problem here is that we have a circular dependency because of the `walletIds` in account. A better way would be to drop that field and have `Wallets.getWalletForAccount` just query the `wallet.accountId` (with an appropriate index).
I don't see a reason to keep the field on account - then this can clearly be part of the wallet use cases.

src/app/accounts/set-default-walletid.ts
```diff
@@ -0,0 +1,22 @@
+import { InvalidWalletId } from "@domain/errors"
+import { AccountsRepository } from "@services/mongoose"
+
+export const setDefaultWalletId = async ({
```
bodymindarts, line 4: Where does this get called? Can it work with a persisted wallet?

nicolasburtey: only in tests for now

src/app/accounts/index.ts
nicolasburtey: > The problem here is that we have a circular dependency because of the walletIds in account. A better way would be to drop that field and have Wallets.getWalletForAccount just query the wallet.accountId (with an appropriate index).

the circular dependency has been removed. 

maybe @dolcalmi and @vindard could have another look.

note for @vindard: in many of the payments function, we're passing around `WalletId`. I think we could save a lot of back and forth to mongodb is we were passing a `Wallet`. some optimization to consider in the future.

src/app/accounts/add-wallet.ts
```diff
@@ -0,0 +1,30 @@
+import { AccountsRepository, WalletsRepository } from "@services/mongoose"
+
+export const addWallet = async ({
```
bodymindarts, line 3: `addWallet` is no longer needed.
Since there is no update to Account (as setting `defaultWalletId` happens elsewhere) it collapses to:

```
export const addWallet = async (
{  accountId,
   type,
   currency,
 }: {
   accountId: AccountId
   type: WalletType
   currency: WalletCurrency
 }): Promise<Wallet | ApplicationError> => {
  return await WalletsRepository().persistNew({
     accountId,
     type,
     currency,
   })
}
  
   ```
   
   In other use cases `WalletRepository` should be called directly - there is no code saved by having this wrapper.
   The only time to introduce this trivial wrapper is to expose the use case to GQL (which isn't until later).

**Comment by bodymindarts** [comment]
Other than the changes [here](https://github.com/GaloyMoney/galoy/pull/954) and removing / collapsing `add-wallet` I have no further comments and would approve.

**Comment by nicolasburtey** [comment]
> Other than the changes [here](https://github.com/GaloyMoney/galoy/pull/954) and removing / collapsing `add-wallet` I have no further comments and would approve.

I feel this added PR make this PR bigger, and a lot of the changes are unrelated to the current PR. 

I have merged it anyhow because the code is done, but I'm not sure this is a good practice? I have received multiple feedbacks on prior PR with the insight that my PR were too big, and this was because I was fixing side things when working on a PR, and this seems to be the same things happening here?

File: src/app/accounts/add-wallet.ts (hunk 15 above)
Comment (nicolasburtey) [reply src/app/accounts/add-wallet.ts:3]: there is an additional safety check on whether `accountId` is valid. I've also start build with it with this PR https://github.com/GaloyMoney/galoy/pull/935 so removing it to add it back in the next PR is not very relevant IMO

**Comment by bodymindarts** [comment]
Ah yes I didn’t consider that side constraint. 

You are right that my refactoring bloats the PR. Perhaps I should have done it against main. OTOH now we have a 1 approval policy so with us collaborating on the branch I didn’t consider other reviewers. Still… next time

**Review by bodymindarts — DISMISSED** [review]

**Comment by nicolasburtey** [comment]
> Ah yes I didn’t consider that side constraint. You are right that my refactoring bloats the PR. Perhaps I should have done it against main. OTOH now we have a 1 approval policy so with us collaborating on the branch I didn’t consider other reviewers. Still… next time

I think that's the type of PR that may need more eyes on the code (for catching eventual issue but also for the awareness of how the code is moving)

File: src/domain/ledger/activity-checker.ts [hunk 16]
```diff
@@ -0,0 +1,38 @@
+import { toSats } from "@domain/bitcoin"
+
+const MS_PER_HOUR = (60 * 60 * 1000) as MilliSeconds
+const MS_PER_DAY = (24 * MS_PER_HOUR) as MilliSeconds
+const MS_PER_30_DAYS = (30 * MS_PER_DAY) as MilliSeconds
+
+export const ActivityChecker = ({
```
vindard, line 7: I would either:
- rename this to `MonthlyActivityChecker`, or

- pass the `monthlyVolumeThreshold` argument into a sub-function called `above30DayThreshold`/`aboveMonthlyThreshold` here instead of into the top-level function
 
just to make this concept a bit clearer.

src/app/accounts/add-wallet.ts
```diff
@@ -0,0 +1,23 @@
...
+  if (account instanceof Error) return account
+
+  const wallet = await WalletsRepository().persistNew({
+    accountId,
+    type,
+    currency,
+  })
+  if (wallet instanceof Error) return wallet
+
+  return wallet
```
vindard, line 22: I could be tempted to remove the `account = ...` line in a future update if I didn't understand that we'd like to check that the account exists first. This could either be a good place for a code comment, or we could alternatively structure code like below.

We should also add a test calling this function with an `accountId` that doesn't exist.

```suggestion
  return !(account instanceof Error)
    ? WalletsRepository().persistNew({
        accountId,
        type,
        currency,
      })
    : account
```

src/app/accounts/update-default-walletid.ts
```diff
@@ -0,0 +1,26 @@
...
+  if (walletIds instanceof Error) return walletIds
+
+  if (!walletIds.some((wid) => wid === walletId)) return new InvalidWalletId()
+
+  account.defaultWalletId = walletId
+
+  const result = await AccountsRepository().update(account)
+  if (result instanceof Error) return result
+
+  return result
```
vindard, line 25:
```suggestion
  return AccountsRepository().update(account)
```

src/services/mongoose/schema.types.d.ts
```diff
@@ -67,7 +74,8 @@ interface UserType {
   created_at: string
   onchain: OnChainObjectForUser[]
   twoFA: TwoFAForUser
-  walletId: WalletId
+  walletIds: WalletId[]
```
vindard, line 77: Probably don't need this `walletIds` anymore

vindard (changes requested):
Left a few optional comments. Only required change is to maybe add a test for `addWallet` that passes in an invalid `accountId`.

nicolasburtey:
> Left a few optional comments. Only required change is to maybe add a test for `addWallet` that passes in an invalid `accountId`.

I'm wondering if this check should be from the addWallet, or directly in the repository. 

This is really a db consistency so maybe that make sense to be part of the wallet repo instead of add wallet?

bodymindarts:
> > Left a few optional comments. Only required change is to maybe add a test for `addWallet` that passes in an invalid `accountId`.
> 
> I'm wondering if this check should be from the addWallet, or directly in the repository.
> 
> This is really a db consistency so maybe that make sense to be part of the wallet repo instead of add wallet?

I like this idea. The wallet repository could do The check in the persist new function. We have 1 repo calling another one in that case but I think this is a good situation for it.

src/app/accounts/add-wallet.ts
nicolasburtey: moved it to the persistNew in WalletsRepository

vindard:
> Thinking about it more it’s really stupid that we have to take care of it in application code at all. It should be a foreign key constraint on the Schema. Can mongoose do that? Any relational db can

It looks like you can in a creative way: https://egorpanok.com/2018-11-11-how-to-emulate-a-foreign-key-constraint-check-on-mongo-with-mongoose/

vindard (dismissed):
Approving because my comments have been resolved, and it looks like most others as well.

The only outstanding comment I can see is the contraint on the Schema mentioned [here](https://github.com/GaloyMoney/galoy/pull/921#issuecomment-1013934179), but this can probably happen in another PR if we think this is ok.

nicolasburtey:
> > Thinking about it more it’s really stupid that we have to take care of it in application code at all. It should be a foreign key constraint on the Schema. Can mongoose do that? Any relational db can
> 
> It looks like you can in a creative way: https://egorpanok.com/2018-11-11-how-to-emulate-a-foreign-key-constraint-check-on-mongo-with-mongoose/

interesting solution

if we don't use the Wallet mongoose object directly but goes through the Wallet repository, the last commit achieve the same effect

bodymindarts (approved):

PR #923, opened 2022-01-11
refactor: update language

dolcalmi:
- Move update language from core to app layer
- Add integration and unit tests

### DIFF
--- a/test/integration/app/users/update-language.spec.ts
+++ b/test/integration/app/users/update-language.spec.ts
@@ -1,0 +1,38 @@
+import { Users } from "@app"
+import { InvalidLanguageError } from "@domain/errors"
+
+import { getUserIdByTestUserIndex } from "test/helpers"
+
+let userId0: UserId
+
+beforeAll(async () => {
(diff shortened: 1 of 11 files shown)


test/integration/app/users/update-language.spec.ts
```diff
@@ -0,0 +1,38 @@
+import { Users } from "@app"
+import { InvalidLanguageError } from "@domain/errors"
+
+import { getUserIdByTestUserIndex } from "test/helpers"
+
+let userId0: UserId
+
+beforeAll(async () => {
+  userId0 = await getUserIdByTestUserIndex(0)
```
nicolasburtey, line 9: maybe we can create the user also before?

I'm trying for every test I'm working on to add this at the beginning of the test, so that we could eventually whip out the db between every tests for more consistency, for the tests that don't depend on global consistency with lnd/bitcoin

nicolasburtey (approved):
just a suggestion about tests, but otherwise PR looks good to me

PR #926, opened 2022-01-11
feat: dedicated migration image

bodymindarts:
For our next db migration we would like to have at least some level of
automation. Here we are adding an image to execute the migrations and
test-run them as part of the `integration-in-ci` make task.

Next step will be to add a job to the chart that can be used to execute the migration using this image.

Note I have used
```
FROM nodejs:16-alpine
```
for the image instead of a distroless one because `mongodb-migrate` depends on `/bin/sh` which isn't present in distroless. I haven't investigated wether or not there is a good alternative to alpine for this.

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -14,4 +14,4 @@ jobs:
     - uses: actions/setup-node@v2
     - run: yarn install
     - name: Run integration tests
-      run: docker compose up -d && make integration-in-ci
+      run: docker compose build && docker compose up -d && make integration-in-ci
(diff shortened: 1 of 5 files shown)


dolcalmi:
:thinking: we need a way to run only the migrations added after initial deploy (for every new customer) or... we need a different migration package/structure

bodymindarts:
This is not an issue - migration will run on initial standup when DB is empty so it will be a no-op.

nicolasburtey:
> This is not an issue - migration will run on initial standup when DB is empty so it will be a no-op.

it's not a no-op because migrate-mongo will actually update the `changelog` collection as the result of this run.

nicolasburtey (approved):

PR #928, opened 2022-01-11
chore: improve function naming

nicolasburtey:


src/app/accounts/send-balance-to-accounts.ts
```diff
@@ -3,7 +3,7 @@ import { getCurrentPrice } from "@app/prices"
 import { getBalanceForWalletId } from "@app/wallets"
 import { NotificationsService } from "@services/notifications"
 
-export const sendBalanceToAccounts = async (logger: Logger) => {
+export const sendDefaultWalletBalanceToUsers = async (logger: Logger) => {
```
bodymindarts, line 6: File name needs updating

bodymindarts (approved):

PR #929, opened 2022-01-11
chore: improve tracing for cron

bodymindarts:
I think there is something unexpected going on with the rebalancing. This PR should help get better trace data to investigate.


src/servers/cron.ts
```diff
@@ -17,43 +18,62 @@ import { SpecterWallet } from "@core/specter-wallet"
...
+          },
+          async () => {
+            try {
+              baseLogger.info(`starting ${task.name}`)
+              await task()
+              results.push(true)
+            } catch (error) {
+              baseLogger.error({ error }, `issue with task ${task.name}`)
+              results.push(false)
+              return error
```
vindard, line 68: Do we need this return if the containing `await asyncRunInSpan` isn't being assigned?

vindard (approved):
Just one little nit

src/servers/cron.ts
bodymindarts: Its so that `asyncRunInSpan` will auto add the error via `span.recordException`

PR #933, opened 2022-01-12
chore: Update mobile versions

daviroo:
This will prompt users on ios who haven't yet upgraded to the latest version of the app.

image


krtk6160 (approved):

nicolasburtey (changes requested):
default.yaml should not be updated on galoy but from the deployments side

ie: we don't want the config change of a single instance of Galoy to be published on this repository

PR #935, opened 2022-01-12
feat: invoice with fiat denomination

nicolasburtey:


src/domain/fiat/index.ts
```diff
@@ -0,0 +1,25 @@
...
+export const toFiat = (amount: number): FiatAmount => {
+  return amount as FiatAmount
+}
+
+export const checkedToFiat = (amount: number): FiatAmount | ValidationError => {
+  if (!(amount && amount > 0)) return new InvalidFiatAmount()
+  return toFiat(amount)
+}
+
+export const satsToFiatOptionPricing = ({
```
bodymindarts, line 13: This is going to be a service call. Should be represented in the domain as an `IPricingService` or something like that.

src/domain/fiat/index.types.d.ts
```diff
@@ -0,0 +1 @@
+type FiatAmount = number & { readonly brand: unique symbol }
```
bodymindarts, line 1: I have doubts about `domain/fiat` as a useful grouping. `fiat` isn't an entity or aggregate in the system which is typically the boundary for which behaviour is defined.
We do have the `bitcoin/{onchain,offchain}` modules which define a bunch of primitives for handling bitcoin stuff - perhaps it is, or will become something similar. But until it becomes more than a single primitive type I would probably group it somewhere else.

src/domain/wallet-invoices/index.types.d.ts
```diff
@@ -3,13 +3,20 @@ type WalletInvoice = {
...
+  readonly fiat: FiatAmount | null
+  readonly currency: WalletCurrency
   paid: boolean
 }
 
 type WalletInvoiceValidator = {
   validateToSend(fromWalletId: WalletId): true | ApplicationError
 }
 
+type WalletFactoryArgs = {
```
bodymindarts, line 15: This should be `WalletInvoiceFactoryConfig` I think.
We haven't made it explicit yet but I have been following the pattern `Config` for constructors and `Args` for functions on types.

src/domain/wallet-invoices/wallet-invoice-factory.types.d.ts
```diff
@@ -1,8 +1,12 @@
 type WalletInvoiceFactory = {
-  create(registeredInvoice: RegisteredInvoice): WalletInvoice
-  createForRecipient(registeredInvoice: RegisteredInvoice): WalletInvoice
+  createForSelf(
+    registeredInvoice: RegisteredInvoice,
+  ): (fiat: FiatAmount | null) => WalletInvoice
```
bodymindarts, line 4: From scanning the code I have not seen why the currying is necessary. All calls I can find are of the form:
```
  const walletInvoice = walletInvoiceCreateFn(registeredInvoice)(fiat)
```
While I think currying can be very useful I don't understand the need for it here.

src/migrations/20220122121713-add-default-currency-walletinvoices.ts
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  async up(db) {
+    console.log("Begin up migration")
+    const result = await db
+      .collection("invoiceusers")
```
bodymindarts, line 5: Should we use this as an opportunity to rename the collection inline with the domain types to `walletinvoices`?

src/services/mongoose/wallet-invoices.ts
```diff
@@ -144,5 +148,7 @@ const walletInvoiceFromRaw = (result): WalletInvoice => ({
   walletId: result.walletId as WalletId,
   selfGenerated: result.selfGenerated,
   pubkey: result.pubkey as Pubkey,
-  paid: result.paid,
+  paid: result.paid as boolean,
+  fiat: result.fiat as FiatAmount,
```
bodymindarts, line 152: This is supposed to be the fiat denomination correct? What about the display amount - are we going to persist that as well?

EG for Bitcoin Jungle we may have USD invoices that should still be displayed in colon.

test/integration/app/wallets/invoice.spec.ts
```diff
@@ -20,45 +24,205 @@ import {
...
+
+describe("Wallet - addInvoice Fiat", () => {
+  it("add a self generated fiat invoice", async () => {
+    const price = await getCurrentPrice()
+    if (price instanceof Error) return price
+
+    const fiatInput = 100
+    const sats = fiatInput / price
+
+    const lnInvoice = await Wallets.addInvoiceByWalletId({
```
bodymindarts, line 142: I'm pretty sure this should be a different use case or? `addFiatInvoiceByWalletId`. Or is it the same mutation that creates both fiat and etc invoices?

src/app/wallets/add-invoice-for-wallet.ts
```diff
@@ -55,58 +72,157 @@ export const addInvoiceForRecipient = async ({
   memo = "",
   descriptionHash,
 }: AddInvoiceForRecipientArgs): Promise<LnInvoice | ApplicationError> => {
-  const recipientWalletIdChecked = checkedToWalletId(recipientWalletId)
-  if (recipientWalletIdChecked instanceof Error) return recipientWalletIdChecked
+  const walletIdChecked = checkedToWalletId(recipientWalletId)
```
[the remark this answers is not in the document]
nicolasburtey: so:
1 - the context of recipient is useful mainly when there is also a sender, which is not the case inside this function. 
2 - shorter name making function one liner instead of many lines, which help for the readability

src/domain/fiat/index.types.d.ts
nicolasburtey: we'll have some logic in here shortly with the dealer. I have been thinking of naming it dealer instead of fiat otherwise, any thoughts on this?

test/integration/app/wallets/invoice.spec.ts
nicolasburtey: TDB. 

but that might be make sense to have 2 endpoints.

nicolasburtey: actually not sure. the currency is embedded in the wallet, so there is no need to pass this independently. 

the client should call `addInvoice({walletId, amount})` and depending of the currency of the wallet, the amount would either be fiat or sats. 

this logic on knowing whether to know if the wallet should be a fiat wallet or btc wallet should be done on the client

src/domain/wallet-invoices/index.types.d.ts
nicolasburtey: not sure this is what I have seen elsewhere. @vindard what do you think?

src/domain/wallet-invoices/wallet-invoice-factory.types.d.ts
nicolasburtey: > From scanning the code I have not seen why the currying is necessary

it's because the `price` is not been fetched initially in the top level use case function (`addInvoiceByWalletId` or `addInvoiceNoAmountByWalletId`). we're doing it an inner function. agree this may be not the most readable code

nicolasburtey: the main things is there are 3 different variables:
- with/without amount
- fiat or btc
- forSelf for forRecipient

we could write the function 8 times but the current code/currying reflect the refactoring of those multi variables inputs.

src/domain/wallet-invoices/index.types.d.ts
vindard: Juan and I have been back and forth on this in the past as well. I think framing the top level function args as `..Config` and the inner function args as `...Args` makes sense and helps to separate the abstractions better actually 🤔

src/domain/wallet-invoices/wallet-invoice-factory.types.d.ts
dolcalmi: @nicolasburtey will this be updated in regards to what we talk about in daily?

src/domain/wallet-invoices/index.types.d.ts
dolcalmi: yep should have Config suffix

src/domain/fiat/index.ts
dolcalmi: it should be DealerService, right?

dolcalmi (changes requested):
looks good according to the roadmap, just a couple of changes/comments

src/domain/fiat/index.ts
nicolasburtey: yes fiat and dealer are really the same "thing" in the context of the galoy backend

src/domain/wallet-invoices/wallet-invoice-factory.types.d.ts
nicolasburtey: yup: 347996c5c74db94e104286538314b716d1106110

src/services/mongoose/schema.ts
```diff
@@ -25,13 +25,26 @@ const dbMetadataSchema = new Schema({
...
+      validator: function (v) {
+        return v.match(WalletIdRegex)
+      },
+    },
+  },
 
   // usd equivalent. sats is attached in the invoice directly.
   // optional, as BTC wallet doesn't have to set a sat amount when creating the invoice
-  usd: Number,
+  fiat: Number,
```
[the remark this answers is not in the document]
nicolasburtey: schema should have short name I think (they're been repeated with a document db), but I'll name it fiatAmount at the domain level.

test/unit/domain/ledger/activity-checker.spec.ts
```diff
@@ -38,7 +40,7 @@ describe("ActivityChecker", () => {
       getVolumeFn,
     })
     expect(
-      checker.aboveThreshold(["walletId" as WalletId, "walletId2" as WalletId]),
+      checker.aboveThreshold([uuidv4() as WalletId, "walletId2" as WalletId]),
```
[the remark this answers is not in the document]
nicolasburtey: no type check here actually, so reverting my change.

src/services/mongodb/index.ts
```diff
@@ -60,7 +60,7 @@ export const setupMongoConnection = async (syncIndexes = false) => {
     if (syncIndexes) {
       await User.syncIndexes()
       await Transaction.syncIndexes()
-      await InvoiceUser.syncIndexes()
+      await WalletInvoice.syncIndexes()
```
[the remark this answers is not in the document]
nicolasburtey: I'm not sure to get your point?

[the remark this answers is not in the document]
dolcalmi: nevermind, we only query by id(hash)

dolcalmi (dismissed):

dolcalmi:
the only missing change was `satsToFiatOptionPricing` but will be removed/updated in the next PRs

bodymindarts (approved):

PR #937, opened 2022-01-12
fix: only paid should be mutable

nicolasburtey:
I don't think that's a good idea to expose the other field on the update. they should be immutable.

### DIFF
--- a/src/domain/wallet-invoices/index.types.d.ts
+++ b/src/domain/wallet-invoices/index.types.d.ts
@@ -13,7 +13,7 @@ type WalletInvoiceValidator = {
 interface IWalletInvoicesRepository {
   persistNew: (invoice: WalletInvoice) => Promise<WalletInvoice | RepositoryError>
 
-  update: (invoice: WalletInvoice) => Promise<WalletInvoice | RepositoryError>
+  markAsPaid: (paymentHash: PaymentHash) => Promise<WalletInvoice | RepositoryError>
 
   findByPaymentHash: (
     paymentHash: PaymentHash,
--- a/src/services/mongoose/wallet-invoices.ts
+++ b/src/services/mongoose/wallet-invoices.ts
@@ -15,45 +15,34 @@ export const WalletInvoicesRepository = (): IWalletInvoicesRepository => {
     paid,
   }: WalletInvoice): Promise<WalletInvoice | RepositoryError> => {
     try {
-      await new InvoiceUser({
+      const invoiceUser = await new InvoiceUser({
         _id: paymentHash,
         walletId,
         selfGenerated,
         pubkey,
         paid,
       }).save()
-      return {
-        paymentHash,
-        walletId,
-        selfGenerated,
-        pubkey,
-        paid,
-      } as WalletInvoice
+      return walletInvoiceFromRaw(invoiceUser)
     } catch (err) {
       return new UnknownRepositoryError(err)
     }
   }
 
-  const update = async ({
-    paymentHash,
-    walletId,
-    selfGenerated,
-    pubkey,
-    paid,
-  }: WalletInvoice): Promise<WalletInvoice | RepositoryError> => {
+  const markAsPaid = async (
+    paymentHash: PaymentHash,
+  ): Promise<WalletInvoice | RepositoryError> => {
     try {
-      const data = { walletId, selfGenerated, pubkey, paid }
-      const doc = await InvoiceUser.updateOne({ _id: paymentHash }, { $set: data })
-      if (doc.nModified !== 1) {
+      const invoiceUser = await InvoiceUser.findOneAndUpdate(
+        { _id: paymentHash },
+        { paid: true },
+        {
+          new: true,
+        },
+      )
+      if (!invoiceUser) {
         return new RepositoryError("Couldn't update invoice for payment hash")
       }
-      return {
-        paymentHash,
-        walletId,
-        selfGenerated,
-        pubkey,
-        paid,
-      }
+      return walletInvoiceFromRaw(invoiceUser)
     } catch (err) {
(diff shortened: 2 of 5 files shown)


dolcalmi:
We have discussed this type of implementations before. As far as I remember we did it initially this way but our repo model require the whole object. cc @bodymindarts

src/domain/wallet-invoices/index.types.d.ts
```diff
@@ -13,7 +13,13 @@ type WalletInvoiceValidator = {
 interface IWalletInvoicesRepository {
   persistNew: (invoice: WalletInvoice) => Promise<WalletInvoice | RepositoryError>
 
-  update: (invoice: WalletInvoice) => Promise<WalletInvoice | RepositoryError>
```
bodymindarts, line 16: This change is redundant - field mutability is enforced on the type level https://github.com/GaloyMoney/galoy/blob/main/src/domain/wallet-invoices/index.types.d.ts#L1-L7

test/integration/services/mongoose/wallet-invoices.spec.ts
```diff
@@ -48,6 +49,7 @@ describe("WalletInvoices", () => {
     const invoiceToUpdate = persistResult as WalletInvoice
     invoiceToUpdate.paid = true
     const updatedResult = await repo.update(invoiceToUpdate)
+    console.log({ updatedResult })
```
bodymindarts, line 52: Is this accidentally committed debug output?

src/services/mongoose/wallet-invoices.ts
```diff
@@ -161,3 +139,11 @@ export const WalletInvoicesRepository = (): IWalletInvoicesRepository => {
     deleteUnpaidOlderThan,
   }
 }
+
+const walletInvoiceFromRaw = (result): WalletInvoice => ({
+  paymentHash: result.id as PaymentHash,
```
bodymindarts, line 144: I think this should be `_id` right? The schema doesn't define an `id` field:
```
const invoiceUserSchema = new Schema({
  _id: String, // hash of invoice
  walletId: String,

  // usd equivalent. sats is attached in the invoice directly.
  // optional, as BTC wallet doesn't have to set a sat amount when creating the invoice
  usd: Number,

  timestamp: {
    type: Date,
    default: Date.now,
  },

  selfGenerated: {
    type: Boolean,
    default: true,
  },

  pubkey: {
    type: String,
    require: true,
  },

  paid: {
    type: Boolean,
    default: false,
  },
})

```
But there is some inconsistency going on as `_id` should be reserved for mongo `ObjectId`s but here its a `string`... hmm

src/services/mongoose/wallet-invoices.ts
```diff
@@ -91,14 +75,8 @@ export const WalletInvoicesRepository = (): IWalletInvoicesRepository => {
...
-    for await (const invoice of pending) {
-      yield {
-        paymentHash: invoice.id as PaymentHash,
-        walletId: invoice.walletId as WalletId,
-        selfGenerated: invoice.selfGenerated,
-        pubkey: invoice.pubkey as Pubkey,
-        paid: invoice.paid,
-      } as WalletInvoice
+    for await (const invoiceUser of pending) {
+      yield walletInvoiceFromRaw(invoiceUser)
```
bodymindarts, line 79: Perhaps we can converge to `...FromRecord` for these translation methods.

src/services/mongoose/wallet-invoices.ts
```diff
@@ -161,3 +139,11 @@ export const WalletInvoicesRepository = (): IWalletInvoicesRepository => {
     deleteUnpaidOlderThan,
   }
 }
+
+const walletInvoiceFromRaw = (result): WalletInvoice => ({
```
bodymindarts, line 143: Need to define `InvoiceUserRecord` and type this correctly.

bodymindarts (changes requested):

src/services/mongoose/wallet-invoices.ts
nicolasburtey: > I think this should be `_id` right? The schema doesn't define an `id` field:
> 
> ```
> const invoiceUserSchema = new Schema({
>   _id: String, // hash of invoice
>   walletId: String,
> 
>   // usd equivalent. sats is attached in the invoice directly.
>   // optional, as BTC wallet doesn't have to set a sat amount when creating the invoice
>   usd: Number,
> 
>   timestamp: {
>     type: Date,
>     default: Date.now,
>   },
> 
>   selfGenerated: {
>     type: Boolean,
>     default: true,
>   },
> 
>   pubkey: {
>     type: String,
>     require: true,
>   },
> 
>   paid: {
>     type: Boolean,
>     default: false,
>   },
> })
> ```
> 
> But there is some inconsistency going on as `_id` should be reserved for mongo `ObjectId`s but here its a `string`... hmm

it shouldn't be `_id` if we consider `_id` are `ObjectId`

bodymindarts: agree - it should be `id` if anything but that would require a migration or? My point is as the code currently stands:
```
paymentHash: result.id as PaymentHash,
```
shouldn't work because the schema doesn't define `id` so I guess its using mongooses virtual field.

src/services/mongoose/wallet-invoices.ts
nicolasburtey: should be done, but not really the goal of this PR

src/services/mongoose/wallet-invoices.ts
nicolasburtey: should be done, but not really the goal of this PR

bodymindarts (approved):

PR #939, opened 2022-01-13
feat: Add reimbursement script

krtk6160:


nicolasburtey (approved):

bodymindarts (approved):

src/app/wallets/intraledger-send-payment.ts
```diff
@@ -55,6 +55,9 @@ export const intraledgerPaymentSendWalletId = async ({
 
   if (senderWalletId === recipientWalletId) return new SelfPaymentError()
 
+  const recipientWallet = await WalletsRepository().findById(recipientWalletId)
```
dolcalmi, line 58: where is this value is used?

krtk6160: Only being used below to check if the wallet exists or not

PR #940, opened 2022-01-13
chore: we have moved on using arrow functions

nicolasburtey:
settings this as a eslint-rule


vindard (approved):
This looks reasonable to me

PR #949, opened 2022-01-14
refactor: bootstrap default wallet in createUser

bodymindarts:
When removing bootstrap related code from `addWallet` it collapses simply to a `WalletsRepository.persistNew` so I have deleted that use case completely.

Also renamed `setDefaultWalletId` to `updateDefaultWalletId` as it can be changed more than 1 time.

### DIFF
--- a/src/app/users/create-user.ts
+++ b/src/app/users/create-user.ts
@@ -1,8 +1,11 @@
-import { addWallet } from "@app/accounts"
 import { checkedToPhoneNumber } from "@domain/users"
 import { WalletCurrency, WalletType } from "@domain/wallets"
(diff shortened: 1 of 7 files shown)


src/app/users/create-user.ts
```diff
@@ -34,12 +37,17 @@ export const createUser = async ({
   const account = await AccountsRepository().findByUserId(user.id)
   if (account instanceof Error) return account
 
-  const wallet = await addWallet({
+  const wallet = await WalletsRepository().persistNew({
```
nicolasburtey, line 40: the `addWallet` still exists I guess?

nicolasburtey (approved):

PR #950, opened 2022-01-14
fix: query for number of businesses

nicolasburtey:


bodymindarts (approved):

src/servers/exporter.ts
```diff
@@ -131,7 +131,7 @@ const main = async () => {
       walletRoles[role].set(balance)
     }
 
-    business_g.set(await User.count({ title: { $exists: true } }))
+    business_g.set(await User.count({ title: { $ne: null } }))
```
dolcalmi, line 134: needs to have exists and $ne

dolcalmi (changes requested):

dolcalmi (approved):

src/servers/exporter.ts
nicolasburtey: I don't think that's necessary. the query I did with `{ title: { $ne: null } }` seemed to return the right number

PR #954, opened 2022-01-14
refactor: unit test business logic around activity

bodymindarts:
There was a lot of business logic in the `accounts/active-accounts` use case that I extracted into an `ActivityChecker` in `domain/ledger` and included a unit test. The integration test you had wasn't actually checking the logic but - as an integration test - only if everything is wired together correctly. If the types are correct there is no need for this kind of integration test. Typechecker + unit test is preferable.

Also moved `listWalletsByAccountId` under app as it is accessed from GraphQL layer - so it is a top level use case. Also it has a dependency on the `WalletsRepository` which should not happen from `@domain`. `@domain` only depends on the interfaces.


src/domain/ledger/activity-checker.ts
```diff
@@ -0,0 +1,38 @@
+import { toSats } from "@domain/bitcoin"
+
+const MS_PER_HOUR = (60 * 60 * 1000) as MilliSeconds
+const MS_PER_DAY = (24 * MS_PER_HOUR) as MilliSeconds
```
nicolasburtey, line 4: this is duplicated from app.ts. I'll dedupe it.

nicolasburtey (approved):

PR #959, opened 2022-01-16
refactor: remove cron main span

dolcalmi:
- Remove cron main span
- Remove return on catch
- Move rebalance as last task


nicolasburtey (approved):
LGTM

PR #962, opened 2022-01-16
refactor: - some error throw fix

nicolasburtey:
- update docker-compose
- fix some test

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,6 +11,7 @@ services:
     - lnd2
     - lnd-outside-1
     - lnd-outside-2
+    - otel-agent
   otel-agent:
     image: otel/opentelemetry-collector-contrib:0.37.1
     command: ["--config=/etc/otel-agent-config.yaml", "--log-level=debug"]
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -11,6 +11,7 @@ services:
     - lnd2
     - lnd-outside-1
     - lnd-outside-2
+    - otel-agent
```
bodymindarts, line 14: I don't think `otel-agent` is a dependency of integration tests is it?

bodymindarts (approved):
Approving but think `otel-agent` was mistakenly added.

docker-compose.yml
nicolasburtey: it was added on purpose to get telemetry when doing `make integration`

nicolasburtey: won't do anything unless the env var are being set. but maybe that would be useful to get this in the CI server to see the difference between the tests over time. like if we introduce regression we would could it?

nicolasburtey: like this trace is interesting/intriguing: https://ui.honeycomb.io/galoy/datasets/nicolas-dev/result/oi9y9hjhAqF?tab=raw the consistent 3s (timeout?) could highlight a bug

PR #779, opened 2021-12-03
chore: update all scripts for docker compose v2

bodymindarts:
Due to high demand for compatibility with docker compose v2 I think its time we upgraded our scripts.

PR #964, opened 2022-01-16
chore: Updated Dev Documentation In DEV.md And Node Engine Minimum Version

kelvinator07:
This PR fixes 2 things
1 - A typo in the dev documentation for verifying the docker compose version 
2 - The node engine in package.json to allow to run the project on node version greater than or equal to v14. (Base Version Specified On The Dev Documentation)

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -14,7 +14,7 @@ $ jq --version
 jq-1.6
 $ docker --version
 Docker version 20.10.7, build f0df350
-$ docker compose version
+$ docker-compose --version
 Docker Compose version 2.0.0
 ```
 
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "build-docs": "npx spectaql spectaql-config.yml -1"
   },
   "engines": {
-    "node": "16"
+    "node": ">=14"
   },
   "dependencies": {
     "@google-cloud/storage": "^5.8.5",


DEV.md
```diff
@@ -14,7 +14,7 @@ $ jq --version
 jq-1.6
 $ docker --version
 Docker version 20.10.7, build f0df350
-$ docker compose version
+$ docker-compose --version
```
krtk6160, line 17: The original command works fine for me:
image

kelvinator07: Doesn't Work For Me.. Thats Strange
Screenshot 2022-01-17 at 10 40 34

bodymindarts:
I don't think we should relax the node requirements. 14 is no longer under LTS. Instead we should tighten the dev requirements. https://github.com/GaloyMoney/galoy/pull/967

DEV.md
krtk6160: Are you running docker-compose v2?

vindard: Needs docker-compose v2, this was a deliberate change we made some months back when we upgraded docker-compose requirements from v1 to v2 (see PR #779)

kelvinator07:
> I don't think we should relax the node requirements. 14 will not be LTS much longer. Instead we should tighten the dev requirements. #967

Sounds Better, You Might Also Want To Update The Node Engine in Package.json to >=16

DEV.md
kelvinator07: I'm Still On v1, Will Upgrade To v2

kelvinator07:
I Will Be Closing This PR Since All Issues Has Been Resolved. Thanks

PR #974, opened 2022-01-17
feat: make checkSelfWalletIdRateLimits at account level

nicolasburtey:
this will allow for better ddos protection, in case someone wanted to create too many wallets

### DIFF
--- a/src/app/wallets/add-invoice-for-wallet.ts
+++ b/src/app/wallets/add-invoice-for-wallet.ts
@@ -18,23 +18,23 @@ export const addInvoiceByWalletId = async ({
   if (wallet instanceof Error) return wallet
 
   return addInvoice({
-    walletId: wallet.id,
+    wallet,
     amount,
     memo,
   })
 }
 
 export const addInvoice = async ({
-  walletId,
+  wallet,
   amount,
   memo = "",
 }: AddInvoiceArgs): Promise<LnInvoice | ApplicationError> => {
-  const limitOk = await checkSelfWalletIdRateLimits(walletId)
+  const limitOk = await checkSelfWalletIdRateLimits(wallet.accountId)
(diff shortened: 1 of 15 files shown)


src/app/wallets/add-invoice-for-wallet.ts
```diff
@@ -97,7 +107,10 @@ export const addInvoiceNoAmountForRecipient = async ({
   const recipientWalletIdChecked = checkedToWalletId(recipientWalletId)
   if (recipientWalletIdChecked instanceof Error) return recipientWalletIdChecked
 
-  const limitOk = await checkRecipientWalletIdRateLimits(recipientWalletIdChecked)
+  const wallet = await getWallet(recipientWalletIdChecked)
```
bodymindarts, line 110: Instead of calling another use case can you just call the repository directly:
```
  const wallets = WalletsRepository().findById(walletId)
```
No need for the indirection

src/graphql/main/schema.graphql
```diff
@@ -2,6 +2,7 @@ interface Account {
   csvTransactions(walletIds: [WalletId!]!): String!
   defaultWalletId: WalletId!
   id: ID!
+  walletIds: [WalletId!]!
```
bodymindarts, line 5: Why is this being exposed?
Shouldn't the way to query this be:
```
wallets {
  id
}
```

bodymindarts (changes requested):
just 2 things
- use `WalletsRepository().findById(walletId)` instead of the use case `getWallet` 
- remove accidentally(?) exposed `walletIds` from schema

bodymindarts (approved):
Looking over this PR I noticed we're exporting some functions that are not being used anywhere so I created a follow up PR to remove them https://github.com/GaloyMoney/galoy/pull/988

PR #975, opened 2022-01-17
fix: typesafe conversion to/from ObjectId

bodymindarts:


src/services/mongoose/utils.ts
```diff
@@ -0,0 +1,9 @@
+import { Types } from "mongoose"
+
+export const toObjectId = <T extends string>(id: T): Types.ObjectId => {
```
nicolasburtey, line 3: there is not much benefits to do `<T extends string>` here IMO. would be better to do `export const toObjectId = (id: AccountId | UserId):`?

nicolasburtey (approved):

dolcalmi (approved):

PR #982, opened 2022-01-18
fix: WalletInvoice.update should not create new records

bodymindarts:
due to concern that we may inadvertently circumvent the `readonly` type assertions by creating a new object that we pass to the `WalletInvoiceRepository` we should also make sure that it is a true `update` instead of an `upsert`.

### DIFF
--- a/src/services/mongoose/wallet-invoices.ts
+++ b/src/services/mongoose/wallet-invoices.ts
@@ -31,15 +31,10 @@ export const WalletInvoicesRepository = (): IWalletInvoicesRepository => {
   const update = async ({
     paymentHash,
     paid,
-  }: {
-    paymentHash: PaymentHash
-    paid: boolean
-  }): Promise<WalletInvoice | RepositoryError> => {
+  }: WalletInvoice): Promise<WalletInvoice | RepositoryError> => {
     try {
       const data = { paid }
-      const invoiceUser = await InvoiceUser.findOneAndUpdate({ _id: paymentHash }, data, {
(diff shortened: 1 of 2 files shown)


src/services/mongoose/wallet-invoices.ts
```diff
@@ -31,15 +31,10 @@ export const WalletInvoicesRepository = (): IWalletInvoicesRepository => {
...
+  }: WalletInvoice): Promise<WalletInvoice | RepositoryError> => {
     try {
       const data = { paid }
-      const invoiceUser = await InvoiceUser.findOneAndUpdate({ _id: paymentHash }, data, {
-        new: true,
-      })
+      const invoiceUser = await InvoiceUser.findOneAndUpdate({ _id: paymentHash }, data)
       if (!invoiceUser) {
         return new RepositoryError("Couldn't update invoice for payment hash")
       }
```
nicolasburtey, line 40: we need the new here. otherwise the data returned back the line below will be the old record

nicolasburtey (changes requested):
`new` doesn't mean it's a new record, it means whether the old data or updated data (new) should be sent back from mongodb

src/services/mongoose/wallet-invoices.ts
dolcalmi: https://mongoosejs.com/docs/tutorials/findoneandupdate.html#getting-started 
```
You should set the new option to true to return the document after update was applied.
```

PR #985, opened 2022-01-18
docs: add docs for testing migrations

bodymindarts:


Makefile
```diff
@@ -40,6 +40,11 @@ reset-deps: clean-deps start-deps
 
 test: unit integration
 
+test-migrate:
+	docker compose down -v
```
nicolasburtey, line 44: not sure about this one. I think the best way to test is a migrate is to run the test on the commit before the changes, then checkout the current branch, then try the migration.

running migration on an empty db will most frequently not be very useful.

bodymindarts: This PR is just about moving from _nothing_ documented to _something_ documented - I agree there could be better forms of testing migrations and for critical ones this is not enough. But running them against an empty DB is a starting point that has some use.

dolcalmi (approved):

nicolasburtey (approved):

PR #987, opened 2022-01-19
feat: Add fee update script

krtk6160:


src/app/accounts/update-account-withdraw-fee.ts
```diff
@@ -0,0 +1,18 @@
...
+}: {
+  id: AccountId
+  fee: number
+}): Promise<Account | ApplicationError> => {
+  const accountsRepo = AccountsRepository()
+
+  const account = await accountsRepo.findById(id)
+  if (account instanceof Error) return account
+
+  account.withdrawFee = fee as WithdrawFee
```
bodymindarts, line 15: This should be validated IMO. `checkedToWithdrawFee`

src/debug/update-fee.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const updateFee = async (operations: Array<feeUpdateOperation>) => {
+  await setupMongoConnection()
+  console.log("Mongoose connection ready")
+  const walletRepo = WalletsRepository()
+  for (const { walletId, fee } of operations) {
+    const wallet = await walletRepo.findById(walletId)
+    if (wallet instanceof RepositoryError) {
+      console.error(`Could not fetch wallet for walletId: ${walletId}`, { wallet })
+      break
```
nicolasburtey, line 25: we could continue instead?

src/debug/update-fee.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+    console.log(`Fee for walletId: ${walletId} set to ${fee}`)
+  }
+  return mongoose.connection.close()
+}
+
+// Populate the array below with the wallet ids and fees to be set
+
+const feeUpdateOperations: Array<feeUpdateOperation> = [
+  { walletId: "" as WalletId, fee: 0 as Satoshis },
```
nicolasburtey, line 43: I think we should pass a json instead of hardcoding this

src/debug/update-fee.ts
```diff
@@ -0,0 +1,52 @@
...
+]
+
+// For example:
+// feeUpdateOperations = [
+//   { walletId: "70767cc2-9cab-4bab-b78a-5e6b2d947163" as WalletId, fee: 13 as Satoshis },
+//   { walletId: "70767cc2-9cab-4bab-b78a-5e6b2d947163" as WalletId, fee: -1 as Satoshis },
+// ]
+
+updateFee(feeUpdateOperations).then(console.log).catch(console.error)
```
nicolasburtey, line 52: use a "if __main__" here (not sure about the exact syntax, but there will example across our repo), that will be a cleaner.

krtk6160: Are you suggesting to rename the `updateFee` function to main.

Something like this? ->
image

src/debug/update-fee.ts
krtk6160: Yep, fixed

src/debug/update-fee.ts
krtk6160: You mean read a file off disk?

src/app/accounts/update-account-withdraw-fee.ts
```diff
@@ -0,0 +1,22 @@
...
+  id,
+  fee,
+}: {
+  id: AccountId
+  fee: number
+}): Promise<Account | ApplicationError> => {
+  const checkedFee = checkedToWithdrawFee(fee)
+  if (checkedFee instanceof Error) return checkedFee
+
+  const accountsRepo = AccountsRepository()
```
bodymindarts, line 14:
```
const account = await AccountsRepository().findById(id)
```
=> can delete 2 lines.

src/core/types.d.ts
```diff
@@ -20,6 +20,11 @@ type FeeRates = {
   withdrawFeeFixed: number
 }
 
+type WithdrawFeeRange = {
```
bodymindarts, line 23: Don't add any code to `src/core` we are deleting all that code. It should be defined close to the `checkedToXxx` is defined.

src/debug/fee-update-operations.json
```diff
@@ -0,0 +1,6 @@
+{
+  "feeUpdateOperations": [
+    { "walletId": "11d2a204-9981-420f-a148-020b7d52607f", "fee": 13 },
```
bodymindarts, line 3: These aren't real walletIds right?

src/domain/accounts/index.ts
```diff
@@ -65,3 +68,11 @@ export const checkedToContactAlias = (alias: string): ContactAlias | ValidationE
   }
   return alias as ContactAlias
 }
+
+export const checkedToWithdrawFee = (fee: number): WithdrawFee | ValidationError => {
+  const { min, max } = getWithdrawFeeRange()
```
bodymindarts, line 73: `domain` layer should be pure functions - inject the config.

bodymindarts (changes requested):

src/debug/fee-update-operations.json
krtk6160: They're valid but are from local env

src/app/accounts/update-account-withdraw-fee.ts
krtk6160: We're using accountsRepo in two places. One is here, and one is a few lines below for updating the account.
If we use `await AccountsRepository().findById(id)` then the update call with have to be changed to `return AccountsRepository().update(account)` which would mean two initializations of `AccountsRepository`.

Is this what you're suggesting?

src/debug/update-fee.ts
nicolasburtey: > Or are you referring to the method of determining whether a module is being run as the main one or via an import?

this one

src/debug/fee-update-operations.json
nicolasburtey: I guess they're not valid then, because they're been created randomly

src/app/accounts/update-account-withdraw-fee.ts
bodymindarts: No I missed that its being used twice.

src/debug/fee-update-operations.json
krtk6160: The term "valid" would imply that the value has passed the validation checks, which is true for this value.
The validation check being a regex.

src/domain/accounts/index.ts
krtk6160: I checked to see if we were importing config in domain prior to this change, and we are in `domain/rate-limit` and `domain/wallets`. Is that planned for removal?

src/core/types.d.ts
krtk6160: Fixed

src/debug/update-fee.ts
krtk6160: > > Or are you referring to the method of determining whether a module is being run as the main one or via an import?
> 
> this one

I can't find it in our codebase. Why do you think we need it?

src/domain/accounts/index.ts
```diff
@@ -4,8 +4,11 @@ import {
   InvalidAccountStatusError,
   InvalidUsername,
   InvalidContactAlias,
+  InvalidWithdrawFeeError,
 } from "@domain/errors"
 
+import { getWithdrawFeeRange } from "@config/app"
```
bodymindarts, line 10: is this being used?

src/domain/wallets/index.types.d.ts
```diff
@@ -169,3 +169,8 @@ type WithdrawalFeeCalculator = {
   onChainWithdrawalFee({ onChainFee, walletFee }: OnChainWithdrawalFeeArgs): Satoshis
   onChainIntraLedgerFee(): Satoshis
 }
+
+type WithdrawFeeRange = {
+  min: number
```
bodymindarts, line 174: These should be type `Satoshis` or?

krtk6160: Fixed

src/domain/accounts/index.ts
krtk6160: Oops, removed

src/graphql/error-map.ts
```diff
@@ -94,6 +94,10 @@ export const mapError = (error: ApplicationError): CustomApolloError => {
       message = "An amount is required to complete payment"
       return new ValidationInternalError({ message, logger: baseLogger })
 
+    case "InvalidWithdrawFeeError":
```
bodymindarts, line 97: This should never be exposed to the enduser. If we get here the app is mis-configured. Returning the `ValidationInternalError` is notifying the caller that they have made a mistake. This is a bad state and should probably be part of the `unhandled` errors further down.

krtk6160: Fixed

bodymindarts (approved):

PR #990, opened 2022-01-19
refactor: simplify updateLnPayments use-case method

vindard:
## Description

Builds on #848

This PR simplifies the logic for the `updateLnPayments` method by getting rid of some of the complexity around parallelizing across multiple lnd instances (see [this thread](https://github.com/GaloyMoney/galoy/pull/848#discussion_r786639296)).


nicolasburtey (approved):
haven't dive into the algo (I'll guess you'll double check this works on staging, and we can check in prod if that's useful) but that looks much simpler/cleaner, as seen by the diff.

PR #992, opened 2022-01-20
fix: api v1

nicolasburtey:


bodymindarts (changes requested):
Typo in your find and replace

vindard:
Addressed in #993

PR #994, opened 2022-01-21
fix: continue-token error handling in updateLnPayment

vindard:
## Description

_Builds on #848_

This PR introduces changes to break the 'while' loop whenever an inconsistent pagination token comes back from lnd in the `updateLnPayments` method. This inconsistency can happen in two places here, and in its current state we don't handle this inconsistency well and can potentially end up with an infinite loop if `lnd` misbehaves.

I found this when testing some inconsistencies in the `lightning` package (see https://github.com/alexbosworth/lightning/pull/55).

### DIFF
--- a/src/app/lightning/update-ln-payments.ts
+++ b/src/app/lightning/update-ln-payments.ts
@@ -19,8 +19,11 @@ export const updateLnPayments = async (): Promise<true | ApplicationError> => {
   for (const key of pubkeys) {
     const pubkey = key as Pubkey
 
-    let after: PagingStartToken | PagingContinueToken = undefined
-    while (processedLnPaymentsHashes.length < incompleteLnPayments.length) {
+    let after: PagingStartToken | PagingContinueToken | PagingStopToken = undefined
+    while (
+      processedLnPaymentsHashes.length < incompleteLnPayments.length &&
+      after !== false
+    ) {
       // Fetch from Lightning service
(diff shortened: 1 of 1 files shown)


src/app/lightning/update-ln-payments.ts
```diff
@@ -19,8 +19,11 @@ export const updateLnPayments = async (): Promise<true | ApplicationError> => {
   for (const key of pubkeys) {
     const pubkey = key as Pubkey
 
-    let after: PagingStartToken | PagingContinueToken = undefined
-    while (processedLnPaymentsHashes.length < incompleteLnPayments.length) {
+    let after: PagingStartToken | PagingContinueToken | PagingStopToken = undefined
+    while (
+      processedLnPaymentsHashes.length < incompleteLnPayments.length &&
+      after !== false
```
dolcalmi, line 25: is this not the same of using only `after`

dolcalmi (changes requested):

src/app/lightning/update-ln-payments.ts
vindard: `undefined` is a valid value for `after`, represents a paging start token. `false` is specifically used as the stop paging value (Alex does it as well in his [ln-accounting package](https://github.com/alexbosworth/ln-accounting/blob/master/records/get_all_payments.js#L115))

dolcalmi (approved):

PR #998, opened 2022-01-21
refactor: payment input validation

bodymindarts:
This PR introduces a (unit-tested) `PaymentInputValidator` to take care of some common validation that must happen for all payment use cases.

Its currently only introduced into OnChainPay (as it is the simplest place to introduce it) but will be introduced in the other use cases in subsequent PRs.

### DIFF
--- a/src/app/wallets/pay-on-chain.ts
+++ b/src/app/wallets/pay-on-chain.ts
@@ -98,6 +98,7 @@ export const payOnChainByWalletId = async ({
     address,
     checkedAmount,
     memo,
+    sendAll,
   })
   const checkedAddress = checkedToOnChainAddress({
     network: BTC_NETWORK,
--- a/src/app/wallets/pay-on-chain.ts
+++ b/src/app/wallets/pay-on-chain.ts
@@ -31,17 +31,17 @@ const { dustThreshold } = getOnChainWalletConfig()
 export const payOnChainByWalletIdWithTwoFA = async ({
   senderAccount,
   senderWalletId,
-  amount,
+  amount: amountRaw,
   address,
   targetConfirmations,
   memo,
   sendAll,
   twoFAToken,
 }: PayOnChainByWalletIdWithTwoFAArgs): Promise<PaymentSendStatus | ApplicationError> => {
-  const checkedAmount = sendAll
+  const amount = sendAll
     ? await LedgerService().getWalletBalance(senderWalletId)
-    : checkedToSats(amount)
-  if (checkedAmount instanceof Error) return checkedAmount
+    : checkedToSats(amountRaw)
+  if (amount instanceof Error) return amount
 
   const user = await getUser(senderAccount.ownerId)
   if (user instanceof Error) return user
@@ -50,7 +50,7 @@ export const payOnChainByWalletIdWithTwoFA = async ({
   const twoFACheck = twoFA?.secret
     ? await checkAndVerifyTwoFA({
         walletId: senderWalletId,
-        amount: checkedAmount,
+        amount,
         twoFASecret: twoFA.secret,
         twoFAToken,
       })
@@ -71,7 +71,7 @@ export const payOnChainByWalletIdWithTwoFA = async ({
 export const payOnChainByWalletId = async ({
   senderAccount,
   senderWalletId,
-  amount,
+  amount: amountRaw,
   address,
   targetConfirmations,
   memo,
@@ -79,7 +79,7 @@ export const payOnChainByWalletId = async ({
 }: PayOnChainByWalletIdArgs): Promise<PaymentSendStatus | ApplicationError> => {
   const checkedAmount = sendAll
     ? await LedgerService().getWalletBalance(senderWalletId)
-    : checkedToSats(amount)
+    : checkedToSats(amountRaw)
   if (checkedAmount instanceof Error) return checkedAmount
 
   const validator = PaymentInputValidator(WalletsRepository().findById)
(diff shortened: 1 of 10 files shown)


src/domain/wallets/payment-input-validator.ts
```diff
@@ -0,0 +1,50 @@
+import { checkedToSats } from "@domain/bitcoin"
+import {
+  SelfPaymentError,
+  InvalidWalletId,
+  InvalidAccountStatusError,
+} from "@domain/errors"
+import { AccountStatus } from "@domain/accounts"
+
+export const PaymentInputValidator = (
+  getWalletFn: PaymentInputValidatorConfig,
```
nicolasburtey, line 10: why passing a function here as the argument? does it have to do with boundaries or unit testing?

test/unit/domain/wallets/payment-input-validator.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+    },
+    contacts: [],
+  }
+
+  it("returns the correct types when everything is valid", async () => {
+    const getWalletFn: PaymentInputValidatorConfig = (walletId: WalletId) => {
+      const wallet = {
+        senderWalletId: dummySenderWallet,
+        recipientWalletId: dummyRecipientWallet,
+      }[walletId]
```
nicolasburtey, line 44: I don't understand what this function is doing. I understand that getWalletFn take a `WalletId` and return a `Wallet`. I'm not sure why there is a notion of `senderWalletId` and `recipientWalletId` here (the wording would apply we're getting 2 wallets and applying it in a context of a payment)

src/app/wallets/pay-on-chain.ts
```diff
@@ -75,19 +73,34 @@ export const payOnChainByWalletIdWithTwoFA = async ({
...
+    senderWalletId,
+    recipientWalletId: null,
+  })
+  if (validationResult instanceof Error) return validationResult
+
   const onchainLogger = baseLogger.child({
     topic: "payment",
     protocol: "onchain",
     transactionType: "payment",
     address,
```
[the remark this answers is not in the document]
bodymindarts: Out of scope for this PR, I just moved this logic - but can do it in follow up.

src/domain/wallets/payment-input-validator.ts
bodymindarts: Domain layer should does not depend on service layer. Domain layer is made up of pure functional components. App layer is where we depend on external services (ie. not purely functional) and compose the domain layer primitives to use cases.

The more business logic we can extract into domain layer the better. Because they are pure functions we can unit test them which is simpler, faster and more isolated than the currently dominating integration test style we have for most of the functionality.

So it has to do with both boundaries and testing.

Boundaries because this is business logic that we would prefer to have in domain layer.
Testing because extracting into domain layer makes it a pure function.

src/domain/wallets/payment-input-validator.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    if (senderAccount.status !== AccountStatus.Active) {
+      return new InvalidAccountStatusError()
+    }
+    const senderWallet = await getWalletFn(senderWalletId)
+    if (senderWallet instanceof Error) return senderWallet
+
+    if (senderWallet.accountId !== senderAccount.id) return new InvalidWalletId()
+
+    if (recipientWalletId != null) {
```
dolcalmi, line 29:
```suggestion
    if (recipientWalletId) {
```

test/unit/domain/wallets/payment-input-validator.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+    const result = await validator.validatePaymentInput({
+      amount: 2,
+      senderWalletId: "senderWalletId",
+      senderAccount: dummyAccount,
+      recipientWalletId: "recipientWalletId",
+    })
+    if (result instanceof Error) throw result
+    const { amount, senderWallet } = result
+    expect(amount).toBe(2)
+    expect(senderWallet).toBe(dummySenderWallet)
```
dolcalmi, line 59: please use `expect.objectContaining` to validate all returned attributes

test/unit/domain/wallets/payment-input-validator.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    const validator: PaymentInputValidator = PaymentInputValidator(getWalletFn)
+    const result = await validator.validatePaymentInput({
+      amount: -1,
+      senderWalletId: "senderWalletId",
+      senderAccount: dummyAccount,
+      recipientWalletId: "recipientWalletId",
+    })
+    expect(result instanceof Error).toBe(true)
```
dolcalmi, line 79:
```suggestion
    expect(result).toBeInstanceOf(Error)
```

please change in other tests and please use specific error instance

test/unit/domain/wallets/payment-input-validator.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+    const validator: PaymentInputValidator = PaymentInputValidator(getWalletFn)
+    const result = await validator.validatePaymentInput({
+      amount: 2,
+      senderWalletId: "senderWalletId",
+      senderAccount: dummyAccount,
+      recipientWalletId: null,
+    })
+    if (result instanceof Error) throw result
+    const { recipientWallet } = result
+    expect(recipientWallet).toBe(null)
```
dolcalmi, line 144: same here, please validate result, not only `recipientWallet`

dolcalmi (changes requested):
Still pending to test with graphql

src/app/wallets/pay-on-chain.ts
```diff
@@ -33,24 +33,21 @@ import { getOnChainFeeByWalletId } from "./get-on-chain-fee"
 const { dustThreshold } = getOnChainWalletConfig()
 
 export const payOnChainByWalletIdWithTwoFA = async ({
+  senderAccount,
```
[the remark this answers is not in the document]
bodymindarts: Yes this is exactly the rule I'm using here.

dolcalmi (approved):

PR #999, opened 2022-01-22
refactor: make fees more explicit

nicolasburtey:
refactor: split the ledger in several sub module

also, rename the current with USD when it's intended to be a WalletCurrency (ie: store of value), and DisplayCurrency for what could be converted in any currency

a lot of code is moving around but no change of functionality


src/domain/wallets/index.types.d.ts
```diff
@@ -166,11 +166,11 @@ type DepositFeeCalculator = {
 }
 
 type OnChainWithdrawalFeeArgs = {
-  onChainFee: Satoshis
-  walletFee: Satoshis
+  minerFee: Satoshis
+  bankFee: Satoshis
```
vindard, line 170: Can be `dealerFee`... or are we also introducing a new `Bank` concept?

nicolasburtey: It was the bankFee elsewhere already. it's not a new concept. it's just the fee the bankowner charges for withdrawal on top of the miner fees.

vindard:
I committed the relevant draft changes from this morning here as a reference: https://github.com/GaloyMoney/galoy/commit/358359770e9465389f408be9d16321f7d96eb2ab
- New args type for TxReceive
- Accounting for dealer-involved entries

src/app/wallets/get-on-chain-fee.ts
```diff
@@ -38,49 +52,22 @@ export const getOnChainFee = async ({
...
   // avoids lnd balance sniffing attack
-  if (balance < amount) return new InsufficientBalanceError("Balance is too low")
+  if (balance < amountChecked) return new InsufficientBalanceError("Balance is too low")
 
   const onChainService = OnChainService(TxDecoder(BTC_NETWORK))
   if (onChainService instanceof Error) return onChainService
 
-  const onChainFee = await onChainService.getOnChainFeeEstimate({
-    amount,
+  const minerFee = await onChainService.getOnChainFeeEstimate({
```
dolcalmi, line 60: maybe `networkFee` should be a more generic name (it works for lightning too)

src/app/wallets/pay-on-chain.ts
```diff
@@ -310,37 +312,38 @@ const executePaymentViaOnChain = async ({
...
-          ? withdrawalFeeCalculator.onChainWithdrawalFee({
-              onChainFee: onChainTxFee,
-              walletFee: toSats(senderAccount.withdrawFee),
-            })
-          : estimatedFee
-        const sats = toSats(amountToSend + fee)
-        const usd = sats * usdPerSat
-        const usdFee = fee * usdPerSat
+        const sats = toSats(amountToSend + totalFee)
+        const usdDisplay = (sats * usdPerSat) as FiatAmount
```
dolcalmi, line 333: this need a helper similar to `toSats`, please avoid direct cast in use cases

src/app/wallets/update-pending-invoices.ts
```diff
@@ -105,79 +105,76 @@ const updatePendingInvoice = async ({
...
-
-      const notificationsService = NotificationsService(logger)
-      notificationsService.lnInvoicePaid({
-        paymentHash,
-        recipientWalletId: walletId,
-        amount: roundedDownReceived,
-        usdPerSat,
-      })
+    const usdDisplay = (roundedDownReceived * usdPerSat) as FiatAmount
+    const usdFeeLightningLiquidity = (feeLightningLiquidity * usdPerSat) as FiatAmount // TODO: toFiatFeeDisplay()
```
dolcalmi, line 154: we should standardize names. currently have two: bankFee (always in wallet currency) and networkFee (always sats), `usdFeeLightningLiquidity` is really confusing.

src/domain/wallets/deposit-fee-calculator.ts
```diff
@@ -7,6 +7,6 @@ export const DepositFeeCalculator = (): DepositFeeCalculator => {
 
   return {
     onChainDepositFee,
-    lnDepositFee: () => toSats(0),
+    lnDepositFee: () => toSats(0), // TODO: implement
```
dolcalmi, line 10: we also need to update tests, if you change for example to `1` tests dont pass

src/app/wallets/get-on-chain-fee.ts
```diff
@@ -1,35 +1,49 @@
...
 export const getOnChainFee = async ({
-  wallet,
+  walletId,
+  withdrawFee,
   amount,
   address,
   targetConfirmations,
 }: GetOnChainFeeArgs): Promise<Satoshis | ApplicationError> => {
-  const withdrawalFeeCalculator = WithdrawalFeeCalculator()
+  const amountChecked = checkedToSats(amount)
```
bodymindarts, line 23: I don't like that we are naming the 'correct' type `amountChecked` and the raw types `amount` it should be the other way around. Correct types should be the norm and have the simpler name `amount` - unchecked are the special case and should have the special name `amountRaw`.


This is a general observation - not just relating to this PR.

src/app/wallets/get-on-chain-fee.ts
```diff
@@ -1,35 +1,49 @@
...
+  if (amountChecked instanceof Error) return amountChecked
+
+  const targetConfsChecked = checkedToTargetConfs(targetConfirmations)
+  if (targetConfsChecked instanceof Error) return targetConfsChecked
+
+  const walletIdChecked = checkedToWalletId(walletId)
+  if (walletIdChecked instanceof Error) return walletIdChecked
+
+  const wallets = WalletsRepository()
```
bodymindarts, line 32: duplication of line 37 `const walletsRepo = WalletsRepository()`

src/app/wallets/pay-on-chain.ts
```diff
@@ -310,37 +312,38 @@ const executePaymentViaOnChain = async ({
...
-        if (onChainTxFee instanceof Error) {
-          logger.fatal({ err: onChainTxFee }, "impossible to get fee for onchain payment")
-          onChainTxFee = toSats(0)
+        if (minerFee instanceof Error) {
+          logger.error({ err: minerFee }, "impossible to get fee for onchain payment")
+          totalFee = estimatedFee
+        } else {
+          totalFee = withdrawFeeCalculator.onChainWithdrawalFee({
+            minerFee,
+            bankFee: toSats(senderAccount.withdrawFee),
```
bodymindarts, line 328: why is `senderAccount.withdrawFee` not already typed correctly?

src/graphql/root/query/on-chain-tx-fee-query.ts
```diff
@@ -15,15 +15,16 @@ const OnChainTxFeeQuery = GT.Field({
...
     const { walletId, address, amount, targetConfirmations } = args
 
     for (const input of [walletId, address, amount, targetConfirmations]) {
       if (input instanceof Error) throw input
     }
 
-    const fee = await Wallets.getOnChainFeeByWalletId({
+    const fee = await Wallets.getOnChainFee({
       walletId,
+      withdrawFee: domainAccount.withdrawFee,
```
bodymindarts, line 27: I think this is an example of where we can pass in the fully formed `domainAccount` into the use case. Then the use case can also validate that the `wallet` has the correct `accountId`.

src/app/wallets/get-on-chain-fee.ts
nicolasburtey: that is a good point. this would also bring shorter variable name within the function, which would avoid some line return/make the code more compact.

src/app/wallets/pay-on-chain.ts
nicolasburtey: we discuss yesterday whether fees should have their own types or not. we had not reached a conclusion.

src/graphql/root/query/on-chain-tx-fee-query.ts
nicolasburtey: this is actually what I have done initially, then I realize we only need the one value from the `Account` and this is the `withdrawFee`, so why having this use case depend on the whole `Account` when we just need a single value?

I have not thought about the validation, but almost by definition, it's been fetched from the `sessionContext`, so do we need validation?

bodymindarts: I think the "dummer" the resolvers the better. Knowing that an Account has a `withdrawFee` field is already more coupling than necessary.
Having an account in the context is a performance optimisation - it should just be passed along. The use cases / domain is where internal knowledge of the entities should be located.
Consider if we evolve the way we are representing fees. They could be dynamic (see liquidity leacher) for example. At that point a change in the domain would require a change to this resolver. But we know that fees will remain coupled to an account - so if we just pass the account then the change is only located to domain code (as it should be).

Finally we do need to validate that the user has access to the wallet. We consider the account valid when it comes from the context but we need to check that the wallet is actually one connected to the account.

src/app/wallets/pay-on-chain.ts
nicolasburtey: I have removed the fee specific types

bodymindarts (approved):
Can't find anymore serious red flags... will probably be iterated on again anyhow when introducing the `display` / `settlement` / `satoshi` amounts as we paired with on @vindard machine the other day.

src/app/wallets/get-on-chain-fee.ts
nicolasburtey: that's a great idea

nicolasburtey: I'm not too sure this should be networkFee at the end, at least at this level. The reason being that routingFee and minerFee are different. One we always know how much we will pay in advance (minerFee), the other this is not always known (always know if we probe first succesfully)

src/app/wallets/pay-on-chain.ts
nicolasburtey: fixed

src/domain/wallets/index.types.d.ts
nicolasburtey: I think we could introduce the notion of `"Internal"` and `"external"` fees. the first one would be for fee the bankowner charges, the seconds are fees paid to the network (miner fees or routing fees for instance). 

it would make it more clear how much fees we are charging.

because for lightning for instance now we have inboundLiquidityFees (not yet implemented). those fees will be internal, and I think it would be more clear if it's named `InternalInboundLiquidityFees`, otherwise it may not be obvious to someone looking at the code that is a fee charged by the bankowner only 

(there are actually no fee for inbound on lightning, but this may not be obvious to everybody)

for sending on lightning, then there would be something like "InternalBankFee" and "ExternalRoutingFee"

thoughts @vindard @dolcalmi @bodymindarts  ?

src/services/ledger/index.ts
```diff
@@ -370,23 +369,21 @@ export const LedgerService = (): ILedgerService => {
...
-        currency,
+        currency: WalletCurrency.Btc,
       }
 
-      fiat // TODO handle the case where currency = USD
-
       const entry = MainBook.entry(description)
       entry
-        .credit(liabilitiesWalletId, sats - feeLightningLiquidity, metadata)
+        .credit(liabilitiesWalletId, sats - usdFeeInboundLiquidity, metadata)
```
bodymindarts, line 381: This shouldn't type check (don't know how to enforce). Subtracting `Satoshi - FiatAmount` seems like its clearly an error.

nicolasburtey: seems because they are both number that is working 😒

maybe an option would be to have a .add() to those objects (for everything related to Satoshis/Fiat), and by convention not use the + - / any more.

bodymindarts: I think we should have some generic calculation functions:
```
const add = <T extends number> (a: T, b:T) => {
  return a + b
}
const sub = <T extends number> (a: T, b:T) => {
  return a - b
}
```
 etc.
 
see https://stackoverflow.com/questions/54470329/how-to-make-a-generic-add-operator-in-typescript-that-works-with-numbers-and-str

src/domain/fiat/display-currency.ts
```diff
@@ -0,0 +1,5 @@
+export const toDisplayCurrency =
+  (price: UsdPerSat) =>
```
bodymindarts, line 2: Price here should probably be `DisplayCurrencyConversionRate` or something like that.

src/domain/fiat/index.ts
```diff
@@ -1,30 +1,30 @@
...
+    throw new NonIntegeterUsdAmount(`${amount} type ${typeof amount} is not an integer`)
+  return amount as UsdAmount
 }
 
-export const checkedToFiat = (amount: number): FiatAmount | ValidationError => {
-  if (!(amount && amount > 0)) return new InvalidFiatAmount()
-  if (Number.isInteger(amount))
-    return new NonIntegeterFiatAmount(`${amount} is not an integer`)
-  return toFiat(amount)
+export const checkedtoUsd = (amount: number): UsdAmount | ValidationError => {
```
bodymindarts, line 11: The type `UsdAmount` should probably be `UsdCents` or even `MiliCents` or whatever the lowest denominator is we want to support.

src/domain/fiat/index.ts
```diff
@@ -1,30 +1,30 @@
...
+  if (!(amount && amount > 0)) return new InvalidUsdAmount()
+  if (!Number.isInteger(amount))
+    return new NonIntegeterUsdAmount(`${amount} type ${typeof amount} is not an integer`)
+  return toUsd(amount)
 }
 
-export const satsToFiatOptionPricing = ({
+export const satstoUsdOptionPricing = ({
```
bodymindarts, line 18: `camelCase` `satsto` -> `satsTo`

src/domain/fiat/display-currency.ts
bodymindarts: Also why currying? Cannot find a place in the code where it is needed.

src/domain/fiat/index.types.d.ts
```diff
@@ -1 +1,2 @@
-type FiatAmount = number & { readonly brand: unique symbol }
+type UsdAmount = number & { readonly brand: unique symbol }
```
bodymindarts, line 1: how about `UsdCents`?

src/domain/fiat/index.types.d.ts
```diff
@@ -1 +1,2 @@
...
+type DisplayCurrencyAmount = number & { readonly brand: unique symbol }
```
bodymindarts, line 2: How about `DisplayCurrencyBaseAmount`

src/domain/ledger/fee-reimbursement.ts
```diff
@@ -1,9 +1,15 @@
...
 export const FeeReimbursement = (prepaidFee: Satoshis): FeeReimbursement => {
-  const getReimbursement = ({ actualFee }: { actualFee: Satoshis }): Satoshis | null => {
+  const getReimbursement = ({
+    actualFee,
+  }: {
+    actualFee: Satoshis
+  }): Satoshis | FeeDifferenceError => {
     const feeDifference = toSats(prepaidFee - actualFee)
-    if (feeDifference < 0 || feeDifference > prepaidFee) return null
+    if (feeDifference < 0 || feeDifference > prepaidFee) return new FeeDifferenceError()
```
bodymindarts, line 12: This is not an error - it is expected behaviour that there may be no fee reimbursement.
As we don't want to use `null` for absence anymore perhaps we should use `undefined` or `0` - but I don't think it should be an error.

src/domain/ledger/index.types.d.ts
```diff
@@ -80,23 +81,23 @@ type OnChainTxArgs = TxArgs & {
 
 type AddLnTxReceiveArgs = LnTxArgs & {
   currency: WalletCurrency
-  fiat: FiatAmount | undefined
-  usdFeeInboundLiquidity: FiatAmount
+  usd: UsdAmount | undefined
```
bodymindarts, line 84: Isn't this only needed for USD wallets?
Callers of this function already know (via the WalletCurrency) wether or not we are receiving in USD or BTC - so we should also segregate those call paths to the ledger - also because there is different logic (dealer entries) on the ledger side.

ie. instead of `use: UsdAmount | undefined` there should be 
`AddLnTxReceiveArgs` and `AddLnTxReceiveInUsdArgs` or something like that.

src/domain/fiat/display-currency.ts
nicolasburtey: the idea is that the price will be a "config" of this (to be expanded) function. this is why it's curried.

src/domain/ledger/fee-reimbursement.ts
nicolasburtey: fee should not be negative or more than the prepaidFee. this is an error; it would mean something had gone wrong somewhere. 

the only non error is if fee = 0, which mean there is no reimbursement.

src/domain/ledger/index.types.d.ts
nicolasburtey: should be done in another PR when we introduce the dealer?

vindard (commented):
Still reviewing, one comment so far

src/services/ledger/intraledger.ts
```diff
@@ -0,0 +1,162 @@
+import { LedgerTransactionType, toLiabilitiesWalletId } from "@domain/ledger"
+import { LedgerError, UnknownLedgerError } from "@domain/ledger/errors"
+
+import { MainBook } from "./books"
+
+import { translateToLedgerJournal } from "."
+
+export const intraledger = {
+  addOnChainIntraledgerTxSend: async ({
```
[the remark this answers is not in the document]
nicolasburtey: I could use `Transfer` yes

src/core/user-wallet.ts
```diff
@@ -139,13 +139,13 @@ export class UserWallet {
   }) {
     return {
       fee,
-      feeUsd: fee ? UserWallet.satsToUsd(fee) : undefined,
+      feeUsd: fee ? UserWallet.satstoCents(fee) : undefined,
       sats,
-      usd: usd ?? UserWallet.satsToUsd(sats),
+      usd: usd ?? UserWallet.satstoCents(sats),
```
bodymindarts, line 144: camelcase

src/domain/errors.ts
```diff
@@ -35,7 +35,8 @@ export class RewardAlreadyPresentError extends DomainError {}
 export class ValidationError extends DomainError {}
 export class ContactNotExistentError extends DomainError {}
 export class InvalidSatoshiAmount extends ValidationError {}
-export class InvalidFiatAmount extends ValidationError {}
+export class InvalidUsdCents extends ValidationError {}
+export class NonIntegeterUsdCents extends ValidationError {}
```
bodymindarts, line 39: Typo `Integer`

src/domain/wallets/index.types.d.ts
bodymindarts: I like the distinction of `Internal`/`External` fees as a naming convention.

src/servers/trigger.ts
```diff
@@ -64,7 +64,9 @@ export async function onchainTransactionEventHandler(
       "payment completed",
     )
 
-    const walletId = await LedgerService().getWalletIdByTransactionHash(tx.id)
+    // FIXME a tx.id should return a walletId[] instead, in case
+    // multiple UXTO goes to the wallet within the same tx
+    const walletId = await LedgerService().getWalletIdByTransactionHash(txHash)
```
bodymindarts, line 69: we shouldn't be using the txHash as the id for onchain transactions - rather txOutPoint.

bodymindarts (dismissed):

src/servers/trigger.ts
nicolasburtey: txOutPoint would mean there is an index? we don't have this here (maybe we should), we're really only storing the txHash. btw we are still storing addresses in an array. Is that the intention? I thought this was removed.

nicolasburtey:
> I like the distinction of Internal/External fees as a naming convention.

I'll work on this naming convention in subsequent PR.

src/servers/trigger.ts
bodymindarts: the addresses are in an array for backwards compatibility but there is never more than 1.

`TxOutPoint` is the way `UTXO`s are referred to in bitcoin. They are a combination of the `txHash` and the index of the specific `UTXO` within the `tx`. 
It is also the bases for the channel ID in lightning (which is the hashed `txOutPoint`).

bodymindarts (approved):

src/servers/trigger.ts
nicolasburtey: my point is there is no index in this string. so it's not a `TxOutPoint` but really a `TxHash`

bodymindarts: Right and my point is our logic is flawed / outdated we should be looping through all `UTXO`s in the tx and calling `LedgerService().getWalletIdByTxOutPoint(utxo)` for each.
Just something to be aware of.

bodymindarts: Or returning `WalletId[]` would also work the way we're doing it now.

PR #1000, opened 2022-01-22
refactor: split config file

nicolasburtey:


src/config/index.ts
```diff
@@ -0,0 +1,49 @@
+export * from "./error"
+export * from "./process"
+export * from "./yaml"
+
+export const MS_PER_HOUR = (60 * 60 * 1000) as MilliSeconds
+export const MS_PER_DAY = (24 * MS_PER_HOUR) as MilliSeconds
+
+export const SECS_PER_2_MINS = (60 * 2) as Seconds
+export const SECS_PER_5_MINS = (60 * 5) as Seconds
+export const SECS_PER_DAY = (60 * 60 * 24) as Seconds
```
bodymindarts, line 10: Some of these are only used in 1 place - like this one is in `src/domain/bitcoin/lightning/invoice-expiration.ts`.
I think its better to have most of these defined locally where they are used because:
- they are not really config (we would never want to _configure_ how many seconds are in a day) - they are constants
- by co-locating them where they are being used we simplify the dependency graph

I actually think that goes for most of the settings in this file.

bodymindarts (approved):
I'm approving because I think this is a step in a good direction.
I think we need to move all the constants out of the `config` folder and define them where they are being used.

bodymindarts:
Unit tests are failing

nicolasburtey:
merging this, as this is a no-op but touching a lot of file so better so merge without waiting to avoid conflits.

PR #1002, opened 2022-01-23
fix: ip update

nicolasburtey:
I think the original if + then condition didn't make sense. it's like with `const result = await usersIp.update(userIP)
` as if we are updating userIp with the exact same value we just fetched.


nicolasburtey:
ok I think the intent was to update the `lastConnection` string. not sure the previous code was actually doing this.

PR #1003, opened 2022-01-23
feat: instrument lock

nicolasburtey:
lock should be within a span as it's a critical component of the backend and it should be easy to follow when a code is within a lock or not


bodymindarts:
Instead of instrumenting just the lock we could generally add instrumentation to the service layer like we go for app layer. Then we don’t need to add individual spans but all the exported functions would be wrapped

PR #1005, opened 2022-01-23
feat: using medici 5

nicolasburtey:


src/services/ledger/schema.ts
```diff
@@ -1,162 +1,178 @@
...
+  feeUsd?: number
+  username?: string
+  pubkey?: string
+}
+
+const transactionSchema = new Schema<ITransaction>(
+  {
+    hash: {
+      type: Schema.Types.String,
+      ref: "InvoiceUser",
```
nicolasburtey, line 47: @dolcalmi is that intended?

PR #1007, opened 2022-01-24
feat: Mitigate Rewards Spammers

leesalminen:
This PR adds a configurable blocklist of ASNs and allowlist of Countries for earning rewards. This is intended to mitigate rewards spammers, who register many accounts to earn sats.


src/app/accounts/add-earn.ts
```diff
@@ -33,6 +39,13 @@ export const addEarn = async ({
   if (validatedPhoneMetadata instanceof Error)
     return new InvalidPhoneMetadataForRewardError(validatedPhoneMetadata.name)
 
+  const userIP = await UsersIpRepository().findById(user.id)
```
nicolasburtey, line 42: the user is already be prefetched during the authentication, and could be passed from graphql layer to the app layer. this would avoid another request to mongodb. you can look for `domainUser` to see some other example pattern.

src/config/yaml.ts
```diff
@@ -181,6 +181,9 @@ export const getBuildVersions = (): {
   }
 }
 
+export const getBlacklistedASNs = (): string => yamlConfig?.blacklistedASNs
```
nicolasburtey, line 184: this should return an `array`, not a `string`?

src/graphql/error-map.ts
```diff
@@ -162,6 +162,10 @@ export const mapError = (error: ApplicationError): CustomApolloError => {
       message = "Unsupported phone carrier for rewards."
       return new ValidationInternalError({ message, logger: baseLogger })
 
+    case "InvalidIpMetadataForRewardError":
+      message = "Unsupported IP address for rewards."
```
nicolasburtey, line 166: I'm wondering if `Invalid` is not better than `Unsupported`? or maybe some other word could be used. `Unsupported` feels off to me in term of vocabulary.

src/domain/users/ip-metadata-validator.ts
```diff
@@ -0,0 +1,33 @@
+import { MissingIpMetadataError, InvalidIpMetadataTypeError } from "@domain/errors"
+import { getBlacklistedASNs, getWhitelistedCountries } from "@config"
+
+export const IpMetadataValidator = (): IpMetadataValidator => {
+  const validate = (userIP: UserIPs): true | ApplicationError => {
```
nicolasburtey, line 5: we often use `validate` in the context of syntax validation. I'm wondering if something like `isApproved` or `approve` would not be more appropriate?

src/app/accounts/add-earn.ts
nicolasburtey: oh nevermind, we don't fetch user here but userIp

nicolasburtey, line 42: I think we use `userIp` instead of `userIP` in the rest of the codebase

leesalminen: Taking a look at `src/services/mongoose/users.ts`, the `lastIPs` property on the `user` object is not returned in the result set when fetching a user from the graphql layer. 

I implemented it this way as I assumed there was a reason for not wanting to fetch `lastIPs` in `Users.getUserForLogin`. 

Just want to confirm with you that you would like me to start returning  the `lastIPs` property on the user object in the graphql layer and pass it down, as opposed to fetching it as needed. 

Happy to implement this way, just wanted to check with you first to make sure.

src/config/yaml.ts
leesalminen: Correct, thank you. Pushed a fix for this to type as `Array<string>`.

src/graphql/error-map.ts
leesalminen: I was mirroring the verbiage used in `InvalidPhoneMetadataForRewardError`, which returns a message of `Unsupported phone carrier for rewards.`. 

IMHO, `Unsupported` is better than `Invalid` for both the phone metadata and IP metadata error. To me, `Invalid` would mean that the IP address / phone number doesn't exist or is incorrectly formatted, while `Unsupported` means that the application doesn't support rewards for that (syntactically correct) IP address / phone number.

Of course, just my opinion and I'm known to be terrible at naming things :). 

I'd be happy to change it, but would ask if you want to change the verbiage for both IP metadata & phone metadata, or just the IP metadata.

src/domain/users/ip-metadata-validator.ts
leesalminen: I do agree here, I think `approve` would be a superior method name to `validate`. In this instance, I was mirroring the naming convention in `PhoneMetadataValidator().validate()`.

Would you like mot change both the `PhoneMetadataValidator()` and `IpMetadataValidator()` to have a method name of `approve`?

src/app/accounts/add-earn.ts
leesalminen: I pulled this from `src/app/users/get-user.ts`'s `updateUserIPsInfo` method, which uses the variable name of `userIP` in the same context.

Grepping around the code base, it looks like `userIP` is being used as variable naming in a caller, while `userIp` is used as the variable name in the callee's signature.

Happy to change, if desired. Just let me know how you would like to proceed.

src/domain/users/index.types.d.ts
```diff
@@ -64,6 +64,10 @@ type PhoneMetadataValidator = {
   validate(phoneMetadata: PhoneMetadata | null): true | ApplicationError
 }
 
+type IpMetadataValidator = {
+  validate(userIP: UserIPs): true | ApplicationError
```
bodymindarts, line 68: `ApplicationError` is the catch all error we return from the `app` use cases and should not be present in the `domain` layer. Domain layer primitives should return specific errors in their signature. I know its not consistent anywhere but since this is adding code lets stick to it.

Having domain return `ApplicationError` creates a circular dependency from app -> domain -> app where we only want it to go app -> domain. You could also fix this in the `PhoneMetadataValidator` in line 64 while you're at this code.

src/domain/users/ip-metadata-validator.ts
bodymindarts: Hmm so we are considering differentiating validation based on value construction vs validation based on value association.

I don't think we have made this differentiation so far - I don't mind but if change the verb `validate` we should also change the noun `Validator`.
As `approve` is a bit opaque (just like `validate`) perhaps a more explicit name like `isIpAllowedToLogIn` would also fix the ambiguity.

src/domain/users/ip-metadata-validator.ts
```diff
@@ -0,0 +1,33 @@
...
+    const blacklistedASNs = getBlacklistedASNs()
+    const whitelistedCountries = getWhitelistedCountries()
```
bodymindarts, line 7: I don't think we should be using loaded terms like `blacklist` / `whitelist`. There are many OSS projects that avoid them and since there is no cost associated with it I would follow this standard.
`blockList`, `denyList` and `allowList` / `acceptList` are more descriptive alternatives.

src/graphql/error-map.ts
bodymindarts: I think a word like `blocked` might also work. `Your phone carrier is blocked from receiving rewards`, `Your IP is blocked from receiving rewards`

bodymindarts (changes requested):

src/app/accounts/add-earn.ts
nicolasburtey: ok I thought we had some consistency on this but it seems we don't, so no need to change it

src/domain/users/ip-metadata-validator.ts
leesalminen: Very good point, I've just pushed a change for this.

src/graphql/error-map.ts
leesalminen: I like the text that @bodymindarts proposes. That seems to be the most descriptive way of telling the user this. Change pushed.

src/domain/users/index.types.d.ts
leesalminen: Understood. I've changed both the `PhoneMetadataValidator` and `IpMetadataValidator` to use `ValidationError` instead of `ApplicationError`.

default.yaml
```diff
@@ -133,6 +133,10 @@ ipRecording:
 
 # blacklistedIPs: ['1.2.3.4', '5.6.7.8']
```
nicolasburtey, line 134: not fully related to this PR, but I think this comment can be removed (and maybe the one above as well?)

we initially had in mind to do some filtering at the app the level, but this should really be done at the nginx/load balancer level

default.yaml
```diff
@@ -133,6 +133,10 @@ ipRecording:
...
 
+blockListedASNs: []
```
nicolasburtey, line 136: could you give some example of the expected syntax here? same for countries. if we use an iso code, would could name it.

src/domain/users/ip-metadata-validator.ts
```diff
@@ -0,0 +1,33 @@
+import { InvalidIpMetadataTypeError } from "@domain/errors"
+import { getBlockListedASNs, getAllowListedCountries } from "@config"
+
+export const IpMetadataValidator = (): IpMetadataValidator => {
+  const validate = (userIP: UserIPs): true | ValidationError => {
+    const blockListedASNs = getBlockListedASNs()
```
nicolasburtey, line 6: if we want this to be more unit testable, we should have the blockListedASNs and allowListedCountries passed as a parameter to the IpMetadataValidator wrapper function.

src/domain/users/ip-metadata-validator.ts
```diff
@@ -0,0 +1,33 @@
...
+    const allowListedCountries = getAllowListedCountries()
+
+    const lastIP = userIP.lastIPs.sort(
+      (a, b) => b.lastConnection.getTime() - a.lastConnection.getTime(),
+    )[0]
+
+    if (!lastIP) return true
+
+    if (lastIP.proxy && lastIP.proxy == "yes") return new InvalidIpMetadataTypeError()
```
nicolasburtey, line 15: the statement `lastIP.proxy == "yes"` feels a bit off. if proxy is a boolean, we should type it that way, and convert from a string to a boolean at the service level.

src/services/mongoose/users-ips.ts
```diff
@@ -56,7 +56,7 @@ export const UsersIpRepository = (): IUsersIPsRepository => {
 
 const userIPsFromRaw = (result: UserIPsType): UserIPs => {
   return {
-    id: result.id as UserId,
+    id: fromObjectId<UserId>(result._id) as UserId,
```
bodymindarts, line 59: casting `as UserId` shouldn't be needed. That is what the `<UserId>` generic is for.

PR #1008, opened 2022-01-24
feat: Add locale configuration for push notifications

leesalminen:
This PR adds a new yaml config setting for the instance's locale. It modifies how push notification amounts are formatted by setting the formatting to the set locale using `Number.toLocaleString`.


src/core/types.d.ts
```diff
@@ -47,6 +47,11 @@ type TwoFAConfig = {
   threshold: number
 }
 
+type LocaleConfig = {
```
nicolasburtey, line 50: `src/core` is deprecated, the legacy code is living there and this directory will soon be removed. we're now using domain driven design. 

maybe you can create a `src/domain/localization` and add this type here. 

I think having some unit tests on this PR would also be useful.

leesalminen: Understood, I've moved to `src/domain/localization` as requested.

I will push some unit tests.

leesalminen:
TODO:: add unit tests.

PR #1010, opened 2022-01-24
feat: add support for the x-forwarded-for header if x-real-ip is not set

leesalminen:
This PR adds support for using the `x-forwarded-for` request header to determine the user's IP if `x-real-ip` isn't set. `x-forwarded-for` can contain a comma separated string of IPs, and we only want to keep the first one.


nicolasburtey (approved):

leesalminen:
Determined with Justin that this isn't needed. the issue lies with our infra configuration.

PR #1022, opened 2022-01-27
feat: backfill lnPayment objects from lnd

vindard:
## Background

_This PR builds on #848_ 

In #848 we add payments to the `lnpayments` collection for any new payments made. All the old payments in our `lnd` are still unpersisted though.

The ideal way to do this would be to use lnd's `listpayments` api to fetch all payments from `lnd` and persist to the `lnPayments` collection but we can't fetch past index `716895` currently because of a likely corruption in our lnd db (see https://github.com/lightningnetwork/lnd/issues/5658).

### Description

This PR instead uses our `medici_transactions` collection and the `TrackPaymentsV2` api in lnd to fetch transactions and populate our `lnpayments` collection. It allows us to persist the following for each of our payment statuses:
- `"pending"`: the `paymentHash` and `sentFromPubkey` properties
- `"failed"`: the `paymentHash`, `sentFromPubkey` and `status` properties
- `"settled"`: all properties except the `attempts` property that contains paths information

In all cases above the `isCompleteRecord` value is left at false because ideally we would still like to try to recover our lnd db's functionality and be able to run `listpayments` to get complete objects.

## After this PR

We'd need to get `listpayments` working again. We can:
- isolate which payment hash is the problematic one based on, likely at index `716893`, and delete this payment
- test again to see if `listpayments` is fixed
- if not, do another iteration of lnd-fetching with `TrackPaymentsV2` where we use incomplete hashes to
  - query lnd and delete payments with indices past the corruption point
  - mark the `lnPayments` record as complete since we can no longer complete its state

**Note:** if `listpayments` is broken and we populate hashes in `lnPayments` collection past the corruption point with this PR, then that method will always fail at that corruption point when we run the `updateLnPayments` method from cron. We would need to double-check that the failure is handled properly from the cron job.


nicolasburtey:
> After this PR

after we have saved all the payment into mongodb, I would just do a `lncli "deleteallpayments"` , basically to reset the whole set in lnd (after checking there is no in flight payment when we do this). this might help clean up any bad state.

vindard:
Closing this PR in favor of a new approach starting at PR #1040

PR #1023, opened 2022-01-27
refactor: separate 'listSettledAndFailed' method

vindard:
## Description

When testing on our prod lnd payments set, I noticed that the `token` returned by the `getPayments` vs the `getFailedPayments` calls to lnd were not always consistent for the same start token as previously assumed. Because of this, we won't be able to paginate those two methods together, and so this PR is to separate the calling of the joined `listSettledAndFailedPayments` into two separate calls wherever it is currently used.

For context, this is the reason that both methods paginate differently on bigger databases with selectively cleaned payments already removed: [`getPayments`](https://github.com/alexbosworth/lightning/blob/3eaa5541e8f542ebd15e7d2da5e685aba6ea99f1/lnd_methods/offchain/get_payments.js#L135) vs. [`getFailedPayments`](https://github.com/alexbosworth/lightning/blob/3eaa5541e8f542ebd15e7d2da5e685aba6ea99f1/lnd_methods/offchain/get_failed_payments.js#L130)

### DIFF
--- a/src/app/lightning/update-ln-payments.ts
+++ b/src/app/lightning/update-ln-payments.ts
@@ -13,17 +13,20 @@ export const updateLnPayments = async (): Promise<true | ApplicationError> => {
   const lndService = LndService()
   if (lndService instanceof Error) return lndService
 
-  const listFns = [lndService.listSettledPayments, lndService.listFailedPayments]
+  const listFns: ListLnPayments[] = [
+    lndService.listSettledPayments,
+    lndService.listFailedPayments,
+  ]
   const pubkeys = lndService
     .listActivePubkeys()
     .filter((pubkey) => pubkeysFromPayments.has(pubkey))
 
   for (const listFn of listFns) {
-    for (const key of pubkeys) {
+    for (const pubkey of pubkeys) {
       processedLnPaymentsHashes = await updateLnPaymentsByFunction({
         processedLnPaymentsHashes,
         incompleteLnPayments,
-        pubkey: key as Pubkey,
+        pubkey,
         listFn,
       })
     }
--- a/src/app/lightning/update-ln-payments.ts
+++ b/src/app/lightning/update-ln-payments.ts
@@ -1,3 +1,4 @@
+import { UnknownLightningServiceError } from "@domain/bitcoin/lightning"
 import { LndService } from "@services/lnd"
 import { baseLogger } from "@services/logger"
 import { LnPaymentsRepository } from "@services/mongoose/ln-payments"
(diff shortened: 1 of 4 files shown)


dolcalmi (approved):

src/app/lightning/update-ln-payments.ts
```diff
@@ -12,63 +12,80 @@ export const updateLnPayments = async (): Promise<true | ApplicationError> => {
...
   if (lndService instanceof Error) return lndService
+
+  const listFns = [lndService.listSettledPayments, lndService.listFailedPayments]
   const pubkeys = lndService
     .listActivePubkeys()
     .filter((pubkey) => pubkeysFromPayments.has(pubkey))
 
-  for (const key of pubkeys) {
-    const pubkey = key as Pubkey
+  for (const listFn of listFns) {
```
nicolasburtey, line 21: `listFn` no longer in a list

src/app/lightning/update-ln-payments.ts
```diff
@@ -12,63 +12,80 @@ export const updateLnPayments = async (): Promise<true | ApplicationError> => {
...
+    for (const key of pubkeys) {
```
nicolasburtey, line 22: maybe use `pubkey` instead of `key`?

src/app/lightning/update-ln-payments.ts
```diff
@@ -12,63 +12,80 @@ export const updateLnPayments = async (): Promise<true | ApplicationError> => {
...
+      processedLnPaymentsHashes = await updateLnPaymentsByFunction({
+        processedLnPaymentsHashes,
+        incompleteLnPayments,
+        pubkey: key as Pubkey,
```
nicolasburtey, line 26: the type assignment should not be necessary. if `lndService.listActivePubkeys()` is not typing the result to `Pubkey`, this should be added there, not here

src/app/lightning/update-ln-payments.ts
```diff
@@ -12,63 +12,80 @@ export const updateLnPayments = async (): Promise<true | ApplicationError> => {
...
-      }
-      if (after === results.endCursor) break
-      after = results.endCursor
+const updateLnPaymentsByFunction = async ({
+  processedLnPaymentsHashes,
+  incompleteLnPayments,
+  pubkey,
+  listFn,
+}): Promise<PaymentHash[]> => {
+  let after: PagingStartToken | PagingContinueToken | PagingStopToken = undefined
```
nicolasburtey, line 40: I think this is the type of value that we could also show as an event with opentemeletry. 

like this we would get a clear sense of the evolution of this longer running loop from honeycomb 

could be something like `addEventToCurrentSpan("updateLnPayments", {after})`

vindard:
@dolcalmi any idea why the `updateLnPayments` [traces](https://ui.honeycomb.io/galoy/datasets/arvin-dev?query=%7B%22time_range%22%3A600%2C%22granularity%22%3A0%2C%22breakdowns%22%3A%5B%5D%2C%22calculations%22%3A%5B%7B%22op%22%3A%22COUNT%22%7D%5D%2C%22filters%22%3A%5B%7B%22column%22%3A%22code.function%22%2C%22op%22%3A%22%3D%22%2C%22value%22%3A%22updateLnPayments%22%7D%5D%2C%22filter_combination%22%3A%22AND%22%2C%22orders%22%3A%5B%5D%2C%22havings%22%3A%5B%5D%2C%22limit%22%3A1000%7D) look different when run from tests (first 2 peaks) vs from cron (last peak)

src/app/lightning/update-ln-payments.ts
vindard: Done: [sample trace](https://ui.honeycomb.io/galoy/datasets/arvin-dev/result/rg6zNdKvS2D/trace/iUcHXZVbfhZ?span=1ee6b89ae9d5bdfb)

dolcalmi (approved):

PR #1030, opened 2022-01-28
feat: create 2 wallets for the dealer

nicolasburtey:
important note: they are under the same account. the dealer would programaticaly get the balance of both throught the normal graphql api

### DIFF
--- a/src/migrations/20220203152515-bootstrap_accounts.ts
+++ b/src/migrations/20220203152515-bootstrap_accounts.ts
@@ -1,0 +1,89 @@
+/* eslint @typescript-eslint/no-var-requires: "off" */
+// @ts-expect-error ts complain about "Cannot redeclare block-scoped variable 'crypto'"
+const crypto = require("crypto")
+
+const adminUsers = [
+  {
+    role: "dealer",
+    needUsdWallet: true,
+  },
+  { role: "funder" },
+  { role: "bankowner" },
+]
+
+const getRandomInvalidPhone = () => {
+  return `+abc${Math.floor(Math.random() * 999_999_999_999_999)}`
+}
+
+const defaultWallet = { __v: 0, type: "checking", onchain: [] }
+
+module.exports = {
+  async up(db /*, client */) {
+    for (const userMeta of adminUsers) {
+      const role = userMeta.role
+      let user = await db.collection("users").findOne({ role })
+
+      let userId
+      if (!user) {
+        const phone = getRandomInvalidPhone()
+        const created_at = new Date()
+        const result = await db.collection("users").insertOne({ role, phone, created_at })
+        userId = result.insertedId
+
+        user = await db.collection("users").findOne({ role })
+      } else {
+        console.log({ role }, "user exist")
+        userId = user._id
+      }
+
+      const _accountId = userId
+      // const accountId = Types.ObjectId(userId)
+      let wallet = await db.collection("wallets").findOne({ _accountId, currency: "BTC" })
+
+      if (!wallet) {
+        const currency = "BTC"
+        // @ts-expect-error: ts complain about Crypto (capital C) doesn't have randomUUID
+        const id = crypto.randomUUID()
+
+        const result = await db
+          .collection("wallets")
+          .insertOne({ _accountId, currency, id, ...defaultWallet })
+
+        wallet = await db.collection("wallets").findOne({ _accountId, currency: "BTC" })
+
+        console.log({ result, currency }, "wallet has been created")
+      } else {
+        console.log({ role, walletExist: wallet }, "wallet exist")
+      }
+
+      if (user.defaultWalletId === undefined) {
+        const result = await db
+          .collection("users")
+          .findOneAndUpdate({ role }, { $set: { defaultWalletId: wallet.id } })
+
+        console.log({ result, role }, "update defaultWalletId")
+      } else {
+        console.log({ role }, "no need to update defaultWalletId")
+      }
+
+      if (userMeta.needUsdWallet) {
+        const currency = "USD"
+        wallet = await db.collection("wallets").findOne({ _accountId, currency })
+
+        if (!wallet) {
+          // @ts-expect-error: ts complain about Crypto (capital C) doesn't have randomUUID
+          const id = crypto.randomUUID()
+          const result = await db
+            .collection("wallets")
+            .insertOne({ _accountId, currency, id, ...defaultWallet })
+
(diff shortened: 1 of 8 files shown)


src/services/mongodb/index.ts
```diff
@@ -16,12 +22,25 @@ export const ledger = loadLedger({
...
-  dealerWalletResolver: async () => {
-    const { defaultWalletId } = await User.findOne(
-      { role: "dealer" },
-      { defaultWalletId: 1 },
-    )
-    return defaultWalletId
+  dealerBtcWalletResolver: async () => {
+    const user: UserRecord = await User.findOne({ role: "dealer" }, { id: 1 })
+    // FIXME remove the use of UserRecord when role if part of the AccountRepository
+    const accountId = user._id as AccountId
```
bodymindarts, line 28: I think this needs to go through `fromObjectId<AccountId>`

src/servers/exporter.ts
```diff
@@ -73,14 +74,18 @@ const business_g = new client.Gauge({
...
+  getFunderWalletId,
+  getBankOwnerWalletId,
+]
 const walletRoles = {}
 
-for (const role of roles) {
-  walletRoles[role] = new client.Gauge({
-    name: `${prefix}_${role}_balance`,
-    help: "funder balance BTC",
+for (const walletName in walletIdsFn) {
```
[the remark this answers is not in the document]
nicolasburtey: you're right, something was wrong in my code. I have refactored it also to make it cleaner.

src/servers/exporter.ts
```diff
@@ -114,21 +119,20 @@ const main = async () => {
     const userCount = await User.countDocuments()
     userCount_g.set(userCount)
 
-    for (const index in roles) {
-      const role = roles[index]
-      const walletId = await accountRoles[index]()
+    for (const walletIdFn in walletIdsFn) {
```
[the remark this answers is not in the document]
nicolasburtey: you're right, something was wrong in my code. I have refactored it also to make it cleaner.

test/helpers/user.ts
```diff
@@ -90,6 +92,17 @@ export const createUserWallet = async (index: number) => {
         }
       : null
     userRepo = await createUser({ phone, phoneMetadata })
+    if (userRepo instanceof Error) throw userRepo
+
+    if (entry.additionalWallets) {
+      for (const { currency } of entry.additionalWallets) {
```
vindard, line 98: `currency` is type `any` here coming out of `entry.additionalWallets`. Could potentially be a good candidate for a `checkedToCurrency` validator.

vindard (commented):
The `walletsInit` improvement is a lot cleaner now! Just left one nit about an `any` type, but otherwise looks good to me.

vindard (approved):

nicolasburtey:
this wallet will have to be create manually for now. what do you think @bodymindarts ?

test/helpers/user.ts
nicolasburtey: that is true but it's for the tests only, and I think the best way to manage this is to have a yaml validator, which will be a big PR by itself, so I guess to keep it as is for now.

bodymindarts:
> this wallet will have to be create manually for now. what do you think @bodymindarts ?

What do you mean manually? When standing up a new environment or in existing environments or both?

nicolasburtey:
> > this wallet will have to be create manually for now. what do you think @bodymindarts ?
> 
> What do you mean manually? When standing up a new environment or in existing environments or both?

I guess both until we automate it

bodymindarts:
> I guess both until we automate it

the app should be able to startup in a consistent way from scratch. If the wallet isn't present it should be created on startup IMO.

src/migrations/20220203152515-bootstrap_accounts.ts
```diff
@@ -0,0 +1,95 @@
+/* eslint @typescript-eslint/no-var-requires: "off" */
+// @ts-expect-error ts complain about "Cannot redeclare block-scoped variable 'crypto'"
+const crypto = require("crypto")
+
+const adminUsers = [
```
nicolasburtey, line 5: re-import because migrate mongo use export.default, not sure how to switch to import/export.

maybe you have some clue @samerbuna @jotapea ?

src/migrations/20220203152515-bootstrap_accounts.ts
```diff
@@ -0,0 +1,90 @@
...
+        wallet = await db.collection("wallets").findOne({ _accountId, currency })
+
+        if (!wallet) {
+          // @ts-expect-error: ts complain about Crypto (capital C) doesn't have randomUUID
+          const id = crypto.randomUUID()
+          const result = await db
+            .collection("wallets")
+            .insertOne({ _accountId, currency, id, ...defaultWallet })
+
+          wallet = await db.collection("wallets").findOne({ _accountId, currency: "BTC" })
```
vindard, line 81: What's the purpose of this line?

vindard (approved):

src/migrations/20220203152515-bootstrap_accounts.ts
```diff
@@ -0,0 +1,90 @@
...
+
+module.exports = {
+  async up(db /*, client */) {
+    for (const userMeta of adminUsers) {
+      const role = userMeta.role
+      let user = await db.collection("users").findOne({ role })
+
+      let userId
+      if (!user) {
+        const phone = getRandomInvalidPhone()
```
vindard, line 29: Just double-checking, what's the thinking behind adding "invalid phone" admin users?

nicolasburtey: phone is a mandatory field (there is an index for it), so we need a phone number. but we don't want anyone to log on this user, so having an invalid phone number is the way to do it. 

I think this will be refactor with the new auth library @bodymindarts

src/migrations/20220203152515-bootstrap_accounts.ts
nicolasburtey: check if a wallet already exists. we don't do anything in that case.

nicolasburtey: this script will be used both for new cluster, but also existing cluster that have those users (partially) setup

src/migrations/20220203152515-bootstrap_accounts.ts
bodymindarts: Eventually yes - first Kratos needs phone / code auth support (supposedly released in Q2?)

vindard (approved):

dolcalmi (approved):

PR #1035, opened 2022-01-31
refactor: prepare ledger for multicurrency wallets

nicolasburtey:


src/app/cold-storage/rebalance-to-cold-wallet.ts
```diff
@@ -58,16 +59,16 @@ export const rebalanceToColdWallet = async (): Promise<boolean | ApplicationErro
 
   const description = `deposit of ${rebalanceAmount} sats to the cold storage wallet`
 
-  const usd = rebalanceAmount * usdPerSat
-  const usdFee = fee * usdPerSat
+  const amountDisplayCurrency = DisplayCurrencyConversionRate(usdPerSat)(rebalanceAmount)
+  const feeDisplayCurrency = DisplayCurrencyConversionRate(usdPerSat)(fee)
```
vindard, line 63: Similar point from before, it looks like this is always called as `DisplayCurrencyConversionRate(usdPerSat)(amount)`. Currying may be redundant here.

src/app/wallets/ln-send-payment.ts
```diff
@@ -306,35 +306,32 @@ const executePaymentViaIntraledger = async ({
...
-          usd,
-          usdFee,
+          sats: amount,
+          amountDisplayCurrency,
           recipientWalletId,
-          pubkey: lndService.defaultPubkey(),
-          senderUsername: senderAccount.username,
+          recipientWalletCurrency: recipientWallet.currency,
           recipientUsername: null,
+          pubkey: lndService.defaultPubkey(),
```
vindard, line 334: Just thinking about this again, the `defaultPubkey` may not be relevant here because there's no send happening and there's no guarantee that it's the same as the recipient invoice's destination pubkey. 

It might be more relevant to use the `recipientPubkey` value at this point (for when we eventually move to multiple lnds).

src/app/wallets/pay-on-chain.ts
```diff
@@ -153,59 +153,56 @@ const executePaymentViaIntraledger = async ({
...
+          `Payment amount '${amount}' exceeds balance '${balance}'`,
         )
 
       const onchainLoggerOnUs = logger.child({ balance, onUs: true })
       const journal = await LockService().extendLock(
         { logger: onchainLoggerOnUs, lock },
         async () =>
-          LedgerService().addOnChainIntraledgerTxTransfer({
-            senderWalletId,
+          LedgerService().addOnChainIntraledgerTxSend({
```
vindard, line 195: Looks like we lost the `addOnChainIntraledgerTxTransfer` change somehow in this diff

vindard (commented):
Adding some initial comments

src/domain/wallets/index.types.d.ts
```diff
@@ -179,20 +179,30 @@ type PaymentInputValidatorConfig = (
...
   amount: Satoshis
   senderWallet: Wallet
-  recipientWallet?: Wallet
+}
+
+type ValidatePaymentInputRecipientArgs = {
+  recipientWalletId: string
+  senderWallet: Wallet
+}
+type ValidatePaymentInputRecipientRet = {
```
vindard, line 197: Minor nit, we've so far been using the `Result` suffix in places like this. Can probably standardise on one or the other

src/graphql/root/mutation/intraledger-payment-send.ts
```diff
@@ -22,30 +22,20 @@ const IntraLedgerPaymentSendMutation = GT.Field({
   args: {
     input: { type: GT.NonNull(IntraLedgerPaymentSendInput) },
   },
-  resolve: async (
-    _,
-    args,
-    { domainAccount, logger }: { domainAccount: Account; logger: Logger },
-  ) => {
+  resolve: async (_, args, { domainAccount, logger }) => {
```
vindard, line 25: Not sure if this was an intentional reversion?

src/graphql/root/mutation/intraledger-payment-send.ts
```diff
@@ -65,7 +55,8 @@ const IntraLedgerPaymentSendMutation = GT.Field({
       memo,
       amount,
       senderWalletId: walletId,
-      senderAccount: domainAccount,
+      // FIXME: domainAccount should not be undefined if we reach this stage. Improve type
```
vindard, line 58: Looks like another unintentional reversion

src/app/wallets/ln-send-payment.ts
nicolasburtey: just changing the order of property here to group sender/recipient- not changing the logic on the pubkey. is there are some things to improve it, would you mind creating a PR for it?

src/domain/wallets/index.types.d.ts
nicolasburtey: @bodymindarts

nicolasburtey: result: 26 occurences
ret: 6 occurences

nicolasburtey: I like ret thought. few characters means fewer line return.

src/graphql/root/mutation/intraledger-payment-send.ts
nicolasburtey: yes now the graphql domain is automatically been typed

nicolasburtey: this is the PR that brought those changes: https://github.com/GaloyMoney/galoy/pull/891

src/graphql/root/mutation/intraledger-payment-send.ts
nicolasburtey: it is intentional. just the result of the same PR (we need to improve the types at the graphql layer)

nicolasburtey: basically we know the domainAccount is not undefined here, but typescript doesn't know it.

src/app/cold-storage/rebalance-to-cold-wallet.ts
nicolasburtey: not using the currying now, but could be useful in the future. we can also see the first parameter is the `Input` and the second is the `Args`

src/domain/wallets/index.types.d.ts
bodymindarts: I prefer Ret for return type. I associate Result with the ‘Result<T,E>’ type (which we aren’t using but is common in many languages)

src/app/cold-storage/rebalance-to-cold-wallet.ts
vindard: Yup, I'm curious how different this pattern is from things like our `WalletInvoiceValidator` that takes an invoice in the constructor and then has method like `validateToSend` that takes args

src/app/wallets/ln-send-payment.ts
vindard: Yup it's a separate thing, I can PR that

src/app/cold-storage/rebalance-to-cold-wallet.ts
bodymindarts: I'm with @vindard - don't see a need for currying here. Wouldn't introduce the pattern until its really necessary. Would just stick with the constructor `XxxConfig` and method `xxxArgs` type patterns.

src/app/wallets/intraledger-send-payment.ts
```diff
@@ -157,15 +157,15 @@ const executePaymentViaIntraledger = async ({
   logger: Logger
 }): Promise<PaymentSendStatus | ApplicationError> => {
   const validator = PaymentInputValidator(WalletsRepository().findById)
-  const validationResult = await validator.validatePaymentInput({
+  const validationSender = await validator.validatePaymentInput({
```
bodymindarts, line 160: I think this can be renamed back to `validationResult`

dolcalmi (dismissed):

bodymindarts (approved):

src/graphql/root/mutation/intraledger-payment-send.ts
nicolasburtey: not sure to understand your point @dolcalmi ?

src/app/wallets/ln-send-payment.ts
```diff
@@ -306,35 +306,32 @@ const executePaymentViaIntraledger = async ({
...
+          senderWalletCurrency: senderWallet.currency,
+          senderUsername: senderAccount.username,
           paymentHash,
           description,
-          sats,
-          fee: lnFee,
-          usd,
-          usdFee,
+          sats: amount,
+          amountDisplayCurrency,
```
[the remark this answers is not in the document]
nicolasburtey: correct. but this will comes in another PR. currently the assumption is that wallets are still all BTC wallets. I'm working on the plumbing to send the currency up to the ledger right now.

src/app/wallets/pay-on-chain.ts
```diff
@@ -153,59 +153,56 @@ const executePaymentViaIntraledger = async ({
...
+            senderWalletId: senderWallet.id,
+            senderWalletCurrency: senderWallet.currency,
+            senderUsername: senderAccount.username,
             description: "",
-            sats,
-            fee,
-            usd,
-            usdFee,
+            sats: amount,
+            amountDisplayCurrency,
```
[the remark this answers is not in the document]
nicolasburtey: correct. but this will comes in another PR. currently the assumption is that wallets are still all BTC wallets. I'm working on the plumbing to send the currency up to the ledger right now.

src/graphql/root/mutation/intraledger-payment-send.ts
nicolasburtey: ok I have understood what to do to remove the type error

src/services/ledger/receive.ts
```diff
@@ -11,114 +13,173 @@ import { translateToLedgerJournal } from "."
...
+      return addReceiptFee({
+        metadata,
+        walletId,
+        walletCurrency,
+        sats,
         fee,
-        usdFee,
+        description,
+      })
+    } else {
```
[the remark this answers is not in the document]
nicolasburtey: I think there is more clarity here with the else: it's really one branch ( fee > 0`) or, else, the other. 

the case where we don't want to have a else for me is if we have a scenario where we want to check some uncommon condition, and where we could return early if we don't get what we want.

src/services/ledger/receive.ts
```diff
@@ -11,114 +13,173 @@ import { translateToLedgerJournal } from "."
...
+      return addReceiptFee({
+        metadata,
+        walletId,
+        walletCurrency,
+        sats,
         fee: feeInboundLiquidity,
-        feeUsd: feeInboundLiquidityDisplayCurrency,
+        description,
+      })
+    } else {
```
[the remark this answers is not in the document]
nicolasburtey: I think there is more clarity here with the else: it's really one branch ( fee > 0`) or, else, the other. 

the case where we don't want to have a else for me is if we have a scenario where we want to check some uncommon condition, and where we could return early if we don't get what we want.

src/services/ledger/send.ts
```diff
@@ -66,38 +61,34 @@ export const send = {
...
+    if (bankFee > 0) {
+      return addSendInternalFee({
+        walletId,
+        walletCurrency,
+        metadata,
+        description,
+        sats,
+        fee: bankFee,
+      })
+    } else {
```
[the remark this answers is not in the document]
nicolasburtey: I think there is more clarity here with the else: it's really one branch ( fee > 0`) or, else, the other. 

the case where we don't want to have a else for me is if we have a scenario where we want to check some uncommon condition, and where we could return early if we don't get what we want.

src/services/ledger/admin.ts
```diff
@@ -71,28 +68,25 @@ export const admin = {
...
-        usd,
-        currency: WalletCurrency.Btc,
-      }
+    const metadata: AddColdStorageSendLedgerMetadata = {
+      type: LedgerTransactionType.ToHotWallet,
+      pending: false,
+      hash: txHash,
+      payee_addresses: [payeeAddress],
+      fee,
+      feeUsd: feeDisplayCurrency,
```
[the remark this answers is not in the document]
nicolasburtey: when currency = "BTC", sats is a duplicated to the amount passed in the ledger

src/graphql/root/mutation/intraledger-payment-send.ts
nicolasburtey: ok got it actually

bodymindarts (approved):

PR #1037, opened 2022-02-01
feat: persist preimage on payment send/update

vindard:
## Description

This is prep PR for another branch where we fix the population of the `revealedPreImage` field we return in the API ([currently hardcoded as `null`](https://github.com/GaloyMoney/galoy/blob/61b13001697fb32100bbc92ecd9b967b81b190dd/src/domain/wallets/tx-history.ts#L189)).

In this PR, we persist the `revealedPreImage` asynchronously to the `lnpayments` collection in two places that we have access to it in the codebase:
- on payment send
- on pending payment settle


src/domain/bitcoin/lightning/payments/index.types.d.ts
```diff
@@ -23,6 +28,6 @@ interface ILnPaymentsRepository {
     lnPaymentPartial: LnPaymentPartial,
   ): Promise<LnPaymentPartial | RepositoryError>
   update(
-    lnPayment: PersistedLnPaymentLookup,
+    lnPayment: PersistedLnPaymentLookup | LnPaymentUpdatePreImage,
```
dolcalmi, line 31: not sure about this pattern... Have we used it elsewhere?

src/services/lnd/schema.ts
```diff
@@ -4,26 +4,11 @@ import * as mongoose from "mongoose"
...
-  },
-  milliSatsFee: {
-    type: Number,
-    required: true,
-  },
+  confirmedAt: Date,
+  destination: String,
+  revealedPreImage: String,
+  roundedUpFee: Number,
+  milliSatsFee: Number,
```
dolcalmi, line 11: shouldn't this match non optional field in https://github.com/alexbosworth/ln-service#getpayments ? (I mean for the `required` flag)

dolcalmi (changes requested):
looks good, just few questions

vindard:
Converted to draft since we decided to persist these in a new `paymentsmetadata` collection to be part of the Ledger service instead

vindard:
Note to self, the transactions listed from these two queries should match after tests are run:
- On `medici_transaction`:
  `{type: "payment", void_reason: {$exists: false}, memo: {$ne: "Payment canceled"}}`

- On `medici_transaction_metadatas`:
  `{revealedPreImage: {$exists: true}}`

src/services/ledger/domain/index.types.d.ts
```diff
@@ -0,0 +1,22 @@
...
+type PersistNewLedgerTransactionMetadataArgs = {
+  id: LedgerTransactionId
+  ledgerTxMetadata: LedgerTransactionMetadata
+}
+
+interface ITransactionsMetadataRepository {
+  updateByHash(
+    ledgerTxMetadata: LedgerTransactionMetadataWithHash,
+  ): Promise<true | RepositoryError>
+  persistnew(
```
bodymindarts, line 18: `camelCase`

src/services/ledger/domain/index.types.d.ts
```diff
@@ -0,0 +1,22 @@
...
+type LedgerTransactionMetadataWithHash = Omit<LedgerTransactionMetadata, "hash"> & {
+  hash: PaymentHash | OnChainTxHash
+}
+
+type PersistNewLedgerTransactionMetadataArgs = {
+  id: LedgerTransactionId
+  ledgerTxMetadata: LedgerTransactionMetadata
+}
+
+interface ITransactionsMetadataRepository {
```
bodymindarts, line 14: `persistAll`

src/services/ledger/intraledger.ts
```diff
@@ -165,7 +169,16 @@ const addIntraledgerTxTransfer = async ({
       .debit(senderLiabilitiesWalletId, sats, debitMetadata)
 
     const savedEntry = await entry.commit()
-    return translateToLedgerJournal(savedEntry)
+    const journalEntry = translateToLedgerJournal(savedEntry)
+
+    journalEntry.transactionIds.map((_id) =>
```
bodymindarts, line 174: Avoid N+1

src/services/ledger/receive.ts
```diff
@@ -178,7 +204,16 @@ const addReceiptFee = async ({
       .credit(bankOwnerPath, fee, metadata)
 
     const savedEntry = await entry.commit()
-    return translateToLedgerJournal(savedEntry)
+    const journalEntry = translateToLedgerJournal(savedEntry)
+
+    journalEntry.transactionIds.map((_id) =>
+      txMetadataRepo.persistnew({
+        id: _id,
```
bodymindarts, line 211: `_id` => by convention `_` represents a mongoose / MongoDB native id.
`id` without `_` represents a domain representation
and the conversion happens with `from/toObjectId<Type>`

src/services/ledger/receive.ts
```diff
@@ -178,7 +204,16 @@ const addReceiptFee = async ({
       .credit(bankOwnerPath, fee, metadata)
 
     const savedEntry = await entry.commit()
-    return translateToLedgerJournal(savedEntry)
+    const journalEntry = translateToLedgerJournal(savedEntry)
+
+    journalEntry.transactionIds.map((_id) =>
```
bodymindarts, line 209: perhaps add a type hint to the lambda to be explicit what kind of ID is here.

src/services/ledger/domain/index.types.d.ts
```diff
@@ -0,0 +1,22 @@
+type LedgerTransactionMetadata = {
+  hash?: PaymentHash | OnChainTxHash
```
bodymindarts, line 2: Isn't `journalId` missing and or `ledgerTxId` is missing

src/services/ledger/domain/index.types.d.ts
```diff
@@ -0,0 +1,22 @@
...
+  revealedPreImage?: RevealedPreImage
+}
+type LedgerTransactionMetadataWithHash = Omit<LedgerTransactionMetadata, "hash"> & {
+  hash: PaymentHash | OnChainTxHash
+}
+
+type PersistNewLedgerTransactionMetadataArgs = {
+  id: LedgerTransactionId
```
bodymindarts, line 10: `id` is foreign key => `ledgerTransactionId`
=> aggregate root in DDD speak.

src/services/ledger/schema.ts
```diff
@@ -160,3 +160,20 @@ transactionSchema.index({
 })
 
 export const Transaction = mongoose.model("Medici_Transaction", transactionSchema)
+
+const transactionMetadataSchema = new Schema({
```
bodymindarts, line 164: missing `_ledgerTransactionId`?

src/services/ledger/schema.types.d.ts
```diff
@@ -0,0 +1,5 @@
+interface TransactionMetadataRecord {
+  id: string
```
bodymindarts, line 2: `id: string` => `Record.id` should be type `ObjectId`?

bodymindarts (dismissed):

src/services/ledger/schema.ts
vindard: It looks like for the Schemas where the `_id` is mongo-generated we never define this property in the Schema definition.

src/services/ledger/schema.types.d.ts
```diff
@@ -0,0 +1,6 @@
+interface TransactionMetadataRecord {
+  _id: ObjectId
+
+  hash?: string
```
bodymindarts, line 4: `hash` and `revealedPreImage` are specified as optional here but not in the schema:
```
  {
     hash: String,
     revealedPreImage: {
       type: String,
       index: true,
     },
   },
   ```
   Or is optional the default?

bodymindarts (approved):

src/services/ledger/schema.types.d.ts
vindard: Would have to confirm, but I believe optional is the default and we add `required: true` to make it non-optional

PR #1039, opened 2022-02-02
refactor: separate testUsers to adminUsers

nicolasburtey:
also get users to a ref instead of an index number

mostly a no-op. just make separation a bit cleaner.


test/integration/servers/trigger.spec.ts
```diff
@@ -48,18 +48,18 @@ let userType3: UserRecord
...
-  await createUserWallet(12)
+  await createUserWalletFromUserRef("A")
+  await createUserWalletFromUserRef("D")
+  await createUserWalletFromUserRef("F")
 
-  walletId0 = await getDefaultWalletIdByTestUserIndex(0)
-  walletId3 = await getDefaultWalletIdByTestUserIndex(3)
-  walletId12 = await getDefaultWalletIdByTestUserIndex(12)
+  walletId0 = await getDefaultWalletIdByTestUserRef("A")
+  walletId3 = await getDefaultWalletIdByTestUserRef("D")
```
UncleSamtoshi, line 56: Now that we are using references, names with indexes like `walletId3` no longer make sense

default.yaml
```diff
@@ -43,48 +43,29 @@ coldStorage:
 lndScbBackupBucketName: "lnd-static-channel-backups"
 
 test_accounts:
-- phone: "+16505554321" # user0
-  code: 321321
-- phone: "+16505554322" # user1
-  code: 321432
+- ref: A
+  phone: "+16505554321" # user0
```
UncleSamtoshi, line 47: comments such `# user0` are no longer applicable

test/helpers/user.ts
```diff
@@ -1,78 +1,85 @@
...
   if (account instanceof Error) throw account
   return account
 }
 
-export const getUserIdByTestUserIndex = async (index: number) => {
-  const user = await getUserByTestUserIndex(index)
+export const getUserIdByTestUserRef = async (ref: string) => {
+  const user = await getUserByTestUserRef(ref)
   return user.id
 }
```
UncleSamtoshi, line 33: what is the value if having this function versus just getting the user and accessing the id property?

UncleSamtoshi: several of the functions seem to fall into this category

test/helpers/user.ts
```diff
@@ -1,78 +1,85 @@
...
-export const getUserRecordByTestUserIndex = async (index: number) => {
-  const entry = yamlConfig.test_accounts[index]
+export const getUserRecordByTestUserRef = async (ref: string) => {
+  const entry = yamlConfig.test_accounts.find((item) => item.ref === ref)
   const phone = entry.phone as PhoneNumber
 
   return User.findOne({ phone }) as UserRecord
 }
 
+const adminUsers = [
```
UncleSamtoshi, line 61: Should we consolidate the config for users to be either all yaml or all json instead of split across default.yml and test/helpers/user.ts

UncleSamtoshi (changes requested):

test/integration/servers/trigger.spec.ts
nicolasburtey: fixed

test/helpers/user.ts
nicolasburtey: probably. this can be done in another PR (working on the adminUsers for this PR here which is why I've refactoring this for now)

test_accounts is also used in prod. so the current default.yaml also serve as a documentation which is helpful I think until we have some documentation around this as well

test/helpers/user.ts
nicolasburtey: diff looks like this: https://github.com/GaloyMoney/galoy/pull/1043

not sure I'm a huge fan of it. I don't like doing `()` around the `await`

UncleSamtoshi: thats fair, it does look cleaner the way it is

test/helpers/user.ts
UncleSamtoshi: sounds good

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -196,117 +195,117 @@ describe("UserWallet - Lightning Pay", () => {
...
+    let userType1 = await getUserRecordByTestUserRef("B")
 
-    expect(userType0.contacts).toEqual(
-      expect.arrayContaining([expect.objectContaining({ id: username1 })]),
+    expect(userTypeA.contacts).toEqual(
+      expect.arrayContaining([expect.objectContaining({ id: usernameB })]),
     )
-    const contact0 = userType0.contacts.find(
-      (userContact) => userContact.id === username1,
+    const contact0 = userTypeA.contacts.find(
```
UncleSamtoshi, line 270: naming

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -196,117 +195,117 @@ describe("UserWallet - Lightning Pay", () => {
...
-    const finalBalance1 = await getBTCBalance(walletId1)
+    const finalBalance1 = await getBTCBalance(walletIdB)
     const txResult = await Wallets.getTransactionsForWalletId({
-      walletId: walletId1,
+      walletId: walletIdB,
     })
     if (txResult.error instanceof Error || txResult.result === null) {
       throw txResult.error
     }
     const userTransaction1 = txResult.result
```
UncleSamtoshi, line 246: name

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -196,117 +195,117 @@ describe("UserWallet - Lightning Pay", () => {
...
     const { result: userTransaction0, error } = await Wallets.getTransactionsForWalletId({
-      walletId: walletId0,
+      walletId: walletIdA,
     })
     if (error instanceof Error || userTransaction0 === null) {
       throw error
     }
 
-    const finalBalance1 = await getBTCBalance(walletId1)
+    const finalBalance1 = await getBTCBalance(walletIdB)
```
UncleSamtoshi, line 239: name

UncleSamtoshi:
Looking good but there are still some issues with naming. I marked a couple of them, but not all of them.

UncleSamtoshi (changes requested):

nicolasburtey:
@UncleSamtoshi I've fixed the one you mentioned. possible there are still some variables not renamed but we can renamed them as we see them

UncleSamtoshi:
Sounds good! I will be touching these files as well, so I can do some cleanup.

UncleSamtoshi (dismissed):

nicolasburtey:
@UncleSamtoshi can you re-approve? had to rebase

UncleSamtoshi (approved):

PR #1040, opened 2022-02-02
chore: migrate lnd payments db via listpayments

vindard:
## Description

This PR adds a debug function to be run **_once_** in our prod environment. The function will use the `getPayments` method (`listpayments` api from `lnd`) to iterate over all our `lnd` payments and:
- populate a new lnPayment object where an object didn't exist before for a given _payment hash_ in the `lnpayments` collection
- skip `isCompleteRecord: true` objects in `lnpayments` collection
- skip existing objects in `lnpayments` that are still pending in `lnd`
- update `isCompleteRecord: false` objects in `lnpayments` that are no longer pending in `lnd`

### Handling db corruption

Our lnd payments db currently has corrupt payments that we can't query using the `getPayments` method (see https://github.com/lightningnetwork/lnd/issues/5658).

Our approach with this function is to hijack the `getPayments` pagination to paginate 1 payment at a time and to skip corrupt records by decrementing the pagination counter.

We insist on using this `getPayments` method instead of the `trackpaymentsv2` api (`getPayment` method) for our first pass because we can also access path and failure details via the `attempts` property that is only available via this method.

We will populate the corrupted payments using `trackpaymentsv2` in follow-up PR #1041 

### Observability

Sample trace [available here](https://ui.honeycomb.io/galoy/datasets/arvin-dev/result/yta3dMqa4TN/trace/zbd77PGRuCL?span=90d81ac216ca5c71)


vindard:
Open question, should I add telemetry for this function?

vindard:
Converted to draft to sort out final commit (https://github.com/GaloyMoney/galoy/pull/1040/commits/ccaf2ac29606d8d147ea93ae72d6900a32101330)

Makefile
```diff
@@ -71,3 +71,7 @@ check-code:
 	yarn tsc-check
 	yarn eslint-check
 	yarn build
+
+debug:
```
krtk6160, line 75: I don't think we need to add a command to the Makefile.
Can just run the script via yarn directly as we've been doing so far.

vindard: Yup I need to remove the Makefile entry and hardcore `active: true` line. Just had those in for easy testing locally.

dolcalmi (approved):

PR #1041, opened 2022-02-02
chore: migrate lnd payments db via trackpaymentv2

vindard:
## Description

**_Note: we will only run this after the #1040 function has been run_**

In this PR we query using all the payment hashes that exist in our `medici_transactions` table. For each hash, we check if the hash exists in the `lnpayments` collection and if not we query lnd using the `trackpaymentsv2` api (`getPayment`) and add a new object to the `lnpayments` collection.

### Limitations

This function is only intended to be used after we populate everything we could using the `listpayments` api from PR #1040. It should pull any missing lnpayments we still have and add them to our collection from lnd.

These payments will be incomplete because of the limitations of the `trackpaymentsv2` api but we will still mark them as `isCompleteRecord` since it represents all the data we can possibly get from `lnd` at this time. Specifically missing will be:
- for `settled` payments, no `attempts` value and `paymentRequest` value
- for `failed` payments, everything will be missing except the `status`

### Observability

Sample trace ~[here](https://ui.honeycomb.io/galoy/datasets/arvin-dev/result/Hu5WTXt4zBG/trace/mficXRPCJG4?span=fff95382b42e0d38)~ [here](https://ui.honeycomb.io/galoy/datasets/arvin-dev/result/tuL7oqxbLqq)
_Note: it looks like the 'Cancelled on client' errors for `trackpaymentsv2` is normal_


src/debug/migrate-ln-payments-trackpaymentsv2.ts
```diff
@@ -0,0 +1,131 @@
...
+      "migrateLnPayment.paymentHash":
+        paymentHash instanceof Error ? paymentHash.name : paymentHash,
+    },
+    async (): Promise<true | LightningServiceError> => {
+      if (paymentHash instanceof Error) return paymentHash
+
+      // Check if hash exists in lnPayments repo
+      const lnPaymentsRepo = LnPaymentsRepository()
+      const lnPaymentExists = await lnPaymentsRepo.findByPaymentHash(paymentHash)
+      if (!(lnPaymentExists instanceof CouldNotFindError)) {
```
dolcalmi, line 57: probably and edge case but what happens if it returns another error?

src/services/lnd/index.ts
```diff
@@ -247,7 +247,7 @@ export const LndService = (): ILightningService | LightningServiceError => {
       return payment
     }
 
-    return new PaymentNotFoundError("Payment hash not found")
+    return new PaymentNotFoundError(JSON.stringify({ paymentHash, pubkey }))
```
dolcalmi, line 250: do we need to return this info in the error? is this not known by the caller? (in case we want to do some tracing)

src/services/lnd/index.ts
```diff
@@ -563,6 +563,8 @@ const lookupPaymentByPubkeyAndHash = async ({
   } catch (err) {
     const errDetails = parseLndErrorDetails(err)
     switch (errDetails) {
+      case KnownLndErrorDetails.SentPaymentNotFound:
+        return new PaymentNotFoundError(JSON.stringify({ paymentHash, pubkey }))
```
dolcalmi, line 567: same here

dolcalmi (approved):

src/debug/migrate-ln-payments-trackpaymentsv2.ts
```diff
@@ -0,0 +1,131 @@
...
+  return true
+}
+
+const migrateLnPayment = async (
+  paymentHash: PaymentHash | LightningServiceError,
+): Promise<true | LightningServiceError> =>
+  asyncRunInSpan(
+    "debug.migrateLnPayment",
+    {
+      [SemanticAttributes.CODE_FUNCTION]: "debug.migrateLnPayment",
```
dolcalmi, line 47: also, just for future spans... please add `debug` as namespace, `migrateLnPayment` as code function and `debug.migrateLnPayment` as a span name

src/services/lnd/index.ts
vindard: I can't remember the details now, but I do remember there was some strange edge-case somewhere where we didn't have this info and I needed it to show up in the error

src/services/lnd/index.ts
vindard: https://github.com/GaloyMoney/galoy/pull/1041#discussion_r816157413

src/debug/migrate-ln-payments-trackpaymentsv2.ts
vindard: Did you mean like this? https://github.com/GaloyMoney/galoy/pull/1041/commits/cc9e9900ceafcdf08375d62af5698ee3aef8e245

dolcalmi: yes

dolcalmi (dismissed):

dolcalmi (approved):

PR #1044, opened 2022-02-02
feat: populate preimage in ledger transactions from lnpayments

vindard:
## Description

This is the first proposal to populate the `revealedPreImage` property in `WalletTransaction` objects that eventually get returned out via the api under `settlementVia`.

The approach here was to:
- record the preimage on successful send to `lnpayments` collection
- record the preimage on update of settled pending  payment to `lnpayments` collection
- add the `revealedPreImage` property to `LedgerTransaction` type and implement with a join from the LedgerService requests to the `lnpayments` collection.

### Feedback

We decided that this approach isn't ideal for 2 reasons:
- the dependency from LedgerService on `lnpayments` collection is dirty
- the join approach may be unnecessary premature optimization

Our first alternative discussed was to simply add a new `revealedPreImage` property in the metadata for transactions.
- this solves the join complexity
- this is bad because we would have to edit transactions after-the-fact (on pending payment update) and we'd ideally like to have transactions be immutable

Our second alternative was to create a new `transactions_metadata` collection to store mutable metadata related to transactions to:
- this would live within the Ledger service and solves the crossing-boundaries issue
- this would solve the "immutable transactions" issue because we would only update this other collection
- we will do 2 separate calls instead of a single join for simplicity to start, and will re-visit alternatives if performance becomes an issue

## Conclusion

This PR will be closed in favour a new one where we implement the new `transactions_metadata` collection and write/read the `revealedPreImage` property from there.

### DIFF
--- a/src/app/wallets/ln-send-payment.ts
+++ b/src/app/wallets/ln-send-payment.ts
@@ -455,11 +455,17 @@ const executePaymentViaLn = async ({
           })
 
       // Fire-and-forget update to 'lnPayments' collection
-      LnPaymentsRepository().persistNew({
-        paymentHash: decodedInvoice.paymentHash,
-        paymentRequest: decodedInvoice.paymentRequest,
-        sentFromPubkey: rawRoute ? pubkey : lndService.defaultPubkey(),
-      })
+      if (!(payResult instanceof LnAlreadyPaidError)) {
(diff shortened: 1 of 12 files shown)


vindard:
Note: the first part of this work was originally targeted in PR #1037 but that PR has since been rebased to prep for the new "metadata collection" approach.

PR #1046, opened 2022-02-02
chore: update price proto

dolcalmi:
Add currency param to price proto.


src/services/price/protos/price.proto
```diff
@@ -8,4 +8,6 @@ message PriceResponse {
   float price = 1;
 }
 
-message PriceQuery {}
+message PriceQuery {
+  string currency = 1;
```
nicolasburtey, line 12: shouldn't this be an `enum`?

PR #1047, opened 2022-02-03
feat: Add dealer price service

sebastienverreault:


src/domain/dealer-price/index.types.d.ts
```diff
@@ -0,0 +1,18 @@
...
+interface IDealerPriceService {
+  getExchangeRateForImmediateUsdBuy(
+    amountInSatoshis: Satoshis,
+  ): Promise<UsdCents | DealerPriceServiceError>
+  getExchangeRateForImmediateUsdSell(
+    amountInUsd: UsdCents,
+  ): Promise<Satoshis | DealerPriceServiceError>
+  getExchangeRateForFutureUsdBuy(
+    amountInSatoshis: Satoshis,
+    timeToExpiryInMinutes: Minutes,
```
nicolasburtey, line 12: I think we should have seconds timeframe instead of minutes here.

src/services/dealer-price/index.ts
```diff
@@ -0,0 +1,122 @@
...
+    } catch (error) {
+      baseLogger.error(
+        { error },
+        "GetExchangeRateForImmediateUsdBuy unable to fetch price",
+      )
+      return new UnknownDealerPriceServiceError()
+    }
+  }
+
+  const getExchangeRateForFutureUsdBuy = async function (
```
nicolasburtey, line 77: should we use the wording Option instead of Future? Option hint at the optionality of the transaction. it's not like a `forward` contract where the transaction will happen in the future no matter what.

src/services/dealer-price/proto/services/price/v1/price_service_pb.d.ts
```diff
@@ -0,0 +1,245 @@
+// package: services.price.v1
+// file: services/price/v1/price_service.proto
+
+/* tslint:disable */
+/* eslint-disable */
+
+import * as jspb from "google-protobuf"
+
+export class GetExchangeRateForImmediateUsdBuyRequest extends jspb.Message {
+  getAmountInSatoshis(): number
```
nicolasburtey, line 10: what is the goal of this function that don't take any input argument? versus the function below?

nicolasburtey: if this is because the argument is optional, we probably want to make it mandatory (not an expert in grpc but I guess we could do that)

nicolasburtey (commented):
skimmed over the PR, have some questions

src/domain/dealer-price/index.types.d.ts
sebastienverreault: Ok I don't mind but the at 1 minute time frame the expiry is so small that I can't price and option with non-zero premium. I will change it but it is likely too granular.

src/services/dealer-price/index.ts
sebastienverreault: I'm not using the instrument's name "Futures" here, just saying it is for *later* settlement. Option/Forward/Futures are too specific and the action is really about getting the result of conversion according to a rate for now or later...
If you have a better wording, I'm ready to change them all.

src/services/dealer-price/proto/services/price/v1/price_service_pb.d.ts
sebastienverreault: These are 4 generated code files in that folder from the one proto file I defined. It exist to statically strongly type the grpcs. Will also validate versioning if ever it evolves.

src/services/dealer-price/index.ts
nicolasburtey: my point is that: `GetExchangeRateForImmediateUsdBuyRequest` is like a market order on how it will be handle in the backend. 

the dealer can assume that every time this function is been called, there will be an associated transaction

for the `getExchangeRateForFutureUsdBuy`, it's really a quote. there won't be necessarily a transaction associated to it. 

whether it's `quote` or `option` or something else, I'm not sure, but here the wording doens't really show the different between the 2 types of request (in my opinion at least) which is why I like the `option`. maybe you prefer `quote`?

sebastienverreault: Yeah not sure, I just wanted to convey that one is a request for the immediate usd buy/sell "exchange rate" or quote while the other is for a future/later/delayed buy/sell "exchange rate" or quote and saying Option like the instrument is kind of leaking the implementation.
The original naming was confusing me so I changed them but really don't mind swapping for something else.
Just leave the 4 names in the comment and I will replace them.

src/domain/dealer-price/index.types.d.ts
sebastienverreault: My point was that, given the conditions this operates in, there is objectively no material differences in option prices for changes in expiry in seconds. So 0s, 15s, 60s will all have the same premium and therefore exchange rate returned. So I felt that by using seconds we imply granularity and control that does not materially exist.

nicolasburtey: have you modeled the implied price during high volatility event (with vol at > 300% to 1000%?). I think in 99% of the time, the time value of the option will be close to zero... unless the market is moving quickly, which is precisely when the option premium should go drastically up.

src/services/dealer-price/proto/services/price/v1/price_service_pb.d.ts
nicolasburtey: yeah looked at it and it seems there is no way in GRPC to make an argument mandatory (unlike Graphql) so this is why those 2 methods are being generated

src/services/dealer-price/index.ts
```diff
@@ -0,0 +1,116 @@
...
+  credentials.createInsecure(),
+)
+
+const clientGetExchangeRateForImmediateUsdBuy = util.promisify(
+  client.getExchangeRateForImmediateUsdBuy.bind(client),
+)
+const clientGetExchangeRateForImmediateUsdSell = util.promisify(
+  client.getExchangeRateForImmediateUsdSell.bind(client),
+)
+const clientGetExchangeRateForFutureUsdBuy = util.promisify(
```
nicolasburtey, line 33: maybe `clientGetQuoteRateForFutureUsdBuy` is a good name?

if you don't like the use of option, I think the wording `quote` convey the optionality?

src/services/dealer-price/index.ts
```diff
@@ -0,0 +1,116 @@
...
+
+const serverPort = process.env.PRICE_SERVER_PORT ?? "50055"
+const client = new PriceServiceClient(
+  `localhost:${serverPort}`,
+  credentials.createInsecure(),
+)
+
+const clientGetExchangeRateForImmediateUsdBuy = util.promisify(
+  client.getExchangeRateForImmediateUsdBuy.bind(client),
+)
```
vindard, line 29: We could type the response at the declaration for these here instead of casting them below

```suggestion
const clientGetExchangeRateForImmediateUsdBuy = util.promisify<
  GetExchangeRateForImmediateUsdBuyRequest,
  GetExchangeRateForImmediateUsdBuyResponse
>(client.getExchangeRateForImmediateUsdBuy.bind(client))
)
```

src/services/dealer-price/index.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  const getExchangeRateForImmediateUsdSell = async function (
+    amountInUsd: UsdCents,
+  ): Promise<Satoshis | DealerPriceServiceError> {
+    try {
+      const response = (await clientGetExchangeRateForImmediateUsdSell(
+        new GetExchangeRateForImmediateUsdSellRequest().setAmountInUsd(amountInUsd),
+      )) as GetExchangeRateForImmediateUsdSellResponse
+      return response.getPriceInSatoshis() as Satoshis
```
vindard, line 67: For `Satoshi`-casting we've been using the `toSats` method to make it easier if we wanted to add any validation later on across the codebase.

```suggestion
      return toSats(response.getPriceInSatoshis())
```

src/services/dealer-price/index.ts
```diff
@@ -0,0 +1,116 @@
...
+          amountInSatoshis,
+        ),
+      )) as GetExchangeRateForImmediateUsdBuyResponse
+      return response.getPriceInUsd() as UsdCents
+    } catch (error) {
+      baseLogger.error(
+        { error },
+        "GetExchangeRateForImmediateUsdBuy unable to fetch price",
+      )
+      return new UnknownDealerPriceServiceError()
```
vindard, line 56: For `new Unknown...Error` we've so far been passing the caught error in for better visibility higher up in the code.
```suggestion
      return new UnknownDealerPriceServiceError(error)
```

vindard (commented):
Few little nits to bring some of the typing/error-handling in line with the rest of the codebase

vindard (changes requested):
We will need to add the new `DealerPriceServiceError` types to these files:
- Imported into `src/app/errors.ts`
- Added to switch in `src/graphql/error-map.ts`

src/services/dealer-price/index.ts
sebastienverreault: well spotted thank you for all the above!

src/services/dealer-price/index.ts
```diff
@@ -0,0 +1,179 @@
...
+  const getExchangeRateForImmediateUsdBuy = async function (
+    amountInSatoshis: Satoshis,
+  ): Promise<UsdCents | DealerPriceServiceError> {
+    try {
+      const response = await clientGetExchangeRateForImmediateUsdBuy(
+        new GetExchangeRateForImmediateUsdBuyRequest().setAmountInSatoshis(
+          amountInSatoshis,
+        ),
+      )
+      return response.getAmountInUsd() as UsdCents
```
vindard, line 73: I just noticed from another PR that we have a `toCents` method now as well (similar to `toSats`). Sorry I missed this before 🙈

vindard (changes requested):

src/services/dealer-price/index.ts
sebastienverreault: Yeah no worries I had to do a second pass to convert all to Cents on that interface anyways so very timely! Thx

src/domain/dealer-price/index.types.d.ts
```diff
@@ -0,0 +1,34 @@
...
+  ): Promise<Satoshis | DealerPriceServiceError>
+  getSatsFromCentsForFutureBuy(
+    amountInUsd: UsdCents,
+    timeToExpiryInSeconds: Seconds,
+  ): Promise<Satoshis | DealerPriceServiceError>
+  getSatsFromCentsForFutureSell(
+    amountInUsd: UsdCents,
+    timeToExpiryInSeconds: Seconds,
+  ): Promise<Satoshis | DealerPriceServiceError>
+  getCentsPerBtcExchangeMidRate(): Promise<UsdCents | DealerPriceServiceError>
```
nicolasburtey, line 33: I think the type here is not `UsdCents`. It should be a `Ratio`. `RatioCentsPerSats` maybe?

nicolasburtey:
just one comment, but looks good otherwise @sebastienverreault

src/domain/wallets/amount-converter.ts
```diff
@@ -91,11 +100,12 @@ export const AmountConverter = ({
 
       if (walletCurrency === WalletCurrency.Btc) {
         return new NotImplementedError(
-          "unsupposed use cas: walletCurrency = Btc, send with cents",
+          "unsupported use cas: walletCurrency = Btc, send with cents",
```
nicolasburtey, line 103:
```suggestion
          "unsupported use case: walletCurrency = Btc, send with cents",
```

src/domain/wallets/amount-converter.ts
```diff
@@ -43,11 +47,15 @@ export const AmountConverter = ({
 
       if (walletCurrency === WalletCurrency.Btc) {
         return new NotImplementedError(
-          "unsupposed use cas: walletCurrency = Btc, receive from cents",
+          "unsupported use cas: walletCurrency = Btc, receive from cents",
```
nicolasburtey, line 50:
```suggestion
          "unsupported use case: walletCurrency = Btc, receive from cents",
```

nicolasburtey (approved):

PR #1054, opened 2022-02-05
fix: user specific price subscriptions

UncleSamtoshi:
Currently all users receive eachothers price subscription publishes. This fix scopes immediatePublishes to a single user.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,7 @@
     "redlock": "^4.2.0",
     "subscriptions-transport-ws": "^0.11.0",
     "twilio": "^3.65.0",
+    "uuid": "^8.3.2",
     "uuid-by-string": "^3.0.4"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -94,7 +94,6 @@
     "redlock": "^4.2.0",
     "subscriptions-transport-ws": "^0.11.0",
     "twilio": "^3.65.0",
-    "uuid": "^8.3.2",
     "uuid-by-string": "^3.0.4"
   },
   "devDependencies": {
--- a/src/graphql/root/subscription/price.ts
+++ b/src/graphql/root/subscription/price.ts
@@ -1,3 +1,5 @@
+import crypto from "crypto"
+
 import { GT } from "@graphql/index"
 import ExchangeCurrencyUnit from "@graphql/types/scalar/exchange-currency-unit"
 import PricePayload from "@graphql/types/payload/price"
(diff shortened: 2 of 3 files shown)


src/graphql/root/subscription/price.ts
```diff
@@ -39,34 +40,36 @@ const PriceSubscription = {
...
+      pubsub.publishImmediate(immediateEventName, {
         errors: [{ message: "Unsupported exchange amount" }],
       })
     } else {
       const satUsdPrice = await Prices.getCurrentPrice()
       if (!(satUsdPrice instanceof Error)) {
-        pubsub.publishImmediate(eventName, { satUsdCentPrice: 100 * satUsdPrice })
+        pubsub.publishImmediate(immediateEventName, { satUsdCentPrice: 100 * satUsdPrice })
       }
+      return pubsub.asyncIterator([immediateEventName, eventName])
```
samerbuna, line 69: This could be just the `immediateEventName`

src/graphql/root/subscription/price.ts
```diff
@@ -39,34 +40,36 @@ const PriceSubscription = {
...
     }
 
-    return pubsub.asyncIterator(eventName)
+    return pubsub.asyncIterator(immediateEventName)
```
samerbuna, line 72: I think this one should have both events.

UncleSamtoshi: I think it needs to just be the immediateEventName because in the case that the user received an immediate error for wrong inputs, they should continue to receive data for something they didnt ask for

src/graphql/root/subscription/price.ts
UncleSamtoshi: If we dont include eventName then they wont get updates every 30 seconds from the trigger

src/graphql/root/subscription/price.ts
samerbuna: I see but that means the subscription will only work if there's a valid initial response from the Prices repo, and would otherwise fail. Might be okay.

UncleSamtoshi: Its not inside the if statement, so I think even if the first price repo call is invalid it will still work in the future.

samerbuna (approved):

package.json
```diff
@@ -94,6 +94,7 @@
     "redlock": "^4.2.0",
     "subscriptions-transport-ws": "^0.11.0",
     "twilio": "^3.65.0",
+    "uuid": "^8.3.2",
```
dolcalmi, line 97: this library was removed recently, please use `crypto` package

dolcalmi (changes requested):

package.json
nicolasburtey: yup, we're aiming at reducing the number of library we depend on

nicolasburtey (approved):

PR #1055, opened 2022-02-05
fix: remove-deprecated-code

nicolasburtey:


vindard:
> I probably have removed some tests here, including some checks that might have been useful, so this PR should not be merged as is.

I added back the useful things. This should be ok now.

bodymindarts (approved):

PR #1056, opened 2022-02-05
fix: bug conversion?

nicolasburtey:
paymentAmount is type as Satoshis. If we want to go to MilliSats, then we want to divide by 1000, not the way around?


nicolasburtey:
nevermind. the actual code is good

PR #1059, opened 2022-02-06
feat: use ids instead of offsets for transaction cursors

samerbuna:


samerbuna:
@daviroo this is the PR that changes pagination to use id-based cursors and also makes edges types non nullable. I've tested this on the web-wallet and it's working great. We should also check to make sure that this does not break the current pagination we have for transactions on the mobile wallet.

daviroo (approved):
I tested this locally and it works ok with the current query for transactions in the mobile app.

PR #1061, opened 2022-02-06
feat: add AmountConverter

nicolasburtey:


src/domain/fiat/index.types.d.ts
```diff
@@ -1,2 +1,47 @@
...
+  cents: UsdCents
+}
+
+type XOR<T1, T2> =
+  | (T1 & { [k in Exclude<keyof T2, keyof T1>]?: never })
+  | (T2 & { [k in Exclude<keyof T1, keyof T2>]?: never })
+
+type CentsXORSats = XOR<AmountFromSatoshis, AmountFromCents>
+
+type OrderType = "immediate" | "quote"
```
bodymindarts, line 33: Can this be a `const` object (ie. the same pattern we use for all other string constants).

src/domain/fiat/index.types.d.ts
```diff
@@ -1,2 +1,47 @@
...
+
+type GetAmountsSendOrReceiveArgs = {
+  walletCurrency: WalletCurrency
+  order: OrderType
+} & CentsXORSats
```
bodymindarts, line 38: You are unifying 2 different call signatures into 1 quasi overloaded function. Cents XOR sats. In cases where:
- the calling code knows which variant is being called (it does)
- the function must use conditionals to separate code paths depending on which is being called (it does)
- there is no or very little shared code between the 2 code paths (there is none)
it is generally preferable to segregate the 2 signatures into separate functions.

In this case it would:
- simplify the type
- remove a conditional
- probably result in less code overall

src/domain/fiat/index.types.d.ts
```diff
@@ -1,2 +1,47 @@
...
+
+type GetAmountsSendOrReceiveRet =
+  | NotReachableError
+  | NotImplementedError
+  | {
```
bodymindarts, line 43: Up to know we have typically defined the 'happy path' type first and the error types after.

Also shouldn't it be a `Promise<>`?

bodymindarts (approved):
I'm approving to get all the renaming (use -> display currency) into the code and move forward.

I think the `AmountConverter.getAmountsReceive` function should be split into 2 - there is no reason for the additional cyclomatic complexity that unifying the call signature brings.

src/domain/fiat/index.types.d.ts
nicolasburtey: > looks like so far we define the type and then wrap it in Promise<Ret> in the function signature
indeed what I have done here

test/integration/app/wallets/invoice.spec.ts
```diff
@@ -131,27 +131,29 @@ describe("Wallet - addInvoice BTC", () => {
...
+describe("Wallet - addInvoice USD", () => {
+  it("add a self generated USD invoice", async () => {
+    const displayCurrencyPerSat = await getCurrentPrice()
+    if (displayCurrencyPerSat instanceof Error) return displayCurrencyPerSat
 
-    const fiatInput = 100
-    const sats = fiatInput / price
+    const centsInput = 10000
+
+    const usdPerCent = displayCurrencyPerSat * 100
```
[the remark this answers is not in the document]
nicolasburtey: don't think we should have things in `src/config/index.ts` related to tests: we don't want to load those variables in memory for production when there won't be used. 

they could be a in helper function under test tough

nicolasburtey:
> I think the AmountConverter.getAmountsReceive function should be split into 2 - there is no reason for the additional cyclomatic complexity that unifying the call signature brings.

when you say split in 2, on what variables do you have in mind? "quote" versus "immediate"?

bodymindarts:
> when you say split in 2, on what variables do you have in mind? "quote" versus "immediate"?

No on the `sats` vs `cents` - they are mutually exclusive and follow different code paths in the function.

nicolasburtey:
> > when you say split in 2, on what variables do you have in mind? "quote" versus "immediate"?
> 
> No on the `sats` vs `cents` - they are mutually exclusive and follow different code paths in the function.

but then the complexity will be back in the app layer I think. need to complete the integration to see what really make sense, and eventually do a refactor once the integration has been completely one time

bodymindarts:
> but then the complexity will be back in the app layer I think. need to complete the integration to see what really make sense, and eventually do a refactor once the integration has been completely one time

Perhaps I don't understand the way you intend to use this - with an XOR type the app layer must already know wether it has sats or cents. Right? So unifying the interface in this way doesn't change anything in the app layer.

bodymindarts (approved):

PR #1065, opened 2022-02-07
fix: Price server return SatPerUsd, not UsdPerSat

nicolasburtey:
price returned for instance is 0.00042

this is: 0.00042 = Sat/Usd, or 0.00042 Usd = 1 Sat


bodymindarts:
> this is: 0.00042 = Sat/Usd, or 0.00042 Usd = 1 Sat

if `0.00042 Usd = 1 Sat` then that is `usd/sat` or `2380.95 sat/usd` (read 2380 satoshis for 1 US dollar)
Isn't it?

nicolasburtey:
0.00042 * Usd = 1 * Sat
0.00042 = 1 * Sat / Usd
0.00042 = Sat / Usd

I guess the question is, is it : `0.00042 = Sat / Usd` or `0.00042 Sat / Usd` (the signed are inversed if we don't have the = between the value and currencies)

PR #1066, opened 2022-02-07
fix: onchain can be null in prod

bodymindarts:


src/services/mongoose/wallets.ts
```diff
@@ -17,7 +17,8 @@ export interface WalletRecord {
   _accountId: Types.ObjectId
   type: string
   currency: string
-  onchain: OnChainMongooseType[]
+  // TODO: even though mongoose should set a default this value has been observed to be null in prod
+  onchain: OnChainMongooseType[] | null
```
nicolasburtey, line 21: use undefined?

bodymindarts: I have updated the type to reflect reality in production - there it is `null`.

bodymindarts:
I am closing this in favour of fixing the data inconsistency in production. I could not find any other environment (both staging and clients) that had this inconsistency.

PR #1067, opened 2022-02-07
ci: release after tests pass, fix blank `appVersion` and added slack

sandipndev:
https://github.com/GaloyMoney/charts/pull/577 was a bad PR caused because `version` was not present as an input to the `bump-image-in-chart` job, this PR fixes it.

It also adds the `release` job to the `galoy` group for visibility and uses the `on_failure` hook to notify if release went wrong.

Have updated the pipeline.

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -344,6 +344,7 @@ jobs:
       - name: migrate-edge-image
       - name: pipeline-tasks
       - name: charts-repo
+      - name: version
       outputs:
       - name: charts-repo
       params:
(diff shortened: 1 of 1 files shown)


bodymindarts (dismissed):

bodymindarts:
I think these commits should be prefixed with `ci` instead of `fix`

sandipndev:
Yes, I thought Squash Commits only take message on PR but I guess, the description will keep these messages.
Can change the commit messages, just a moment.

PR #1069, opened 2022-02-08
refactor: separate e2e tests from integration

UncleSamtoshi:
- Move server tests into a new e2e testing category
- Create a way for initializing the state of the testing system
- Update CI pipeline and GitHub actions to test e2e


Makefile
```diff
@@ -51,6 +51,15 @@ watch-unit:
 watch-compile:
 	$(BIN_DIR)/tsc --watch  --noEmit --skipLibCheck
 
+e2e:
+	yarn build && \
+	yarn test:e2e
+	
+e2e-in-ci:
+	. ./.envrc && \
+	yarn build && \
```
bodymindarts, line 60: Why do we need to `build` here?

We are `build`ing in the `integration-in-ci` task because we are also test-executing the cronjob (which relies on the code having been compiled).
Is there a dependency in this job on code that was compiled?

ci/pipeline.yml
```diff
@@ -109,6 +110,35 @@ jobs:
...
+    timeout: 12m
+    tags: ["galoy-staging"]
+    config:
+      platform: linux
+      image_resource: #@ task_image_config()
+      inputs:
+      - name: bundled-deps
+      - name: pipeline-tasks
+      - name: repo
+        path: galoy
```
bodymindarts, line 130: We need to give this a different path so that the `docker compose` prefixes on the docker machine don't collide.
This should probably be `galoy-e2e` and the integration test job can be updated to `galoy-integration` path.

ci/pipeline.yml
```diff
@@ -109,6 +110,35 @@ jobs:
...
+      params:
+        REPO_PATH: galoy
```
bodymindarts, line 132: Mirror the path (see comment above) `galoy-e2e`

ci/tasks/test-e2e.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
bodymindarts, line 1: We are re-using this `sync-to-docker-vm` code in many places. Its probably a candidate to extract into a helper script and bring in from `concourse-shared` repo cc @sandipndev 

Would leave it like this for this PR as that is an orthogonal task.

bodymindarts:
tests are failing locally on Mac:
```
  ● Test suite failed to run

    pid must be a number

      25 |
      26 | export const killServer = async (serverPid: PID): Promise<void> =>
    > 27 |   killAsync(serverPid, "SIGKILL")
         |   ^
      28 |

      at Object.<anonymous>.module.exports (node_modules/tree-kill/index.js:16:29)
      at killServer (test/helpers/integration-server.ts:27:3)
      at test/e2e/servers/graphql-main-server/galoy-pay-testing.spec.ts:68:19

Test Suites: 1 failed, 1 passed, 2 of 4 total
Tests:       12 failed, 15 passed, 27 total
Snapshots:   0 total
Time:        67.667 s
Ran all test suites.
✨  Done in 72.14s.
```

ci/pipeline.yml
UncleSamtoshi: Not quite sure what you mean by this, don't they both have the same repository input with `path: galoy`

bodymindarts: They do - which is a problem because the path determines the scope of the `docker compose up` container naming. If they have the same path the 2 contexts will interfere with each other

bodymindarts: We have a shared singleton docker vm where things get run - its not isolated between concourse jobs

bodymindarts (approved):

PR #1071, opened 2022-02-09
test: check transactions are sorted in reverse chronologically

vindard:
## Description

This is an implicit condition I discovered (internal to Medici) in PR #1044 that I thought we should capture explicitly in our tests.

### DIFF
--- a/test/integration/02-user-wallet/02-tx-ordering.spec.ts
+++ b/test/integration/02-user-wallet/02-tx-ordering.spec.ts
@@ -1,0 +1,35 @@
+import { Wallets } from "@app"
+
+import {
+  createUserWalletFromUserRef,
+  getDefaultWalletIdByTestUserRef,
+} from "test/helpers"
+
+let walletIdA: WalletId
+
+beforeAll(async () => {
+  await createUserWalletFromUserRef("A")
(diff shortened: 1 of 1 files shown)


test/integration/02-user-wallet/02-tx-ordering.spec.ts
```diff
@@ -0,0 +1,71 @@
+import { Wallets } from "@app"
```
nicolasburtey, line 1: test in `02-user-wallet` are meant to be part of the "globally consistent tests suite", ie with the check that the double entry ledger is valid (`checkIsBalanced`)

I don't think this test needs to be in 02 folder for the reason we're not checking anything related to the double entry ledger aspect here

vindard: Makes sense, but I'm also bootstrapping off of that by the time this test runs the wallet would have had a number of send/receive intraledger/offchain/onchain transactions already recorded to test against.

Is this enough of a reason to keep it here, or should I move this and recreate some of those transactions explicitly?

nicolasburtey: you can decide

vindard:
Closing this as stale... we already implicitly test for this

PR #1077, opened 2022-02-11
feat: combine record from new metadata collection into domain LedgerTransaction type

vindard:
## Description

_Builds on #1037_

This PR adds records from the new `medici_transaction_metadata` collection to `LedgerTransaction` objects returned from various queries in the Ledger service.

### Challenge
**How do we deal with partial failures since we are now doing 2 queries to build a single `LedgerTransaction`?:**
- ~my current approach is to only error on the first request to main collection, and to concatenate empty objects if the 2nd metadata query errors~
- ~a better approach may be to re-use the `PartialResult` type to indicate if the LedgerTransaction being returned is a partial because of some failure on the second request. Do we want to do this?~

- We decided to report & not return errors for the metadata request

### Open `TODO`s
- [x] Report errors on metadata request errors
- [x] Double-check unit testing for translation methods
- [x] Include metadata fetch in integration test

### DIFF
--- a/src/domain/wallets/tx-history.ts
+++ b/src/domain/wallets/tx-history.ts
@@ -138,7 +138,7 @@ export const fromLedger = (
             settlementVia: {
               type: SettlementMethod.IntraLedger,
               counterPartyWalletId: recipientWalletId as WalletId,
-              counterPartyUsername: username || null,
+              counterPartyUsername: username,
             },
           }
           return walletTransaction
@@ -169,7 +169,7 @@ export const fromLedger = (
             settlementVia: {
               type: SettlementMethod.IntraLedger,
               counterPartyWalletId: recipientWalletId as WalletId,
-              counterPartyUsername: username || null,
+              counterPartyUsername: username,
             },
           }
           return walletTransaction
@@ -201,7 +201,7 @@ export const fromLedger = (
         settlementVia: {
           type: SettlementMethod.IntraLedger,
           counterPartyWalletId: recipientWalletId as WalletId,
-          counterPartyUsername: username || null,
+          counterPartyUsername: username,
         },
       }
       return walletTransaction
--- a/src/domain/wallets/tx-history.ts
+++ b/src/domain/wallets/tx-history.ts
@@ -66,6 +66,7 @@ export const fromLedger = (
       paymentHash,
       txHash,
       pubkey,
+      revealedPreImage,
       username,
       address,
       pendingConfirmation,
@@ -184,7 +185,7 @@ export const fromLedger = (
             },
             settlementVia: {
               type: SettlementMethod.Lightning,
-              revealedPreImage: null,
+              revealedPreImage,
             },
           }
           return walletTransaction
--- a/src/services/ledger/index.ts
+++ b/src/services/ledger/index.ts
@@ -22,6 +22,7 @@ import {
   BigIntFloatConversionError,
   ErrorLevel,
   paymentAmountFromNumber,
+  WalletCurrency,
 } from "@domain/shared"
 import { toObjectId } from "@services/mongoose/utils"
 import {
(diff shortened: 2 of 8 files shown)


src/domain/wallets/tx-history.ts
```diff
@@ -118,7 +119,7 @@ export const fromLedger = (
             settlementVia: {
               type: SettlementMethod.IntraLedger,
               counterPartyWalletId: recipientWalletId as WalletId,
-              counterPartyUsername: username || null,
+              counterPartyUsername: username,
```
dolcalmi, line 122: are we sure we dont have null values in db?

dolcalmi: did you test this at graphql level?

src/services/ledger/index.ts
```diff
@@ -60,30 +63,34 @@ export const LedgerService = (): ILedgerService => {
...
+      txnRecord = results[0]
+      if (results.length !== 1) {
+        return new CouldNotFindTransactionError()
       }
-      return new CouldNotFindTransactionError()
     } catch (err) {
       return new UnknownLedgerError(err)
     }
+
+    return translateTxRecordsToLedgerTxs([txnRecord])[0]
```
dolcalmi, line 81: are you sure this cant throw an  exception? I guess is better to include it in the try block

dolcalmi (changes requested):

src/services/ledger/index.ts
vindard: Fixed

src/domain/wallets/tx-history.ts
vindard: How do you mean? We're casting falsy values to `undefined` as they're being read out of the database either way [here](https://github.com/GaloyMoney/galoy/pull/1077/files#diff-463de0ec8400f9be4647cb42dee40f257b5055c84cbe9328daa260760f4a1904R422)

src/services/ledger/index.ts
```diff
@@ -305,31 +317,78 @@ export const LedgerService = (): ILedgerService => {
...
+  }
+  return { ...fromTxnRecord, ...fromTxnMetadata }
+}
+
+const translateTxRecordsToLedgerTxs = async (
+  txnRecords: TransactionRecord[],
+): Promise<LedgerTransaction[]> => {
+  const txnsMetadata = await Promise.all(
+    txnRecords.map(async (tx): Promise<LedgerTransactionMetadata | undefined> => {
+      const txnMetadataResult = await txnMetadataRepo.findById(
```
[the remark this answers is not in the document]
vindard: We could, it would be the same as implementing a similar join [to this](https://github.com/GaloyMoney/galoy/pull/1044/files#diff-463de0ec8400f9be4647cb42dee40f257b5055c84cbe9328daa260760f4a1904R285) here in-code (it look slike a `View` is just an aggregation pipeline [computed on-demand](https://www.mongodb.com/docs/manual/core/views/)).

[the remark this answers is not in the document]
vindard: We could, it would be the same as implementing a similar join [to this](https://github.com/GaloyMoney/galoy/pull/1044/files#diff-463de0ec8400f9be4647cb42dee40f257b5055c84cbe9328daa260760f4a1904R285) here in-code (it looks like a `View` is just an aggregation pipeline [computed on-demand](https://www.mongodb.com/docs/manual/core/views/)).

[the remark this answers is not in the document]
vindard: Ok I implemented here as an option: https://github.com/GaloyMoney/galoy/compare/pass-preimage-to-ledger-transactions...optimize-metadata-join

It's not ideal though because we're moving away from using the `Mainbook.ledger` query which means we'd have to ensure it's faithfully reproduced.

I'd be in favour of merging this version of the code and then profiling how much longer the extra `transactions` gql query takes in staging before we consider optimizing it (or reverting if there's no good solution) like we suggested in #1044

vindard:
Taken over by PR #1427 where the metadata collection fetch is implemented more selectively.

PR #1078, opened 2022-02-11
feat: use empty string for default language for easier client defaulting

samerbuna:


daviroo (changes requested):
I had a quick look at the logic we use in the app  and I think this shouldn't effect the current behaviour.  I would prefer to test the app against this branch before it moves to production so I'll reject until I have tested (should be later tonight).

daviroo (approved):
Tested with the app in local dev.  Works good 👍

PR #1079, opened 2022-02-12
feat: move limits to UsdCents

nicolasburtey:
The PR change how the limits for the app to use Usd instead of Btc.

Beyond the e2e and cron tests issues, this PR is ready for review

Prior to this PR, every limit was done in sats. With the introduce of cents, it make more sense to change the limit and have everything communicated in cents/USD. so that if the wallet USD, then we need to conversion, and if the wallet is BTC, we do a conversion with the current price. 

Some comments:
- I had to _abuse_ a little bit the `getCurrentPrice`/`DisplayCurrencyConverter`. this is supposed to be used only for DisplayCurrency. because the assumption is displayCurrency will be Usd in the foreseeing future, I mix the types. 
The way we should handle that is that the Dealer should also return the mid bid/ask price. The dealer is one returning the UsdCents. but there is no api for that, and using the Dealer would create a dependency on the dealer being ready, which is why I've used `DisplayCurrencyConverter.fromCentstoSats` and `DisplayCurrencyConverter.fromSatsToCents` for now. 
- memoSatsThreshold anti spam filter still needs to be USD denominated.
- I introduce the pattern of add/sub to make sure we're not mixing UsdCents and Satoshis. The type is not great, because we can't chain it. I have not found the way to do `varSat1.add(varSat2)` instead of `add(varSats1, varSat2)`


package.json
```diff
@@ -144,16 +143,16 @@
     "grpc-tools": "^1.11.2",
     "grpc_tools_node_protoc_ts": "^5.3.2",
     "http-terminator": "^3.0.3",
-    "jest": "^27.5.1",
-    "jest-junit": "^13.0.0",
+    "jest": "^27.0.6",
```
bodymindarts, line 146: Is this downgrade on purpose?

package.json
```diff
@@ -144,16 +143,16 @@
...
+    "jest-junit": "^12.2.0",
     "migrate-mongo": "=8.2.2",
     "newman": "^5.2.4",
     "nodemon": "^2.0.15",
     "pino-pretty": "^5.1.0",
     "prettier": "^2.5.1",
     "spectaql": "^0.11.0",
     "tree-kill": "^1.2.2",
-    "ts-jest": "^27.1.3",
+    "ts-jest": "^27.0.3",
```
bodymindarts, line 155: downgrade?

src/app/wallets/check-limit-helpers.ts
```diff
@@ -1,43 +1,57 @@
...
 
 export const checkIntraledgerLimits = async ({
   amount,
   walletId,
+  walletCurrency,
   account,
+  dCConverter,
 }: {
-  amount: Satoshis
+  amount: CurrencyBaseAmount
```
bodymindarts, line 16: Shouldn't this be USDCents?

src/app/wallets/update-pending-invoices.ts
```diff
@@ -139,21 +140,39 @@ const updatePendingInvoice = async ({
       return true
     }
 
-    const invoicePaid = await walletInvoicesRepo.markAsPaid(paymentHash)
-    if (invoicePaid instanceof Error) return invoicePaid
-
     const displayCurrencyPerSat = await getCurrentPrice()
     if (displayCurrencyPerSat instanceof Error) return displayCurrencyPerSat
 
+    // TODO: this should be a in a mongodb transaction session with the ledger transaction below
```
bodymindarts, line 146: Currently we are treating the ledger as a discrete 'sub domain'. We should not have transactions across sub domain boundaries. If we need transactional consistency guarantees (ie atomic updates) across these entities then we have a design flaw.

My feeling is we should execute the `markAsPaid` call _after_ the ledger update (as ledger is the source of truth). And I don't see an issue with it having eventual consistency.

Not a TODO for this PR - just reflecting on the comment

src/app/wallets/update-pending-invoices.ts
```diff
@@ -139,21 +140,39 @@ const updatePendingInvoice = async ({
...
+    if (walletCurrency === WalletCurrency.Usd && !cents) {
+      const converter = AmountConverter({
+        dCConverter,
+        dealerFns: DealerPriceService(),
+      })
+      const walletCurrency = WalletCurrency.Usd
+      const amounts = await converter.getAmountsReceive({
+        walletCurrency,
+        sats: roundedDownReceived,
+        order: "immediate",
```
bodymindarts, line 171: Please don't use magic strings. Reference constants like `OrderType.Immediate`

src/domain/accounts/index.types.d.ts
```diff
@@ -11,6 +11,27 @@ type DepositFeeRatio = number & { readonly brand: unique symbol }
 
 type ContactAlias = string & { readonly brand: unique symbol }
 
+type AccountLimitsArgs = {
+  level: AccountLevel
+  accountLimits?: {
+    intraLedger: {
+      level: {
+        [l: number]: UsdCents
```
bodymindarts, line 19: What about level 2?

src/domain/fiat/index.ts
```diff
@@ -15,3 +15,8 @@ export const OrderType = {
   Locked: "immediate",
   Active: "quote",
 } as const
+
+export const CENTS_PER_USD = 100
+
+export const add = <T extends number>(arg0: T, arg1: T): T => (arg0 + arg1) as T
+export const sub = <T extends number>(arg0: T, arg1: T): T => (arg0 - arg1) as T
```
bodymindarts, line 22: Careful - our `checkedToSats` method ensures that `sats > 0`. Casting here breaks that validation. Need to consider where all we are depending on sats being `>0` and wether or not this can be an issue or else we need to re-check in this method. Actually perhaps we should recheck regardless and throw an error (instead of returning) so that we will get visibility incase this breaks down at runtime.

bodymindarts (commented):
Looks okay overall. I'm confused about the `CurrencyBaseAmount` - what does this type signify? Do we have 4 currency types? `Satoshis`, `USDCents`, `DisplayCurrencyBase` and `CurrencyBaseAmount`?.

I think we should eventually move to a more complex amount representation where we encode the currency in the value. It would increase type safety and require less values to be passed around. Something like :

```
type MoneyAmount<T> = {
  currency: T
  baseUnit: BigInt
}
```

Could be part of the `BigInt` refactoring.

Other than that I don't like the whole logic around the limit checking. I once tried to do a refactoring to simplify it but got stuck and gave up. I think a core issue is due to it being combined with the 2FA even though that is a different use case.
That being said I think its fine to the way you've coded it here following the pattern of the existing implementation. Eventually it'll need a cleanup.

All in all I think there are just a few small issues to take care of before merging.

package.json
nicolasburtey: merge issue

package.json
nicolasburtey: merge issue

src/app/wallets/check-limit-helpers.ts
nicolasburtey: > `Satoshis`, `USDCents`, `DisplayCurrencyBase` and `CurrencyBaseAmount`?.

so `CurrencyBaseAmount` is not really an additional types as it is for the 3 others:

```
type CurrencyBaseAmount = Satoshis | UsdCents
```

This means that the value attached to it can either be a Satoshis or a UsdCents.

There are many scenario in which this arise. For instance, when we read an entry from the ledger, we can't say it's a `Satoshis` any more, it's currency dependent. so it should either be resolves to UsdCents or Satoshis later down where more context is available. 

It's the same logic with the Limit checkers. The amount passed can be either, and the function will decided based on the currency wallet which denomination should be used for the amount based on walletCurrency.

Does that make sense?

nicolasburtey:
> I think we should eventually move to a more complex amount representation where we encode the currency in the value. It would increase type safety and require less values to be passed around. Something like :
> 
> ```
> type MoneyAmount<T> = {
>   currency: T
>   baseUnit: BigInt
> }
> ```

I like the idea of the interface a lot. (don't have to do it with the BigInt refactoring, I would actually avoid that because the bigInt refactor will be a big change so better to mix them up)

src/domain/fiat/index.ts
nicolasburtey: I tried to make it so that we could do `A.add(B).sub(C)` but wasn't able to code the type correctly. maybe you'll know how to do it?

nicolasburtey: I'll do the check you suggest for now.

src/domain/accounts/index.types.d.ts
nicolasburtey: the syntax says there can be any limits (as is), as long as the key is a number, this will pass.

src/app/wallets/check-limit-helpers.ts
bodymindarts: I think I understand - but this is IMO the wrong way to model it. We should be using a generic for these situations:
```
<T extends Satoshis | UsdCents> someFunction({amount}: {amount: T})
```

Not saying we have to introduce that now in this PR - but when we do the `BigInt` refactoring this should be in scope IMO.

src/domain/fiat/index.ts
bodymindarts: > I tried to make it so that we could do A.add(B).sub(C) but wasn't able to code the type correctly. maybe you'll know how to do it?

This is an additional indication for a more 'complex` amount value that would host that function:

```
type MoneyAmount<T extends UsdCents | Satoshis> = {
  currency: T extends UsdCents ? "USD" : "BTC"
  baseAmount: T
  add: (other: MoneyAmount<T>) -> MoneyAmount<T>
  sub: (other: MoneyAmount<T>) -> MoneyAmount<T>
}
```
etc.

nicolasburtey: ie: circular dependancies are not allowed

src/app/wallets/check-limit-helpers.ts
nicolasburtey: > `<T extends Satoshis | UsdCents> someFunction({amount}: {amount: T})`

we don't necessarily know which type is it as some point in the case. at least I have designed it as is.

ie: you get an amount from graphql, but the type can be either `Satoshis` or `UsdCents`. I guess this could be types at the same time we look into the wallet currency type. might be a better design indeed.

bodymindarts (dismissed):

bodymindarts (approved):

bodymindarts (approved):

PR #1086, opened 2022-02-14
chore: add cold storage instrumentation

dolcalmi:
When we query loop-out onchain transactions bitcoind returns a 500 error contributing to create false alerts because auto-instrumentation made by HttpInstrumentation.

- Add helper method to add instrumentation to service response
- Add parent attributes to HttpInstrumentation (this will allow to filter `Invalid or non-wallet transaction id` 500 errors from cron alerts)

### DIFF
--- a/src/services/cold-storage/index.ts
+++ b/src/services/cold-storage/index.ts
@@ -4,9 +4,11 @@ import { BTC_NETWORK, getBitcoinCoreRPCConfig, getColdStorageConfig } from "@con
 import {
   InsufficientBalanceForRebalanceError,
   InvalidCurrentColdStorageWalletServiceError,
+  InvalidOrNonWalletTransactionError,
   UnknownColdStorageServiceError,
 } from "@domain/cold-storage/errors"
 import { checkedToOnChainAddress } from "@domain/bitcoin/onchain"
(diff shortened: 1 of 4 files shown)


src/services/cold-storage/index.ts
```diff
@@ -104,6 +106,9 @@ export const ColdStorageService = async (): Promise<
       const { amount } = await bitcoindCurrentWalletClient.getTransaction(txHash)
       return amount < 0
     } catch (err) {
+      if (err && err.message === "Invalid or non-wallet transaction id") {
```
bodymindarts, line 109: Can't this be simplified to `err?.message === ""`

src/services/tracing.ts
```diff
@@ -124,6 +124,18 @@ registerInstrumentations({
           requestHeaders: ["apollographql-client-name", "apollographql-client-version"],
         },
       },
+      applyCustomAttributesOnSpan: (span, _request, response) => {
```
bodymindarts, line 127: Where is this being used? And why is the span being passed in not the active one?

bodymindarts (approved):

dolcalmi:
closed in favor of https://github.com/GaloyMoney/galoy/pull/1099

PR #1090, opened 2022-02-15
fix: yarn deps are not in sync

dolcalmi:
Fix yarn.lock issue. If you want to test current behavior please delete `yarn.lock` then `yarn install` and try to compile the app.

This PR fix tracing issue with graphql,  updates/removes some dependencies, update jest config and update log level in lnd.conf to improve performance


src/services/tracing.ts
```diff
@@ -30,7 +29,7 @@ propagation.setGlobalPropagator(new W3CTraceContextPropagator())
 
 // FYI this hook is executed BEFORE the `formatError` hook from apollo
 // The data.errors field here may still change before being returned to the client
-const gqlResponseHook = (span: Span, data: graphqlTypes.ExecutionResult) => {
+const gqlResponseHook = (span: Span, data) => {
```
dolcalmi, line 32: @opentelemetry/instrumentation-graphql use a different graphql version to validate types

bodymindarts (approved):

PR #1091, opened 2022-02-16
feat: add send/receive USD over lightning

nicolasburtey:


src/app/wallets/send-intraledger.ts
```diff
@@ -178,6 +182,11 @@ const executePaymentViaIntraledger = async ({
 
   const { amount, senderWallet, recipientWallet } = validationResult
 
+  // TODO Usd use case
+  assert(recipientWallet.currency === WalletCurrency.Btc)
```
vindard, line 186: From these top level use-cases we should probably still be returning some sort of error instead of throw with `assert`

src/app/wallets/add-invoice-for-wallet.ts
```diff
@@ -185,26 +185,16 @@ const addInvoiceFiatDenomiation = async ({
...
 
-  // TODO: ensure we don't get a stalled price // fail otherwise
-  const price = await getCurrentPrice()
-  if (price instanceof Error) return price
-
-  const dCConverter = DisplayCurrencyConverter(price)
-
   const dealer = DealerPriceService()
 
-  const amountConverter = AmountConverter({ dealerFns: dealer, dCConverter })
```
vindard, line 196: Can we delete `AmountConverter` and associated unit tests now that it seems we no longer use it?

vindard (commented):
Adding a few more comments

src/domain/bitcoin/lightning/invoice-expiration.ts
```diff
@@ -1,8 +1,11 @@
-import { SECS_PER_2_MINS, SECS_PER_DAY } from "@config"
+const SECS_PER_2_MINS = (60 * 2) as Seconds
```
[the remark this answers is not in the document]
nicolasburtey: static values are not really config

src/app/wallets/add-invoice-for-wallet.ts
nicolasburtey: yes I think it's safe to remove them. Maybe I would still keep the unit test to test the mock of the dealer

src/app/wallets/send-on-chain.ts
```diff
@@ -182,8 +181,13 @@ const executePaymentViaIntraledger = async ({
...
   // TODO Usd use case
-  assert(recipientWallet.currency === WalletCurrency.Btc)
-  assert(senderWallet.currency === WalletCurrency.Btc)
+  if (
+    !(
+      recipientWallet.currency === WalletCurrency.Btc &&
+      senderWallet.currency === WalletCurrency.Btc
+    )
+  )
+    return new NotImplementedError("USD intraledger")
```
nicolasburtey, line 190: we should probably have a bracket here. (when a `if` and the result are not on the same line, using bracket avoid merging issue. (ie: https://embeddedgurus.com/barr-code/2014/03/apples-gotofail-ssl-security-bug-was-easily-preventable/)

src/domain/bitcoin/lightning/invoice-expiration.ts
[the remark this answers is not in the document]
vindard: We have a bunch of other stativ values in there I think. Maybe we can make a `constants` file in an upcoming PR

src/app/wallets/add-invoice-for-wallet.ts
vindard: Ok we should leave for now then

vindard (approved):
Looks fine to me otherwise, approving in the interest of speed and that we'd need to test on staging.

PR #1097, opened 2022-02-17
fix: convert wallet volume to cents before comparing

vindard:
## Description

Limits are denominated in cents now, so wallet volumes must be converted (in addition to payment amount) before doing limit checks.

Before this, the defined cents limit would be counted as sats for any payments (a much lower threshold).


bodymindarts (approved):

vindard:
**Some tests from staging environment**

For the tests below, the external limits were exhausted first and then the intraledger limits from the same sender were tested.
- `span.num_events` with a value of `3` failed, `2` succeeded
- Base currency here is `cents` (`threshold` as well)
- `txVolume.outgoing` is in sats
- 

### External withdrawal limits

image


### Intraledger limits
image

PR #1099, opened 2022-02-17
chore: update error structure

dolcalmi:
- Update error structure, now all domain errors extends `DomainError`
- Add error level to `DomainError` (default to Info and Critical for Unknown extended errors)
- Update tracing service to support error level.
- All catch (throws) are recorded as `critical`
- Add first example on how to handle an exception with `isWithdrawalTransaction` (500 error from bitcoind)

### DIFF
--- a/src/services/tracing.ts
+++ b/src/services/tracing.ts
@@ -22,6 +22,7 @@ import {
   SpanStatusCode,
   SpanOptions,
   TimeInput,
+  Exception,
 } from "@opentelemetry/api"
 import { tracingConfig } from "@config"
 
@@ -50,13 +51,14 @@ const gqlResponseHook = (span: Span, data) => {
 const recordGqlErrors = ({ errors, span, subPathName }) => {
   const subPath = subPathName ? `${subPathName}.` : ""
 
-  span.recordException({
-    name: `graphql.${subPath}execution.error`,
-    message: JSON.stringify(errors),
-  })
-  span.setStatus({
-    code: SpanStatusCode.ERROR,
-  })
+  recordException(
+    span,
+    {
+      name: `graphql.${subPath}execution.error`,
+      message: JSON.stringify(errors),
+    },
+    "critical",
+  )
   const firstErr = errors[0]
   if (firstErr.message != "") {
     span.setAttribute(`graphql.${subPath}error.message`, firstErr.message)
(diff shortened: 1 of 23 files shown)


src/services/tracing.ts
```diff
@@ -194,18 +196,31 @@ export const addEventToCurrentSpan = (
   }
 }
 
+export const recordException = (span: Span, exception: Exception, level?: ErrorLevel) => {
+  const errorLevel = level || exception["level"] || "warn"
+  span.setAttribute("error.level", errorLevel)
+  span.recordException(exception)
+  if (errorLevel === "critical") span.setStatus({ code: SpanStatusCode.ERROR })
```
bodymindarts, line 203: I think `recordException` call automatically calls `setStatus`.

Also call the global value like `ErrorLevel.Critical` instead of `"critical"`.

src/services/tracing.ts
```diff
@@ -194,18 +196,31 @@ export const addEventToCurrentSpan = (
   }
 }
 
+export const recordException = (span: Span, exception: Exception, level?: ErrorLevel) => {
+  const errorLevel = level || exception["level"] || "warn"
+  span.setAttribute("error.level", errorLevel)
```
bodymindarts, line 201: We should set `error.level` on all spans if it exists (not just critical). That would be a great way to search on honeycomb.

bodymindarts (dismissed):

src/services/tracing.ts
dolcalmi: https://github.com/open-telemetry/opentelemetry-specification/issues/1008 I thought the same but seems we should set status

src/services/tracing.ts
dolcalmi: do you mean setStatus to error? because `const errorLevel = level || exception["level"] || "warn"` assign `warns` if exception does not have a level

dolcalmi: I did the update and also changed the default to `info` https://github.com/GaloyMoney/galoy/pull/1099/commits/807ca0979bba7e46574b50e350a5ca226d5877d1

src/services/tracing.ts
dolcalmi: > Also call the global value like `ErrorLevel.Critical` instead of `"critical"`.

Fixed https://github.com/GaloyMoney/galoy/pull/1099/commits/807ca0979bba7e46574b50e350a5ca226d5877d1

bodymindarts (approved):

PR #1101, opened 2022-02-18
Add services tracing wrappers

dolcalmi:
- Update missing params in asyncRunInSpan implementations
- Add tracing wrappers to main services (lnd offchain, lnd onchain, ledger and phone provider)


bodymindarts (approved):
Can you also do lock and remove the lock tracing from services index

PR #1102, opened 2022-02-19
feat: dual currency lightning with intraledger settlement

nicolasburtey:


src/app/wallets/send-lightning.ts
```diff
@@ -370,22 +371,95 @@ const executePaymentViaIntraledger = async ({
...
+    // side effect: spread is paid by the sender
+    cents = centsInvoice
+    sats = toSats(amount)
+  } else {
+    sats = toSats(amount)
+  }
+
+  const amountDisplayCurrency = sats
+    ? dCConverter.fromSats(sats)
+    : dCConverter.fromCents(cents!)
```
vindard, line 441: Potential idea to get around typing issue here
```suggestion
const amountDisplayCurrency = sats
  ? dCConverter.fromSats(sats)
  : cents
  ? dCConverter.fromCents(cents)
  : new NotImplementedError("Should never get here")
if (amountDisplayCurrency instanceof Error) return amountDisplayCurrency
```

vindard (commented):
Looks ok generally

There's just one potential discrepancy that I'm seeing that seems disproportional. The medici transactions generated from the `it("receives payment from outside to USD wallet with amount", ...)` test show a significant difference between the actual amounts in cents and the display amounts. Just checking if this is expected behaviour.

I changed the amount from `2500` to `15000` in the test to see if it scales linearly and it looks like it does.

image

src/app/wallets/send-lightning.ts
nicolasburtey: could be an option, but still hacky IMO. I think we will want to refactor those use cases in different function so that we have a stronger typing, and those would not be necessary any more

nicolasburtey:
> Looks ok generally
> 
> There's just one potential discrepancy that I'm seeing that seems disproportional. The medici transactions generated from the `it("receives payment from outside to USD wallet with amount", ...)` test show a significant difference between the actual amounts in cents and the display amounts. Just checking if this is expected behaviour.
> 
> I changed the amount from `2500` to `15000` in the test to see if it scales linearly and it looks like it does.
> 
> image

this commit may be fixing it? https://github.com/GaloyMoney/galoy/pull/1102/commits/704442315adadfccecbc6e632ff580d9c7e1c213

vindard (approved):

PR #1110, opened 2022-02-21
feat: handle expired invoices

vindard:
### Description

Includes two change around handling expired invoices:
1. Handle failing to delete an already expired invoice
2. Handle failing to pay an expired invoice


nicolasburtey (approved):

nicolasburtey:
one thing to add in the documentation is why USD invoice time is small (2 minutes) compared to traditional BTC invoice. @vindard can dive in with you if you need more understanding of why that is.

PR #1111, opened 2022-02-21
fix: htlc stuck after output channel close

dolcalmi:
add additional validation to resolve payment status.

If payment is pending:
- Check payment timeout against current block height
- If transaction has expired then check that all the output channels (in paths) are closed

### DIFF
--- a/src/services/lnd/index.ts
+++ b/src/services/lnd/index.ts
@@ -33,13 +33,19 @@ import {
   GetInvoiceResult,
   getPayments,
   getFailedPayments,
+  getClosedChannels,
+  getWalletInfo,
 } from "lightning"
 
 import { wrapAsyncFunctionsToRunInSpan } from "@services/tracing"
 import { timeout } from "@utils"
 
-import { TIMEOUT_PAYMENT } from "./auth"
+import { LocalCacheService } from "@services/cache"
+import { SECS_PER_5_MINS } from "@config"
+import { CacheKeys } from "@domain/cache"
+
 import { getActiveLnd, getLndFromPubkey, getLnds } from "./utils"
+import { TIMEOUT_PAYMENT } from "./auth"
 
 export const LndService = (): ILightningService | LightningServiceError => {
   const activeNode = getActiveLnd()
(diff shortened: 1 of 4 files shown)


src/services/lnd/index.ts
```diff
@@ -623,3 +625,51 @@ const translateLnPaymentLookup = (p): LnPaymentLookup => ({
...
+    })
+
+    const failed = pending.paths
+      .map((p) => {
+        const [first] = p.hops
+        return closedChannels.find((c) => c.id === first.channel)
+      })
+      .every((s) => !!s)
+
+    if (failed) return PaymentStatus.Failed
```
nicolasburtey, line 671: do we have all this logic just to return `PaymentStatus.Failed` instead of `PaymentStatus.Pending`?

shouldn't we focus on making sure payment can't be pending after some time, and settle it properly in the ledger?

I dont think this approach is the right one, because we may ends up with many pending that could be pending forever in the ledger, and will have to go through this loop forever because they have not been settled properly. 

this will, over time, introduce more and more discrepancy in our ledger.

nicolasburtey (changes requested):
I had a look at the PR, and I think the change we're doing it as the wrong level. returning the correct return status as a hack is not the way to go. we need to find/fix the root cause otherwise it will introduce discrepancy between the ledger and lnd, that will be even harder to debug because it's been "cosmetically" fix at the API level.

src/services/lnd/index.ts
dolcalmi: do we have all this logic just to return PaymentStatus.Failed instead of PaymentStatus.Pending? **yes**

shouldn't we focus on making sure payment can't be pending after some time, and settle it properly in the ledger? **this method is used by updatePendingPayments so it will update ledger when return failed. I think is a clean and easy way to remove it later if lnd is fixed**

nicolasburtey: ok that make sense then. thanks for the clarification.

src/services/lnd/index.ts
```diff
@@ -633,3 +635,51 @@ const translateLnPaymentLookup = (p): LnPaymentLookup => ({
...
+        return channels
+      },
+    })
+
+    const failed = pending.paths
+      .map((p) => {
+        const [first] = p.hops
+        return closedChannels.find((c) => c.id === first.channel)
+      })
+      .every((s) => !!s)
```
vindard, line 679: Just double-checking, `failed` is a boolean that is only true if the first hop of every `path` in the `paths` array maps to a closed channel in our lnd?

vindard (approved):
Looks good to me. The meat of this is to conditionally convert a payment status from `pending` to `failed` if we notice the first hop in all paths is from a channel that's already closed in our lnd, which I think makes sense for now.

PR #1113, opened 2022-02-22
feat: containerize tests in ci

UncleSamtoshi:
- add ports through docker-compose.override.yml
- containerize integration and e2e tests
- update CI to use containerized tests


bodymindarts:
This is great - tested to be much more reliable. I suggest going over the details in code review this evening.

.github/workflows/integration-test.yml
```diff
@@ -14,4 +14,4 @@ jobs:
     - uses: actions/setup-node@v2
     - run: yarn install
     - name: Run integration tests
-      run: docker compose up integration-deps -d && make integration-in-ci
+      run:  make integration-in-container
```
bodymindarts, line 17: whitespace (`run:  ` has 2 ` `)

Makefile
```diff
@@ -55,8 +55,13 @@ e2e:
 	yarn build && \
 	yarn test:e2e
 
-e2e-in-ci:
-	. ./.envrc && \
+e2e-in-container:
```
bodymindarts, line 58: don't like the distinction `-in-container` vs `in-docker` ... need more explicit names to differentiate.

.env.ci
```diff
@@ -0,0 +1,41 @@
+JEST_TIMEOUT=90000
```
bodymindarts, line 1: Can we add a comment to the top of this file without breaking it?
Would be good to describe when its used how it needs to be kept up to date etc.

Makefile
```diff
@@ -81,3 +90,7 @@ check-code:
 	yarn tsc-check
 	yarn eslint-check
 	yarn build
+
+create-tmp-env-file:
```
bodymindarts, line 94: I think this is only ever used in `ci` - perhaps we can reflect that in the name.

bodymindarts (commented):
Excellent - I think other than some naming this is really good.

bodymindarts (dismissed):

bodymindarts (approved):

PR #1120, opened 2022-02-23
fix: onchain addresses null exception

dolcalmi:
- Fix https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/k3psRSkCssd/trace/C7dUF9Wh4cV?span=22ee5be24a693e55
- Enable onchain tests (not sure why were disabled)

### DIFF
--- a/src/services/mongoose/wallets.ts
+++ b/src/services/mongoose/wallets.ts
@@ -117,7 +117,8 @@ const resultToWallet = (result: WalletRecord): Wallet => {
   const accountId = fromObjectId<AccountId>(result._accountId)
   const type = result.type as WalletType
   const currency = result.currency as WalletCurrency
-  const onChainAddressIdentifiers = result.onchain.map(({ pubkey, address }) => {
+  const onChain = result.onchain || []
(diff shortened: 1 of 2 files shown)


src/services/mongoose/wallets.ts
```diff
@@ -117,7 +117,8 @@ const resultToWallet = (result: WalletRecord): Wallet => {
   const accountId = fromObjectId<AccountId>(result._accountId)
   const type = result.type as WalletType
   const currency = result.currency as WalletCurrency
-  const onChainAddressIdentifiers = result.onchain.map(({ pubkey, address }) => {
+  const onChain = result.onchain || []
```
bodymindarts, line 120: I'm not sure about this. Isn't it masking an underlying bug / data inconsistency.
- if this compiled before then the type must be wrong
- if the type is correct but the runtime data is wrong then we need to fix it there

nicolasburtey (approved):

src/services/mongoose/wallets.ts
dolcalmi: yep, it is a problem with mongoose default value... I will merge this and probably create a migration script to fix users collection (probably an issue with old users)

PR #1123, opened 2022-02-23
feat: add account generation script

krtk6160:
This script has already been run on a live environment and works as expected

### DIFF
--- a/src/debug/create-wallets.ts
+++ b/src/debug/create-wallets.ts
@@ -21,7 +21,11 @@ import { checkedToWalletId, WalletCurrency, WalletType } from "@domain/wallets"
 import { createToken } from "@services/jwt"
 import { baseLogger } from "@services/logger"
 import { setupMongoConnection } from "@services/mongodb"
-import { AccountsRepository, WalletsRepository } from "@services/mongoose"
+import {
+  AccountsRepository,
+  UsersRepository,
+  WalletsRepository,
+} from "@services/mongoose"
 import { User } from "@services/mongoose/schema"
 import { createObjectCsvWriter } from "csv-writer"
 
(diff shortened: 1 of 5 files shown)


src/debug/create-wallets.ts
```diff
@@ -0,0 +1,107 @@
...
+  if (args.length === 5) {
+    const numWallets = parseInt(args[2])
+
+    const disbursementAmount = checkedToSats(parseInt(args[4]))
+    if (disbursementAmount instanceof Error) return disbursementAmount
+
+    const disburserWalletId = checkedToWalletId(args[3])
+    if (disburserWalletId instanceof Error) return disburserWalletId
+
+    const wallets = await generateWallets(numWallets)
```
nicolasburtey, line 95: probably need to dump this in a csv

src/debug/create-wallets.ts
```diff
@@ -0,0 +1,107 @@
...
+    const account = await User.create({ phone })
+
+    const btcWallet = await WalletsRepository().persistNew({
+      accountId: account._id,
+      type: WalletType.Checking,
+      currency: WalletCurrency.Btc,
+    })
+    if (btcWallet instanceof Error) return btcWallet
+
+    const usdWallet = await WalletsRepository().persistNew({
```
nicolasburtey, line 59: maybe usd wallet the default wallet

src/debug/create-wallets.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    console.log("verify", jwt.verify(jwtToken, JWT_SECRET))
+  }
+
+  return wallets
+}
+
+const main = async () => {
+  const args = process.argv
+  if (args.length === 5) {
```
nicolasburtey, line 86: can we also have an arg to set the default level of the account when they are generated?

src/debug/create-wallets.ts
krtk6160: Done

src/debug/create-wallets.ts
krtk6160: done

src/debug/create-wallets.ts
krtk6160: done

bodymindarts (approved):

PR #1125, opened 2022-02-23
feat: add 'accountUpdateDefaultWalletId' mutation

vindard:


src/graphql/root/mutation/account-update-default-wallet-id.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    const result = await Accounts.updateDefaultWalletId({
+      walletId,
+      accountId: domainAccount.id,
+    })
+
+    if (result instanceof Error) {
+      const appErr = mapError(result)
+      return { errors: [{ message: appErr.message || appErr.name }] }
```
dolcalmi, line 34:
```suggestion
      return { errors: [mapError(result)] }
```

dolcalmi (approved):
just a minor comment.. feel free to merge

src/graphql/root/mutation/account-update-default-wallet-id.ts
dolcalmi: mapError already do message/name coalescing
```js
const errorName = error.name as ApplicationErrorKey
let message = error.message || errorName || ""
```

vindard: With the current destructuring I get:
```
{
    "data": {
        "accountUpdateDefaultWalletId": {
            "errors": [
                {
                    "message": "Unknown error occurred (code: InvalidWalletId)"
                }
            ],
            "account": null
        }
    }
}
```

And with the direct `mapError(result)` suggested I get:
```
{
    "errors": [
        {
            "message": "Unknown error occurred (code: InvalidWalletId)",
            "code": "UNKNOWN_CLIENT_ERROR"
        }
    ],
    "data": null
}
```

PR #1127, opened 2022-02-23
chore: improve UserUpdateUsername returns

vindard:
## Description

Improve the responses from the `UserUpdateUsername` mutation:
- on success, includes the username instead of `null` in the response
- on "is immutable" edge case, still returns a user object as feedback (didn't before this)

### DIFF
--- a/src/graphql/root/mutation/user-update-username.ts
+++ b/src/graphql/root/mutation/user-update-username.ts
@@ -1,4 +1,5 @@
 import { Accounts } from "@app"
+import { UsernameIsImmutableError } from "@domain/accounts"
 import { mapError } from "@graphql/error-map"
 import { GT } from "@graphql/index"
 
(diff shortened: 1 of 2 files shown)


src/graphql/root/mutation/user-update-username.ts
```diff
@@ -30,7 +31,9 @@ const UserUpdateUsernameMutation = GT.Field({
 
     if (result instanceof Error) {
       const appErr = mapError(result)
-      return { errors: [{ message: appErr.message || appErr.name }] }
+      return result instanceof UsernameIsImmutableError
+        ? { errors: [{ message: appErr.message || appErr.name }], user: domainUser }
+        : { errors: [{ message: appErr.message || appErr.name }] }
```
dolcalmi, line 36: this is not necessary `message: appErr.message || appErr.name` please use only `[mapError(result)]`

src/graphql/types/object/graphql-user.ts
```diff
@@ -35,7 +35,7 @@ const GraphQLUser = GT.Object({
       type: Username,
       description: "Optional immutable user friendly identifier.",
       resolve: async (source, args, { domainAccount }) => {
-        return domainAccount?.username
+        return domainAccount?.username || source.username
```
dolcalmi, line 38: why is this needed?

dolcalmi (changes requested):

src/graphql/root/mutation/user-update-username.ts
vindard: I tested this, see: https://github.com/GaloyMoney/galoy/pull/1125#discussion_r813879550

bodymindarts (approved):

PR #1132, opened 2022-02-24
feat: add new default-wallet-id query to support wallet currencies

vindard:
## Description

This PR adds a new query to:
1. return the wallet currency for a user's `defaultWalletId`
2. return the `defaultWalletId(ByCurrency)` based on an optional specified currency

### Context

We need this to know what type of wallet we are working with in places like for our `galoy-pay` repo. Right now, galoy-pay breaks if the user's default wallet is `USD` and they try to set an amount in the interface because the `lnInvoiceSendOnBehalfOfRecipient` mutation does not support USD wallets. We would need to know the wallet type to be able to switch on:
- `lnInvoiceSendOnBehalfOfRecipient`
- `lnUsdInvoiceSendOnBehalfOfRecipient`


vindard:
Question: would it be better to call this query `accountDefaultPublicWallet`?

samerbuna (approved):

samerbuna:
> Question: would it be better to call this query `accountDefaultPublicWallet`?

I think having public in the name might be confusing

PR #1136, opened 2022-02-25
fix: round cents down if not rounded int

vindard:
## Description

We've noticed that we get [this dealer error](https://ui.honeycomb.io/galoy/datasets/galoy-hack/result/CMXV2Fnj8FY/trace/8nPgLTPFKQ4?span=e8bc65978cbe0d3c) for wallets with a non-rounded-integer balance.

This is an attempt to fix this.

### DIFF
--- a/src/app/wallets/get-lightning-fee.ts
+++ b/src/app/wallets/get-lightning-fee.ts
@@ -99,7 +99,11 @@ const feeProbe = async ({
   let balanceInSats: Satoshis
   if (walletCurrency === WalletCurrency.Usd) {
     const dealer = DealerPriceService()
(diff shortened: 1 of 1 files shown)


UncleSamtoshi (approved):

vindard:
@sebastienverreault temporary fix here for a serialization error (this [trace](https://ui.honeycomb.io/galoy/datasets/galoy-hack/result/CMXV2Fnj8FY/trace/8nPgLTPFKQ4?span=e8bc65978cbe0d3c)) that happens at [this line](https://github.com/GaloyMoney/galoy/blob/7382794e2ae28ea4d578bb0708867a52f9a54767/src/app/wallets/get-lightning-fee.ts#L102) in the code when the balance was `5711.33865006808` cents in the USD wallet.

PR #1148, opened 2022-03-01
refactor: validate config

UncleSamtoshi:
- validate yml config using json schema

### DIFF
--- a/src/config/schema.ts
+++ b/src/config/schema.ts
@@ -1,0 +1,257 @@
+import { JTDDataType } from "ajv/dist/types/jtd-schema"
+
+export type ConfigSchema = JTDDataType<typeof configSchema>
+
+const buildNumberConfigSchema = {
+  type: "object",
+  properties: {
+    minBuildNumber: { type: "number" },
+    lastBuildNumber: { type: "number" },
+  },
+  required: ["minBuildNumber", "lastBuildNumber"],
+  additionalProperties: false,
+}
+
+const accountLimitConfigSchema = {
+  type: "object",
+  properties: {
+    level: {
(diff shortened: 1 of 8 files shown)


src/config/schema.ts
```diff
@@ -0,0 +1,272 @@
...
+      type: "object",
+      properties: {
+        LOW_BOUND_RATIO_SHORTING: { type: "number" },
+        LOW_SAFEBOUND_RATIO_SHORTING: { type: "number" },
+        HIGH_SAFEBOUND_RATIO_SHORTING: { type: "number" },
+        HIGH_BOUND_RATIO_SHORTING: { type: "number" },
+        LOW_BOUND_LEVERAGE: { type: "number" },
+        LOW_SAFEBOUND_LEVERAGE: { type: "number" },
+        HIGH_SAFEBOUND_LEVERAGE: { type: "number" },
+        HIGH_BOUND_LEVERAGE: { type: "number" },
```
dolcalmi, line 81: if these are not env variables should keep camel case standard but... :thinking:  are we using this values? this seems to be more a dealer config than backend config

dolcalmi (changes requested):
looks good.. just a minor comment

src/config/schema.ts
nicolasburtey: probably not used any more. would probably make sense to look at all the value in the yaml to see if they're still used as part of this PR

UncleSamtoshi: It does not appear to be used by the Galoy app, so I will remove it.

dolcalmi (approved):

PR #1153, opened 2022-03-02
fix: minimize chance of small negative user balance from fee estimation

vindard:
## Description

There is an edge-case where onchain fee estimation can be lower than the actual fee used for an onchain txn in the case where the fee environment is very volatile.

In these cases, it is possible for our balance check to be bypassed by a small amount which amounts to the difference the fee estimate is off by in sats. The highest we've seen this be off by so far is 3586 sats.

To help address this two measures are proposed in this PR:
1. Bring the estimate check closer to the actual send to reduce the effects from fee volatility

1. **(Feature change!)** Add a fee buffer when doing the balance check to help mitigate from the edge instances where the fee can still differ

   This 2nd measure effectively means that a user would not be able to empty their wallet via onchain, but can still do so over Lightning/Intraledger

### DIFF
--- a/default.yaml
+++ b/default.yaml
@@ -137,6 +137,7 @@ onChainWallet:
   scanDepth: 360
   scanDepthOutgoing: 2
   scanDepthChannelUpdate: 8
+  feeBuffer: 5000
 
 # TODO
 # carrierRegexFilter: ""
--- a/src/app/wallets/send-on-chain.ts
+++ b/src/app/wallets/send-on-chain.ts
@@ -310,32 +310,37 @@ const executePaymentViaOnChain = async ({
   })
   if (withdrawalLimitCheck instanceof Error) return withdrawalLimitCheck
 
-  const estimatedFee = await getOnChainFee({
-    walletId: senderWallet.id,
-    account: senderAccount,
-    amount,
-    address,
-    targetConfirmations,
-  })
+  const getFeeEstimate = () =>
+    getOnChainFee({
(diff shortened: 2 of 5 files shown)


dolcalmi (approved):

bodymindarts:
The function we are calling from the `OnChainService`is [`sendToOnChainAddress`](https://github.com/GaloyMoney/galoy/blob/main/src/services/lnd/onchain-service.ts#L148).

That method accepts a [`fee_tokens_per_vbyte`](https://github.com/alexbosworth/lightning/blob/master/lnd_methods/onchain/send_to_chain_address.d.ts#L29-L43).
Why don't we simply pass the original estimation on to that function - then we don't need a buffer.

bodymindarts (changes requested):
See above

bodymindarts (approved):

vindard:
> The function we are calling from the `OnChainService`is [`sendToOnChainAddress`](https://github.com/GaloyMoney/galoy/blob/main/src/services/lnd/onchain-service.ts#L148).
> 
> That method accepts a [`fee_tokens_per_vbyte`](https://github.com/alexbosworth/lightning/blob/master/lnd_methods/onchain/send_to_chain_address.d.ts#L29-L43). Why don't we simply pass the original estimation on to that function - then we don't need a buffer.

Juan and I looked at this as well and noticed that we probably can't enforce which UTXOs it selects so transaction size (and final fee) can still be variable.

PR #1155, opened 2022-03-02
refactor: implement ln receive-side methods within entry builder

vindard:
## Description

Continuing from #1152, this is WIP for swapping into receive ledger use-cases.

### DIFF
--- a/src/services/ledger/domain/entry-builder.ts
+++ b/src/services/ledger/domain/entry-builder.ts
@@ -1,4 +1,5 @@
 import { AmountCalculator, WalletCurrency } from "@domain/shared"
+import { Payment } from "bitcoinjs-lib"
 
 import { lndLedgerAccountId } from "./accounts"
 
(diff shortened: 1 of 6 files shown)


src/services/ledger/domain/entry-builder.ts
```diff
@@ -166,18 +176,29 @@ const EntryBuilderCreditWithUsdDebit = ({
...
+  const creditAccount = ({
     accountId,
     amount,
   }: {
     accountId: LedgerAccountId
-    amount: PaymentAmount<D>
+    amount: PaymentAmount<"USD">
   }) => {
-    entry.debit(dealerBtcAccountId, Number(amount.amount), {
+    entry.credit(dealerBtcAccountId, Number(debitAmount.amount), {
```
vindard, line 197: Should we explore `debit`/`credit` wrappers where the `dealerBtcAccountId` lives in a `dealerBtcAccount` object that also has a `currency` property and there is some currency validation between the amount and ledger-account objects?

For e.g. it's still easy to pass a typed amount to the wrong account for the amount's currency at this step.

bodymindarts: Yeah this could certainly improve safety - not sure how this will get integrated but certainly an idea as we improve the outer layer of the ledger subdomain

UncleSamtoshi (approved):

PR #1156, opened 2022-03-02
fix: surface test failures in gh actions

UncleSamtoshi:


dolcalmi (approved):
this needs to be merged by @nicolasburtey or @bodymindarts 

It is ok that integration is failing (main is failing too)

PR #1161, opened 2022-03-03
Entry builder wip

bodymindarts:


vindard:
Given #1152, do we still need this work?

PR #1164, opened 2022-03-03
chore: add instrumentation for balance and onchain fees

vindard:
## Description

This builds from debugging the issue for PR #1153

### DIFF
--- a/src/services/ledger/index.ts
+++ b/src/services/ledger/index.ts
@@ -10,6 +10,7 @@ import {
   toLiabilitiesWalletId,
   toWalletId,
 } from "@domain/ledger"
+import { ErrorLevel } from "@domain/errors"
 import {
   CouldNotFindTransactionError,
   LedgerError,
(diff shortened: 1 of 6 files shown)


src/services/ledger/index.ts
```diff
@@ -179,6 +184,16 @@ export const LedgerService = (): ILedgerService => {
       const { balance } = await MainBook.balance({
         account: liabilitiesWalletId,
       })
+      if (balance < 0) {
+        addAttributesToCurrentSpan({
+          "getWalletBalance.error.invalidBalance": `${balance}`,
+          "error": true,
+          "error.level": ErrorLevel.Warn,
+        })
+        addEventToCurrentSpan("exception", {
```
bodymindarts, line 193: Have you considered how to query this? Isn't `exception` usually associated with an error and recorded as `span.recordException`?

bodymindarts (dismissed):

src/services/ledger/index.ts
vindard: Addressed here: https://github.com/GaloyMoney/galoy/pull/1164/commits/1a84907f32ea963633b3d1caff6ed6b1fbba71a6

Do we want exceptions to only be recorded when an entire use-case returns an error and terminates (current behaviour), or is it ok to have exceptions recorded mid-process like this?

src/services/ledger/index.ts
```diff
@@ -179,6 +183,14 @@ export const LedgerService = (): ILedgerService => {
       const { balance } = await MainBook.balance({
         account: liabilitiesWalletId,
       })
+      if (balance < 0) {
+        recordExceptionInCurrentSpan({
```
dolcalmi, line 187: this should pass critical (level)

dolcalmi (changes requested):

dolcalmi (approved):

PR #1165, opened 2022-03-04
fix: rounding to integer currency amounts

UncleSamtoshi:


UncleSamtoshi:
@sebastienverreault I was a little confused on the wording of the Dealer API. This PR assumes that "Buy" and "Sell" always refer to USD.

vindard:
As I understand it BTC is our reference currency and we're "buying & selling" USD

src/app/wallets/send-lightning.ts
```diff
@@ -607,9 +607,9 @@ const executePaymentViaLn = async ({
         // TODO: test properly. move this to domain
         const baseCentsWithoutFee = toCents(amount)
         const satsWithoutFee = toSats(rawRoute.tokens - rawRoute.safe_fee)
-        const ratio = satsWithoutFee / baseCentsWithoutFee
+        const centsPerSat = baseCentsWithoutFee / satsWithoutFee
```
dolcalmi, line 610: if toCents and toSats are only casting then are not needed  in lines 608-609

src/services/dealer-price/index.ts
```diff
@@ -95,7 +95,7 @@ export const DealerPriceService = (): IDealerPriceService => {
           amountInSatoshis,
         ),
       )
-      return toCents(response.getAmountInCents())
+      return toCents(Math.floor(response.getAmountInCents()))
```
dolcalmi, line 98: if we will persist big int then should be parsed here

src/services/dealer-price/index.ts
```diff
@@ -111,7 +111,7 @@ export const DealerPriceService = (): IDealerPriceService => {
           amountInSatoshis,
         ),
       )
-      return toCents(response.getAmountInCents())
+      return toCents(Math.ceil(response.getAmountInCents()))
```
dolcalmi, line 114: same here

src/services/dealer-price/index.ts
```diff
@@ -131,7 +131,7 @@ export const DealerPriceService = (): IDealerPriceService => {
           .setAmountInSatoshis(amountInSatoshis)
           .setTimeInSeconds(timeToExpiryInSeconds),
       )
-      return toCents(response.getAmountInCents())
+      return toCents(Math.floor(response.getAmountInCents()))
```
dolcalmi, line 134: same here

src/services/dealer-price/index.ts
```diff
@@ -148,7 +148,7 @@ export const DealerPriceService = (): IDealerPriceService => {
           .setAmountInSatoshis(amountInSatoshis)
           .setTimeInSeconds(timeToExpiryInSeconds),
       )
-      return toCents(response.getAmountInCents())
+      return toCents(Math.ceil(response.getAmountInCents()))
```
dolcalmi, line 151: and here

dolcalmi (changes requested):

src/services/dealer-price/index.ts
UncleSamtoshi: Are you suggesting returning a UsdPaymentAmount? That change could be made but would expand the scope of the PR.

dolcalmi (dismissed):

dolcalmi (approved):

sebastienverreault:
> @sebastienverreault I was a little confused on the wording of the Dealer API. This PR assumes that "Buy" and "Sell" always refer to USD.

Yeah cross-currency buy/sell are confusing so I tried to make it explicit in the name of the endpoint what the action was, but if you have suggestions to further clarify I'm all for it. 

I wrote this down to make it clear for myself also: [Price Service](https://github.com/GaloyMoney/dealer/blob/main/PRICE-SERVICE.md).

If the service was for [immediate | future] exchange rates it would be clearer, but that was not the requirements.

PR #1170, opened 2022-03-08
Ledger facade

bodymindarts:


src/services/ledger/domain/entry-builder.ts
```diff
@@ -1,8 +1,8 @@
 import { AmountCalculator, WalletCurrency } from "@domain/shared"
 
-import { coldStorageAccountId, lndLedgerAccountId } from "./accounts"
+import { coldStorageAccountDescriptor, coldStorageAccountId, lndLedgerAccountDescriptor, lndLedgerAccountId } from "./accounts"
 
-const ZERO_SATS = {
+export const ZERO_SATS = {
```
bodymindarts, line 5: should not be exporter... rather import the one from `@domain/shared`

src/services/ledger/index.types.d.ts
```diff
@@ -36,29 +36,30 @@ type AddOnchainSendLedgerMetadata = NonIntraledgerLedgerMetadata & {
 type AddColdStorageLedgerMetadata = NonIntraledgerLedgerMetadata & {
   hash: OnChainTxHash
   payee_addresses: OnChainAddress[]
-  currency: WalletCurrency
+  currency?: WalletCurrency
```
bodymindarts, line 39: Why did this become optional?

src/services/ledger/index.types.d.ts
```diff
@@ -36,29 +36,30 @@ type AddOnchainSendLedgerMetadata = NonIntraledgerLedgerMetadata & {
...
 type AddColdStorageReceiveLedgerMetadata = AddColdStorageLedgerMetadata
 
 type AddColdStorageSendLedgerMetadata = AddColdStorageLedgerMetadata
 
-type IntraledgerLedgerMetadata = LedgerMetadata & {
-  memoPayer: string | undefined
-  username: Username | undefined
+type IntraledgerBaseMetadata = LedgerMetadata & {
+  usd: DisplayCurrencyBaseAmount // to be renamed amountDisplayCurrency
+  memoPayer: string | undefined | null
```
bodymindarts, line 48: I think we decided to standardise on `undefined` instead of `null`... can we fix this when its constructed or is there a reason its like this?
Also why not `memoPayer?: string` instead of the `undefined`

test/unit/services/ledger/domain/entry-builder.spec.ts
```diff
@@ -23,14 +24,59 @@ describe("EntryBuilder", () => {
...
+    )
+    Object.values<any>(journal.credits).forEach((entry) =>
+      entry.metadata.currency === WalletCurrency.Usd
+        ? (usdCredits += entry.amount)
+        : (btcCredits += entry.amount),
+    )
+
+    expect(usdCredits).toEqual(usdDebits)
+    expect(btcCredits).toEqual(btcDebits)
+    console.log(usdCredits)
```
bodymindarts, line 47: remove console.log

bodymindarts (commented):
Also need to move the types out of the src files. The only time I think it is okay to have them in the source files is if they are private and local.

src/services/ledger/index.types.d.ts
UncleSamtoshi: Not sure how that happened, I will remove that.

src/services/ledger/index.types.d.ts
UncleSamtoshi: I've switched to using undefined.

UncleSamtoshi:
Moved to PR #1198

PR #1174, opened 2022-03-09
refactor: remove 2fa

bodymindarts:
Removing all 2fa related code because:
- this feature has never made it to production
- we are introducing an auth solution that will support 2fa verified sessions (Ory Kratos) so the logic regarding that doesn't need to be in the backend
- removing this code simplifies the cleanup of the USD related code by a large factor (as there is less complexity to consider during the re-design).


nicolasburtey:
how does the introduction of Ory Kratos changes most of the code here? 

it's only change the source of the repository, the rest should be the same?

bodymindarts:
I think the main thing here is that currently the 2fa limits code is coupled to the 'regular' limit checking even though these code paths serve 2 different use cases. USD features implies changes to the limit checking logic. It will be easier to clean up without needing to take the unrelated and unused 2fa feature into account. Since this commit is in the git history it would be trivial to bring it back when / if we want / need.

nicolasburtey:
I'm not sure to understand your point about the coupling. can you point to a code example?

Generally, this code is working AFAIK (it has been tested at least) and is really needed for the users. of course removing the features will make the code easier, but I'm not sure to understand why we want to remove it for USD to add it back just after. 

If there is some coupling then why not fixing it instead of removing everything.

The fact it's not been used is because it's has not been implemented in the frontend yet. it's not related to the backend.

PR #1181, opened 2022-03-11
feat: implement a kratos auth browser endpoint

samerbuna:


src/app/users/create-user.ts
```diff
@@ -9,12 +9,31 @@ import { WalletCurrency } from "@domain/shared"
 import { WalletType } from "@domain/wallets"
 import { TwilioClient } from "@services/twilio"
 
+export const setupWallet = async (userId: UserId): Promise<Wallet | Error> => {
```
bodymindarts, line 12: no need to export this right?
Also return should be `| ApplicationError`

src/app/users/create-user.ts
```diff
@@ -35,20 +54,18 @@ export const createUser = async ({
...
   if (wallet instanceof Error) return wallet
 
-  account.defaultWalletId = wallet.id
+  return user
+}
 
-  const result = await AccountsRepository().update(account)
-  if (result instanceof Error) return result
+export const createKratosUser = async ({ userId }: { userId: string }) => {
+  const user = await UsersRepository().persistNewKratosUser({ userId })
```
bodymindarts, line 64: need to validate or at least cast the `KratosUserId`

src/app/users/generate-2fa.ts
```diff
@@ -8,6 +8,8 @@ export const generate2fa = async (
   const user = await UsersRepository().findById(userId)
   if (user instanceof Error) return user
 
+  if (!user.phone) return new Error("Phone is not defined")
```
bodymindarts, line 11: Thes needs to be an error that derives from `DomainError`

src/app/users/login.ts
```diff
@@ -70,6 +73,48 @@ export const login = async ({
...
+  }
+
+  {
+    const limitOk = await checkfailedLoginAttemptPerEmailAddressLimits(emailAddressValid)
+    if (limitOk instanceof Error) return limitOk
+  }
+  const userRepo = UsersRepository()
+  let user = await userRepo.findByKratosUserId(userId)
+
+  if (user instanceof CouldNotFindUserFromIdError) {
```
bodymindarts, line 104: Error should be `CouldNotFindUserFromKratosIdError`

src/domain/users/index.types.d.ts
```diff
@@ -68,6 +71,8 @@ type PhoneMetadataValidator = {
 interface IUsersRepository {
   findById(userId: UserId): Promise<User | RepositoryError>
   findByPhone(phone: PhoneNumber): Promise<User | RepositoryError>
+  findByKratosUserId(userId: string): Promise<User | RepositoryError>
```
bodymindarts, line 74: `string` => `KratosUserId` everywhere

bodymindarts (changes requested):

src/domain/users/index.ts
```diff
@@ -15,6 +17,11 @@ import {
 // for a more precise "regex"
 const PhoneNumberRegex = /^\+\d{7,14}$/i // FIXME {7,14} to be refined
 
+const KratosUserIdRegex =
+  /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i
+
+const EmailAddressRegex = /^[\w-.+]+@([\w-]+\.)+[\w-]{2,4}$/i
```
dolcalmi, line 23: this regex is giving a `pattern error`, it works if I remove `-` form `[\w-.+]`. could you please verify?

src/servers/auth-router.ts
```diff
@@ -0,0 +1,61 @@
...
+      ip,
+    })
+
+    if (authToken instanceof Error) {
+      throw authToken
+    }
+
+    res.send({ authToken })
+  } catch (error) {
+    console.dir(error, { depth: 0 })
```
dolcalmi, line 56: isn't console dir only for web? :thinking: if so please use baseLogger

src/services/mongoose/users.ts
```diff
@@ -57,6 +74,21 @@ export const UsersRepository = (): IUsersRepository => {
     }
   }
 
+  const persistNewKratosUser = async ({
+    userId,
+  }: {
+    userId: string
```
dolcalmi, line 80: maybe we should change param name to something like `kratosUserId` or `oauthUserId` because `userId` is used in other service methods (this apply to all kratos related methods)

src/graphql/types/object/graphql-user.ts
```diff
@@ -27,7 +27,7 @@ const GraphQLUser = GT.Object({
     },
 
     phone: {
-      type: GT.NonNull(Phone),
+      type: Phone,
```
dolcalmi, line 30: could this affect mobile app?

src/domain/users/index.types.d.ts
```diff
@@ -58,7 +62,7 @@ type User = {
 
 type NewUserInfo = {
   phone: PhoneNumber
-  phoneMetadata: PhoneMetadata | null
+  phoneMetadata?: PhoneMetadata
```
dolcalmi, line 65: this breaks PhoneMetadataValidator null validation `if (phoneMetadata === null || !phoneMetadata.carrier)`.

dolcalmi (changes requested):
looks good, just minor comments

src/domain/users/index.types.d.ts
dolcalmi: if we're not catching this in unit tests, maybe it's worth adding an additional test

src/domain/users/index.ts
samerbuna: Where are you getting that error?

src/graphql/types/object/graphql-user.ts
samerbuna: Not for phone-authenticated users.

bodymindarts (approved):

bodymindarts (approved):

PR #1182, opened 2022-03-11
chore: make all usernames lowercase

UncleSamtoshi:
To improve query times, we will remove the use of case insensitive regex and replace it by having all usernames be lowercase. The change has three parts:
1. This PR: Ensure all new usernames are lowercase
2. Future PR: Migrate all existing usernames to lowercase
3. Future PR: Remove the use of regex when finding a user by username

### DIFF
--- a/src/domain/accounts/index.ts
+++ b/src/domain/accounts/index.ts
@@ -51,10 +51,11 @@ export const checkedAccountStatus = (status: string) => {
 export const UsernameRegex = /(?!^(1|3|bc1|lnbc1))^[0-9a-z_]{3,50}$/i
 
 export const checkedToUsername = (username: string): Username | ValidationError => {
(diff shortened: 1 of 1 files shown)


nicolasburtey (changes requested):
the ability to have uppercase and lowercase is meant to be a feature, not a bug.

it's similar with twitter. I can have "nicolas" or "Nicolas", but other user can't have the handle "NICOLAS".

this helps for the readability. 

yet. if someone search for a username, they will find "nicolas" whichever case they're using for their search.

this should be managed by mongodb. I guess if this has a negative performance impact, we will want to measure it, because (I think, but we could measure) it shouldn't have a large impact.

UncleSamtoshi:
Understood. I think the change is unnecessary anyways as it looks like our performance impact is a recent event and not systemic.

PR #1184, opened 2022-03-12
feat: dynamic fees

nicolasburtey:
feat: implement higher fees on swap out imbalance

note: this works only for bitcoin wallet as of now. I think this is a good first step because:
1/ we can better understand if this approach works starting with bitcoin wallet
2/ usd wallet won't be plugged to onchain on the first iteration

there is a now two different method that can be set in the yaml for onchain withdrawal fees: either "flat" (same behavior as before) or "proportionalOnImbalance"


src/domain/ledger/imbalance-calculator.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (lnImbalance instanceof Error) return lnImbalance
+
+    const onChainImbalance = await getImbalance({
+      since,
+      walletId,
+      volumeFn: volumeOnChainFn,
+    })
+    if (onChainImbalance instanceof Error) return onChainImbalance
+
+    return (lnImbalance - onChainImbalance) as Imbalance
```
mjdietzx, line 44: > Some users will have outsized effects on the bank’s loop imbalance. For example, a merchant that routinely receives offchain payments, and periodically withdraws bitcoin from their account onchain, will impart a consistently negative rate of change on the bank’s loop imbalance. Unless this usage is balanced by users with opposite effects on the bank’s loop imbalance, the bank will eventually have to swap out to rebalance its onchain and offchain liquidity and regain some inbound liquidity.

It doesn't seem like there is any additional fee for this case as-is though? Bc when you calculate the dynamic fee:

```js
const aboveThreshold = Math.max(imbalance - thresholdImbalance, 0)
```

Any `imbalance` that is negative (as in the case quoted above) `aboveThreshold` will default to `0`. Maybe a `Math.abs(imbalance)` is needed in `WithdrawalFeeCalculator`?

src/domain/ledger/imbalance-calculator.ts
```diff
@@ -0,0 +1,50 @@
...
+    volumeFn,
+    walletId,
+    since,
+  }: {
+    volumeFn: GetVolumeSinceFn
+    walletId: WalletId
+    since: Date
+  }) => {
+    const volume_ = await volumeFn({
+      walletId,
```
mjdietzx, line 21: I do wonder if the ideal metric to measure and charge based on is the user's wallet-scoped imbalance.

> To accomplish this in a competitive way, a bank must, among other things, maintain suﬀicient onchain and offchain liquidity. We define the loop imbalance as the difference between the bank’s on-chain liquidity and offchain outbound liquidity.

To me the metric it seems you really care about is the bank's overall _loop imbalance_.

As-is, the way you calculate this dynamic fee seems to econimically incentivize all your users to behave in a homogenous/net-balanced way during the rolling 30 day window. It seems you have a diverse set of users who behave differently (ie b2c merchants vs consumers vs b2b service-providers/producers, etc..) Don't you want to incentivize them all balancing out as a whole, not necessarily individually? That seems to be what your system needs to function smoothly/efficiently?

I'm just wondering if a component of this calculation could be how much this specific transaction will impact the bank's overall _loop imbalance_ (if this txn will improve it maybe this component of the fee should be 0 (or even negative in the future), if this txn further imbalances it maybe this additional fee kicks in?) But maybe this kind of data is not accessible here, in this scope or permission level?

mjdietzx (commented):
This is an awesome idea! And the first time I've looked at your codebase. Nice and simple PR that implements a powerful concept

I am very unfamiliar with your product/platform, so have very little context other than quickly reading `Pricing liquidity for Lightning wallets` and going through this PR. The questions I left are pretty high-level/conceptual, and may indicate that I am totally misunderstanding something. But I just wanted to get a little involved, and would be interested to learn more

src/domain/ledger/imbalance-calculator.ts
nicolasburtey: the user get higher fees only in the case of positive (swap out) imbalance.

swap out means the net inbound flow for lightning is more than net inbound flow for onchain (ie: user send funds on lightning and withdraw it onchain). 

if a user does the opposite then this is great, and there is no fees for that. 

the formula make it so that until the imbalance reach `thresholdImbalance`, there is no additional fees. `aboveThreshold` would be 0 with the formula above.

src/domain/ledger/imbalance-calculator.ts
nicolasburtey: > To me the metric it seems you really care about is the bank's overall loop imbalance.

that is correct, but at the level of individual users, the approach in this PR currently only is adding fees for user that are "looping out" our swaping out above some threshold.

This is based on empirical evidence (which are highlighted in the paper) that this is the flow that needs higher fees for a more balanced overall bank imbalance.

test/unit/domain/ledger/imbalance-calculator.spec.ts
```diff
@@ -0,0 +1,90 @@
+import { toSats } from "@domain/bitcoin"
+import { ImbalanceCalculator } from "@domain/ledger/imbalance-calculator"
+import { WalletCurrency } from "@domain/shared"
+
+let btcWallet: Wallet
+
+beforeAll(() => {
+  btcWallet = {
```
bodymindarts, line 8: this doesn't mutate iso there is no need for a `beforeAll` - can just assign it as const.
Actually I cannot even see you using anything other than the `wallet.id` so perhaps you only need that.

src/domain/wallets/withdrawal-fee-calculator.ts
```diff
@@ -1,13 +1,61 @@
...
+    minerFee,
+  }: OnChainWithdrawalFlatFeeArgs): Promise<WithdrawalFeeCalculatorRes> => ({
+    totalFee: toSats(minerFee + minBankFee),
+    bankFee: minBankFee,
+  })
+
+  const onChainWithdrawalProportionalOnImbalanceFee = async ({
+    minerFee,
+    minBankFee,
+    imbalanceCalculatorFn,
```
bodymindarts, line 20: I don't see the need to add this fn here. IMO it would be better to simply call the `WithdrawalFeeCalculator` with the already computed imbalance. That would make it more isolated and only depend on values.

In some cases dependency injection is unavoidable but if we can avoid it and just work with values we should. OTOH if you do inject it it should be injected via the `OnChainWithdrawalConfig` constructor args as I don't see this function ever changing at runtime.

bodymindarts: Also this introduces a circular dependency on the module level from `ledger -> wallet -> ledger`. We need to keep the dependency graph uni-directional `ledger -> wallet`

bodymindarts (dismissed):

src/app/wallets/get-on-chain-fee.ts
```diff
@@ -34,10 +41,19 @@ export const getOnChainFee = async ({
...
+  const withdrawFeeCalculator = WithdrawalFeeCalculator({
+    feeRatio: getFeesConfig().withdrawRatio,
+    thresholdImbalance: getFeesConfig().withdrawThreshold,
+  })
+
   const payeeWallet = await walletsRepo.findByAddress(address)
+  if (payeeWallet instanceof UnknownRepositoryError) return payeeWallet
 
-  const isIntraLedger = !(payeeWallet instanceof Error)
+  const isIntraLedger = !(
```
dolcalmi, line 50: if you are doing validation here with specific error (although specific error validation is not needed)  then line 59 is not necessary.

dolcalmi (changes requested):

src/domain/ledger/imbalance-calculator.ts
mjdietzx: Makes total sense - thanks for the explanation

src/domain/wallets/withdrawal-fee-calculator.ts
```diff
@@ -1,10 +1,25 @@
...
-    bankFee,
-  }: OnChainWithdrawalFeeArgs): Satoshis => toSats(minerFee + bankFee)
+    minBankFee,
+    imbalance,
+  }: {
+    minerFee: Satoshis
+    minBankFee: Satoshis
+    imbalance: SwapOutImbalance
+  }) => {
+    const aboveThreshold = Math.max(imbalance - thresholdImbalance, 0)
```
jireva, line 16: Do I understand correctly that `imbalance` here refers to the 30 day rolling sum of the user's imbalance?

src/app/wallets/send-on-chain.ts
```diff
@@ -359,23 +369,46 @@ const executePaymentViaOnChain = async ({
...
+          volumeOnChainFn: LedgerService().onChainTxBaseVolumeSince,
+          sinceDaysAgo: getFeesConfig().withdrawDaysLookback,
+        })
+
+        const imbalance = await imbalanceCalculator.getSwapOutImbalance(senderWallet.id)
+        if (imbalance instanceof Error) return imbalance
+
+        const fees = withdrawFeeCalculator.onChainWithdrawalFee({
+          minerFee,
+          minBankFee: toSats(senderAccount.withdrawFee),
```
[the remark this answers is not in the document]
nicolasburtey: indeed. we should probably not set any value on the account by default.

src/domain/wallets/withdrawal-fee-calculator.ts
nicolasburtey: yes @jireva. what do you think about this? it's a config so can be changed easily.

jireva: I'm concerned that charging for imbalance this way is not ideal.  Consider a user that accumulates a large imbalance over a few days, then starts making smaller, more balanced, transactions.  Their first few transaction will pay for that imbalance but their onchain payment fees will remain high for 30 days.

(30 days itself seems reasonable. I was coding up some simulations to see how it (and other durations) would compare to tracking the full history.)

dolcalmi (changes requested):

src/domain/wallets/withdrawal-fee-calculator.ts
jireva: I would recommend adding one more input to `WithdrawalFeeCalculator`: `feesPaid`(?), to track the rolling sum of fees paid by the user (in the same window as `imbalance`) and using that to make sure that the imbalance fee is not charged multiple times:
```
const aboveThreshold = Math.max(imbalance - thresholdImbalance, 0)
const bankFee = toSats(Math.max(minBankFee, Math.ceil(aboveThreshold * feeRatio) - Math.ceil(feesPaid)))
```

Another, more complex, possibility is to make the fee depend on the amount of the current transaction (that way, small transactions never get charged big fees). This requires two more inputs: `amount`, the amount of the current onchain payment and `maxFeeRatio`, the maximum fee ratio to apply to users that have an imbalance above the threshold.
And the resulting `bankFee` calculation would be something like:
```
const aboveThreshold = Math.max(imbalance - thresholdImbalance, 0)
const amountIfAboveThreshold = (aboveThreshold > 0) ? amount : 0
const bankFee = toSats(Math.max(minBankFee, Math.min(Math.ceil(amountIfAboveThreshold * maxFeeRatio), Math.ceil(aboveThreshold * feeRatio) - Math.ceil(feesPaid))))
```

nicolasburtey: I don't think we want to add `feesPaid` because it will add a lot of overhead:
- we need to make another time consuming aggregate query to our database. also I'm not sure fees are currently simply queriable just for bankFee (I think it may be but I'm not sure)
- fee paid in absolute is not the best relevant metrics, because if a user would send 1,000 payments+ on chain in a month, even if this is small payment, there would be fees because of the min fees being paid.

but you're right on the amount. this should definitively be taken into consideration here. I'll update the formula

jireva (dismissed):

src/app/wallets/send-on-chain.ts
```diff
@@ -359,23 +372,47 @@ const executePaymentViaOnChain = async ({
           return txHash
         }
 
-        let totalFee: Satoshis
+        let minerFee: Satoshis
 
-        const minerFee = await onChainService.lookupOnChainFee({
+        const minerFee_ = await onChainService.lookupOnChainFee({
```
bodymindarts, line 377: I don't understand this. Higher up you are calling `const estimatedFee = await getFeeEstimate()` which is already calling
```
  const minerFee = await onChainService.getOnChainFeeEstimate({
    amount: amountChecked,
    address,
    targetConfirmations: targetConfsChecked,
  })
```
in its code path - why the duplicate lookups?

bodymindarts (commented):
Don't have time to check it all but I think we probably have to extract some duplication between the `send-on-chan` and `get-on-chain-fee` use cases / files. Probably like we discussed at the last code review with a `on-chain-fee-internal` file or something like that.

src/app/wallets/send-on-chain.ts
```diff
@@ -359,23 +372,47 @@ const executePaymentViaOnChain = async ({
...
-          totalFee = withdrawFeeCalculator.onChainWithdrawalFee({
-            minerFee,
-            bankFee: toSats(senderAccount.withdrawFee),
+        if (minerFee_ instanceof Error) {
+          const defaultMinersFee = toSats(5000)
+          logger.error({ err: minerFee_ }, "impossible to get fee for onchain payment")
+          addAttributesToCurrentSpan({
+            "payOnChainByWalletId.errorGettingMinerFee": true,
           })
+          minerFee = estimatedFee > defaultMinersFee ? estimatedFee : defaultMinersFee
```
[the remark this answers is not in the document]
nicolasburtey: simplified the logic. we don't really need `defaultMinersFee`

src/app/wallets/send-on-chain.ts
nicolasburtey: the first is an estimation, the second is the real fees. the issue is because we don't know in advance how much fees we're really going to pay. this is a limitation in how we use lnd currently. we only have an estimation on the fees that are doing to be paid, but can't have certainty on it.

the way to solve this would be to use psbt but this would implies a lot of the logic would have to be a on our side. this is not advisable to do it now, IMO should be done when we do a refactoring of our send onchain payment stack with a queue to do batch payment

nicolasburtey:
> Don't have time to check it all but I think we probably have to extract some duplication between the send-on-chan and get-on-chain-fee use cases / files. Probably like we discussed at the last code review with a on-chain-fee-internal file or something like that.

Agree, but can do that as a follow on PR. I'd prefer to merge/deploy this sooner than later as this would help to reduce the loops.

src/domain/primitives/index.types.d.ts
```diff
@@ -6,4 +6,5 @@ type WalletId = string & { readonly brand: unique symbol }
 type AccountId = string & { readonly brand: unique symbol }
 type Seconds = number & { readonly brand: unique symbol }
 type MilliSeconds = number & { readonly brand: unique symbol }
+type Days = number & { readonly brand: unique symbol }
```
bodymindarts, line 9: Just a note - we should be using `bigint` everywhere we don't allow fractions.

bodymindarts (approved):

PR #1188, opened 2022-03-15
feat: hold invoice

nicolasburtey:
this PR make invoice been settled solely by trigger, remove the need to query lnd from a main query from the user.

This is a [big performance optimization](https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/EGqhArMX7R2).

### DIFF
--- a/src/domain/bitcoin/lightning/index.ts
+++ b/src/domain/bitcoin/lightning/index.ts
@@ -21,9 +21,6 @@ export const PaymentSendStatus = {
   AlreadyPaid: { value: "already_paid" },
 } as const
 
-export const lnPaymentStatusEvent = (paymentHash: PaymentHash) =>
-  `LN-PAYMENT-STATUS-${paymentHash}`
-
 const sha256 = (buffer: Buffer) => createHash("sha256").update(buffer).digest("hex")
 const randomSecret = () => randomBytes(32)
 
--- a/src/domain/bitcoin/lightning/index.ts
+++ b/src/domain/bitcoin/lightning/index.ts
@@ -1,3 +1,5 @@
+import { createHash, randomBytes } from "crypto"
+
 export { decodeInvoice } from "./ln-invoice"
 export {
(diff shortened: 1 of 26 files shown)


nicolasburtey:
we could revisit once this is merged: https://github.com/lightningnetwork/lnd/issues/6338

nicolasburtey:
from Alex:

> I don't think there would be any issue from the LND perspective, although you can of course cancel the single invoice subscription after it's settled or expired/canceled so I would use that to minimize overall subscriptions

maybe we could also limit the BTC lightning invoice to something like 4h instead of the current 24h?

src/domain/bitcoin/lightning/index.ts
```diff
@@ -18,3 +20,16 @@ export const PaymentSendStatus = {
   Pending: { value: "pending" },
   AlreadyPaid: { value: "already_paid" },
 } as const
+
+export const lnPaymentStatusEvent = (paymentHash: PaymentHash) =>
```
vindard, line 24: This isn't used anywhere

src/graphql/error-map.ts
```diff
@@ -271,6 +271,10 @@ export const mapError = (error: ApplicationError): CustomApolloError => {
       message = "Invalid walletId for account."
       return new ValidationInternalError({ message, logger: baseLogger })
 
+    case "SecretDoesNotMatchAnyExistingHodlInvoiceError":
+      message = `Unknown error occurred (code: ${error.name})`
```
vindard, line 275: This message is a duplicate. The case can be added to the unhandled list below Ln 280

vindard (changes requested):
Initial review, 2 small nits

src/graphql/error-map.ts
nicolasburtey: yup thanks

nicolasburtey:
closing, as this PR has been replaced by #1395

PR #1191, opened 2022-03-17
feat: add qr code functionality to wallet generation script

krtk6160:


package.json
```diff
@@ -151,6 +152,7 @@
     "nodemon": "^2.0.15",
     "pino-pretty": "^5.1.0",
     "prettier": "^2.5.1",
+    "qrcode": "^1.5.0",
```
dolcalmi, line 155: not sure if we want to add additional packages for debug scripts. This can be done in an external script/project

dolcalmi (changes requested):

package.json
krtk6160: This seemed acceptable since we are doing the same [here](https://github.com/GaloyMoney/galoy/blob/main/src/debug/create-custom-invoices.ts#L8) with the `bip66` package, but now that you mentioned this that's probably just leftover.

Should we just remove them from `package.json` and add a step to the instructions of each script to install those packages manually?
Not sure I understand what you mean by "external script/project". Are you suggesting to wrap up the debug scripts in an npm package?

nicolasburtey: I guess the goal is to move using `monorepo` @dolcalmi ?

nicolasburtey:
@krtk6160 closing this as I think this is outdated and no longer useful. feel free to open as needed

PR #1192, opened 2022-03-17
feat: fake funding transaction script

krtk6160:


src/debug/create-txn.ts
```diff
@@ -0,0 +1,97 @@
...
+
+const main = async () => {
+  const args = process.argv
+  if (args.length === 3) {
+    const sats = parseInt(args[2])
+    const checkedSats = checkedToSats(sats)
+    if (checkedSats instanceof Error) return checkedSats
+
+    return consolidateFunds(checkedSats)
+  } else {
```
dolcalmi, line 86: else is not necessary

src/debug/create-txn.ts
```diff
@@ -0,0 +1,97 @@
...
+import {
+  AccountsRepository,
+  UsersRepository,
+  WalletsRepository,
+} from "@services/mongoose"
+
+const consolidateFunds = async (sats: Satoshis) => {
+  await setupMongoConnection()
+
+  const phone = "+1928282928" as PhoneNumber
```
dolcalmi, line 27: shouldn't be this from config or cmd param?

src/debug/create-txn.ts
krtk6160: removed

PR #1193, opened 2022-03-17
feat: delete payments by hash from lnd

vindard:
## Description

_Follows from #1041_

This PR adds a new method to the cronjob to delete all payments from lnd that have been completely persisted in our `lnpayments` collection. It only deletes payments older than 2 weeks.


src/app/lightning/delete-ln-payments.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+const checkAndDeletePaymentForHash = async ({
+  paymentHash,
+  pubkey,
+}: {
+  paymentHash: PaymentHash
+  pubkey: Pubkey
+}): Promise<boolean | ApplicationError> =>
+  asyncRunInSpan(
```
dolcalmi, line 30: if we are using Lightning in cron then we are wrapping this method twice. I think you can remove this wrapper and use the automatic wrapper from app index

src/servers/cron.ts
```diff
@@ -43,12 +43,19 @@ const main = async () => {
     if (result instanceof Error) throw result
   }
 
+  const deleteLndPaymentsBefore2Weeks = async () => {
+    const timestamp2Weeks = new Date(Date.now() - TWO_WEEKS_IN_MS)
```
dolcalmi, line 47: I think is better to use 1 or 2 months. 2 weeks is too short to handle current support cases

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -1352,6 +1352,120 @@ describe("UserWallet - Lightning Pay", () => {
...
+    // Check that payment no longer exists
+    const retrievedDeletedPayment = await lndService.lookupPayment({ paymentHash })
+    expect(retrievedDeletedPayment).toBeInstanceOf(PaymentNotFoundError)
+
+    // Check that deleting missing payment doesn't return error
+    const deletedAttempt = await lndService.deletePaymentByHash({ paymentHash })
+    expect(deletedAttempt).not.toBeInstanceOf(Error)
+  })
+
+  it("runs delete-payment cronjob", async () => {
```
dolcalmi, line 1397: this test should be in cron path (test)

dolcalmi (changes requested):
We also need to add fallback for admin LightningPaymentQuery (for deleted payments)

src/servers/cron.ts
nicolasburtey: I'll vote for 2 months. I think that as long as we only keep the payments data in a bounded time window, we won't let bolt.db grown to an unsustainable size.

src/app/lightning/delete-ln-payments.ts
vindard: The extra wrapper here is useful for getting more granular with certain attributes like `deleted`

- Trace without extra wrapper: https://ui.honeycomb.io/galoy/datasets/arvin-dev/result/v631Ah4Exxj/trace/yF7iqYvUk19?span=a36a1938aa52cf30
- Trace with extra wrapper: https://ui.honeycomb.io/galoy/datasets/arvin-dev/result/o9hyBy1s1pN/trace/JGYUFWQi3Uu?span=2abeaa8603d0ec78

vindard:
> We also need to add fallback for admin LightningPaymentQuery (for deleted payments)

Good catch, [resolved](https://github.com/GaloyMoney/galoy/pull/1193/commits/66c9eac2cf2444528adef06f2e0f9475278c646e)

test/integration/02-user-wallet/02-send-lightning.spec.ts
vindard: Done https://github.com/GaloyMoney/galoy/pull/1193/commits/2a33a6323a15a564912bdde4cb0948e16a1d4867

dolcalmi (dismissed):

dolcalmi (approved):

PR #1195, opened 2022-03-18
test: organise usd-send-lighting integration tests

vindard:
## Description

Separate our integration tests that involve usd wallets into distinct sections

### DIFF
--- a/test/integration/02-user-wallet/02-send-lightning.spec.ts
+++ b/test/integration/02-user-wallet/02-send-lightning.spec.ts
@@ -474,258 +474,6 @@ describe("UserWallet - Lightning Pay", () => {
     expect(imbalanceFinal).toBe(imbalanceInit - amountInvoice)
   })
 
-  // TODO: add probing scenarios
-  it("pay amountless invoice from usd wallet", async () => {
-    const dealerUsdWalletId = await getDealerUsdWalletId()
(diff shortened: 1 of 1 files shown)


dolcalmi (approved):

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -1352,6 +1100,263 @@ describe("UserWallet - Lightning Pay", () => {
   })
 })
 
+describe("USD Wallets - Lightning Pay", () => {
+  // TODO: add probing scenarios
+  describe("Lightning invoices with amounts", () => {
```
bodymindarts, line 1105: IMO this file has more tests than necessary. We do not (and indeed should not) exhaustively test all scenarios via integration tests. Most of the branching happens in the PaymentsFlowBuilder which is exhaustively unit tested. If we trust those tests (we should) the we can assume that the business logic is correct and we only need to have a test that touches the app-layer code.

My sense is that 2 tests perhaps 1 intraledger with amount and 1 not intraledger and no amount - or something like that - should be enough.

nicolasburtey: I agree long term but while we're actively developing the USD I think those integration tests are helpful. 

a good way to make sure we are not missing anything would be to do some code coverage, and make sure with both integration and unit test we have full coverage of the critical (ie: payment related) parts

PR #1196, opened 2022-03-18
add an expiration time when signing JWTs

mjdietzx:
When going through code I noticed the JWTs issued don't have an expiration. There is no default value for `expiresIn` so current behavior is that these JWTs never expire (unless/until we rotate `JWT_SECRET`). So in this PR I try to improve security in the case that a JWT is stolen/leaks.

Without much context I simply configured JWTs to expire after 1 week. This is a total guess on my part. On one hand security best-practices would suggest this should be much shorter (eg 15 minutes), on the other hand this likely affects user experience - probably don't want users being logged out automatically too soon, and refreshing tokens is additional work and I'm not sure where that falls on priorities.

I also realize there may be a reason that these tokens don't expire that I just don't know about. If that's the case please lmk. Bc I'm not sure if this change is actually wanted I'll leave this as a draft for now until I have some initial feedback (this PR could have just as easily been an issue but I figured I'll get up to speed w/ the codebase faster if I just implement this)


nicolasburtey:
you're spot on the fact it's best practice have shorten/not infinite token lifetime. also a best practive would be to have 2 tokens to be able to have a separate authentication servers. 

> I also realize there may be a reason that these tokens don't expire that I just don't know about. If that's the case please lmk.

yes there is a reason. currently the authentication is done by phone+text.
- it's very costly to send text, specially in a country like ES (it's like 100x the price compared to sending a text in US, pricing detail should be public on twilio)
- phone is very practical in developing country because it's easy for people to login back. but of course the risk is high (sim jack, lost phone number), but it's bearable if you consider most use the wallet as a spending wallet, not saving wallet
- this solution is transitory, we're currently replacing with the ory.sh auth stack. the first thing is to bring email in additional to phone as a way to login which is not only more secure, but also will "encourage" periodic login/logout without incurring exponential cost. if you last at recent merged PR, you'll see that this work is already ongoing.

bodymindarts:
Moving to JWT's that expire and can be refreshed / revoked is part of our ongoing re-write of our auth stack.
As a first step we are almost finished integrating with https://www.ory.sh/kratos/. Our next step (that will address the JWT issues) will be to add https://www.ory.sh/hydra/

mjdietzx:
Thanks for the explanation gents - makes sense! This PR seems moot so closing it. I'll take a look at the recent/ongoing ory.sh work in meantime

bodymindarts:
Most of that work is in the web wallet (eg https://github.com/GaloyMoney/web-wallet/pull/78)

PR #1197, opened 2022-03-18
feat: add a setting to make captcha mandatory

nicolasburtey:


src/config/yaml.ts
```diff
@@ -29,6 +29,10 @@ try {
 export const yamlConfig = merge(defaultConfig, customConfig)
 
 const ajv = new Ajv()
+
+// TODO: fix errors
+// const ajv = new Ajv({ allErrors: true, strict: "log" })
+
```
samerbuna, line 35: Please clean

samerbuna (approved):
The `userLogin` mutation should not be blocked. Only the `userRequestAuthCode`.

src/graphql/root/mutation/user-login.ts
```diff
@@ -24,6 +25,11 @@ const UserLoginMutation = GT.Field({
     input: { type: GT.NonNull(UserLoginInput) },
   },
   resolve: async (_, args, { logger, ip }) => {
+    const isCaptchaMandatory = getCaptcha().mandatory
+    if (isCaptchaMandatory) {
+      return { errors: [{ message: "use captcha endpoint to login" }] }
+    }
+
```
samerbuna, line 32: I don't think we need that here. This mutation is still used.

src/config/yaml.ts
nicolasburtey: the comment is voluntary. the current schema.ts validation has issue that are been shown with `strict: "log"`, but me or someelse should cleaned up later

nicolasburtey:
> The userLogin mutation should not be blocked. Only the userRequestAuthCode.

do you get on top of your head why we have 2 userRequestAuthCode mutation then?

samerbuna:
> do you get on top of your head why we have 2 userRequestAuthCode mutation then?

Not sure what you mean, we have that mutation used in both the main and admin schemas.

samerbuna (approved):

PR #1198, opened 2022-03-23
refactor: add LedgerFacade with new entry builder

UncleSamtoshi:
Continues the work of PR #1170:
- add ledger facade
- rewrite ledger entry-builder to work with expected PaymentFlow updates

### DIFF
--- a/src/app/wallets/send-intraledger.ts
+++ b/src/app/wallets/send-intraledger.ts
@@ -230,7 +230,7 @@ const executePaymentViaIntraledger = async ({
           amountDisplayCurrency,
           recipientWalletId,
           recipientWalletCurrency: recipientWallet.currency,
-          recipientUsername,
+          recipientUsername: recipientUsername || undefined,
           memoPayer,
         }),
       )
--- a/src/app/wallets/send-on-chain.ts
+++ b/src/app/wallets/send-on-chain.ts
@@ -241,7 +241,7 @@ const executePaymentViaIntraledger = async ({
             recipientWalletId: recipientWallet.id,
             recipientWalletCurrency: recipientWallet.currency,
             recipientUsername: recipientAccount.username,
-            memoPayer: memo ?? null,
+            memoPayer: memo || undefined,
           }),
       )
       if (journal instanceof Error) return journal
--- a/src/domain/ledger/index.types.d.ts
+++ b/src/domain/ledger/index.types.d.ts
@@ -124,14 +124,14 @@ type AddColdStorageTxSendArgs = {
 type IntraledgerTxArgs = {
   senderWalletId: WalletId
   senderWalletCurrency: WalletCurrency
-  senderUsername: Username | null
+  senderUsername?: Username
   description: string
   sats?: Satoshis
   cents?: UsdCents
   recipientWalletId: WalletId
   recipientWalletCurrency: WalletCurrency
-  recipientUsername: Username | null
-  memoPayer: string | null
+  recipientUsername?: Username
+  memoPayer?: string
 }
 
 type AddIntraLedgerTxSendArgs = IntraledgerTxArgs & {
@@ -139,7 +139,7 @@ type AddIntraLedgerTxSendArgs = IntraledgerTxArgs & {
 }
 
 type SendIntraledgerTxArgs = IntraledgerTxArgs & {
-  recipientUsername: Username | null
+  recipientUsername?: Username
   shareMemoWithPayee: boolean
   metadata:
     | AddLnIntraledgerSendLedgerMetadata
--- a/src/services/ledger/domain/entry-builder.ts
+++ b/src/services/ledger/domain/entry-builder.ts
@@ -52,13 +52,11 @@ const EntryBuilderFee = <M extends MediciEntry>({
       })
     }
 
-    const usdWithOutFee = calc.sub(usdWithFee, usdProtocolFee)
-    const btcWithOutFee = calc.sub(btcWithFee, btcProtocolFee)
     return EntryBuilderDebit({
       metadata,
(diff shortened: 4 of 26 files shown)


src/domain/ledger/index.types.d.ts
```diff
@@ -124,22 +124,22 @@ type AddColdStorageTxSendArgs = {
...
-  senderUsername: Username | null
+  senderUsername?: Username
   description: string
   sats?: Satoshis
   cents?: UsdCents
   recipientWalletId: WalletId
   recipientWalletCurrency: WalletCurrency
-  recipientUsername: Username | null
-  memoPayer: string | null
+  recipientUsername?: Username
```
UncleSamtoshi, line 133: @dolcalmi Any thoughts on changing these two metadata value types from null to undefined?

UncleSamtoshi: This is the place where transactions are brought from mongoose into our domain https://github.com/GaloyMoney/galoy/blob/21b76e1e261ea818f03d42a182b044a50397da8c/src/services/ledger/index.ts#L308

UncleSamtoshi: The current [LedgerTransaction type](https://github.com/GaloyMoney/galoy/blob/e9a9686ee704bd6e7cc09834d09c266be9c58739/src/domain/ledger/index.types.d.ts#L24) actually already uses undefined, but I'm not sure how a null value persisted in the database becomes an undefined value.

bodymindarts (dismissed):

src/app/wallets/send-intraledger.ts
```diff
@@ -230,7 +230,7 @@ const executePaymentViaIntraledger = async ({
           amountDisplayCurrency,
           recipientWalletId,
           recipientWalletCurrency: recipientWallet.currency,
-          recipientUsername,
+          recipientUsername: recipientUsername ?? undefined,
```
dolcalmi, line 233: are you sure we want to exclude `""`? else use `||`

src/app/wallets/send-on-chain.ts
```diff
@@ -241,7 +241,7 @@ const executePaymentViaIntraledger = async ({
             recipientWalletId: recipientWallet.id,
             recipientWalletCurrency: recipientWallet.currency,
             recipientUsername: recipientAccount.username,
-            memoPayer: memo ?? null,
+            memoPayer: memo ?? undefined,
```
dolcalmi, line 244: are you sure we want to exclude `""`? else please use `||`

UncleSamtoshi: I think it makes sense to exclude ```""```. Can you think of any scenario where we would want it?

src/services/ledger/domain/entry-builder.ts
```diff
@@ -1,194 +1,195 @@
...
         entry,
         metadata,
         staticAccountIds,
-        btcAmount: amount,
-        usdAmount: debitAmount,
-      })
+        btcAmount: calc.sub(btcWithFee, btcWithOutFee),
+        usdAmount: calc.sub(usdWithFee, usdWithOutFee),
+      }
+      withdrawUsdFromDealer(dealerData)
```
dolcalmi, line 158: this is a bad practice (it is really difficult to follow what is the purpose of the method) or we need to change the method name or we need to update entry in this line.

src/services/ledger/domain/entry-builder.ts
```diff
@@ -1,194 +1,195 @@
...
         entry,
         metadata,
         staticAccountIds,
-        btcAmount: debitAmount,
-        usdAmount: amount,
-      })
+        btcAmount: btcWithFee,
+        usdAmount: usdWithFee,
+      }
+      withdrawUsdFromDealer(dealerData)
```
dolcalmi, line 172: same here

src/services/ledger/domain/index.types.d.ts
```diff
@@ -3,6 +3,11 @@ type LedgerAccountId = string & { [ledgerAccountId]: never }
 
 type TxMetadata = any
```
dolcalmi, line 4: as far as I remember we did not have this alert. could you please add an eslint exception?

src/services/ledger/domain/legacy-entry-builder.ts
```diff
@@ -0,0 +1,247 @@
+import { AmountCalculator, WalletCurrency } from "@domain/shared"
```
dolcalmi, line 1: do we plan to remove this file? else, please rename

dolcalmi (dismissed):

src/services/ledger/domain/legacy-entry-builder.ts
UncleSamtoshi: Yes we plan on removing this file.

src/services/ledger/domain/entry-builder.ts
UncleSamtoshi: @bodymindarts This is Juan's only hold up with the PR. I believe the concern is that when we are using a separate method (now named ```AddUsdToBtcConversionToEntry```)  to make changes to the entry, we are obscuring that the entry is being changed. @dolcalmi by returning the entry from ```AddUsdToBtcConversionToEntry``` and changing the name, does this address your problem?

bodymindarts: I think I understand @dolcalmi comment and think that the renaming is the easiest way to address it. I wouldn't use a capitalised name however (as this typically indicates a kind of constructor). `addUsdToBtcConvertionSoEntry` seems good however.

bodymindarts (approved):

PR #1200, opened 2022-03-26
refactor: add optional hash property to 'transactions' selection in 'me'

vindard:
### Description

This PR adds an optional `hash` argument that can be used to fetch only the transactions that match a hash for a given user's wallets.


src/graphql/types/abstract/wallet.ts
```diff
@@ -24,7 +28,7 @@ const IWallet = GT.Interface({
       description: dedent`Transactions are ordered anti-chronologically,
       ie: the newest transaction will be first`,
       type: TransactionConnection,
-      args: connectionArgs,
+      args: { ...connectionArgs, hash: { type: PaymentHash } },
```
vindard, line 31: How can I add multiple types here? This should be a union of `PaymentHash | OnChainTxHash` (cc @samerbuna, @dolcalmi)

vindard: Yea I think I'll actually pause this one because #1210 may be a cleaner way to get at this information

PR #1209, opened 2022-03-28
feat: automated deployment on a raspiblitz

openoms:
Sets up the backend and the web-wallet on testnet with a single [install script](https://github.com/openoms/galoy/blob/self-hosting/scripts/deploy-raspiblitz.sh) and copies the necessary extra files from [galoy/scripts/assets directory](https://github.com/openoms/galoy/tree/self-hosting/scripts/assets).

Backend
- [x] set up all dependencies
- [x] install [the galoy API](https://github.com/GaloyMoney/galoy)
- [x] modded [Makefile](https://github.com/openoms/galoy/blob/self-hosting/scripts/assets/galoy-Makefile) and [docker-compose.selfhosted.yaml](https://github.com/openoms/galoy/blob/self-hosting/docker-compose.selfhosted.yml)
- [x] configure [galoy](https://github.com/openoms/galoy/blob/self-hosting/.envrc.selfhosted)
- [x] self-signed HTTPS endpoint for the [galoy-API](https://github.com/openoms/galoy/blob/self-hosting/scripts/assets/galoy-api_ssl.conf)

Web-wallet
moved to https://github.com/GaloyMoney/web-wallet/pull/87

Testing
- [x] running on amd64 (headless server)
- [ ] running on aarch64 (RPi4 with SSD)


scripts/assets/galoy-API_ssl.conf
```diff
@@ -0,0 +1,19 @@
+## galoy-admin-API_ssl.conf
```
nicolasburtey, line 1: I think we're using only lowercase in our repo for filename to avoid issues with the fact not all OS are case sensitives

.envrc
```diff
@@ -2,11 +2,12 @@ export DOCKER_HOST_IP=${DOCKER_HOST_IP:-127.0.0.1}
...
 export NODE_ENV=development
 
-export LND1_TLS=$(base64 dev/lnd/tls.cert | tr -d '\n\r')
-export LND2_TLS=$(base64 dev/lnd/tls.cert | tr -d '\n\r')
-export LNDONCHAIN_TLS=$(base64 dev/lnd/tls.cert | tr -d '\n\r')
-export TLSOUTSIDE1=$(base64 dev/lnd/tls.cert | tr -d '\n\r')
-export TLSOUTSIDE2=$(base64 dev/lnd/tls.cert | tr -d '\n\r')
+LND_TLS=$(base64 ~/.lnd/tls.cert | tr -d '\n\r')
+export LND1_TLS=${LND_TLS:-$(base64 dev/lnd/tls.cert | tr -d '\n\r')}
+export LND2_TLS=${LND_TLS:-$(base64 dev/lnd/tls.cert | tr -d '\n\r')}
```
nicolasburtey, line 7: you may need only one TLS here (ie: not needed to set LND2_TLS), although I'm not sure how the backend would react. this may need some changes in the yaml also

the general idea here is that our backend can work with multiple lnd, and we run 2 by default, but on a raspiblitz we can just depend on one

I'm not recommending looking at it now, but after there is a first end-to-end integration working

openoms: yes, currently just using the tls.cert and the macaroon of the one LND instance if it exists. Alternatives could be added for other bases as well in the longer term.
I there is no local file it should fall back to the docker container (in the case of the macaroons) or the predefined value in case of the TLS cert here.

scripts/assets/galoy-API_ssl.conf
openoms: fixed in

openoms:
Testing can be done with:
```
# download 
https://raw.githubusercontent.com/openoms/galoy/self-hosting/scripts/deploy-raspiblitz.sh -O deploy-raspiblitz.sh
# run
bash deploy-raspiblitz.sh
```

Requirements:
RaspiBlitz v1.7.2+ patched to 'dev'
LND on Testnet activated"

openoms:
A few issues when trying to install on a Raspberry Pi 4 (arm64):
The mongodb container not compatible:
```
 ₿ docker container logs -f --details galoy-mongodb-1
 standard_init_linux.go:228: exec user process caused: exec format error
 standard_init_linux.go:228: exec user process caused: exec format error
```
would need to find an alternative to https://hub.docker.com/r/bitnami/mongodb

due to the `grpc_tools` package is not being available for arm: https://github.com/grpc/grpc-node/issues/1405
would either need to 
* avoid the installation completely (currently needed for the services: `trigger.js` and `cron.js`)
* override the `yaml.lock` file to use the equivalent of: `npm install grpc-tools --target_arch=x64`
* test to install all yarn dependencies for amd64

Because of these and also the scalability and hardware reliability issues with the RPi will concentrate on working with the `amd64` platform for now.

nicolasburtey:
can this be close or do you still work on it @openoms ?

openoms:
Closing for now as working on the kubernetes based deployment from the charts.

PR #1218, opened 2022-03-31
chore: adding log to debug new fees issues

nicolasburtey:
some fees are underestimated when using proportional onchain fees. this leads to some balance that may be slightly 0. 
this PR helps getting more info on the issue


bodymindarts (approved):

nicolasburtey:
@vindard @dolcalmi should I get this merged in regards to the fee/balance issue we have seen recently? or do you think the issue is coming from somewhere else (I'll close it in that case)?

nicolasburtey:
Closing for now

PR #1229, opened 2022-04-06
chore(deps): upgrade mongoose

bodymindarts:


krtk6160 (approved):

src/services/mongoose/accounts.ts
```diff
@@ -62,8 +62,8 @@ export const AccountsRepository = (): IAccountsRepository => {
     try {
       const accounts: UserRecord[] = await User.find(
         {
-          title: { $exists: true, $ne: null },
-          coordinates: { $exists: true, $ne: null },
+          title: { $exists: true, $ne: undefined },
```
dolcalmi, line 65: did we do a migration forr this?

yarn.lock
```diff
@@ -1868,6 +1868,13 @@
   dependencies:
     bson "*"
 
+"@types/bson@1.x || 4.0.x":
```
dolcalmi, line 1871: please remove and re-create yarn.lock with `yarn install`

src/services/mongoose/accounts.ts
bodymindarts: no - migration for this... I guess we should check if this is an issue

src/graphql/root/mutation/device-notification-token-create.ts
```diff
@@ -28,7 +28,7 @@ const DeviceNotificationTokenCreateMutation = GT.Field<
       const user = await User.findOne({ _id: toObjectId<UserId>(uid) })
       if (!user) throw Error("find user issue")
 
-      user.deviceToken.addToSet(deviceToken)
+      user.deviceToken.push(deviceToken)
```
dolcalmi, line 31: if you change this you need to validate that device token is not present in the array

dolcalmi (changes requested):

src/graphql/root/mutation/device-notification-token-create.ts
bodymindarts: I see... perhaps there is another method `addToSet` is no longer present.

src/services/mongoose/accounts.ts
bodymindarts: I have checked on bbw that the queries:
```
{ title: { $exists: true, $ne: null }, coordinates: { $exists: true, $ne: null} }
```
and 
```
{ title: { $exists: true, $ne: undefined }, coordinates: { $exists: true, $ne: undefined } }
```
yield identical results so I'd say we do not need a migration for this.

krtk6160 (approved):

dolcalmi (approved):

PR #1232, opened 2022-04-07
feat: add rewards ip config

dolcalmi:
Add ip country and asn validation to rewards

### DIFF
--- a/src/domain/users-ips/index.types.d.ts
+++ b/src/domain/users-ips/index.types.d.ts
@@ -11,7 +11,7 @@ type IPMetadataValidatorArgs = {
 }
 
 type IPMetadataValidator = {
-  validateForReward(ipMetadata?: IPType): true | ApplicationError
+  validateForReward(ipMetadata?: IPType): true | ValidationError
 }
 
 interface IUsersIPsRepository {
--- a/src/domain/users-ips/index.types.d.ts
+++ b/src/domain/users-ips/index.types.d.ts
@@ -3,6 +3,13 @@ type UserIPs = {
   lastIPs: IPType[]
 }
 
+type IPMetadataValidatorArgs = {
+  denyIPCountries: string[]
+  allowIPCountries: string[]
+  denyASNs: string[]
+  allowASNs: string[]
+}
+
 type IPMetadataValidator = {
   validateForReward(ipMetadata?: IPType): true | ApplicationError
 }
--- a/src/domain/users-ips/index.types.d.ts
+++ b/src/domain/users-ips/index.types.d.ts
@@ -3,6 +3,10 @@ type UserIPs = {
   lastIPs: IPType[]
 }
 
+type IPMetadataValidator = {
+  validateForReward(ipMetadata?: IPType): true | ApplicationError
+}
+
 interface IUsersIPsRepository {
   update(userIp: UserIPs): Promise<true | RepositoryError>
   findById(userId: UserId): Promise<UserIPs | RepositoryError>
--- a/src/domain/users-ips/ip-metadata-validator.ts
+++ b/src/domain/users-ips/ip-metadata-validator.ts
@@ -10,12 +10,7 @@ export const IPMetadataValidator = ({
   allowIPCountries,
   denyASNs,
   allowASNs,
-}: {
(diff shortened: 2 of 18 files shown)


test/unit/domain/users-ips/ip-metadata-validator.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+  ip: "89.187.173.251" as IpAddress,
+  firstConnection: new Date(),
+  lastConnection: new Date(),
+  asn: "AS60068",
+  provider: "ISP",
+  country: "United States",
+  isoCode: "US",
+  region: "Florida",
+  city: "Miami",
+  proxy: false,
```
vindard, line 26: Could also use a name like `isViaProxy` or something similar to be clearer here

test/unit/domain/users-ips/ip-metadata-validator.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+
+describe("IPMetadataValidator", () => {
+  describe("validateForReward", () => {
+    it("returns true for empty config", () => {
+      const validator =
+        IPMetadataValidator(defaultConfig).validateForReward(defaultIpInfo)
+      expect(validator).toBe(true)
+    })
+
+    it("returns true for an allowed country", () => {
```
vindard, line 37: Can it return false/error for a country not in a defined allowed-list? Can probably add test-cases for:
- country not in defined allow-list fails
- country not in defined deny-list passes

src/domain/users-ips/index.types.d.ts
```diff
@@ -3,6 +3,17 @@ type UserIPs = {
...
 
+type IPMetadataValidatorArgs = {
+  denyIPCountries: string[]
+  allowIPCountries: string[]
+  denyASNs: string[]
+  allowASNs: string[]
+}
+
+type IPMetadataValidator = {
+  validateForReward(ipMetadata?: IPType): true | ApplicationError
```
vindard, line 14: `ValidationError` here instead of `ApplicationError`?

src/domain/users-ips/ip-metadata-validator.ts
```diff
@@ -0,0 +1,37 @@
...
+  InvalidIPMetadataASNError,
+} from "@domain/errors"
+
+export const IPMetadataValidator = ({
+  denyIPCountries,
+  allowIPCountries,
+  denyASNs,
+  allowASNs,
+}: IPMetadataValidatorArgs): IPMetadataValidator => {
+  const validateForReward = (ipMetadata?: IPType): true | ApplicationError => {
```
bodymindarts, line 14: Same here dependency graph is app -> domain. Not the other way around. Domain should return specific errors.

bodymindarts (dismissed):

src/domain/users-ips/index.types.d.ts
dolcalmi: Fixed and `PhoneMetadataValidator` too https://github.com/GaloyMoney/galoy/pull/1232/commits/9333231b2b3e87f0f2eb04110a8f03bcee7ac70b

src/domain/users-ips/ip-metadata-validator.ts
dolcalmi: Fixed and `PhoneMetadataValidator` too https://github.com/GaloyMoney/galoy/pull/1232/commits/9333231b2b3e87f0f2eb04110a8f03bcee7ac70b

test/unit/domain/users-ips/ip-metadata-validator.spec.ts
dolcalmi: Added additional tests in regards to your suggestion https://github.com/GaloyMoney/galoy/pull/1232/commits/2507717d52130b03fb7521f8a35202c44460fb7c

test/unit/domain/users-ips/ip-metadata-validator.spec.ts
dolcalmi: I spent some time checking this and probably require a migration to have consistent data.. do you mind if we include this in another PR?

Issue #1220, opened 2022-04-02
Make the myUpdates subscription work for receivers


samerbuna:
The Galoy API has a `myUpdates` subscription operation that an authenticated user can use to get instant updates about new transactions that happen under their wallets. Currently, when a transaction is settled on the Galoy backend, the `myUpdates` subscription pushes updates to the sender of that transaction. We need it to to push an update to the receiver account as well.

To test the current flow and prepare for the new feature, run the Galoy API locally and then open 3 different browser sessions for http://localhost:4002/graphql

In one session, authenticate as userA (get token and put it in HTTP headers), and as userB in another session. Then run the `myUpdates` subscription for both sessions:

```gql
subscription myUpdates {
  myUpdates {
    me {
      defaultAccount {
        wallets {
          balance
        }
      }
    }
    update {
      __typename
      ... on LnUpdate {
        walletId
        paymentHash
        status
      }
      ... on OnChainUpdate {
        walletId
        txHash
        amount
      }
      ... on IntraLedgerUpdate {
        walletId
        amount
      }
    }
  }
}
```

In the third browser session, using userA token in the HTTP headers as well, make a payment to userB (using `intraLedgerPaymentSend` mutation). You'll see that the subscription operation for userA receives an IntraLedgeUpdate payload while the subscription operation for userB does not.

userB should receive an update. This should work with LN/OnChain payments as well. Both sender and receiver should get updates about the transaction.

nicolasburtey:
has this been resolved @samerbuna ?

PR #1234, opened 2022-04-11
fix: ensure recipients of intraledger payments receive notifications via gql subscription

### LINKED ISSUE #934: subcription may not be working for intraledger tx
to reproduce:

- open https://api.staging.galoy.io/graphql
- get a JWT by login in, add it to the headers
- use the following subscription:

```
subscription myUpdates {
    myUpdates {
      errors {
        message
      }
      me {
        id
        defaultAccount {
          id
          wallets {
            id
            walletCurrency
            balance
          }
        }
      }
      update {
        type: __typename
        ... on LnUpdate {
          paymentHash
          status
        }
        ... on OnChainUpdate {
          txNotificationType
          txHash
          amount
          usdPerSat
        }
        ... on IntraLedgerUpdate {
          txNotificationType
          amount
          usdPerSat
        }
      }
    }
  }
```

from the mobile app, do some quizzes. there is no updates being sent by the subscription

note: 
- it could be that this is only for quizzes question because it using internally another path than the one that originate from a "pay to username" AFAIR
- I have tried invoice and this is working.
- I have not tested the onchain tx.

mjdietzx:
Prior to this commit, a bug in `notificationsService.intraLedgerPaid` resulted in the
`NotificationType.IntraLedgerPayment` being published to the sender twice. Now the
notification is published to the sender, and then the recipeint, correcting this
unintended behavior.

Resolves #934 and #1220.

### DIFF
--- a/src/services/notifications/index.ts
+++ b/src/services/notifications/index.ts
@@ -226,7 +226,7 @@ export const NotificationsService = (logger: Logger): INotificationsService => {
         walletId: WalletId
         type: NotificationType
       }) => {
-        const wallet = await WalletsRepository().findById(senderWalletId)
+        const wallet = await WalletsRepository().findById(walletId)
         if (wallet instanceof Error) return wallet
 
         const account = await AccountsRepository().findById(wallet.accountId)


mjdietzx:
Notice the in-line function (L222)
```js
const publish = async ({
  walletId,
  type,
}: {
```

was ignoring the `walletId` param and using the upper-scope `senderWalletId` instead when finding the `wallet` and then `account`, and therefore `pubsub.publish(accountUpdatedEventName, {` was always directed to the sender's account `ACCOUNT_UPDATE-${sendersAccountId}`

dolcalmi (approved):

PR #1237, opened 2022-04-12
feat: add new Lightning Payment Flow Builder to handle payments

vindard:
**_Note:_** _Converted to draft until #1269 is merged_

## Description

This PR adds a new `Lightning Payment Flow Builder` to be used in both lightning-fee-estimation and send-lightning-payment use-case functions. This is a refactor from PR #1176 where the work was initially done.

The relevant starting diff between these branches is [`payments`..`refactor-payments`](https://github.com/GaloyMoney/galoy/compare/6c30985d9e7c07711c11d2947588bf1c7ce86fb7..f15039a7ca43360960954b0b0380401840d6e6f9).

### Motivation

These changes refactor the complexity of handling currency conversions between BTC and USD wallets into a unit-tested builder object (functional core). This significantly thins out the logic needed in use-case methods (imperative shell) and improves test coverage of the required complexity. 

### Open questions
- [x] ~How will we handle display currency calculation? https://github.com/GaloyMoney/galoy/pull/1237#discussion_r855521054~ Added an explicit conversion step
- [x] ~How will we handle the mobile app expecting "dollars" for the `usd` display amount property which we are now writing `cents` values to?~ We left `usd` and `feeUsd` as-is for now and added new `displayAmount` and `displayFee` properties. These will be migrated away in a future PR.

### Open `TODO`s
- [x] ~Unit test for `NewFeeReimbursement` object~ (https://github.com/GaloyMoney/galoy/pull/1237/commits/64afb07623c313f7622a019b70998e71bf2e01b9)
- [x] ~Swap new use-cases into gql mutations~ (https://github.com/GaloyMoney/galoy/pull/1237/commits/57620ae05ce62bc23d0337ce4490eb1ade01195c)
- [x] ~Add lnUsdNoAmountFeeProbe~ (https://github.com/GaloyMoney/galoy/pull/1237/commits/d1839f7cca7c7768dbec38d0e0684d84c9b18c44)
- [x] ~Get `DealerPriceService` and `NewDealerPriceService` working for e2e tests~ (https://github.com/GaloyMoney/galoy/pull/1237/commits/e5ab7d01a2a80fb8e777a04e20411ec7200e7285)
- [x] ~swap `newReimburseFee` into updatePendingPayments~ (https://github.com/GaloyMoney/galoy/pull/1237/commits/ed09b5cfb3e581891d8f6e6c8aa5347aebce2d3d)
- [x] ~Remove old use-case methods and dependencies~ (https://github.com/GaloyMoney/galoy/compare/62cd76205595beba49ac4547241005f63d8fdb70...6262f89fb204687861774e4a9ef51b3919761d45)
- [x] ~Rename duplicated types/objects to drop the `New` prefixes~ (https://github.com/GaloyMoney/galoy/compare/62cd76205595beba49ac4547241005f63d8fdb70...6262f89fb204687861774e4a9ef51b3919761d45)
- [x] ~post-refactor, decide if midPriceFn should fallback from dealer to PriceService~ re-added fallback since we've isolated the midPrice code paths to no-hedge paths and this is only used for limits checks (https://github.com/GaloyMoney/galoy/pull/1237/commits/694b4a7ab3c5c433b2fffb8fa148702e360912c0)
- [x] ~Decide if to switch PaymentsFlowRepository to redis~ **_decided no, see commit message_** (https://github.com/GaloyMoney/galoy/pull/1237/commits/dc28fd2d7c284c42f7445603705ccfc38327c19e)
- [x] ~Fix: for midPriceFn, PriceService gives us quote in `DollarsPerSat` but current assumption is `CentsPerSat`~ (https://github.com/GaloyMoney/galoy/pull/1237/commits/52a39a37a16b56ee7c4f035d35eb4722328942df)
- [x] ~Add back `sats` to ledger transactions via a db migration and adjust `recordSend` here~ see PR #1269

**Additional considerations:**
- [x] ~introduce fallback to PriceService for midPrice if Dealer errors for non-hedged paths~; _in long term, should this optimization be handled by dealer? (from code review)_ (https://github.com/GaloyMoney/galoy/pull/1237/commits/376d0a8dfc8d6c0ab7f2cea854f12f193f9174e6)
- [x] ~ensure no-dealer-needed code paths work (see https://github.com/GaloyMoney/galoy/pull/1237#issuecomment-1103101186)~ (https://github.com/GaloyMoney/galoy/pull/1237/commits/9722e6289002a3caee0385a0cd520453723cd7c2 & https://github.com/GaloyMoney/galoy/pull/1237/commits/8d1bbc16815946733007b83d350a025d1d342237)

## For Future PRs

- better handle display currencies (https://github.com/GaloyMoney/galoy/pull/1237/commits/5b63cb7aeb75cca7e3ae257f620c094859872d4b & https://github.com/GaloyMoney/galoy/pull/1237#discussion_r865883451)

- Fix `Transaction` object in graphql (see [here](https://github.com/GaloyMoney/galoy/blob/761b4ae476b2259bef48e06fa3c3e5ab90966477/src/graphql/types/object/transaction.ts#L38)) to accommodate CentAmount transactions

- split `withConversion` into midPriceFn vs. dealerFn code paths for readability

- remove old DealerPriceService when wallet-invoice refactor is done

- remove old LimitChecker when payment flow for onchain is done

- remove `feeUsd`, `usd` props from transactions and migrate old values into new ones from https://github.com/GaloyMoney/galoy/issues/1257#issuecomment-1110182267
  - clean up unused properties in `LedgerTransaction` type (https://github.com/GaloyMoney/galoy/pull/1083)

- clean up `paymentRequest` input types in use-case functions and `decodeInvoice` method; should be string coming in and validated `EncodedPaymentRequest` out

- clean up redundant update method in LightningPaymentFlow repo (https://github.com/GaloyMoney/galoy/pull/1237#discussion_r865963180)

- clean up how payment flows get persisted, deleted etc.
   - remove lnd dependency and `pending` property in update-pending
   - simplify TTL calculation
   - consider moving into redis with new simplified TTL determination

- clean up `AccountValidator` method: https://github.com/GaloyMoney/galoy/pull/1237#discussion_r865926844

- create a facade method for FeeReimbursements metadata: https://github.com/GaloyMoney/galoy/pull/1237#discussion_r865924236

**From PR review comments**
- change paymentRequest types to string and add validation step (https://github.com/GaloyMoney/galoy/pull/1237#discussion_r856279468)
- refactor `ByWalletId` to be `ByWalletDescriptor` (https://github.com/GaloyMoney/galoy/pull/1237#discussion_r856221436)
- refactor notifications amount types (https://github.com/GaloyMoney/galoy/pull/1237#discussion_r856222175)


vindard:
These 4 commits fix an issue where we were getting a rounded down 0-amount coming into the PriceRatio for fee reimbursements: https://github.com/GaloyMoney/galoy/compare/c058a7fbf9abec3cb665556697d7ca8f1a22e6e3...4a83c9242deaa6ba472ade6ab9aac973eea26b06

vindard:
These 4 commits fix an issue where we were getting a rounded down 0-amount coming into the PriceRatio for fee reimbursements: https://github.com/GaloyMoney/galoy/compare/c058a7fbf9abec3cb665556697d7ca8f1a22e6e3...55f8167c06ce8e3dc8dea811b1ff9646fdd59205

vindard:
These 4 commits fix an issue where we were getting a rounded down 0-amount coming into the PriceRatio for fee reimbursements: https://github.com/GaloyMoney/galoy/compare/e626ed7946a20b44ae7a908731c171a753faf8ff...b1f73448c36560d68bacf52758bf57a193bbbd14

This commit changes the rounding behaviour of `PriceRatio` to address an issue with fee-reimbursement calculations: https://github.com/GaloyMoney/galoy/pull/1237/commits/5a798d752f8c61103fd113fdf79ec9f27be53e61

vindard:
Added a dealer interface wrapper around the existing Price service (https://github.com/GaloyMoney/galoy/pull/1237/commits/e5ab7d01a2a80fb8e777a04e20411ec7200e7285) to:
- get e2e tests working in absence of a working Dealer Price service in test environment
- have a working instantiation for environments that won't have a dealer service (BTC-only environments)

A cleaner way to address this would be to:
- clean up code-paths so that non-hedging pathways can flow in a way where dealer is not even instantiated
- pass non-dealer midPrice functions directly from PriceService to payment flow builder instead of via a Dealer interface wrapper around the PriceService

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -1,10 +1,12 @@
 import { createHash, randomBytes } from "crypto"
 
-import { Wallets, Lightning } from "@app"
+import { Wallets, Lightning, Payments } from "@app"
+import { getMidPriceRatio } from "@app/payments/helpers"
```
vindard, line 4: How should we handle a case like this where technically the methods aren't needed outside of the `payments` package but I'm importing them in tests to compare against. Should we also expose these in the top level `payments/index`, or is this direct import ok?

src/app/payments/send-lightning.ts
```diff
@@ -0,0 +1,653 @@
...
+        const unitForMsg = senderWallet.currency === WalletCurrency.Btc ? "sats" : "cents"
+        return new InsufficientBalanceError(
+          `Payment amount '${paymentAmount.amount}' ${unitForMsg} exceeds balance '${balance.amount}'`,
+        )
+      }
+
+      const journal = await LockService().extendLock({ logger, lock }, async () => {
+        const lnIntraLedgerMetadata = LedgerFacade.LnIntraledgerLedgerMetadata({
+          // FIXME: display currency
+          amountDisplayCurrency: Number(
```
vindard, line 440: How are we thinking to calculate this?

Right now I'm just passing the USD-equivalent amount straight through (can be from either dealer or price service). Should we explicitly add another step here to convert from the USD-amount to the display currency amount? It'll be a no-op for USD display currency, but will be applicable for CRC (cc @dolcalmi)

src/app/payments/send-lightning.ts
```diff
@@ -0,0 +1,653 @@
...
+          `Payment amount '${paymentAmount.amount}' ${unitForMsg} exceeds balance '${balance.amount}'`,
+        )
+      }
+
+      const lndService = LndService()
+      if (lndService instanceof Error) return lndService
+
+      const journal = await LockService().extendLock({ logger, lock }, async () => {
+        const metadata = LedgerFacade.LnSendLedgerMetadata({
+          // FIXME: display currency
```
vindard, line 561: https://github.com/GaloyMoney/galoy/pull/1237/files#r855521054

default.yaml
```diff
@@ -9,6 +9,10 @@ displayCurrency:
 # username of the account that will be used to fund rewards
 funder: "FunderWallet"
 
+dealer:
+  usd:
+    hedgingEnabled: false
```
nicolasburtey, line 14: more than `hedgingEnabled`, should it be a broader setting `usdAvailable`?

nicolasburtey: (note: I know I was the one using usd specifically, but as we start to see project that could use eur as a store of value, I'm wondering if we should rename usd to fiat in a follow up PR. I initially did the distinction of using usd to separate with the displayCurrency, but that was probably a bit short sighted.)

src/app/payments/get-protocol-fee.ts
```diff
@@ -0,0 +1,129 @@
...
+  })
+}
+
+export const getNoAmountLightningFeeEstimation = async ({
+  walletId,
+  paymentRequest,
+  amount,
+}: {
+  walletId: string
+  paymentRequest: EncodedPaymentRequest
```
nicolasburtey, line 44: `walletId` and `number` has not been validated yet but `paymentRequest` has?

nicolasburtey: is the type correct?

src/app/payments/get-protocol-fee.ts
```diff
@@ -0,0 +1,129 @@
...
+  const builder = await constructPaymentFlowBuilder({
+    senderWallet,
+    invoice,
+    uncheckedAmount,
+    usdFromBtc: dealer.getCentsFromSatsForImmediateBuy,
+    btcFromUsd: dealer.getSatsFromCentsForImmediateSell,
+  })
+  if (builder instanceof Error) return builder
+
+  const usdPaymentAmount = await builder.usdPaymentAmount()
```
nicolasburtey, line 87: we go through this whether the wallet is USD or BTC?

src/app/payments/get-protocol-fee.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  let paymentFlow
+  if (!(await builder.needsRoute())) {
+    const limitCheck = await newCheckIntraledgerLimits({
+      amount: usdPaymentAmount,
+      wallet: senderWallet,
+      priceRatio,
+    })
+    if (limitCheck instanceof Error) return limitCheck
+
```
nicolasburtey, line 103: this should be prior to the `if` to avoid duplication?

src/app/payments/get-protocol-fee.ts
```diff
@@ -0,0 +1,129 @@
+import { decodeInvoice } from "@domain/bitcoin/lightning"
```
nicolasburtey, line 1: just for info, seen the filename I was expecting that `payments/protocol-fee` would also cover the onchain fees.

src/app/payments/helpers.ts
```diff
@@ -0,0 +1,296 @@
...
+
+      const usdPaymentAmount = {
+        amount: BigInt(Math.ceil(Number(amount.amount) * midPriceRatio)),
+        currency: WalletCurrency.Usd,
+      }
+
+      addAttributesToCurrentSpan({
+        "usdFromBtcMidPriceFn.midPriceRatio": midPriceRatio,
+        "usdFromBtcMidPriceFn.incoming.amount": Number(amount.amount),
+        "usdFromBtcMidPriceFn.incoming.unit":
```
nicolasburtey, line 51: the `incoming.unit` comparison feels off to me. 

we're looking at the `amount.currency` which is of type `BtcPaymentAmount`

something is probably off here I think

nicolasburtey: maybe it's just because the code is duplicated below, but in that case code should be dedupe

src/app/payments/helpers.ts
```diff
@@ -0,0 +1,296 @@
...
+  senderWallet,
+  invoice,
+  uncheckedAmount,
+  usdFromBtc,
+  btcFromUsd,
+}: {
+  senderWallet: Wallet
+  invoice: LnInvoice
+  uncheckedAmount?: number
+  usdFromBtc
```
nicolasburtey, line 138: fallback type any?

default.yaml
bodymindarts: I agree this is a valid inquiry - but this PR is already huge... lets try to push as many corrections as possible to follow up PRs and focus here on critical things.

Comments are still useful as reminders of course.

src/domain/ledger/fee-reimbursement.ts
```diff
@@ -1,12 +1,30 @@
...
-    return feeDifference
+export const FeeReimbursement = ({
+  prepaidFeeAmount,
+  priceRatio,
+}: {
+  prepaidFeeAmount: {
+    btc: BtcPaymentAmount
+    usd: UsdPaymentAmount
+  }
+  priceRatio: PriceRatio
```
bodymindarts, line 13: IIsn't passing in the price Ratio redundant? We're already passing in both `btc` and `usd` so we can construct the price ratio in the function right?

src/domain/ledger/index.types.d.ts
```diff
@@ -20,14 +20,14 @@ type LedgerJournal = {
...
 
 // Differentiate fields depending on what 'type' we have (see domain/wallets/index.types.d.ts)
-type LedgerTransaction = {
+type LedgerTransaction<S extends WalletCurrency> = {
   readonly id: LedgerTransactionId
   readonly walletId: WalletId | undefined // FIXME create a subclass so that this field is always set for liabilities wallets
   readonly type: LedgerTransactionType
-  readonly debit: Satoshis
-  readonly credit: Satoshis
+  readonly debit: S extends "BTC" ? Satoshis : UsdCents
```
bodymindarts, line 27: We want to move away from these types completely - I guess you are looking to do this in another PR?

src/domain/ledger/index.types.d.ts
```diff
@@ -211,42 +227,56 @@ interface ILedgerService {
...
 
   getTransactionsByHash(
     paymentHash: PaymentHash | OnChainTxHash,
-  ): Promise<LedgerTransaction[] | LedgerServiceError>
+  ): Promise<LedgerTransaction<WalletCurrency>[] | LedgerServiceError>
 
   getTransactionsByWalletId(
     walletId: WalletId,
-  ): Promise<LedgerTransaction[] | LedgerServiceError>
+  ): Promise<LedgerTransaction<WalletCurrency>[] | LedgerServiceError>
```
bodymindarts, line 238: The `ByWalletId` query should / could probably be moved to `ByWalletDescriptor` so we can return the correct `LedgerTransaction<T>`? (not for this PR).

src/domain/ledger/index.types.d.ts
```diff
@@ -211,42 +227,56 @@ interface ILedgerService {
...
 
   getPendingPaymentsCount(walletId: WalletId): Promise<number | LedgerServiceError>
 
   getWalletBalance(walletId: WalletId): Promise<CurrencyBaseAmount | LedgerServiceError>
 
-  allPaymentVolumeSince(args: IGetVolumeArgs): VolumeResult
+  getWalletBalanceAmount<S extends WalletCurrency>({
+    walletId,
+    walletCurrency,
+  }: {
```
bodymindarts, line 258: this should be type `WalletDescriptor`

src/domain/notifications/index.types.d.ts
```diff
@@ -18,14 +18,14 @@ type OnChainTxPaymentArgs = OnChainTxBaseArgs
 type LnInvoicePaidBitcoinWalletArgs = {
   paymentHash: PaymentHash
   recipientWalletId: WalletId
-  sats: Satoshis
+  sats: Satoshis | bigint
```
bodymindarts, line 21: shouldn't this be `PaymentAmount<T>`?

src/domain/payments/index.types.d.ts
```diff
@@ -0,0 +1,194 @@
...
+  convertFromBtc(convert: BtcPaymentAmount): UsdPaymentAmount
+  usdPerSat(): DisplayCurrencyPerSat
+}
+
+type PaymentFlowState<S extends WalletCurrency, R extends WalletCurrency> = {
+  senderWalletId: WalletId
+  senderWalletCurrency: S
+  settlementMethod: SettlementMethod
+  paymentInitiationMethod: PaymentInitiationMethod
+  paymentHash: PaymentHash
```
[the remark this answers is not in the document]
bodymindarts: For now its just for lightning yeah... onchain USD integration is anot

src/app/payments/send-lightning.ts
```diff
@@ -0,0 +1,653 @@
...
+  if (limitCheck instanceof Error) return limitCheck
+
+  const { paymentHash } = decodedInvoice
+
+  // - get cached route if exists
+  const { rawRoute, outgoingNodePubkey } = paymentFlow.routeDetails()
+
+  // - validate route amount?
+
+  // - amount + fee in display currency?
```
bodymindarts, line 536: What is the purpose of this comment? Is this blocking?

src/services/payment-flow/index.ts
```diff
@@ -0,0 +1,113 @@
+import { UnknownRepositoryError } from "@domain/errors"
+import { CouldNotFindLightningPaymentFlowError, PaymentFlow } from "@domain/payments"
+import { WalletCurrency } from "@domain/shared"
+
+import { PaymentFlowState } from "./schema"
+
+export const PaymentFlowStateRepository = (): IPaymentFlowRepository => {
+  const persistNew = async <S extends WalletCurrency, R extends WalletCurrency>(
```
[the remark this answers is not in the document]
vindard: I left a note on [this commit](https://github.com/GaloyMoney/galoy/commit/dc28fd2d7c284c42f7445603705ccfc38327c19e) about that.

There ended up being a case when updating pending payments where fetching an already constructed payment flow was much easier than trying to reconstruct it from the ledger entries we'd persisted. If we think the redis + TTL has obvious advantages though, I think we can still switch back to that, but I'd do it in a follow-up PR

src/app/payments/get-protocol-fee.ts
vindard: Good point, it looks like we're pushing the validation into `decodeInvoice` in both send-payment and fee-probe use-case methods in which case the object is typed wrongly.

I can make a note to clean this up across those use-cases in a follow-up PR

src/app/payments/get-protocol-fee.ts
vindard: If you mean the limitCheck calc, it's two different functions (`IntraledgerLimits` vs `WithdrawalLimits`)

src/app/payments/helpers.ts
vindard: Not sure I understand. The goal is to get something that looks like this in honeycomb

image

src/app/payments/helpers.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1237/commits/33d67660d29cb0572f11f1394b381704d0868cac

src/app/payments/helpers.ts
```diff
@@ -0,0 +1,280 @@
...
+  if (price instanceof Error) return price
+
+  return (price * CENTS_PER_USD) as CentsPerSatsRatio
+}
+
+export const getMidPriceRatio = async (): Promise<
+  CentsPerSatsRatio | DealerPriceServiceError | PriceServiceError
+> => {
+  let midPriceRatio = usdHedgeEnabled
+    ? await dealer.getCentsPerSatsExchangeMidRate()
```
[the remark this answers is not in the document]
vindard: Changed error level to `Critical` to not mask this fallback: https://github.com/GaloyMoney/galoy/pull/1237/commits/f9c566b93e82fcb038b2791043c3fb7fed59d6ee

src/app/payments/get-protocol-fee.ts
```diff
@@ -0,0 +1,129 @@
...
+export const getLightningFeeEstimation = async ({
+  walletId,
+  paymentRequest,
+}: {
+  walletId: string
+  paymentRequest: EncodedPaymentRequest
+}): Promise<PaymentAmount<WalletCurrency> | ApplicationError> => {
+  const decodedInvoice = decodeInvoice(paymentRequest)
+  if (decodedInvoice instanceof Error) return decodedInvoice
+  if (decodedInvoice.paymentAmount === null) {
```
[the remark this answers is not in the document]
vindard: `0` amounts are co-erced to `null` here already https://github.com/GaloyMoney/galoy/blob/1e971ce675d0c8eec3e3075da698e198b9440e34/src/domain/bitcoin/lightning/ln-invoice.ts#L26

src/domain/payments/payment-flow-builder.ts
```diff
@@ -0,0 +1,507 @@
...
+    username?: Username
+  }): LPFBWithRecipientWallet<S, R> | LPFBWithError => {
+    if (recipientWalletId === state.senderWalletId) {
+      return LPFBWithError(new SelfPaymentError())
+    }
+    if (
+      recipientWalletCurrency === WalletCurrency.Usd &&
+      // This means (usdPaymentAmount === undefined XNOR state.uncheckedAmount === undefined)
+      // XNOR => if both or neither are set we get here - else we're fine
+      !(!usdPaymentAmount != !state.uncheckedAmount)
```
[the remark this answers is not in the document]
vindard: Tested here: https://replit.com/@vindard/Conditional-Equivalence-TS

Changed here: https://github.com/GaloyMoney/galoy/pull/1237/commits/51863b333496f2d0671b8611ae4c07b0e16f8336

vindard:
I added the TTL logic for the Payment Flow repository in these commits: https://github.com/GaloyMoney/galoy/compare/51863b333496f2d0671b8611ae4c07b0e16f8336...f9283b8416e5e5b7535f36a70bc8ee5c904a873e

Expired state is if the `createdAt` property is more than 5 mins in the past and the `paymentSentAndPending` property is set to false. This 2nd property is to override the expired times for when we'd need the payment flows again to update pending payments. The property is marked as false after update so that the cronjob can clean these up as well.

The TTL is enforced in 2 places:
- inside the `find` method where the TTL is checked and the flow is deleted if expired
- in the cronjob where expired flows are deleted

test/jest-e2e.setup.js
```diff
@@ -13,6 +13,8 @@ jest.mock("@services/lnd/unauth", () => {
   return module
 })
 
+jest.mock("@services/dealer-price", () => require("test/mocks/dealer-price"))
```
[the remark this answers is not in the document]
vindard: Yes, for testing the usd flows in _integration_ tests

src/app/payments/send-lightning.ts
vindard: I can remove it from here actually, but the question is repeated here: https://github.com/GaloyMoney/galoy/pull/1237#discussion_r855521054

src/domain/dealer-price/index.types.d.ts
```diff
@@ -32,3 +32,43 @@ interface IDealerPriceService {
   ): Promise<Satoshis | DealerPriceServiceError>
   getCentsPerSatsExchangeMidRate(): Promise<CentsPerSatsRatio | DealerPriceServiceError>
 }
+
+type UsdAmountWithExpiration = {
```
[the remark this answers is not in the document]
vindard: Removed: https://github.com/GaloyMoney/galoy/pull/1237/commits/3a5c1400d99271fcbb4a9d16e337089bd30e9605

src/domain/dealer-price/index.types.d.ts
```diff
@@ -32,3 +32,43 @@ interface IDealerPriceService {
...
+  amount: UsdPaymentAmount
+  expiration: Date
+}
+
+type BtcAmountWithExpiration = {
```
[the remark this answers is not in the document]
vindard: Removed: https://github.com/GaloyMoney/galoy/pull/1237/commits/3a5c1400d99271fcbb4a9d16e337089bd30e9605

src/services/payment-flow/schema.ts
```diff
@@ -0,0 +1,42 @@
...
+    descriptionFromInvoice: String,
+
+    btcPaymentAmount: Number,
+    usdPaymentAmount: Number,
+    inputAmount: Number,
+
+    btcProtocolFee: Number,
+    usdProtocolFee: Number,
+
+    recipientWalletId: String,
```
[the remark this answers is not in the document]
vindard: Mongo assumes all fields are undefined, and then we mark them as `required` sometimes when we want to enforce that they can never be `undefined`.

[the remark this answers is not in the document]
vindard: Undefined is implied, required fields need to be marked explicitly, changed here: https://github.com/GaloyMoney/galoy/pull/1237/commits/08aa8a8d22e7885f24cc7a0af1b523322fa1d38c

src/services/ledger/index.ts
```diff
@@ -199,6 +199,24 @@ export const LedgerService = (): ILedgerService => {
     }
   }
 
+  const getWalletBalanceAmount = async <S extends WalletCurrency>({
+    walletId,
+    walletCurrency,
+  }: {
+    walletId: WalletId
```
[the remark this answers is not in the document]
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1237/commits/511b884c10e209fa657008b92e18af3b4a4cf368

src/domain/ledger/fee-reimbursement.ts
vindard: I had it like this before, but a ratio drawn from the much smaller fee amounts was less sensitive than one derived from the full payment amounts and I was getting some errors at a point based on this. I added the explicit `priceRatio` to calculate the ratio from the full amounts instead outside of the method and then pass that in alongside the fee amounts.

src/domain/ledger/index.ts
```diff
@@ -64,3 +65,10 @@ export const checkedToLedgerTransactionId = (
   }
   return new InvalidLedgerTransactionId(ledgerTransactionId)
 }
+
+export const inputAmountFromLedgerTransaction = (
```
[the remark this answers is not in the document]
vindard: Makes sense... adding a note to revisit this once we've cleared up the things in #1257

src/domain/ledger/index.types.d.ts
vindard: Yes, didn't want to refactor all the other places outside payments that `LedgerTransaction` type is also used quite as yet

src/domain/ledger/index.types.d.ts
vindard: Added to the PR notes in the description here for follow-up

src/domain/ledger/index.types.d.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1237/commits/511b884c10e209fa657008b92e18af3b4a4cf368

src/domain/notifications/index.types.d.ts
vindard: This is one of the places where I started to do the conversion but then it snowballed through the other transactionNotification functions within this so I abandoned for a later PR

src/services/payment-flow/index.ts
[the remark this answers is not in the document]
vindard: See: https://github.com/GaloyMoney/galoy/pull/1237#issuecomment-1108608180

src/graphql/main/schema.graphql
```diff
@@ -91,6 +91,12 @@ input CaptchaRequestAuthCodeInput {
 """
 scalar CentAmount
 
+type CentAmountPayload {
+  amount: CentAmount
+  currencyUnit: ExchangeCurrencyUnit
```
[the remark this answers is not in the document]
vindard: This was originally to support the dual return (usd/btc) from the `LnInvoiceFeeProbe` mutation but I've since added back the WalletCurrency check to that one... which make me realize now that we're missing a `LnUsdInvoiceFeeProbe` mutation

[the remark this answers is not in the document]
vindard: Fixed: https://github.com/GaloyMoney/galoy/pull/1237/commits/f21087523b8e96cebc65b5c63ce448347db7abed & https://github.com/GaloyMoney/galoy/pull/1237/commits/44ca34e462e10d55af6e54152c2bdc91ef293a36

src/app/payments/get-protocol-fee.ts
```diff
@@ -0,0 +1,129 @@
...
+  const usdPaymentAmount = await builder.usdPaymentAmount()
+  if (usdPaymentAmount instanceof Error) return usdPaymentAmount
+  const btcPaymentAmount = await builder.btcPaymentAmount()
+  if (btcPaymentAmount instanceof Error) return btcPaymentAmount
+
+  const priceRatio = PriceRatio({ usd: usdPaymentAmount, btc: btcPaymentAmount })
+  if (priceRatio instanceof Error) return priceRatio
+
+  let paymentFlow
+  if (!(await builder.needsRoute())) {
```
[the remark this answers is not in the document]
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1237/commits/dd17c923811da2513906f90b8a79c83a1a0826fb

src/app/payments/get-protocol-fee.ts
```diff
@@ -0,0 +1,129 @@
...
+    const limitCheck = await newCheckIntraledgerLimits({
```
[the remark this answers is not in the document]
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1237/commits/dd17c923811da2513906f90b8a79c83a1a0826fb

src/app/payments/send-lightning.ts
bodymindarts: I think it needs to be an explicit step that no-ops internally for usd -> usd cases (as suggested)

dolcalmi: in this moment there is no need but we can have an additional step to get display currency amount (probably  is better to pass 'base' and 'quote' to avoid to have a different step for each base currency)

also, you can use `displayCurrency.code` config to pass the quote (in the display currency PR will be a config from the user entity.. like language)

vindard:
**Converted to draft until #1269 is merged**

src/domain/fiat/display-currency.ts
```diff
@@ -23,3 +23,14 @@ export const DisplayCurrencyConverter = (
...
 })
+
+export const NewDisplayCurrencyConverter = (
+  displayCurrencyPrice: DisplayCurrencyBasePerSat,
+): NewDisplayCurrencyConverter => {
+  return {
+    fromBtcAmount: (btc: BtcPaymentAmount): DisplayCurrencyBaseAmount =>
+      (Number(btc.amount) * displayCurrencyPrice) as DisplayCurrencyBaseAmount,
+    fromUsdAmount: (usd: UsdPaymentAmount): DisplayCurrencyBaseAmount =>
+      Number(usd.amount) as DisplayCurrencyBaseAmount,
```
UncleSamtoshi, line 34: Should we throw some kind of Error here if the DisplayCurrency isn't USD?

UncleSamtoshi (approved):

src/domain/fiat/display-currency.ts
vindard: I was thinking about this as well. I think the correct fix though is to have the price service return prices for 2 bases as @dolcalmi intends (`displayCurrencyPerSat` & `displayCurrencyPerCent`) and to then use that 2nd price to correctly implement this function.

I can probably apply that interface here in an incomplete form in the meantime 🤔

src/services/lnd/utils.ts
```diff
@@ -68,6 +72,18 @@ export const deleteFailedPaymentsAttemptAllLnds = async () => {
...
 
+export const deleteExpiredLightningPaymentFlows = async () => {
+  const paymentFlowRepo = PaymentFlowStateRepository(defaultTimeToExpiryInSeconds)
+
+  const deleted = await paymentFlowRepo.deleteExpiredLightningPaymentFlows()
+  if (deleted instanceof Error) {
+    if (!(deleted instanceof CouldNotFindError)) {
+      baseLogger.error({ error: deleted }, "error deleting expired payment flows")
+    }
+    return 0
```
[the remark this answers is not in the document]
vindard: Addressed here: https://github.com/GaloyMoney/galoy/pull/1237/commits/1fc6b561c34417e47c6a5f1a0e8211722047b595

src/app/payments/get-protocol-fee.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  const senderWallet = await WalletsRepository().findById(senderWalletId)
+  if (senderWallet instanceof Error) return senderWallet
+
+  const dealer = NewDealerPriceService()
+  const builder = await constructPaymentFlowBuilder({
+    senderWallet,
+    invoice,
+    uncheckedAmount,
+    usdFromBtc: dealer.getCentsFromSatsForImmediateBuy,
```
bodymindarts, line 83: Shouldn't this be the option price?

src/app/payments/helpers.ts
```diff
@@ -0,0 +1,296 @@
...
+    ? paymentBuilder.withNoAmountInvoice({ invoice, uncheckedAmount })
+    : paymentBuilder.withInvoice(invoice)
+
+  const builderWithSenderWallet = builderWithInvoice.withSenderWallet(senderWallet)
+
+  let builderAfterRecipientStep:
+    | LPFBWithRecipientWallet<WalletCurrency, WalletCurrency>
+    | LPFBWithError
+  if (
+    !(builderWithSenderWallet.isIntraLedger() instanceof Error) &&
```
bodymindarts, line 158: Confused as to why this function can return an Error...

src/app/payments/send-lightning.ts
```diff
@@ -0,0 +1,722 @@
...
+    decodedInvoice,
+  }
+}
+
+const validateNoAmountInvoicePaymentInputs = async ({
+  paymentRequest,
+  amount,
+  uncheckedSenderWalletId,
+  senderAccount,
+}) => {
```
bodymindarts, line 328: no type definition?

src/app/wallets/reimburse-fee.ts
```diff
@@ -1,89 +1,109 @@
...
-  } else {
-    amountDisplayCurrency = DisplayCurrencyConverter(price).fromSats(feeDifference)
+  const {
+    btcPaymentAmount: { amount: satsAmount },
+    usdPaymentAmount: { amount: centsAmount },
+    btcProtocolFee: { amount: satsFee },
+    usdProtocolFee: { amount: centsFee },
+  } = paymentFlow
+
+  const metadata: FeeReimbursementLedgerMetadata = {
```
bodymindarts, line 64: Shouldn't this be returned from a function in the ledger facade? cc @UncleSamtoshi

src/services/payment-flow/index.ts
```diff
@@ -0,0 +1,255 @@
...
+      const result = await PaymentFlowState.findOne({
+        senderWalletId: walletId,
+        paymentHash,
+        inputAmount: Number(inputAmount),
+      })
+      if (!result) return new CouldNotFindLightningPaymentFlowError()
+
+      const paymentFlow: PaymentFlow<S, WalletCurrency> = paymentFlowFromRaw(result)
+      if (isExpired({ paymentFlow, expiryTimeInSeconds })) {
+        await deleteLightningPaymentFlow({ walletId, paymentHash, inputAmount })
```
bodymindarts, line 47: No need for await here or? Could be fire + forget.

src/services/payment-flow/index.ts
```diff
@@ -0,0 +1,255 @@
...
+    try {
+      const result = await PaymentFlowState.findOneAndUpdate(
+        {
+          senderWalletId: paymentFlow.senderWalletId,
+          paymentHash: paymentFlow.paymentHash,
+          inputAmount: Number(paymentFlow.inputAmount),
+        },
+        rawFromPaymentFlow(paymentFlow),
+        {
+          new: true,
```
bodymindarts, line 72: This makes it an upsert or?

src/services/payment-flow/index.ts
```diff
@@ -0,0 +1,255 @@
...
+        return new CouldNotFindLightningPaymentFlowError()
+      }
+
+      return paymentFlow
+    } catch (err) {
+      return new UnknownRepositoryError(err)
+    }
+  }
+
+  const updateLightningPaymentFlow = async <
```
bodymindarts, line 57: Why are these mutable?

src/services/payment-flow/index.ts
```diff
@@ -0,0 +1,255 @@
...
+      if (!result) {
+        return new CouldNotUpdateLightningPaymentFlowError()
+      }
+      return true
+    } catch (err) {
+      return new UnknownRepositoryError(err)
+    }
+  }
+
+  const updatePendingLightningPaymentFlow = async <
```
bodymindarts, line 84: Why 2 different update functions?

src/app/payments/send-lightning.ts
```diff
@@ -0,0 +1,722 @@
...
+
+        if (!(payResult instanceof Error))
+          LedgerFacade.updateMetadataByHash({
+            hash: paymentHash,
+            revealedPreImage: payResult.revealedPreImage,
+          })
+      }
+      if (payResult instanceof LnPaymentPendingError) {
+        paymentFlow.paymentSentAndPending = true
+        paymentFlowRepo.updateLightningPaymentFlow(paymentFlow)
```
vindard, line 684: Can be changed to `updatePendingLightningPaymentFlow` and this old method removed to clean up `update...` interface

src/services/payment-flow/index.ts
vindard: Making a note to clean up in later PR: https://github.com/GaloyMoney/galoy/pull/1237/files#r865962574

src/app/payments/get-protocol-fee.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1237/commits/ac2b75038b876ec0e86ea892faf5aee680997d1b

src/services/payment-flow/index.ts
vindard: No need I think: https://github.com/GaloyMoney/galoy/pull/1237/commits/3bfaa39de68a7c9dbdea197d8044b53fab077817

src/domain/accounts/account-validator.ts
```diff
@@ -0,0 +1,23 @@
+import { InvalidAccountStatusError, InvalidWalletId } from "@domain/errors"
+
+import { AccountStatus } from "."
+
+export const AccountValidator = (): AccountValidator => {
+  const validateAccount = ({
```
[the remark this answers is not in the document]
vindard: Added to notes for future PR

src/domain/shared/index.types.d.ts
```diff
@@ -5,14 +5,25 @@ type ValidationError = import("./errors").ValidationError
...
 
+type ExchangeCurrencyUnit =
+  typeof import("./primitives").ExchangeCurrencyUnit[keyof typeof import("./primitives").ExchangeCurrencyUnit]
+
 type PaymentAmount<T extends WalletCurrency> = {
   currency: T
   amount: bigint
 }
+
+type PaymentAmountPayload<T extends ExchangeCurrencyUnit> = {
```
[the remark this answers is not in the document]
vindard: Moved it across to the `grapqhl` module where it's exposed: https://github.com/GaloyMoney/galoy/pull/1237/commits/081f411c0e5bc1e941d43dd094ccaacf7267f8bf

src/app/payments/helpers.ts
vindard: You're right, this needed to be cleaned up. There is a code path where errors come through, but it's irrelevant because the rest of the logic is resolved via an `LPFBWithError` object.

Cleaned up: https://github.com/GaloyMoney/galoy/pull/1237/commits/ed23db6d17a90c1fc562c19815597c54ffa7e84e

UncleSamtoshi (approved):

PR #1240, opened 2022-04-13
feat: add new/pending account statuses (with history) and a config for initial status

samerbuna:


src/graphql/admin/root/mutation/account-update-status.ts
```diff
@@ -14,23 +14,31 @@ const AccountUpdateStatusInput = GT.Input({
     status: {
       type: GT.NonNull(AccountStatus),
```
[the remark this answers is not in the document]
samerbuna: Do you mean removing the status field and relying on the history to get the current status? That'll require a db migration. I thought we get this change out first and do the migration in another PR.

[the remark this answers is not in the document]
nicolasburtey: yes that is the idea. not duplicating the status at the db layer.

src/app/users/login.ts
```diff
@@ -120,7 +120,16 @@ export const loginWithKratos = async ({
...
 
   const network = BTC_NETWORK
-  return createToken({ uid: user.id, network, kratosUserId: kratosUserIdValid })
+
+  const account = await AccountsRepository().findByUserId(user.id)
+  if (account instanceof Error) return account
+
+  return createToken({
+    uid: user.id,
+    accountStatus: account.status.toUpperCase(),
```
nicolasburtey, line 129: is `accountStatus` really necessary?

This is a mutable value in the backend, so including this value in the token will lead to stale token.

What is the logic for including this in the JWT, versus doing a call to the backend to know the up-to-date status?

src/graphql/admin/schema.graphql
```diff
@@ -25,6 +25,8 @@ enum AccountLevel {
 enum AccountStatus {
   ACTIVE
   LOCKED
+  NEW
+  PENDING
```
nicolasburtey, line 29: is pending needed? is there a difference between `new` and `pending`?

in both case, if the account needs manual approval to get active, the action will be limited on the account between those 2 status, so maybe we don't need to separate them?

nicolasburtey (commented):
some questions/comments

src/app/users/login.ts
samerbuna: My thinking here: a token represents the authority of an account. The status of that account is part of that.

For clients apps enforcing the account status, they'll need to check that status every time they acquire a token.

On tokens being stale, the problem there is that we allow them to live beyond their need.

src/graphql/admin/schema.graphql
samerbuna: New: Account is fresh (no status verifications happened yet). The UI sends the user to the verification flow
Pending: A verification request has been made but the result is not back yet. The UI can either block the user or limit what they can do.

src/app/users/login.ts
bodymindarts: I also have doubts about this - lets discuss the trade offs at standup.

src/graphql/admin/root/mutation/account-update-status.ts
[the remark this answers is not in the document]
samerbuna: Done

src/app/users/login.ts
samerbuna: I removed this and relied on the auth endpoint instead.

src/domain/accounts/index.types.d.ts
```diff
@@ -39,6 +39,13 @@ type AccountContact = {
   transactionsCount: number
 }
 
+type AccountStatusHistory = Array<{
+  status: AccountStatus
+  updatedByUserId: UserId
+  updatedAt?: Date
```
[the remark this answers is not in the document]
samerbuna: Always the default value. No need to pass it when inserting a new status record

bodymindarts (approved):

PR #1245, opened 2022-04-21
feat: allow setting any valid language code as the user preferred

samerbuna:


src/domain/users/index.types.d.ts
```diff
@@ -81,3 +80,12 @@ interface IUsersRepository {
   persistNew({ phone, phoneMetadata }: NewUserInfo): Promise<User | RepositoryError>
   update(user: User): Promise<User | RepositoryError>
 }
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+interface ReadonlyArray<T> {
```
dolcalmi, line 85: if this is a generic implementation probably should be in primitives. this is not a blocker to merge this PR

dolcalmi (approved):
just a comment but not a blocker

PR #1269, opened 2022-04-27
chore: migrate new amount-related properties to old entries

vindard:
**_Note:_** _Converted to draft until #1281 is merged_

## Description: Migration-only (code-change in #1281)

This PR migrates new amount-related properties to past Lightning send payments records.

### To test migration locally
- `git checkout` to `main` branch and run `TEST="01|02-a|02-r|02-s.*l" make reset-integration` to initialize pre-migrated database

- `git checkout` to this branch and edit `Ln 3` in `src/migrations/migrate-mongo-config.js`:
   ```ts
   const address = process.env.MONGODB_ADDRESS ?? "localhost:27017"
   ```

- check mongo pre-migration to make sure `satsAmount` property does not exist in `medici_transactions` collection

- run `node_modules/.bin/migrate-mongo up -f src/migrations/migrate-mongo-config.js`

- check mongo post-migration with `{$or: [{type: "payment"}, {type: "fee_reimbursement"}]}` query to make sure all records returned have a value set for `satsAmount`

## Open `TODO`s
- [x] ~Figure out how to import display currency from default.yaml into migration script (double-check how this actually gets executed in environments and what modules are available to import from) https://github.com/GaloyMoney/galoy/pull/1269#discussion_r861157959~ **Done: https://github.com/GaloyMoney/galoy/pull/1269/commits/7ca4849bfc2943903f0b33292a6707915cac9901**

- [x] ~`BigInt` <-> `Number` question https://github.com/GaloyMoney/galoy/pull/1269#discussion_r861147720~

### DIFF
--- a/src/migrations/20220427130214-add-sats-amount-prop-to-payments.ts
+++ b/src/migrations/20220427130214-add-sats-amount-prop-to-payments.ts
@@ -9,12 +9,12 @@ module.exports = {
       )
       console.log(
         { result: resultDebits },
-        "added 'satsAmount' prop to debit medici_transactions",
+        "added new props to debit medici_transactions",
       )
     } catch (error) {
       console.log(
         { result: error },
-        "Couldn't add 'satsAmount' prop to debit medici_transactions",
+        "Couldn't add new props to debit medici_transactions",
       )
     }
 
@@ -29,40 +29,71 @@ module.exports = {
           $group: {
             _id: "$hash",
             satsAmount: { $first: "$satsAmount" },
+            fee: { $first: "$fee" },
+            usd: { $first: "$usd" },
+            feeUsd: { $first: "$feeUsd" },
           },
         },
       ])
-      for await (const { _id: hash, satsAmount } of allTxns) {
+      for await (const { _id: hash, satsAmount, fee, usd, feeUsd } of allTxns) {
         if (!hash) continue
-        const resultRest = await collection.update({ hash }, [{ $set: { satsAmount } }], {
-          multi: true,
-        })
+        const centsAmount = Math.round((usd - feeUsd) * 100)
+        const centsFee = Math.round(feeUsd * 100)
+
+        const resultRest = await collection.update(
+          { hash },
+          [
+            {
+              $set: {
+                satsAmount,
+                satsFee: fee,
+                centsAmount,
+                centsFee,
+                displayAmount: centsAmount,
+                displayFee: centsFee,
+                displayCurrency: "USD",
+              },
+            },
+          ],
+          {
+            multi: true,
+          },
+        )
         const {
           result: { n, nModified },
         } = resultRest
         console.log(
-          `added 'satsAmount' prop to ${nModified} of ${n}  transactions for hash: ${hash}`,
+          `added new props to ${nModified} of ${n}  transactions for hash: ${hash}`,
         )
       }
     } catch (error) {
       console.log(
         { result: error },
-        "Couldn't add 'satsAmount' prop to rest of medici_transactions",
+        "Couldn't add new props to rest of medici_transactions",
       )
     }
   },
 
   async down(db) {
     try {
-      const result = await db
-        .collection("medici_transactions")
-        .update({}, { $unset: { satsAmount: "" } }, { multi: true })
-      console.log({ result }, "removed 'satsAmount' prop from medici_transactions")
-    } catch (error) {
-      console.log(
-        { result: error },
-        "Couldn't remove 'satsAmount' prop from medici_transactions",
+      const result = await db.collection("medici_transactions").update(
+        {},
+        {
+          $unset: {
+            satsAmount: "",
+            satsFee: "",
+            centsAmount: "",
+            centsFee: "",
+            displayAmount: "",
+            displayFee: "",
+            displayCurrency: "",
+          },
+        },
+        { multi: true },
       )
+      console.log({ result }, "removed new props from medici_transactions")
+    } catch (error) {
+      console.log({ result: error }, "Couldn't remove new props from medici_transactions")
     }
   },
 }
--- a/src/migrations/20220427130214-add-sats-amount-prop-to-payments.ts
+++ b/src/migrations/20220427130214-add-sats-amount-prop-to-payments.ts
@@ -1,0 +1,68 @@
+module.exports = {
+  async up(db) {
+    const collection = db.collection("medici_transactions")
+    try {
+      const resultDebits = await collection.update(
+        { type: "payment", debit: { $gt: 0 } },
+        [{ $set: { satsAmount: { $subtract: ["$debit", "$fee"] } } }],
+        { multi: true },
+      )
+      console.log(
+        { result: resultDebits },
+        "added 'satsAmount' prop to debit medici_transactions",
+      )
+    } catch (error) {
+      console.log(
+        { result: error },
+        "Couldn't add 'satsAmount' prop to debit medici_transactions",
+      )
+    }
+
+    try {
+      const allTxns = await collection.aggregate([
+        {
+          $match: {
+            satsAmount: { $exists: true },
+          },
+        },
+        {
+          $group: {
+            _id: "$hash",
+            satsAmount: { $first: "$satsAmount" },
+          },
+        },
+      ])
+      for await (const { _id: hash, satsAmount } of allTxns) {
+        if (!hash) continue
+        const resultRest = await collection.update({ hash }, [{ $set: { satsAmount } }], {
+          multi: true,
+        })
+        const {
+          result: { n, nModified },
+        } = resultRest
+        console.log(
+          `added 'satsAmount' prop to ${nModified} of ${n}  transactions for hash: ${hash}`,
+        )
+      }
+    } catch (error) {
+      console.log(
+        { result: error },
+        "Couldn't add 'satsAmount' prop to rest of medici_transactions",
+      )
+    }
+  },
+
+  async down(db) {
+    try {
+      const result = await db
+        .collection("medici_transactions")
+        .update({}, { $unset: { satsAmount: "" } }, { multi: true })
+      console.log({ result }, "removed 'satsAmount' prop from medici_transactions")
+    } catch (error) {
+      console.log(
+        { result: error },
+        "Couldn't remove 'satsAmount' prop from medici_transactions",
+      )
+    }
+  },
+}
--- a/src/migrations/20220427130214-add-sats-amount-prop-to-payments.ts
+++ b/src/migrations/20220427130214-add-sats-amount-prop-to-payments.ts
@@ -1,3 +1,6 @@
+/* eslint @typescript-eslint/no-var-requires: "off" */
+const { getDisplayCurrencyConfig } = require("../../lib/config/yaml")
+
 module.exports = {
   async up(db) {
     const collection = db.collection("medici_transactions")
@@ -51,7 +54,7 @@ module.exports = {
                 centsFee,
                 displayAmount: centsAmount,
                 displayFee: centsFee,
-                displayCurrency: "USD",
+                displayCurrency: getDisplayCurrencyConfig()?.code || "USD",
               },
             },
           ],


vindard:
### Notes

- This work was started here but completed in PR #1237 in these commits: https://github.com/GaloyMoney/galoy/compare/30da8da8ff34697f5ae9d310879991713ec0257e...4b80d2e47ec06d379258fb7ae438b71f7089ae43

- The scope of this work is to simply write the new properties to the database. The removal of old properties and refactoring of logic to base from these new properties will be done in follow-up PRs

src/domain/fiat/index.ts
```diff
@@ -4,8 +4,8 @@ import {
   NonIntegerError,
 } from "@domain/errors"
 
-export const toCents = (amount: number): UsdCents => {
-  return amount as UsdCents
+export const toCents = (amount: number | bigint): UsdCents => {
+  return Number(amount) as UsdCents
```
dolcalmi, line 8: if we pass a big int then this conversion is not safe

```
> Number(BigInt("19007199254740992"))
19007199254740990
> BigInt("19007199254740992")
19007199254740992n
```

src/migrations/20220427130214-add-sats-amount-prop-to-payments.ts
```diff
@@ -0,0 +1,99 @@
...
+          [
+            {
+              $set: {
+                satsAmount,
+                satsFee: fee,
+                centsAmount,
+                centsFee,
+                displayAmount: centsAmount,
+                displayFee: centsFee,
+                displayCurrency: "USD",
```
dolcalmi, line 54: we need to get this value from config, this will not work for bitcoin jungle

src/migrations/20220427130214-add-sats-amount-prop-to-payments.ts
```diff
@@ -0,0 +1,99 @@
+module.exports = {
+  async up(db) {
+    const collection = db.collection("medici_transactions")
+    try {
+      const resultDebits = await collection.update(
+        { type: "payment", debit: { $gt: 0 } },
```
dolcalmi, line 6: if we only include payments the db will remain inconsistent

src/services/ledger/receive.ts
```diff
@@ -109,18 +111,38 @@ export const receive = {
...
+    feeDisplayCurrency,
+    amountDisplayCurrency,
+    displayCurrency,
   }: AddLnFeeReeimbursementReceiveArgs): Promise<LedgerJournal | LedgerError> => {
+    const {
+      btcPaymentAmount: { amount: satsAmount },
+      btcProtocolFee: { amount: satsFee },
+    } = paymentFlow
+    const centsAmount = Math.round((amountDisplayCurrency - feeDisplayCurrency) * 100)
+    const centsFee = Math.round(feeDisplayCurrency * 100)
```
dolcalmi, line 128: are we sure we want cents conversion responsibility in the ledger service? shouldn't be part of the price service?

src/services/ledger/schema.ts
```diff
@@ -69,6 +69,24 @@ const transactionSchema = new Schema({
...
+    type: Number,
+  },
+
+  displayAmount: Number,
+  displayFee: {
+    type: Number,
+  },
+  displayCurrency: {
+    type: String,
+    enum: ["USD", "BTC"],
```
dolcalmi, line 87: please dont use enum, there are more than 150 display currencies

src/services/ledger/tx-metadata.ts
```diff
@@ -1,29 +1,52 @@
...
-  pubkey: Pubkey
+  displayCurrency: DisplayCurrency
   feeKnownInAdvance: boolean
 }) => {
+  const {
+    btcPaymentAmount: { amount: satsAmount },
+    btcProtocolFee: { amount: satsFee },
+  } = paymentFlow
+  const centsAmount = Math.round((amountDisplayCurrency - feeDisplayCurrency) * 100)
+  const centsFee = Math.round(feeDisplayCurrency * 100)
```
dolcalmi, line 29: same here

src/services/ledger/tx-metadata.ts
```diff
@@ -106,20 +129,43 @@ export const LnReceiveLedgerMetadata = ({
...
+  feeDisplayCurrency: DisplayCurrencyBaseAmount
   amountDisplayCurrency: DisplayCurrencyBaseAmount
+  displayCurrency: DisplayCurrency
 }) => {
+  const {
+    btcPaymentAmount: { amount: satsAmount },
+    btcProtocolFee: { amount: satsFee },
+  } = paymentFlow
+  const centsAmount = Math.round((amountDisplayCurrency - feeDisplayCurrency) * 100)
+  const centsFee = Math.round(feeDisplayCurrency * 100)
```
dolcalmi, line 151: same here

dolcalmi (dismissed):

src/services/ledger/receive.ts
vindard: This is very temporary here... eventually gets handled by our payment flow builder ([here](https://github.com/GaloyMoney/galoy/pull/1237/files#diff-9152ae6f6a4bd5040a65b4bfa9eb5cd1ffadfc9c069ff3b39cd7dc78e02fde1dR128))

src/services/ledger/tx-metadata.ts
vindard: https://github.com/GaloyMoney/galoy/pull/1269#discussion_r861243181

src/services/ledger/tx-metadata.ts
vindard: https://github.com/GaloyMoney/galoy/pull/1269#discussion_r861243181

src/migrations/20220427130214-add-sats-amount-prop-to-payments.ts
vindard: The migration/new-code covers all transactions that are related to the `hash` (`payment` and `fee_reimbursements`).

I'm planning to migrate the rest of transaction types and redo the domain logic in separate PRs. The scope here is just not to lose info when we start doing USD payments. It's the bare minimum to get #1237 merged in.

src/migrations/20220427130214-add-sats-amount-prop-to-payments.ts
vindard: Good catch. I can change this in other places in the code, but not sure how to import into the migration script here.

Also, it looks like bitcoin jungle is still storing usd on the backend ([here](https://github.com/Bitcoin-Jungle/galoy/blob/f3a927b41101fe07a06df552ea606163328b361c/src/services/ledger/index.ts#L610)). If this is correct, then the migration might be ok like this? They'll likely need a separate migration in the future to migrate their past transactions from usd to crc if that's the case.

src/services/ledger/schema.ts
vindard: Changed to string: https://github.com/GaloyMoney/galoy/pull/1269/commits/4f2636ccad3834e44bbb80239f8466b9c0c5d26e

src/domain/fiat/index.ts
vindard: Oh this is probably a wider problem because we do this a lot in the new code (e.g. [here](https://github.com/GaloyMoney/galoy/pull/1237/files#diff-cbba23c1f04bbbb08b4e3b06a49efcc7be05fa0ebd30e2aa78739812a2e9b6b8R18))

dolcalmi:
```
Of course, you should bear in mind that your BigInt value must be within [Number.MIN_SAFE_INTEGER, Number.MAX_SAFE_INTEGER]
```
if we are constraining the value then I am not sure if the use of BigInt is useful

src/migrations/20220427130214-add-sats-amount-prop-to-payments.ts
dolcalmi: they dont use USD they use CRC but they have to use usd field (they only changed price server to return CRC) but if we want the migration to work with their data then the default value for them should be CRC

src/migrations/20220427130214-add-sats-amount-prop-to-payments.ts
```diff
@@ -0,0 +1,99 @@
...
+        if (!hash) continue
+        const centsAmount = Math.round((usd - feeUsd) * 100)
+        const centsFee = Math.round(feeUsd * 100)
+
+        const resultRest = await collection.update(
+          { hash },
+          [
+            {
+              $set: {
+                satsAmount,
```
UncleSamtoshi, line 48: Were we going to add ```accountingCurrency``` as a rename of ```currency```.

UncleSamtoshi (dismissed):

src/domain/ledger/index.types.d.ts
```diff
@@ -166,9 +166,12 @@ type AddLnFeeReeimbursementReceiveArgs = {
   paymentHash: PaymentHash
   sats: Satoshis
   cents?: UsdCents
-  amountDisplayCurrency: DisplayCurrencyBaseAmount
   journalId: LedgerJournalId
   revealedPreImage?: RevealedPreImage
+  paymentFlow: { btcPaymentAmount: BtcPaymentAmount; btcProtocolFee: BtcPaymentAmount }
```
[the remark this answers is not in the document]
vindard: It becomes a proper `PaymentFlowState`  object [here](https://github.com/GaloyMoney/galoy/pull/1237/files#diff-dcf253ca7d4013a0520a033cd2cca3ff0e8fd22728b453a52d54de869b788fc9R183)

src/migrations/20220427130214-add-sats-amount-prop-to-payments.ts
```diff
@@ -0,0 +1,99 @@
+module.exports = {
+  async up(db) {
+    const collection = db.collection("medici_transactions")
+    try {
+      const resultDebits = await collection.update(
+        { type: "payment", debit: { $gt: 0 } },
+        [{ $set: { satsAmount: { $subtract: ["$debit", "$fee"] } } }],
```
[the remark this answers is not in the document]
vindard: I set this up to mirror what we have in the `PaymentFlowState` object. `btcPaymentAmount` in there is the amount without fees, same with `usdPaymentAmount` mapping directly to `centsAmount`.

src/migrations/20220427130214-add-sats-amount-prop-to-payments.ts
```diff
@@ -0,0 +1,99 @@
...
+            satsAmount: { $first: "$satsAmount" },
+            fee: { $first: "$fee" },
+            usd: { $first: "$usd" },
+            feeUsd: { $first: "$feeUsd" },
+          },
+        },
+      ])
+      for await (const { _id: hash, satsAmount, fee, usd, feeUsd } of allTxns) {
+        if (!hash) continue
+        const centsAmount = Math.round((usd - feeUsd) * 100)
```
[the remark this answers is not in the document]
vindard: For the migration these 2 properties (`usd`, `feeUsd`) are currently persisted as dollars not cents in the db

src/migrations/20220427130214-add-sats-amount-prop-to-payments.ts
vindard: Thinking to do that as part of a follow-up migration after the #1237 PR is merged, see scope bullet point here: https://github.com/GaloyMoney/galoy/pull/1269#issuecomment-1111704927

src/domain/fiat/index.ts
vindard: I think part of the reason too is to avoid having floats in our data

src/migrations/20220427130214-add-sats-amount-prop-to-payments.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1269/commits/7ca4849bfc2943903f0b33292a6707915cac9901

src/domain/fiat/index.ts
vindard: Ran over this with @bodymindarts as well... his thinking is that this is ok as-is for now too

bodymindarts (dismissed):

bodymindarts (approved):

PR #1276, opened 2022-05-02
fix: update decode invoice to use safe tokens

dolcalmi:
Fix issue with msats invoices (probe and payment) updating decode invoice to use safe tokens

https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/sooTn7xzKfk/trace/tTvXGtnMP39?span=8418714adec94963

### DIFF
--- a/src/domain/bitcoin/lightning/ln-invoice.ts
+++ b/src/domain/bitcoin/lightning/ln-invoice.ts
@@ -23,8 +23,8 @@ export const decodeInvoice = (
     return new LnInvoiceMissingPaymentSecretError()
   }
   const paymentSecret: PaymentIdentifyingSecret = decodedInvoice.payment
-  const amount: Satoshis | null = decodedInvoice.tokens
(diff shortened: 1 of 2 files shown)


src/domain/bitcoin/lightning/ln-invoice.ts
```diff
@@ -56,7 +56,7 @@ export const decodeInvoice = (
     description: decodedInvoice.description || "",
     paymentHash: decodedInvoice.id as PaymentHash,
     destination: decodedInvoice.destination as Pubkey,
-    milliSatsAmount: toMilliSatsFromNumber(decodedInvoice.mtokens || 0),
+    milliSatsAmount: toMilliSatsFromNumber(amount ? amount * 1000 : 0),
```
dolcalmi, line 59: this is required to avoid `ExpectedEqualValuesForTokensAndMtokens` exception from lightning library

vindard (approved):
lgtm

PR #1298, opened 2022-05-09
refactor: reimplement payment flow reconstruction logic

vindard:
## Description

This PR moves the reconstruction logic into a separate function, adds unit tests, and removes the fallback to the lnd service to fetch sats amounts for usd invoices.

It also fixes an issue where:
- the `usd` property from ledger txns was not being multiplied by 100 when reconstructing the payment flow
- the unmultiplied float was causing a `RangeError` when `BigInt` was called on it (safeBigInt to be implemented in follow-up PR)


src/domain/payments/payment-flow.ts
```diff
@@ -82,3 +93,64 @@ export const PaymentFlow = <S extends WalletCurrency, R extends WalletCurrency>(
     recipientWalletDescriptor,
   }
 }
+
+export const PaymentFlowFromLedgerTransaction = <
+  S extends WalletCurrency,
+  R extends WalletCurrency,
+>(
+  ledgerTxn: LedgerTransaction<S>,
```
vindard, line 101: This is in the `payments` module. Should it know what a `LedgerTransaction` is, or should this method be in the app layer?

src/app/payments/translations.ts
```diff
@@ -0,0 +1,73 @@
...
+import { PaymentInitiationMethod, SettlementMethod } from "@domain/wallets"
+
+export const PaymentFlowFromLedgerTransaction = <
+  S extends WalletCurrency,
+  R extends WalletCurrency,
+>(
+  ledgerTxn: LedgerTransaction<S>,
+): PaymentFlow<S, R> | ValidationError => {
+  if (ledgerTxn.type !== LedgerTransactionType.Payment) {
+    return new NonLnPaymentTransactionForPaymentFlowError()
```
UncleSamtoshi, line 21: How are on-chain payments updated?

UncleSamtoshi (approved):

src/app/payments/translations.ts
vindard: In trigger on each new block I believe: https://github.com/GaloyMoney/galoy/blob/fc621181ade51e539b87cb41132a70375620dd5d/src/servers/trigger.ts#L119

PR #1302, opened 2022-05-11
refactor: remove legacy core/error

nicolasburtey:


src/servers/graphql-server.ts
```diff
@@ -164,19 +163,6 @@ export const startApolloServer = async ({
     },
     formatError: (err) => {
       try {
-        const exception = err.extensions?.exception as CustomError
-        const errorCode = err.extensions?.code as string
```
nicolasburtey, line 168: this path didn't exist from the type, so I'm assuming it was already `undefined`

bodymindarts (approved):

PR #1303, opened 2022-05-12
feat: segregate span for exporter

nicolasburtey:


UncleSamtoshi:
Are you segregating just for more visibility?

src/servers/exporter.ts
```diff
@@ -189,7 +203,7 @@ const main = async () => {
 
   const port = process.env.PORT || 3000
   logger.info(`Server listening to ${port}, metrics exposed on /metrics endpoint`)
-  await server.listen(port)
```
UncleSamtoshi, line 192: Why did you get rid of ```await``` here?

src/servers/exporter.ts
```diff
@@ -94,85 +94,99 @@ const coldWallets: { [key: string]: client.Gauge<string> } = {}
...
-      lndOffChain_g.set(offChain)
-      lndOpeningChannelBalance_g.set(opening_channel_balance)
-      lndClosingChannelBalance_g.set(closing_channel_balance)
-
-      const userCount = await User.countDocuments()
-      userCount_g.set(userCount)
-
-      for (const wallet of wallets) {
-        let walletId
+      await asyncRunInSpan("getBosScore", {}, async () => {
```
UncleSamtoshi, line 97: Instead of awaiting each individual span, it could be more performant to kickoff each call and then do ```await Promise.all[... each span]```

UncleSamtoshi (approved):

src/servers/exporter.ts
nicolasburtey: This is not a promised function, so the await is a no-op

src/servers/exporter.ts
nicolasburtey: So we are already hammering the db  and request are very slow so I don't think this is a good idea (at least for now)

PR #1311, opened 2022-05-15
feat: made memo sharing sats threshold queryable 

jamespierog:
This PR adds the memo sharing sats threshold to the graphql API. This allows the galoy-pay frontend to determine whether a memo is allowed to be shared in a transaction, and then communicate that information to the user. That way if a user is writing a memo and has less than the minimum sats threshold before sending a payment, they will be notified that the memo will not go through.

The PR also has a fair bit of other miscellaneous code changes due to prettier (in any file that has a massive number of lines changed it's simply because I ran prettier and other checks to refactor my own code).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -48,28 +48,28 @@ services:
     volumes:
     - ${PWD}/dev/bitcoind/bitcoin.conf:/data/.bitcoin/bitcoin.conf
   lnd1:
-    image: lightninglabs/lnd:v0.14.2-beta
+    image: lightninglabs/lnd:v0.14.3-beta
     volumes:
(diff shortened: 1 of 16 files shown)


bodymindarts:
This PR is very noisy as it contains a bunch of non-functional re-formatting. Also the commit history isn't clean.

In regards to the functional changes I would suggest to add this setting to the [`Globals`](https://github.com/GaloyMoney/galoy/blob/main/src/graphql/main/schema.graphql#L136-L144) type instead of making it its own query. WDYT @nicolasburtey?

jamespierog:
Ah yeah I didn't think of using the Globals type, but it makes a lot more sense! I'll clean up the git history and the PR.

PR #1315, opened 2022-05-17
Add nextcloud static channel backups

zoop-btc:
This pull request adds a nextcloud destination for the static channel backups.
Another pull request for the charts will follow and link back. However this can be merged without changes to the helm charts.
Additionally it replaces the dropbox dependency with axios.
Since Nextcloud uses a Webdav interface where the filepath is inside the url the NEXTCLOUD_URL environment variable takes the whole url for example "https://myserver.com/remote.php/dav/files/myuser/backupfolder".
Please see the [official documentation here](https://docs.nextcloud.com/server/latest/developer_manual/client_apis/WebDAV/basic.html#uploading-files).

Inside of the process.ts I have added a check which results in an error if none of the environment variables for any destination are set.
In the backup.ts itself there's also checks which skip whichever method is not defined. I am unsure about my use of logger.debug there for output. I can remove it if you think it's too much logger output.

### DIFF
--- a/src/app/admin/backup.ts
+++ b/src/app/admin/backup.ts
@@ -3,24 +3,36 @@ import {
   DropboxAccessToken,
   GcsApplicationCredentials,
   LND_SCB_BACKUP_BUCKET_NAME,
+  Nextcloudurl,
+  Nextclouduser,
+  Nextcloudpassword
 } from "@config"
 import { Storage } from "@google-cloud/storage"
-import { Dropbox } from "dropbox"
+import axios, { Axios } from "axios";
 
 export const uploadBackup =
   (logger: Logger) =>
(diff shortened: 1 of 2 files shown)


nicolasburtey:
did you close the PR with the expectation to reopen it once the chart is merged?

zoop-btc:
> did you close the PR with the expectation to reopen it once the chart is merged?

I had to open a new one because I rebased and force pushed the nextcloud branch with an appropriate solution.
This one was closed because it was crashing the pods which weren't trigger. Only trigger has the environment variables for the backups assigned in the chart, so an aggressive check was a mistake :)

The new pull request [can be found here](https://github.com/GaloyMoney/galoy/pull/1316).

PR #1316, opened 2022-05-18
feat: add nextcloud static channel backups

zoop-btc:
This pull request adds a nextcloud destination for the static channel backups.
The corresponding pull request for the charts can be found [here](https://github.com/GaloyMoney/charts/pull/978). However this can be merged without changes to the helm charts. If the nextcloud environment variables are not set it will simply skip it.

Additionally it replaces the dropbox dependency with axios.

Since Nextcloud uses a Webdav interface where the filepath is contained inside the url. The NEXTCLOUD_URL environment variable thus takes the whole url for example `https://myserver.com/remote.php/dav/files/myuser/backupfolder`.
Please see the [official documentation here](https://docs.nextcloud.com/server/latest/developer_manual/client_apis/WebDAV/basic.html#uploading-files).

I have added a check in the backup routine which results in an error log entry if none of the environment variables for any destination are set. But it will not crash.
If any of the environment variables for either google cloud, dropbox or nextcloud are not set it will skip them and notify in the logs.

### DIFF
--- a/src/app/admin/backup.ts
+++ b/src/app/admin/backup.ts
@@ -3,9 +3,20 @@ import {
   DropboxAccessToken,
   GcsApplicationCredentials,
   LND_SCB_BACKUP_BUCKET_NAME,
+  Nextcloudurl,
+  Nextclouduser,
+  Nextcloudpassword,
 } from "@config"
 import { Storage } from "@google-cloud/storage"
-import { Dropbox } from "dropbox"
+import axios from "axios"
+import {
+  asyncRunInSpan,
+  SemanticAttributes,
+  addAttributesToCurrentSpan,
+  addEventToCurrentSpan,
+  recordExceptionInCurrentSpan,
+} from "@services/tracing"
+import { ErrorLevel } from "@domain/shared"
 
 export const uploadBackup =
   (logger: Logger) =>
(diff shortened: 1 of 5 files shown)


nicolasburtey:
LGTM @zoop-btc 

you also should remove dropbox from our dependency now (`yarn remove ...`)

bodymindarts:
Did an in person review yesterday and agreed @zoop-btc would add tracing to the backup code paths so that we get notified if one fails as part of this PR. This will also help identify missing configuration.

openoms (approved):
This is a great new feature and looks good (did not test). Agree that it notifies on failure in the logs but doesn't break.

Taking note of the values and the secret format in https://github.com/GaloyMoney/charts/pull/978 for the detailed deployment documentation.

zoop-btc:
I am waiting for this [tracing issue ](https://github.com/GaloyMoney/galoy/issues/1344)to be resolved before this can be upgraded to use tracing. I've found it while calling asyncRunInSpan for the backups.

zoop-btc:
Changed to draft until tracing code is completed.

vindard:
> Changed to draft until tracing code is completed.

Is this still WIP?

nicolasburtey:
> > Changed to draft until tracing code is completed.
> 
> Is this still WIP?

would be good to have some opentelemetry plugged into this file so we could have some alert in case there is backup errors

zoop-btc:
> > > Changed to draft until tracing code is completed.
> > 
> > 
> > Is this still WIP?
> 
> would be good to have some opentelemetry plugged into this file so we could have some alert in case there is backup errors


I added the tracing. It first creates a root span for servers.trigger.uploadBackup. Inside that span the attributes uploadBackup.destination.dropbox /googlecloud / nextcloud will be set to true or false depending on which environment variables are set. If none are set it'll record a critical error event.
<details>
  <summary>_Example of the error_</summary>
  
image

</details>

Then each subsequent uploadDestination is wrapped in a span again and an event is recorded if it's successful or not. The errorlevel of those are warning.

<details>
  <summary>_Example of an error if Dropbox Token is invalid_</summary>
  
image

</details>

I could only test Dropbox and Nextcloud as I don't have a google service account. 

<details>
  <summary>_Here what a successful upload to those two looks like_</summary>

image

</details>

Will make sure that the charts-part of this is also up-to-date and then mark both as ready for review.

src/app/admin/backup.ts
```diff
@@ -3,34 +3,129 @@ import {
...
+        DropboxAccessToken ||
+        GcsApplicationCredentials ||
+        (Nextcloudurl && Nextclouduser && Nextcloudpassword)
+      )
+    ) {
+      const err = new Error(
+        "Missing environment variable for LND static channel backup destination.",
+      )
+      logger.error(err)
+      recordExceptionInCurrentSpan({ error: err, level: ErrorLevel.Critical })
```
dolcalmi, line 38: this will trigger an alert... @nicolasburtey is dropbox required if we enable other storages?

dolcalmi (approved):
Approved with 1 comment @nicolasburtey

src/app/admin/backup.ts
nicolasburtey: we could have an alert if no storage is being set up

nicolasburtey (approved):

PR #1319, opened 2022-05-19
fix: add add'l price check if ratio too imprecise for limits

vindard:
## Description

This is to fix the issue where for small sats amount payments the price ratio calculated is too imprecise to be used for a limits check.

Scenario:
- my transaction volume is 500,000 sats (15,000 cents or $150.00)
- my limit is 400,000 cents (or $4,000.00)
- I send 1 sat which gets rounded to usdPaymentAmount of 1 cent
- Price ratio is 1sat/1cent = 1
- volume in cents gets calculated as 500,000 cents ($5,000.00) even though the real volume is only $150

The fix here is to refetch a price and calculate new PriceRatio based on a fixed minimum amount of sats required for a reasonable precision level.


src/app/payments/helpers.ts
```diff
@@ -305,3 +306,31 @@ export const newCheckTwoFALimits = async ({
     walletVolume,
   })
 }
+
+export const getPriceRatioForLimits = async <
```
dolcalmi, line 310: this should not be in `helpers` (this file should not exists), this is related to limits

src/app/payments/helpers.ts
```diff
@@ -305,3 +306,31 @@ export const newCheckTwoFALimits = async ({
...
+  S extends WalletCurrency,
+  R extends WalletCurrency,
+>(
+  paymentFlow: PaymentFlow<S, R>,
+) => {
+  const minSatsForPrecision = 5000n
```
dolcalmi, line 316: should be a config??

dolcalmi (changes requested):

src/app/payments/helpers.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1319/commits/e45dda0dec76c2eeedd643830e6fb4872e6b82cc

src/app/payments/helpers.ts
vindard: Not sure, it's  more at the use-case level because it calls on external services (won't be able to move to domain)

dolcalmi: you can pass usdFromBtcMidPriceFn and MIN_SATS_FOR_PRICE_RATIO_PRECISION from app layer or should be part of price (app layer)

dolcalmi (approved):
I am approving but `Helpers.ts` is not a good naming for a file in app layer.. Now that file has a lot of responsibilities: payment flow builder initializer, price related functions and limit related functions

PR #1321, opened 2022-05-19
feat: usd wallet migration

krtk6160:


src/migrations/20220519124936-create-usd-wallets.ts
```diff
@@ -0,0 +1,31 @@
...
+const { randomUUID } = require("crypto")
+
+module.exports = {
+  async up(db) {
+    const users = db.collection("users").aggregate([{ $group: { _id: "$_id" } }], {
+      cursor: { batchSize: 100 },
+    })
+
+    let progress = 0
+    for await (const user of users) {
```
nicolasburtey, line 12: it's more an account than a user (we have not segregated account and user yet in the db layer, but have in the domain/service layer)

krtk6160: Since we don't interact with other layers inside migrations, but only the db layer directly via mongoose, should we be using terminology from those other layers here? 
There's no concept of an account in the db layer yet, so it seemed confusing to call a user object an account

PR #1324, opened 2022-05-23
refactor: pubsub service

dolcalmi:
Notification service refactor was bigger than expected so I decided to split it in 3 PRs.

This is the 1st PR that includes pubsub service refactor (pending task for a long time)

### DIFF
--- a/src/services/notifications/index.ts
+++ b/src/services/notifications/index.ts
@@ -1,26 +1,17 @@
-import {
-  getLocale,
-  SAT_USDCENT_PRICE,
-  USER_PRICE_UPDATE_EVENT,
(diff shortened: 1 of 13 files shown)


src/services/notifications/index.ts
```diff
@@ -67,18 +59,22 @@ export const NotificationsService = (logger: Logger): INotificationsService => {
...
+          transaction: {
+            walletId,
+            txNotificationType: type,
+            amount: sats,
+            txHash,
+            displayCurrencyPerSat,
+          },
         },
       })
-      return
```
vindard, line 81: Could this method me a candidate for "default to returning `true` instead of `void` so we always have explicit returns"?

PR #1329, opened 2022-05-26
chore: swap 'paymentAmountFromSats/Cents' to 'paymentAmountFromNumber'

vindard:
## Description

This refactor is to avoid us having to do compound calls like `paymentAmountFromSat(toSats(1))`

### DIFF
--- a/src/domain/payments/index.types.d.ts
+++ b/src/domain/payments/index.types.d.ts
@@ -1,6 +1,6 @@
 type PriceRatio = {
-  convertFromUsd(convert: UsdPaymentAmount): BtcPaymentAmount
-  convertFromBtc(convert: BtcPaymentAmount): UsdPaymentAmount
+  convertFromUsd(convert: UsdPaymentAmount): BtcPaymentAmount | ValidationError
+  convertFromBtc(convert: BtcPaymentAmount): UsdPaymentAmount | ValidationError
   usdPerSat(): DisplayCurrencyBasePerSat
 }
 
--- a/src/domain/payments/index.types.d.ts
+++ b/src/domain/payments/index.types.d.ts
@@ -1,6 +1,6 @@
 type PriceRatio = {
-  convertFromUsd(convert: UsdPaymentAmount): BtcPaymentAmount | ValidationError
-  convertFromBtc(convert: BtcPaymentAmount): UsdPaymentAmount | ValidationError
+  convertFromUsd(convert: UsdPaymentAmount): BtcPaymentAmount
+  convertFromBtc(convert: BtcPaymentAmount): UsdPaymentAmount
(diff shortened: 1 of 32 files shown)


src/domain/payments/index.types.d.ts
```diff
@@ -1,6 +1,6 @@
 type PriceRatio = {
-  convertFromUsd(convert: UsdPaymentAmount): BtcPaymentAmount
-  convertFromBtc(convert: BtcPaymentAmount): UsdPaymentAmount
+  convertFromUsd(convert: UsdPaymentAmount): BtcPaymentAmount | ValidationError
```
bodymindarts, line 2: How can a validation error occur here? This seems wrong.

src/domain/payments/price-ratio.ts
```diff
@@ -15,20 +13,30 @@ export const PriceRatio = ({
...
-    return paymentAmountFromSats(toSats(btcAmount))
+  const convertFromUsd = (
+    convert: UsdPaymentAmount,
+  ): BtcPaymentAmount | ValidationError => {
+    if (convert.amount === 0n) {
+      return paymentAmountFromNumber({ amount: 0, currency: WalletCurrency.Btc })
+    }
+
+    const amount = Math.round(
+      (Number(convert.amount) * Number(btc.amount)) / Number(usd.amount),
```
bodymindarts, line 24: Why is this math still using Number? We should stay in BigInt and multiply by the desired precision. Using `number` for anything to do with money is unreliable.

src/domain/payments/price-ratio.ts
```diff
@@ -15,20 +13,30 @@ export const PriceRatio = ({
...
+    )
+    return paymentAmountFromNumber({ amount: amount || 1, currency: WalletCurrency.Btc })
```
bodymindarts, line 26: why the fallback of 1?

src/domain/payments/index.types.d.ts
vindard: It's because there was math and that `number` type inside that method. Maybe we do need to do the math -> PriceRatio change in this PR as well to keep the types clean.

bodymindarts: yes I think the math fix has to be in here as well.

src/domain/payments/price-ratio.ts
vindard: It's for this test case: https://github.com/GaloyMoney/galoy/blob/3070b6fb9e45626c9833a28e1cb66e0dd708929d/test/unit/domain/payments/price-ratio.spec.ts#L128

bodymindarts (approved):

bodymindarts (approved):

PR #1336, opened 2022-05-27
refactor: add friendly message for insufficient funds error

vindard:
## Description

This PR adds a user-facing message for the error that pops up when all our onchain funds are swapped out and no funds are available in the hot wallet for sending by users (this is usually remedied by a looping out offchain funds back onchain).

We would like to communicate to the end-user in a concise way that:
- our liquidity is currently imbalanced
- it will be re-balanced again soon

We would like to avoid:
- signalling that we somehow don't have enough funds available to fulfill a user request
- implying that we somehow run a fractional reserve system (we don't)

### DIFF
--- a/src/graphql/error-map.ts
+++ b/src/graphql/error-map.ts
@@ -17,6 +17,7 @@ import {
   UsernameError,
   RebalanceNeededError,
   DealerOfflineError,
+  InsufficientLiquidityError,
 } from "@graphql/error"
 import { baseLogger } from "@services/logger"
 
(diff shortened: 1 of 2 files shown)


src/graphql/error-map.ts
```diff
@@ -207,6 +208,10 @@ export const mapError = (error: ApplicationError): CustomApolloError => {
       message = "Unable to find a route for payment."
       return new RouteFindingError({ message, logger: baseLogger })
 
+    case "InsufficientOnChainFundsError":
+      message = "Funds temporarily offline. Will be restored shortly."
```
vindard, line 212: This is what we're looking for feedback on

pretyflaco: Alt2:
Request to send onchain funds received. Onchain replenishment initiated. Please try again in [time until request can be served]

nicolasburtey:
eventually, using a async messaging system/batcher should be the long term solution for this issue.

dissociating the payment initiation with the settlement.

dolcalmi (approved):

PR #1241, opened 2022-04-19
refactor: use wallet-invoice-builder

UncleSamtoshi:
- create invoices with WalletInvoiceBuilder
- receive invoices using WalletInvoiceReceiver

PR #1338, opened 2022-05-27
refactor: update push notifications service

dolcalmi:
Push notification service refactor.

This is the 2nd PR (Notificatios refactor) that includes push notifications service refactor


src/services/notifications/push-notifications.types.d.ts
```diff
@@ -0,0 +1,15 @@
+type SendPushNotificationArgs = {
+  deviceToken: string | string[]
+  title: string
+  body?: string
+  data?: { [key: string]: string }
+}
+
+interface IPushNotificationsService {
```
vindard, line 8: Why not define this in `domain` instead?

src/services/notifications/index.ts
```diff
@@ -520,3 +541,68 @@ export const NotificationsService = (logger: Logger): INotificationsService => {
...
+export const getPushNotificationContent = ({
+  type,
+  amountBaseCurrency,
+  baseCurrency,
+  amountDisplayCurrency,
+  displayCurrency,
+  userLanguage,
+}: {
+  type: NotificationType | "balance"
+  amountBaseCurrency: Satoshis | UsdCents
```
vindard, line 554: We've been trying to shift to using `PaymentAmount<WalletCurrency>` type in new code in place of `Satoshis`/`UsdCents`. This method may be a good candidate for going with the new convention. Single `paymentAmount` property would replace both `amountBaseCurrency` and `baseCurrency` args.

src/services/notifications/index.ts
```diff
@@ -520,3 +541,68 @@ export const NotificationsService = (logger: Logger): INotificationsService => {
...
+} => {
+  const locale = userLanguage || defaultLocale
+  const notificationType = type === "balance" ? type : `transaction.${type}`
+  const title = i18n.__(
+    { phrase: `notification.${notificationType}.title`, locale },
+    { walletCurrency: baseCurrency },
+  )
+  const baseCurrencyName = baseCurrency === WalletCurrency.Btc ? "sats" : ""
+  const baseCurrencySymbol = baseCurrency === WalletCurrency.Usd ? "$" : ""
+  const amount =
```
vindard, line 571: Might consider renaming this to something like `displayedBaseAmount` to clarify that we're explicitly converting from the usual base amount unit (cents) to a unit for display purposes (dollars)

src/services/notifications/index.ts
```diff
@@ -446,58 +489,36 @@ export const NotificationsService = (logger: Logger): INotificationsService => {
     balance: CurrencyBaseAmount
     walletCurrency: WalletCurrency
     userId: UserId
-    price: DisplayCurrencyPerSat | ApplicationError
+    price: DisplayCurrencyPerSat | Error
```
vindard, line 492: Why change from `ApplicationError`?

src/services/notifications/index.ts
```diff
@@ -520,3 +541,68 @@ export const NotificationsService = (logger: Logger): INotificationsService => {
     }),
   }
 }
+
+export const getPushNotificationContent = ({
```
vindard, line 545: Could we also sketch out some different scenarios for this function in unit tests? I'm think for e.g.:
- sat base; usd display
- usd base; usd display
- sat base; crc display
- usd base; crc display

vindard: Also, `get` so far in our codebase implies that this method is reaching out to somewhere external to retrieve something but I believe this is only doing local manipulation? If it's only local only:
- maybe we can use another prefix like `calculate` or `create` (I'm not sure what makes most sense here)
- we can maybe move this into notifications domain to be unit tested

src/domain/notifications/index.types.d.ts
```diff
@@ -26,7 +26,6 @@ type LnInvoicePaidUsdWalletArgs = {
   paymentHash: PaymentHash
   recipientWalletId: WalletId
   cents: UsdCents | bigint
-  displayCurrencyPerSat?: DisplayCurrencyPerSat
```
vindard, line 29: Why remove? How does this work if display currency is CRC?

src/app/wallets/update-pending-invoices.ts
```diff
@@ -190,12 +190,14 @@ const updatePendingInvoice = async ({
         sats: roundedDownReceived,
         displayCurrencyPerSat,
       })
-    } else {
+      return true
+    }
+
+    if (cents) {
```
vindard, line 196: Doesn't this change open up the possibility for now `notificationService` call to happen?

If this change was just for the lint, I'd probably leave as-is for now since @UncleSamtoshi should be cleaning this up in #1241 I believe

vindard (changes requested):
Looks really clean now, but I left a few comments for some small things we can potentially tighten up

src/domain/notifications/index.types.d.ts
dolcalmi: because in this case we dont need display currency, in the next PR we will use an unified API for both (BTC and USD) and we will have again display currency for usd as base

src/app/wallets/update-pending-invoices.ts
dolcalmi: No, it does not open that possibility. Also using ! is worst because if that case exists then it will crash the app. The interface and implementation are specifying that the value is required.

src/services/notifications/index.ts
dolcalmi: ApplicationError is a type defined in app layer (service layer should not depend on app layer), but you are right, we have a problem here. what we really need is to receive a valid value for `displayCurrencyPerSat` so I changed the interface https://github.com/GaloyMoney/galoy/pull/1338/commits/4e304d4a71fd4eb5d775bf0de2d020d0efa013df#diff-efedf3aa6398865cf644d1c13a3f2a6248a28640237c7436f4c1db0b4d3ffdd1L56

src/services/notifications/index.ts
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/1338/commits/4e304d4a71fd4eb5d775bf0de2d020d0efa013df#diff-a6381842e83d5711a96e584a1b5651ebbf4a3cf46e7e66a85d1df9bac42b2a0fR569

src/services/notifications/index.ts
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/1338/commits/487f3ddca353e8af8335487f89cd39fa69d183cd also `DisplayPaymentAmount` was added

src/services/notifications/index.ts
dolcalmi: done here https://github.com/GaloyMoney/galoy/pull/1338/commits/ece539b4f02c5123900446e9406bd6966963f563 

We should not move this to domain because this is only related to the service and it is not used in app layer

test/unit/services/notifications/create-push-notification-content.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import {
+  btcTransactions,
+  usdTransactions,
+  btcTransactionsWithDisplayCurrency,
+  usdTransactionsWithDisplayCurrency,
+} from "./transactions"
+
+describe("Notifications - createPushNotificationContent", () => {
+  test.each(btcTransactions)(
```
vindard, line 11: TIL, we could do this in jest 😅

vindard (approved):
Looks ok to me now

PR #1339, opened 2022-05-31
fix: correct settlementFee value in translation for usd txns

vindard:
## Description

This PR fixes a bug when translating `usd` transactions and refactors the translation unit tests to add a new test case for this scenario.

### DIFF
--- a/src/domain/wallets/index.types.d.ts
+++ b/src/domain/wallets/index.types.d.ts
@@ -49,8 +49,8 @@ type SettlementViaOnChain = {
 type BaseWalletTransaction = {
   readonly id: LedgerTransactionId | OnChainTxHash
   readonly walletId: WalletId | undefined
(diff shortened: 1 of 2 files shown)


test/unit/domain/wallets/tx-history.spec.ts
```diff
@@ -7,178 +7,216 @@ import { toSats } from "@domain/bitcoin"
...
+      settlementAmount,
+      usd,
+      currency,
+    }: {
+      walletId: WalletId
+      settlementAmount: Satoshis | UsdCents
+      usd: number
+      currency: WalletCurrency
+    }): WalletTransaction[] => {
+      const settlementFee =
```
bodymindarts, line 106: Avoid conditionals in the test code - should be straightforward

bodymindarts (approved):

PR #1340, opened 2022-05-31
chore: make 'BtcWallet' naming consistent

vindard:
## Description

We usually camel-case these sorts of names and this is the only instance where it wasn't camel-cased.

_**Note:** would this be considered a breaking API change?_

image


bodymindarts:
Closing because its a breaking API change.

PR #1342, opened 2022-06-01
fix: correct amounts for fee_reimbursement txns in db

vindard:
## Description

This PR corrects the amounts properties being added to `fee_reimbursement` transactions. In the initial changeset, these properties were originally pointed at the original transaction's amount which is incorrect. This PR changes it so that the amounts correspond to the actual transaction they live on.

### Migration

This migration behaves the same for both pre-change and post-change txns in the db since it reads off existing properties (`debit` & `usd`) from the transaction and makes changes based on that. Also we are only writing the new amounts properties and are not reading from them as yet for any business logic or api returns so these changes should not need any other refactors in the codebase.

### DIFF
--- a/src/migrations/20220601191441-fix-reimburse-amounts.ts
+++ b/src/migrations/20220601191441-fix-reimburse-amounts.ts
@@ -1,0 +1,57 @@
+/* eslint @typescript-eslint/ban-ts-comment: "off" */
+// @ts-nocheck
+
+module.exports = {
+  async up(db) {
+    const collection = db.collection("medici_transactions")
+
+    // Fetch all reimbursement txns
+    let reimbursementTxnsCursor
+    try {
+      reimbursementTxnsCursor = collection.find({ type: "fee_reimbursement" })
+    } catch (error) {
+      console.log({ result: error }, "Couldn't find reimbursement txns")
+    }
+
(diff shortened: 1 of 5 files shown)


src/migrations/20220601191441-fix-reimburse-amounts.ts
```diff
@@ -0,0 +1,57 @@
...
+      )
+      if (btcTxn === undefined) {
+        console.log(`Couldn't find usable reimbursements for ${hash}`)
+        continue
+      }
+
+      const update = {
+        satsAmount: btcTxn.credit,
+        centsAmount: Math.floor(btcTxn.usd * 100),
+        displayAmount: Math.floor(btcTxn.usd * 100),
```
dolcalmi, line 38: are we sure this is in cents?

dolcalmi (approved):
code looks good.. just a minor question in migration script

src/migrations/20220601191441-fix-reimburse-amounts.ts
vindard: Yup, it's added as the base unit in the `PaymentFlow` object and in the db (below sample from staging). The idea is to not have floats in the system eventually I believe.

image

Issue #1344, opened 2022-06-03
Tracing context invalid for ln service subscriptions in trigger


zoop-btc:
Hi Galoy Team,

Each time the lnd service triggers a function which subsequently creates a span they will have an invalid trace.parent_id.
I have observed this only in the trigger-server.
For example if lnd receives a new block it will only create a span calling [onchainBlockEventhandler](https://github.com/GaloyMoney/galoy/blob/b60dad7a58c6ca93cc8692bf443b4019208ec12e/src/servers/trigger.ts#L170) going through -> [updateOnChainReceipt ](https://github.com/GaloyMoney/galoy/blob/b60dad7a58c6ca93cc8692bf443b4019208ec12e/src/app/wallets/update-on-chain-receipt.ts#L27)-> [listIncomingTransactions](https://github.com/GaloyMoney/galoy/blob/b60dad7a58c6ca93cc8692bf443b4019208ec12e/src/services/lnd/onchain-service.ts#L72) which gets wrapped [here](https://github.com/GaloyMoney/galoy/blob/b60dad7a58c6ca93cc8692bf443b4019208ec12e/src/services/lnd/onchain-service.ts#L167).
Those spans created here will get passed the ROOT_CONTEXT which is not applicable and each time a block is found they concatenate resulting in big traces.

This can be observed in Honeycomb in this way:
1. Select the trigger-service dataset.
2. Set 'WHERE' to code.function=listIncomingTransactions
3. Check if they all have the same trace.parent_id
4. Open one span and see if there's something like this picture.
![image](https://user-images.githubusercontent.com/101409458/171871052-bffea0cf-337f-448e-955c-29095b765a6b.png)

My proposition is to pass SpanOptions to asyncRunInSpan instead of just Attributes. Meaning we can pass both the `root: true ` option as well as the Attributes to the function. Code change is minimal and existing references updated easily.
Then call asyncRunInSpan inside the trigger entrypoints. [My pull request has done it for the above situation.](https://github.com/GaloyMoney/galoy/pull/1345)

This will be important for my ongoing nextcloud backup pull request as it will make use of traces for notifying about failed backups.

Maybe it would also make sense to make use of the [Context-Manager](https://opentelemetry.io/docs/instrumentation/js/api/context/#context-manager). But since it wasn't implemented this goes above my paygrade.

Thank you regards,
Zoop

zoop-btc:
[...]

PR #1345, opened 2022-06-03
fix: invalid tracing context set as root in trigger

zoop-btc:
This is a proposed fix for [this issue](https://github.com/GaloyMoney/galoy/issues/1344).
It changes the attributes arguments to SpanOptions on the asyncRunInSpan function.
This means when we call it we can optionally set the span to be the root of the trace.
I have called in that way inside the trigger.ts for onchainBlockEventhandler as example.

The main change made:
https://github.com/GaloyMoney/galoy/blob/71f697ba18636b45ab4a6a9ae0a8d2b40a405e07/src/services/tracing.ts#L230-L239

How to set a span to be root with this:
https://github.com/GaloyMoney/galoy/blob/71f697ba18636b45ab4a6a9ae0a8d2b40a405e07/src/servers/trigger.ts#L177-L193

### DIFF
--- a/src/app/lightning/update-ln-payments.ts
+++ b/src/app/lightning/update-ln-payments.ts
@@ -60,15 +60,17 @@ const updateLnPaymentsByFunction = async ({
     const results = await asyncRunInSpan(
       "app.lightning.updateLnPaymentsPaginated",
       {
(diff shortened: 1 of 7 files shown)


dolcalmi (approved):

vindard (approved):
LGTM

PR #1347, opened 2022-06-04
chore: improve types

nicolasburtey:
- make stricter control on values possible for the config yaml
- create a `YamlSchema` type. ajv is supposetly able to derive types itself but I could not make it work, so declaring it explicitly. this at least helped to find one bug (failedLoginAttemptPerEmailAddress instead of failedLoginAttemptEmailAddress)
- got rid of few dozen `noImplcitAny`


src/domain/accounts/const.ts
```diff
@@ -0,0 +1,11 @@
+export const AccountLevel = {
```
bodymindarts, line 1: Up to now we are using a file called `primitives.ts` for these kind of types.

src/domain/payments/index.types.d.ts
```diff
@@ -78,8 +78,11 @@ type LightningPaymentFlowBuilder<S extends WalletCurrency> = {
     uncheckedAmount: number
   }): LPFBWithInvoice<S> | LPFBWithError
   withoutInvoice({
-    uncheckedAmount: number,
-    description: string,
+    uncheckedAmount,
+    description,
+  }: {
+    uncheckedAmount: number
+    description: string | null
```
bodymindarts, line 85: Didn't we say we want to represent absence as undefined?

src/domain/accounts/const.ts
nicolasburtey: updated

PR #1350, opened 2022-06-07
fix: intraledger notification usd amount

dolcalmi:
Fix intraledger notifications amount

Note: this is a temporal fix because in the next PR the notifications interface will change (to receive display currency and display amount)


vindard (dismissed):
Looks good

PR #1352, opened 2022-06-07
Casbin pg

bodymindarts:


bodymindarts:
This PR is mainly a bunch of boilerplate to explore how one could represent authorization in galoy. It is not intended to be merged or continued to work on but it should probably be kept around until initial integration with casbin has been merged.

For the most part the code here is a bit too abstract. It is trying to focus on the `support role` use case but in doing this I realised the very first use case should actually be a kind of `SuperAdmin` role that can add / remove roles on other users. That way we would solve the bootstrapping issue of hard coding 'editors'

Whoever picks up the casbin integration work next should focus on a minimal implementation of the use case of adding / removing roles on users. THEN start defining what other roles there actually are.

src/services/casbin/casbin.conf
```diff
@@ -0,0 +1,14 @@
+[request_definition]
+r = sub, res, act
+
+[policy_definition]
+p = role, scope, res, act
```
nicolasburtey, line 5: what is the intention of scope here @bodymindarts ?

this is a non standard casbin.conf so I'm trying to understand why you went that path instead of using one of the proposed "standard" RBAC conf file

bodymindarts: This is [one of the standard configs](https://casbin.org/docs/en/rbac-with-domains).

Why the 'scope'? I think it is appropriate for our system to consider the `account` as the `domain` / `scope` as described in the example. Consider the following use cases:

A user who should have complete access to all wallets of an account (ie account admin?).
Without the scope every time a new wallet is created in an account we would have to persist an association triplet as follows:
`subject: user/123` - `permission: wallet/read` - `resource: wallet/456` - `allow`.
When using `account` as the scope we could also do the following:
`subject: user/123` - `scope: account/456` - `permission: wallet/read` - `resource: wallet/*` - `allow`.

Or more realistically:
`role: account/admin` - `scope: account/456` - `permission: wallet/read` - `resource: wallet/*` - `allow`.

and then assign the `user/123` to `role: account/admin` in `scope: account/456`.

Then any user with the `account/admin` role would inherit access to any additional wallet created under `account/456` without needing to persist an additional triplet.

Put another way without the `scope` `wallet/*` would give permission to absolutely every wallet in the system _with_ `scope` `wallet/*` can give access to any wallet _of a specific account_ (aka scope / domain).

nicolasburtey:
> This PR is mainly a bunch of boilerplate to explore how one could represent authorization in galoy. It is not intended to be merged or continued to work on but it should probably be kept around until initial integration with casbin has been merged.
> 
> For the most part the code here is a bit too abstract. It is trying to focus on the `support role` use case but in doing this I realised the very first use case should actually be a kind of `SuperAdmin` role that can add / remove roles on other users. That way we would solve the bootstrapping issue of hard coding 'editors'
> 
> Whoever picks up the casbin integration work next should focus on a minimal implementation of the use case of adding / removing roles on users. THEN start defining what other roles there actually are.

I guess you don't see the use of the superAdmin function of Casbin?
https://casbin.org/docs/en/superadmin

this one could be used to bootstrap other users.

if we want to have an "accountant" role with read only across the instance, then the root mode would not work obviously

PR #1354, opened 2022-06-08
chore: small types improvment

nicolasburtey:


src/domain/cold-storage/index.ts
```diff
@@ -8,7 +8,7 @@ export const RebalanceChecker = ({
   const getWithdrawFromHotWalletAmount = ({
     onChainHotWalletBalance,
     offChainHotWalletBalance,
-  }) => {
+  }: RebalanceCheckerArgs): Satoshis => {
```
bodymindarts, line 11: Shouldn't this be `WithdrawFromHotWalletAmountArgs`? Is the type being used somewhere else?

bodymindarts (approved):

src/domain/cold-storage/index.ts
nicolasburtey: good question. will have a look and eventually incorporate the change in my next PR

PR #1355, opened 2022-06-08
fix: round down USD amounts when reimbursing fees

vindard:
## Description

This PR ensures that for reimbursements if there is a non-zero sat fee we should always round down the usd amount that we reimburse.

### DIFF
--- a/src/domain/ledger/fee-reimbursement.ts
+++ b/src/domain/ledger/fee-reimbursement.ts
@@ -22,7 +22,8 @@ export const FeeReimbursement = ({
       currency: WalletCurrency.Btc,
     }
 
-    const feeDifferenceUsdAmount = priceRatio.convertFromBtc(feeDifferenceBtcAmount)
+    const feeDifferenceUsdAmount =
+      priceRatio.convertFromBtcToFloor(feeDifferenceBtcAmount)
 
     return { btc: feeDifferenceBtcAmount, usd: feeDifferenceUsdAmount }
   }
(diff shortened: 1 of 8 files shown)


src/domain/ledger/fee-reimbursement.ts
```diff
@@ -22,7 +22,10 @@ export const FeeReimbursement = ({
       currency: WalletCurrency.Btc,
     }
 
-    const feeDifferenceUsdAmount = priceRatio.convertFromBtc(feeDifferenceBtcAmount)
+    const feeDifferenceUsdAmount =
+      actualFee.amount === 0n
```
dolcalmi, line 26: could you explain this validation? what if a payment is done with only one hop?

vindard: Hops doesn't matter here I believe, this is more related to the final fee amounts we observe after a payment is completed.

This is here because we want to give back the full fee we optimistically take when we can't complete a fee probe. If we use `convertFromBtcToFloor` when there is no actual fee then we will still take 1 cent from users in some rounding scenarios ([this test case](https://github.com/GaloyMoney/galoy/pull/1355/files#diff-7a72dce15a5c6cf035300634963290f4aece5a8a6ad6768ab4ce843b18a4a11fR122))

dolcalmi: I mean... if it is 1 hop then fee is 0 (direct channel)

PR #1359, opened 2022-06-08
refactor: round up usd send fees in payment flow builder

vindard:
## Description

This PR is to round up the routing fees when converting to `usdProtocolFee` when building a payment flow.

The bulk of this PR was:
🛠️  refactoring the payment-flow-builder unit tests with more realistic ratios and appropriate amounts ([here](https://github.com/GaloyMoney/galoy/pull/1359/commits/0359f5b9f357af68af97a5a520cc9767aff1733a))
🛠️  to be able to isolate a failing test ([here](https://github.com/GaloyMoney/galoy/pull/1359/commits/489089e0ad4cb62d8af90ec50578ac746c6b47c0))
✅  to be able to fix with the changes to the fee calculations ([here](https://github.com/GaloyMoney/galoy/pull/1359/commits/90eff7cb32848f6e6eb3d3dbeef239c02efcb768))

### DIFF
--- a/test/unit/domain/shared/amount-calculator.spec.ts
+++ b/test/unit/domain/shared/amount-calculator.spec.ts
@@ -144,5 +144,49 @@ describe("AmountCalculator", () => {
         expect(result.currency).toBe(currency)
       })
     })
+
+    describe("by rounding to ceil", () => {
+      it("no rounding if remainder is 0", () => {
+        const result = calc.divCeil({ amount: pivot, currency }, divisor)
+        expect(result.amount).toEqual(quotient)
+        expect(result.currency).toBe(currency)
+      })
+
+      it("rounds up if remainder is just above pivot", () => {
+        const result = calc.divCeil({ amount: pivot + 1n, currency }, divisor)
+        expect(result.amount).toEqual(quotient + 1n)
+        expect(result.currency).toBe(currency)
+      })
+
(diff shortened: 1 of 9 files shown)


test/unit/domain/shared/amount-calculator.spec.ts
```diff
@@ -57,92 +57,97 @@ describe("AmountCalculator", () => {
...
+        result: {
+          normal: quotient,
+          floor: quotient,
+          ceil: quotient + 1n,
+        },
+      },
+    ]
+
+    for (const describeGroup of describeCases) {
+      describe(`by rounding to ${describeGroup.type}`, () => {
```
dolcalmi, line 138: uhmm not sure how jest behaves with this for syntax so I think is better to use test.each https://jestjs.io/docs/api#testeachtablename-fn-timeout

dolcalmi (changes requested):

test/unit/domain/shared/amount-calculator.spec.ts
vindard: It works well in local testing so far.

I tried `test.each` at first for this, but the test titles are also dependent on the `table` and `test.each` only gets table items in its 2nd `fn` argument.

dolcalmi (approved):

PR #1373, opened 2022-06-14
perf: remove temporarily payment and invoice updates in get balance

dolcalmi:


vindard:
@dolcalmi should we close this now that the issue's been resolved? (resolved by forcing a compaction on lnd)

PR #1377, opened 2022-06-15
refactor: change 'Transaction' gql object amounts type/descriptions

vindard:
## Description

`Transaction` objects now need to reference both `BTC` and `USD` settlement types.

This is a proposed change to the `Transaction` object to allow that. The `settlementAmount` property has been made generic since the currency information for the transaction is carried in the `settlementCurrency` property on the object (similar to how we encode this in the `PaymentAmount` type [in the domain](https://github.com/GaloyMoney/galoy/blob/16c76100cb7ed6241ae7a87982477346e460c175/src/domain/shared/index.types.d.ts#L12))

_**Tbd:** does this break the API?_


samerbuna (approved):
Since the types used here are not interface types, I think this change is okay.

PR #1380, opened 2022-06-16
feat: create endpoint for adding usd wallets

UncleSamtoshi:


src/graphql/admin/root/mutation/account-add-usd-wallet.ts
```diff
@@ -0,0 +1,56 @@
...
+          currency: WalletCurrency.Usd,
+        }),
+      ),
+    )
+    const walletDetails = addWalletResults.map((wallet) => {
+      if (wallet instanceof Error) {
+        return { errors: [{ message: mapError(wallet).message }] }
+      }
+
+      return { errors: [], walletDetails: wallet }
```
dolcalmi, line 49: errors is an array, you should include the errors in only 1 array and not create and array of this structure. @samerbuna could you please validate this?

dolcalmi (changes requested):

src/graphql/admin/root/mutation/account-add-usd-wallet.ts
samerbuna: yes, I agree. This will result in a response that's an array with objects that have the `errors` array. The resp should be an object, not an array.

UncleSamtoshi: I will make this change

UncleSamtoshi:
** I exposed wallets in the admin panel, so now an admin can look at the balance and transactions of users. I'm assuming that an admin should have this permission. Any disagreements @nicolasburtey @bodymindarts ?

nicolasburtey:
> ** I exposed wallets in the admin panel, so now an admin can look at the balance and transactions of users. I'm assuming that an admin should have this permission. Any disagreements @nicolasburtey @bodymindarts ?

I think we should eventually have tier access on the admin panel. only "super admin" should have access to list of transaction of balance. 

staff member doing "level 1 support" does need balance, because it's a security risk. if their account is compromised, then all balance of a deployment could become public.

to do this granular system, we should have a [proper authorization system](https://github.com/GaloyMoney/galoy/pull/1352).

in the meantime, while I didn't do the work on this initially, I think that is ok to expose it

src/graphql/admin/types/object/account.ts
```diff
@@ -20,6 +23,12 @@ const Account: GraphQLObjectType<Account> = GT.Object<Account>({
     level: { type: GT.NonNull(AccountLevel) },
     status: { type: GT.NonNull(AccountStatus) },
     title: { type: GT.String },
+    wallets: {
+      type: GT.NonNullList(Wallet),
+      resolve: async (source) => {
+        return WalletsRepository().listByAccountId(source.id)
```
dolcalmi, line 29: I know some resolvers use services directly but we should avoid this. We need to implement a method to support this at app layer level

src/graphql/admin/types/payload/wallet-details.ts
```diff
@@ -0,0 +1,17 @@
...
+import Wallet from "@graphql/types/abstract/wallet"
+
+const WalletDetailsPayload = GT.Object({
+  name: "WalletDetailsPayload",
+  fields: () => ({
+    errors: {
+      type: GT.NonNullList(IError),
+    },
+    walletDetails: {
+      type: GT.List(Wallet),
```
dolcalmi, line 12: shouldn't be this a non null list too? wallet details always have a value `const walletDetails: Wallet[] = []`

dolcalmi (changes requested):

UncleSamtoshi:
> > ** I exposed wallets in the admin panel, so now an admin can look at the balance and transactions of users. I'm assuming that an admin should have this permission. Any disagreements @nicolasburtey @bodymindarts ?
> 
> I think we should eventually have tier access on the admin panel. only "super admin" should have access to list of transaction of balance.
> 
> staff member doing "level 1 support" does need balance, because it's a security risk. if their account is compromised, then all balance of a deployment could become public.
> 
> to do this granular system, we should have a [proper authorization system](https://github.com/GaloyMoney/galoy/pull/1352).
> 
> in the meantime, while I didn't do the work on this initially, I think that is ok to expose it

I could create a IWalletSummary object that excludes transactions, or I could leave it as is. Thoughts?

src/graphql/admin/types/object/account.ts
UncleSamtoshi: It looks like we already have this method in the app layer, and I will use that instead.

dolcalmi (approved):

PR #1382, opened 2022-06-18
chore: more type improvements

nicolasburtey:


src/graphql/types/scalar/ln-pubkey.ts
```diff
@@ -14,12 +18,12 @@ const LnPubkey = GT.Scalar({
...
 })
 
-function validLnPubkey(value) {
-  // TODO: verify/improve
-  if (value.match(/^[a-f0-9]{66}$/i)) {
-    return value
+function validLnPubkey(value: string): Pubkey | InputValidationError {
+  const pubkey = checkedToPubkey(value)
+  if (pubkey instanceof Error) {
+    return new InputValidationError({ message: "Invalid value for LnPubkey" })
```
dolcalmi, line 24: Shouldn't this be an OutputValidationError?

src/services/lnd/health.ts
```diff
@@ -39,17 +36,19 @@ const isLndUp = async (param: LndParamsUnAuthed): Promise<void> => {
   }
 
   const authParam = authParams.find((p) => p.socket === socket)
-  if (authParam) {
-    authParam.active = active
+  if (!authParam) {
+    throw new Error("unreachable: this should not happen, authParam should not be null")
```
dolcalmi, line 40: not sure about this.. this can crash the service... are we sure to do this?

src/services/lnd/onchain-service.ts
```diff
@@ -23,6 +23,8 @@ import { SECS_PER_5_MINS } from "@config"
 
 import { getActiveOnchainLnd } from "./utils"
 
+import { parseLndErrorDetails } from "."
```
dolcalmi, line 26: in general this can lead to a circular reference, maybe is better to move `parseLndErrorDetails` to utils

dolcalmi (changes requested):
looks good, just minor comments/changes

src/config/yaml.ts
```diff
@@ -117,11 +119,15 @@ export const getLndParams = (): LndParams[] => {
...
+    const pubkey_ = process.env[`${input.name}_PUBKEY`]
+    if (!pubkey_) throw new ConfigError(`missing PUBKEY for ${input.name}`)
+
+    const pubkey = checkedToPubkey(pubkey_)
+    if (pubkey instanceof Error)
+      throw new ConfigError(`wrong PUBKEY formatting for ${input.name}`)
 
     const port = process.env[`${input.name}_RPCPORT`] ?? 10009
-    const type = input.type.map((item) => item as NodeType)
+    const type = input.type.map((item) => item as NodeType) // TODO: verify if validation is done from ya
```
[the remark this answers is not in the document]
nicolasburtey: `yaml.ts`

src/app/index.ts
```diff
@@ -18,9 +18,10 @@ const allFunctions = {
   Users: { ...UsersMod },
   Wallets: { ...WalletsMod },
   Payments: { ...PaymentsMod },
-}
+} as const
 
-for (const subModule in allFunctions) {
+let subModule: keyof typeof allFunctions
```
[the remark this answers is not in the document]
nicolasburtey: yes subModule is really a typed as a string in `for (const subModule in allFunctions) {`

the new code restrict the string so that is can only be the list of string of the function name

src/app/users/login.ts
```diff
@@ -29,14 +25,11 @@ export const login = async ({
   logger,
   ip,
 }: {
-  phone: string
+  phone: PhoneNumber
   code: PhoneCode
   logger: Logger
   ip: IpAddress
 }): Promise<JwtToken | ApplicationError> => {
-  const phoneNumberValid = checkedToPhoneNumber(phone)
```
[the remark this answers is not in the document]
nicolasburtey: the graphql code already does the check now

https://github.com/GaloyMoney/galoy/pull/1382/files#diff-9c3159a2d8de10efb8a453dab443a53751163476970a8ac8eeac0a8ad7a42feeR22

src/app/users/login.ts
```diff
@@ -29,14 +25,11 @@ export const login = async ({
   logger,
   ip,
 }: {
-  phone: string
+  phone: PhoneNumber
```
bodymindarts, line 28: Why are you passing in `PhoneNumber`? Up to now we had discussed to do primitive -> domain type conversion within app layer.
Is this a special case or are we changing the pattern?

src/domain/bitcoin/lightning/index.ts
```diff
@@ -18,3 +20,12 @@ export const PaymentSendStatus = {
   Pending: { value: "pending" },
   AlreadyPaid: { value: "already_paid" },
 } as const
+
+const pubkeyRegex = /^[a-f0-9]{66}$/i
+export const checkedToPubkey = (pubkey: string): Pubkey | ValidationError => {
+  if (pubkey.match(pubkeyRegex)) {
+    return pubkey as Pubkey
+  } else {
+    return new ValidationError("Pubkey conversion error")
```
bodymindarts, line 29: Aren't we using `ValidationError` only as a base class... shouldn't this be `class PubKeyConversionError extends ValidationError`?

bodymindarts (dismissed):

src/config/yaml.ts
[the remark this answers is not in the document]
nicolasburtey: comment fixed

src/graphql/types/scalar/ln-pubkey.ts
nicolasburtey: `parseValue` and `parseLiteral` are always inputs. 
the output are managed by `serialize`

https://stackoverflow.com/questions/41510880/whats-the-difference-between-parsevalue-and-parseliteral-in-graphqlscalartype

on the graphql layer, it's likely we have some check that are related to inputs where the scalar is only used as an output. I din't really spend the time digging into this, this felt very low priority to dive into it.

src/app/users/login.ts
nicolasburtey: the graphql layer provide a way to verify type, and we were already using it for phone, but we still kept the result as a string.

you'll see in the previous code that we were doing two (different (uh)) validations. a first one in the graphql layer with one regex, and another one in the domain with another regex.

https://github.com/GaloyMoney/galoy/pull/1382/files#diff-9c3159a2d8de10efb8a453dab443a53751163476970a8ac8eeac0a8ad7a42feeR22

I have converted some that just required one regex directly (or generally: no async/network call function) in the graphql layer so those types are already cast entering the domain layer

for types conversion that are more complex/require more complex validation I have left them in the domain layer.

src/domain/bitcoin/lightning/index.ts
nicolasburtey: you're right. fixed

src/app/users/login.ts
nicolasburtey: here is an example where the casting were done 3 times: https://github.com/GaloyMoney/galoy/pull/1382/commits/cbb8de8a945ff3acd26e30785abe256f482dcebe#diff-5e943a707e08464a698469ae3b768e75d563aa194102cf6ba398ae017abd2fe1L31

1. at the graphql layer
2. at the requestPhoneCodeWithCaptcha function
3. at the requestPhoneCode function

src/services/lnd/health.ts
nicolasburtey: what would you do?

it's really to silence typescript that I have done this.

authParam is typed LndParamsAuthed | undefined after authParams.find((p) => p.socket === socket), but really it can only be LndParamsAuthed (but typescript doesn't understand this).

maybe I can just cast the type instead?

src/services/lnd/onchain-service.ts
nicolasburtey: https://github.com/GaloyMoney/galoy/pull/1382/commits/af811e72f1549544972e4dea4622a165a99f4dc8

dolcalmi (approved):
Approving but we need to discuss again Justin comments (code review?). I prefer validation in app layer specially if app methods are used in different servers.

ps: this is the only change blocking the merge of this big PR and we can do a phone validation  rollback if we decide to keep app layer type check and validation

PR #1383, opened 2022-06-19
Add onchain block cache

dolcalmi:
- Refactor cache interface
- Add redis cache using `ioredis-cache` package
- Add redis cache to listIncomingTransactions
  - Clean cache when we receive an on-chain tx (trigger)
  - Set cache in updateOnChainReceipt. This improves performance because is only used by trigger subscribeToBlocks and cron 
  - Use getOrSet in getTransactionsForWalletId (most of the requests to lnd will be done by trigger but I added this fallback)


src/domain/cache/index.ts
```diff
@@ -3,6 +3,7 @@ export * from "./errors"
 export const CacheKeys = {
   CurrentPrice: "price:current",
   PriceHistory: "price:history",
+  LastOnChainTransactions: "bitcoin:lastOnChainTxs",
```
nicolasburtey, line 6: it's only mempool tx right? maybe this should be more evident by the name

src/app/wallets/get-transactions-for-wallet.ts
```diff
@@ -42,8 +50,12 @@ export const getTransactionsForWallets = async (
 
   // we are getting both the transactions in the mempool and the transaction that
   // have been mined by not yet credited because they haven't reached enough confirmations
-  const onChainTxs = await onChain.listIncomingTransactions(ONCHAIN_MIN_CONFIRMATIONS)
-  if (onChainTxs instanceof OnChainError) {
+  const onChainTxs = await redisCache.getOrSet({
+    key: CacheKeys.LastOnChainTransactions,
+    ttlSecs: toSeconds(SECS_PER_5_MINS * 2),
```
nicolasburtey, line 55: `SECS_PER_5_MINS * 2` is not great imo. if we want 10 min we could have a const for it?

src/app/wallets/get-transactions-for-wallet.ts
```diff
@@ -1,14 +1,22 @@
...
-import { OnChainService } from "@services/lnd/onchain-service"
+import { OnChainError, TxDecoder, TxFilter } from "@domain/bitcoin/onchain"
+
 import { baseLogger } from "@services/logger"
+import { LedgerService } from "@services/ledger"
+import { RedisCacheService } from "@services/cache"
 import { WalletsRepository } from "@services/mongoose"
+import { OnChainService } from "@services/lnd/onchain-service"
+
+const redisCache = RedisCacheService()
```
nicolasburtey, line 19: what drive your decision to use Redis instead of a local cache?

because cache can be invalidated from trigger?

 do you have a honeycomb query to see how the latency distribution is from `onChain.listIncomingTransactions(ONCHAIN_MIN_CONFIRMATIONS)`?

src/domain/cache/index.ts
nicolasburtey: nevermind I read back the code

dolcalmi: not only mempool, lnd return confirmed and not confirmed transactions

src/app/wallets/get-transactions-for-wallet.ts
dolcalmi: https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/4tkafh5ZGbg it is not only the latency but also the amount of requests to lnd because it is in the main query (specially when lnd has problems)

src/app/wallets/get-transactions-for-wallet.ts
dolcalmi: fixed here https://github.com/GaloyMoney/galoy/pull/1383/commits/523bef1133049ece9c516024caa82d744448a241

src/app/wallets/get-transactions-for-wallet.ts
dolcalmi: this is also a good query to understand in what methods we should improve performance and amount of requests https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/oXTxLx6jKbK

nicolasburtey: > this is also a good query to understand in what methods we should improve performance and amount of requests https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/EGqhArMX7R2

yup. lookup invoice would really benefits for using holdinvoices

src/domain/cache/index.ts
nicolasburtey, line 6: on caveat here is that we call `lastIncomingTransactions` with different depth parameter. I think this cache key here doesn't convey this insights.

maybe cache key could incorporate this somehow?

src/app/wallets/get-transactions-for-wallet.ts
```diff
@@ -42,8 +49,12 @@ export const getTransactionsForWallets = async (
 
   // we are getting both the transactions in the mempool and the transaction that
   // have been mined by not yet credited because they haven't reached enough confirmations
-  const onChainTxs = await onChain.listIncomingTransactions(ONCHAIN_MIN_CONFIRMATIONS)
-  if (onChainTxs instanceof OnChainError) {
+  const onChainTxs = await redisCache.getOrSet({
+    key: CacheKeys.LastOnChainTransactions,
+    ttlSecs: SECS_PER_10_MINS,
```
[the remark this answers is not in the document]
vindard: In theory this shouldn't matter because a partial transactions list is built from the ledger before this code and is early returned if the `OnChainService` errors which would likely happen before this code when we're instantiating the `OnChainService`. This TTL would only affect pending incoming txs I believe, so it's not the same issue we've been observing.

Longer TTL hedges us against a very small window where the `OnChainService` successfully comes up and then a `listIncomingTransactions` call fails. Tradeoff is new pending incoming txs are more delayed in showing up in the API.

If these assumptions are all correct, I think shorter TTL is ok here.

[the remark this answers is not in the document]
nicolasburtey: ok yes thanks for the input. I'm approving it.

nicolasburtey (approved):

PR #1384, opened 2022-06-19
chore: type improvement 6

nicolasburtey:
@dolcalmi if you want some Typescript types challenges: open this PR and see if you can remove my typescript directives and get the correct types :) 

this is the last PR that would activate no-implicity-any in the main codebase

### DIFF
--- a/src/graphql/main/queries.ts
+++ b/src/graphql/main/queries.ts
@@ -30,13 +30,16 @@ const fields = {
   btcPrice: BtcPriceQuery,
   btcPriceList: BtcPriceListQuery,
   onChainTxFee: OnChainTxFeeQuery,
-}
+} as const
 
+/* eslint @typescript-eslint/ban-ts-comment: "off" */
+// @ts-ignore-next-line no-implicit-any error
 const addTracing = (fields) => {
   for (const key in fields) {
     const original = fields[key].resolve
+    // @ts-ignore-next-line no-implicit-any error
     fields[key].resolve = (source, args, context, info) => {
-      const { ip, domainAccount, domainUser } = context
+      const { ip, domainAccount, domainUser } = context as GraphQLContextForUser
       return addAttributesToCurrentSpanAndPropagate(
         {
           [SemanticAttributes.ENDUSER_ID]: domainUser?.id,
--- a/src/graphql/main/queries.ts
+++ b/src/graphql/main/queries.ts
@@ -1,3 +1,5 @@
+import { GraphQLResolveInfo } from "graphql"
+
 import { GT } from "@graphql/index"
 
 import MeQuery from "@graphql/root/query/me"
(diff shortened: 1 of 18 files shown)


samerbuna:
@dolcalmi @vindard can you please review this? I think it's in a good stage to merge (there still a lot more to do, but let's get the check in first)

src/graphql/main/queries.ts
```diff
@@ -30,24 +32,32 @@ const fields = {
...
+            [ACCOUNT_USERNAME]: domainAccount?.username,
+            [SemanticAttributes.HTTP_CLIENT_IP]: ip,
+          },
+          () => original(source, args, context, info),
+        )
+      }
     }
   }
-  return fields
+  return trcFields
```
nicolasburtey, line 60: what does `tcr` stands for?

src/graphql/root/query/on-chain-tx-fee-query.ts
```diff
@@ -28,7 +28,7 @@ const OnChainTxFeeQuery = GT.Field({
 
     const fee = await Wallets.getOnChainFee({
       walletId,
-      account: domainAccount,
+      account: domainAccount as Account,
```
nicolasburtey, line 31: why not having `{ domainAccount: Account }` at the function signature above? we're looking some type guarantee with the casting here

src/graphql/main/queries.ts
samerbuna: trace

src/graphql/root/query/on-chain-tx-fee-query.ts
samerbuna: That type should come from the context, I plan to add the context type to all resolvers in the next PR.

src/servers/middlewares/wallet-id.ts
```diff
@@ -1,29 +1,54 @@
...
   if (walletId instanceof Error) return walletId
 
-  const hasPermissions = await Accounts.hasPermissions(context.domainUser.id, walletId)
+  if (!context.domainUser) {
+    return new Error("Invalid wallet")
+  }
+
+  const hasPermissions = await Accounts.hasPermissions(
+    context.domainUser.id,
+    walletId as WalletId,
```
vindard, line 25: Just a note-to-self that this type is properly applied in the `scalar/wallet-id.ts` scalar object and that this casting will likely be removed later on once the args are properly types here

vindard (commented):
Has conflicts that need resolving, but otherwise looks good to me

src/graphql/root/query/on-chain-tx-fee-query.ts
nicolasburtey: agree but I may be missing something: the casting is still unnecessary here compare to the existing code?

samerbuna: We have 2 context types (one for public and one for authenticated). The default is public, so the type here is `Account | undefined`.  This is an authenticated query, so we either cast here or use the authenticated context type. I faced some issues using the authenticated context type, so I'll probably come up with a different option.

nicolasburtey: my point is that the line above was already making `Account` a not undefined field?

https://github.com/GaloyMoney/galoy/pull/1384/files#diff-4fed0ac3ed1e683af6718bfe6cf2825f4eacedadc4dab1be89b387e8b29b8b6dR19

nicolasburtey (commented):
still have a question on https://github.com/GaloyMoney/galoy/pull/1384/files#r926395160 but going to merge it so we start to get more type enforcement in the core codebase

bodymindarts (approved):

PR #1395, opened 2022-06-24
feat: hodl invoice (on WalletInvoiceBuilder)

vindard:
## Description

This PR re-applies the same changes from #1188 on top of our new WalletInvoiceBuilder abstraction.

### DIFF
--- a/src/app/wallets/get-balance-for-wallet.ts
+++ b/src/app/wallets/get-balance-for-wallet.ts
@@ -1,8 +1,6 @@
 import { LedgerService } from "@services/ledger"
 import { updatePendingPaymentsByWalletId } from "@app/payments"
 
-import { updatePendingInvoicesByWalletId } from "./update-pending-invoices"
-
 export const getBalanceForWallet = async ({
   walletId,
   logger,
@@ -10,16 +8,10 @@ export const getBalanceForWallet = async ({
   walletId: WalletId
   logger: Logger
 }): Promise<CurrencyBaseAmount | ApplicationError> => {
-  const [, updatePaymentsResult] = await Promise.all([
-    updatePendingInvoicesByWalletId({
-      walletId,
-      logger,
-    }),
-    updatePendingPaymentsByWalletId({
-      walletId,
-      logger,
-    }),
-  ])
+  const updatePaymentsResult = await updatePendingPaymentsByWalletId({
+    walletId,
+    logger,
+  })
   if (updatePaymentsResult instanceof Error) return updatePaymentsResult
 
   return LedgerService().getWalletBalance(walletId)
--- a/src/app/wallets/update-pending-invoices.ts
+++ b/src/app/wallets/update-pending-invoices.ts
@@ -46,23 +46,6 @@ export const updatePendingInvoices = async (logger: Logger): Promise<void> => {
   logger.info("finish updating pending invoices")
 }
 
-export const updatePendingInvoicesByWalletId = async ({
-  walletId,
-  logger,
-}: {
-  walletId: WalletId
-  logger: Logger
-}) => {
-  const invoicesRepo = WalletInvoicesRepository()
-
-  const invoices = invoicesRepo.findPendingByWalletId(walletId)
-  if (invoices instanceof Error) return invoices
-
-  for await (const walletInvoice of invoices) {
-    await updatePendingInvoice({ walletInvoice, logger })
-  }
-}
-
 export const updatePendingInvoiceByPaymentHash = async ({
   paymentHash,
   logger,
--- a/src/app/wallets/update-pending-invoices.ts
+++ b/src/app/wallets/update-pending-invoices.ts
@@ -17,29 +17,25 @@ import { WalletInvoiceReceiver } from "@domain/wallet-invoices/wallet-invoice-re
 import * as LedgerFacade from "@services/ledger/facade"
 import { usdFromBtcMidPriceFn } from "@app/shared"
 
-export const updatePendingInvoices = async (logger: Logger): Promise<void> => {
+export const declineHeldInvoices = async (logger: Logger): Promise<void> => {
   const invoicesRepo = WalletInvoicesRepository()
 
-  const walletIdsWithPendingInvoices = invoicesRepo.listWalletIdsWithPendingInvoices()
+  const pendingInvoices = invoicesRepo.yieldPending()
 
-  if (walletIdsWithPendingInvoices instanceof Error) {
+  if (pendingInvoices instanceof Error) {
     logger.error(
-      { error: walletIdsWithPendingInvoices },
+      { error: pendingInvoices },
       "finish updating pending invoices with error",
     )
     return
   }
 
   await runInParallel({
-    iterator: walletIdsWithPendingInvoices,
+    iterator: pendingInvoices,
     logger,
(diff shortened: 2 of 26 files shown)


src/app/wallets/update-pending-invoices.ts
```diff
@@ -94,46 +73,43 @@ const updatePendingInvoice = async ({
...
-    onUs: false,
-  })
-
   if (walletInvoice.paid) {
     pendingInvoiceLogger.info("invoice has already been processed")
     return true
   }
 
+  if (!lnInvoiceLookup.isHeld) {
+    pendingInvoiceLogger.info("invoice is not been held")
```
nicolasburtey, line 109: I guess this means invoice has not been paid? maybe we could clarify it

src/app/wallets/update-pending-invoices.ts
```diff
@@ -238,3 +217,60 @@ const updatePendingInvoice = async ({
...
+  if (lnInvoiceLookup instanceof Error) return lnInvoiceLookup
+
+  if (!lnInvoiceLookup.isHeld) {
+    pendingInvoiceLogger.info({ lnInvoiceLookup }, "invoice is not been held")
+    return false
+  }
+
+  pendingInvoiceLogger.error(
+    { lnInvoiceLookup },
+    "invoice has been held and is now been cancelled",
```
nicolasburtey, line 264: could be useful to show/calculate for how long the invoice has been held.

src/servers/trigger.ts
```diff
@@ -181,7 +181,7 @@ export const onchainBlockEventHandler = async (height: number) => {
 export const invoiceUpdateEventHandler = async (invoice: GetInvoiceResult) => {
```
nicolasburtey, line 181: I think an important test on this PR is to check (manually, on devnet or testnet, most tests doesn't rely on triggered  event), that the event is being trigger when the invoice is been held.

also see what happened if trigger is being restarted after an invoice has been created but not paid.

https://github.com/lightningnetwork/lnd/issues/3120

I went through the code here but not clear to me if there is a different of behavior with held invoice or not (this is what I remember when I first did this PR some months ago)
https://github.com/alexbosworth/lightning/blob/master/lnd_methods/invoices/subscribe_to_invoices.js

src/app/wallets/get-balance-for-wallet.ts
```diff
@@ -1,25 +1,17 @@
...
-    updatePendingInvoicesByWalletId({
-      walletId,
-      logger,
-    }),
-    updatePendingPaymentsByWalletId({
-      walletId,
-      logger,
-    }),
-  ])
+  const updatePaymentsResult = await updatePendingPaymentsByWalletId({
```
nicolasburtey, line 11: note: I think we could move this in a hourly cronjob. Eventually this fallback method is really only used when trigger miss an event. this would be more one step to make the getBalanceForWallet totally independant of lnd

nicolasburtey: note for future PR

src/app/wallets/update-pending-invoices.ts
```diff
@@ -238,3 +217,64 @@ const updatePendingInvoice = async ({
...
+  }
+
+  let heldForMsg = ""
+  if (lnInvoiceLookup.heldAt) {
+    heldForMsg = `for ${elapsedSinceTimestamp(lnInvoiceLookup.heldAt)}s `
+  }
+  pendingInvoiceLogger.error(
+    { lnInvoiceLookup },
+    `invoice has been held ${heldForMsg}and is now been cancelled`,
+  )
```
nicolasburtey, line 269: I think this is what me might want?

```suggestion
  if (lnInvoiceLookup.heldAt) {
    heldForMsg = `for ${elapsedSinceTimestamp(lnInvoiceLookup.heldAt)}s `
      pendingInvoiceLogger.error(
    { lnInvoiceLookup },
    `invoice has been held ${heldForMsg}and is now been cancelled`,
  )
  }
```

src/servers/trigger.ts
```diff
@@ -181,7 +181,7 @@ export const onchainBlockEventHandler = async (height: number) => {
...
   logger.info({ invoice }, "invoiceUpdateEventHandler")
 
-  if (!invoice.is_confirmed) {
+  if (!invoice.is_held) {
```
nicolasburtey, line 184: maybe this if branch as an early return is unnecessary/a duplicate?

we are doing it also in https://github.com/GaloyMoney/galoy/pull/1395/files#diff-f02445547836eff187be3a4b54e41d92b4223dd7c304847b688ef3016eec317fR108

nicolasburtey: I mean un unnecessary optimization that make the flow more complex

nicolasburtey (approved):
I have some comments but I feel the PR can be merged now and we can address those comments/ adding e2e test later.

src/app/wallets/update-pending-invoices.ts
vindard: Oh no, we want that error log to happen whether or not there is a `lnInvoiceLookup.heldAt` value (see [here](https://github.com/GaloyMoney/galoy/pull/1188/files#diff-f02445547836eff187be3a4b54e41d92b4223dd7c304847b688ef3016eec317fR237)).

This conditional is purely for constructing the error log message, which is why I formatted the message to work even if there isn't a value for `heldAt`

It would just look like:
- `invoice has been held and is now been cancelled` without a value
- `invoice has been held for 100s and is now been cancelled` with a value

src/servers/trigger.ts
vindard: You mean we should remove these 2 lines from the trigger?

If we do, it would mean unnecessarily doing 2 external calls until we get to that line again:
- invoices collection mongo lookup
- lnd invoice lookup

vindard: If we decide to do this, it may be better as a separate PR so we can isolate any changed effects on this flow

src/domain/wallet-invoices/wallet-invoice-builder.ts
```diff
@@ -2,17 +2,37 @@ import { invoiceExpirationForCurrency } from "@domain/bitcoin/lightning"
 import { checkedToBtcPaymentAmount, checkedToUsdPaymentAmount } from "@domain/payments"
 import { WalletCurrency, ZERO_SATS } from "@domain/shared"
 
+import { InvalidWalletInvoiceBuilderStateError } from "./errors"
+
 export const WalletInvoiceBuilder = (
   config: WalletInvoiceBuilderConfig,
 ): WalletInvoiceBuilder => {
+  const withSecretAndHash = ({
```
bodymindarts, line 10: I'm not sure we need to expose this step - can the `getSecretAndPaymentHash` function be called from within the `WalletInvoiceBuilder`? Similar to how we call `LnFees` from the `PaymentFlowBuilder`? Its information that is generated within our app - not user input.
Getting rid of that step would remove some code - do you see a downside?

vindard: Good point, made the change here: https://github.com/GaloyMoney/galoy/pull/1395/commits/641e7113456c1a51e57f65318f98c4f226c30e89

bodymindarts (approved):

PR #1396, opened 2022-06-24
feat: loopout automation

ntheile:
This PR adds the ability to do loop outs. 

New Features:
- loopd grpc service 
- fully testable regtest environment for swaps (loopserver, loopd)
- trigger.ts server to listen for swap events
  - on success records swap fees to the ledger (service fee, miner fee, routing fee) 
- cron.ts server to run swap outs on an interval

Quickstart - To test swap out locally in your regtest env
----------------------------
```bash
# 1. Start the loop server
make start-loopd
# 2. Monitor loop outs
loopd1_id=$(docker ps -q -f name="loopd1-1")
docker exec -it $loopd1_id loop -n regtest monitor
# 3. Loop out integration test (open a new terminal)
TEST="swap-out" make integration
# 4. Mine a block (or a few) to finish the loop out
make mine-block
make mine-block
make mine-block
# 5. Successful result should look like this: LOOP_OUT INITIATED => LOOP_OUT PREIMAGE_REVEALED => LOOP_OUT SUCCESS
```

Extra documentation in [swap.md](https://github.com/GaloyMoney/galoy/blob/07ac286ff6a08dd02c1a1dc9f6403afced06dc94/docs/swap.md)


nicolasburtey:
@ntheile a key part that you'll need to add is the routing/loop server fees. this is similar to what we're doing for the miner fees in the [rebalance function](https://github.com/GaloyMoney/galoy/blob/889b5fe0d2ae9f83fb316369120acf06c11a64cb/src/app/cold-storage/rebalance-to-cold-wallet.ts#L73)

because this money is spent, and we use double entry book-keeping, we want to make sure the liabilities (money owe to users) match the assets/utxo, and loop out means there is less assets overall and this should be accounted for

Makefile
```diff
@@ -108,3 +112,7 @@ create-tmp-env-ci:
 # 16 is exit code for critical https://classic.yarnpkg.com/lang/en/docs/cli/audit
 audit:
 	bash -c 'yarn audit --level critical; [[ $$? -ge 16 ]] && exit 1 || exit 0'
+
+mine:
+	docker exec -it galoy-bitcoind-1 /bin/sh -c 'ADDR=$$(bitcoin-cli getnewaddress "") && bitcoin-cli generatetoaddress 6 $$ADDR '
```
dolcalmi, line 117: this does not work if you clone the repo in a different folder name. if we want to keep this in make file then you have to do something like 
```
container_id=$(docker ps -q -f status=running -f name="bitcoind")
docker exec $container_id bitcoin-cli listwallets
```

Makefile
```diff
@@ -23,6 +23,10 @@ start-cron: start-deps
 	. ./.envrc && yarn tsnd --files -r tsconfig-paths/register -r src/services/tracing.ts \
 		src/servers/cron.ts | yarn pino-pretty -c -l
 
+start-loop:
+	chmod +x ./scripts/start-loop.sh
```
dolcalmi, line 27: you can avoid this pushing the file with the correct permissions

default.yaml
```diff
@@ -167,6 +167,11 @@ onChainWallet:
   scanDepthOutgoing: 2
   scanDepthChannelUpdate: 8
 
+loop:
+  maxOutboundLiquidityBalance: 2835916 # 0.03 ($600)
```
dolcalmi, line 171: probably while you are testing but please remove unnecessary comments or please add useful comments

dev/.gitignore
```diff
@@ -1 +1,2 @@
 *.macaroon
+lnd-outside-1
```
dolcalmi, line 2: not sure to understand this

docker-compose.yml
```diff
@@ -111,6 +111,47 @@ services:
...
+      - ${PWD}/dev/lnd/lnd1.macaroon:/root/.lnd/data/chain/bitcoin/regtest/lnd1.macaroon
+    depends_on:
+      - lnd1
+    ports:
+      - "8081:8081"
+    command: 
+      - "loopd"
+      - "--network=regtest"
+      - "--debuglevel=debug"
+      - "--server.host=galoy-loopserver-1:11009"
```
dolcalmi, line 148: you should not use host names dependent on clone folder

scripts/start-loop.sh
```diff
@@ -0,0 +1,39 @@
...
+# remove old macaroons
+rm -rf ./dev/lnd/lnd-outside-1
+rm -rf ./dev/lnd/loop.macaroon
+
+# stop existing docker
+docker compose stop loopserver
+docker compose stop loopclient
+
+# copy macaroons from outside node (for use by mock loopserer)
+docker cp galoy-lnd-outside-1-1:/root/.lnd ./dev/lnd/lnd-outside-1
```
dolcalmi, line 12: same issue here

src/services/loop/loop-rest-client.ts
```diff
@@ -0,0 +1,125 @@
...
+import axios, { AxiosResponse } from "axios"
+
+import { getLoopConfig } from "@config"
+
+// @todo - get from yaml or .envrc
+const loopMacaroon = convertMacaroonToHexString(
+  fs.readFileSync("./dev/lnd/loop.macaroon"),
+)
+const cert = fs.readFileSync("./dev/lnd/tls.cert")
+const key = fs.readFileSync("./dev/lnd/tls.key")
```
dolcalmi, line 14: this will only work in local environments

src/services/loop/loop-rest-client.ts
```diff
@@ -0,0 +1,125 @@
...
+
+export const loopRestClient = {
```
dolcalmi, line 16: Please implement an interface, you can use as example cache service interface (it uses local and redis with the same interface). I think in this specific case probably will include other services in the future.

nicolasburtey: indeed, [bolt](https://boltz.exchange/) could be added as well in a future PR

ntheile: yeah i need to look at how "Ride the lightning" does this because they use both Loop and Boltz for swaps https://github.com/Ride-The-Lightning/RTL/tree/44412d357e205bbe79d14d394889851392d140c1/src/app/shared/components/services

src/services/loop/loop-rest-client.ts
ntheile: yep still WIP

src/services/loop/loop-rest-client.ts
ntheile: Loop and Boltz have slightly different api's and vocabulary but I can come up with a common interface:

Loop Service (RTL)
https://github.com/Ride-The-Lightning/RTL/blob/44412d357e205bbe79d14d394889851392d140c1/src/app/shared/services/loop.service.ts

Boltz Service (RTL)
https://github.com/Ride-The-Lightning/RTL/blob/44412d357e205bbe79d14d394889851392d140c1/src/app/shared/services/boltz.service.ts

Makefile
ntheile: fixed

Makefile
ntheile: fixed

src/services/loop/readme.md
```diff
@@ -1,19 +1,46 @@
...
 
-A successful loop server start returns this
+2. Configure the loop params in `default.yaml`
 
+```yaml
+loop:
+  maxOutboundLiquidityBalance: 2835916 # 0.03 ($600)
+  loopUrl: "https://localhost:8081"
+  loopoutAmount: 945305 # 0.009 ($200)
+  loopProvider: "@services/loop/providers/lightning-labs/loop-swap-provider" # or another custom provider, like Boltz
```
nicolasburtey, line 16: I don't think when we have 2 swap providers that we will want to use one or the other. we should dynamically use both based on which has lowest fee or which is online. 

I know it's a hard problem to solve because we don't know in advance the routing fees.

default.yaml
```diff
@@ -167,6 +167,12 @@ onChainWallet:
   scanDepthOutgoing: 2
   scanDepthChannelUpdate: 8
 
+swap:
+  minOnChainBalance: 20000000 # 0.2
```
nicolasburtey, line 171: I would use 30,000,000 as the default, and swap amount of 25,000,000

docker-compose.yml
```diff
@@ -130,4 +172,4 @@ services:
     env_file:
       - ${TMP_ENV_CI:-.env.ci}
     volumes:
-      - ./:/repo
+      - ./:/repo
```
nicolasburtey, line 175: your IDE may be mis-configured

scripts/start-loop.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
```
nicolasburtey, line 1: should probably more in `dev` than `scripts`?

src/app/swap/swap-out.ts
```diff
@@ -0,0 +1,24 @@
+import { BTC_NETWORK } from "@config"
+import { TxDecoder } from "@domain/bitcoin/onchain"
+import { SwapServiceError } from "@domain/swap/errors"
+import { SwapOutResult } from "@domain/swap/index.types"
+import { SwapService } from "@services/swap"
+import { OnChainService } from "@services/lnd/onchain-service"
+
+export const swapOut = async ({
+  amount,
```
nicolasburtey, line 9: when there is just argument we typically don't use an object, but just pass the scalar directly (ie :`(amount)` instead of `({amount})`)

src/config/schema.ts
```diff
@@ -304,6 +304,19 @@ export const configSchema = {
...
     },
+    swap: {
+      type: "object",
+      properties: {
+        minOnChainBalance: { type: "number" },
+        swapUrl: { type: "string" },
+        swapOutAmount: { type: "number" },
+        swapProviders: {
+          type: "array",
+          items: { type: "string" },
```
nicolasburtey, line 315: you can enforce which string can be passed here with an enum

src/config/schema.ts
```diff
@@ -304,6 +304,19 @@ export const configSchema = {
       ],
       additionalProperties: false,
     },
+    swap: {
+      type: "object",
+      properties: {
+        minOnChainBalance: { type: "number" },
```
nicolasburtey, line 310: should be an integer I think, not a number that could be a float

src/app/index.ts
```diff
@@ -18,6 +19,7 @@ const allFunctions = {
   Users: { ...UsersMod },
   Wallets: { ...WalletsMod },
   Payments: { ...PaymentsMod },
+  Swap: { ...SwapMod },
```
nicolasburtey, line 22: wondering if this should be an app at the root function. I see that more as a service. wdyt @dolcalmi @vindard @bodymindarts ?

.envrc
```diff
@@ -24,11 +25,29 @@ fetch_macaroon() {
...
+  if [ ! -z "${loop_container_id}" ]; then
+    docker cp $loop_container_id:/root/.loop/$NETWORK/loop.macaroon dev/lnd/loop.macaroon 2> /dev/null
+    base64 dev/lnd/loop.macaroon | tr -d '\n\r'
+  fi
+}
+
+fetch_loop_tls() {
+  local loop_container_id=$(docker ps -q -f status=running -f name="loopd")
+  if [ ! -z "${loop_container_id}" ]; then
+    docker cp "$loop_container_id:/root/.loop/regtest/tls.cert" "dev/lnd/loop-tls.cert" 2> /dev/null
```
krtk6160, line 39:
```suggestion
    docker cp "$loop_container_id:/root/.loop/$NETWORK/tls.cert" "dev/lnd/loop-tls.cert" 2> /dev/null
```

.gitignore
```diff
@@ -25,4 +25,4 @@ public/*
 
 coverage
 lib
-artifacts
+artifacts
```
nicolasburtey, line 28: our codebase required line return to avoid those dummy diff (probably a settings in your IDE to set up)

Makefile
```diff
@@ -108,3 +111,8 @@ create-tmp-env-ci:
 # 16 is exit code for critical https://classic.yarnpkg.com/lang/en/docs/cli/audit
 audit:
 	bash -c 'yarn audit --level critical; [[ $$? -ge 16 ]] && exit 1 || exit 0'
+
+mine:
```
nicolasburtey, line 115: when reading mine, I thought "this is mine" and was confusing about the action - versus "mining". maybe `mine-blocks` could be more explicit?

dev/scripts/start-loop.sh
```diff
@@ -0,0 +1,45 @@
...
+# test loop rest api
+curl -k \
+    --request GET \
+    --url     https://localhost:8081/v1/loop/out/quote/500000 \
+    --header  'Content-Type: application/json' \
+    --header  "Grpc-Metadata-macaroon: $LOOP_MACAROON_HEXSTR" \
+    --verbose
+
+# reload env vars
+direnv reload
```
nicolasburtey, line 45: missing end of line

src/app/swap/swap-listener.ts
```diff
@@ -0,0 +1,45 @@
...
+    // TODO log parsedSwapData.message
+    if (type === SwapType.SWAP_OUT && totalFees > 0) {
+      const swapFeeMetadata: SwapFeeLedgerMetadata = {
+        swapId: swapStatus.parsedSwapData.id,
+        swapAmount: toSats(swapStatus.parsedSwapData.amt),
+        htlcAddress: swapStatus.parsedSwapData.htlcAddress,
+        onchainMinerFee,
+        offchainRoutingFee,
+        serviceProviderFee,
+        serviceProvider: "LOOP",
```
nicolasburtey, line 24: we don't typically use all uppercase in our code. I think `loop` would be more appropriate.

src/app/swap/swap-listener.ts
```diff
@@ -0,0 +1,45 @@
...
+        currency: WalletCurrency.Btc,
+        type: LedgerTransactionType.Fee,
+        pending: false,
+      }
+      recordSwapFeeToLedger(swapFeeMetadata)
```
nicolasburtey, line 29: `await`?

maybe no call depend on this but this feels like a function that should not be synchronous.

src/app/swap/swap-listener.ts
nicolasburtey: we do for some `const` actually... nevermind.

src/app/swap/swap-out.ts
```diff
@@ -0,0 +1,69 @@
...
+    currentOnChainHotWalletBalance: onChainBalance,
+    minOnChainHotWalletBalanceConfig,
+    currentOutboundLiquidityBalance: outbound,
+    minOutboundLiquidityBalance,
+  })
+  const isOnChainWalletDepleted = swapChecker.isOnChainWalletDepleted()
+  const isOutboundLiquidityDepleted = swapChecker.isOutboundLiquidityDepleted()
+  logger.info(`SwapApp: isOnChainWalletDepleted: ${isOnChainWalletDepleted},
+    isOutboundLiquidityDepleted: ${isOutboundLiquidityDepleted}`)
+
```
nicolasburtey, line 40: our logger is a json logger. this make it easier to parse in the backend. see https://getpino.io/#/

src/app/swap/swap-out.ts
```diff
@@ -0,0 +1,69 @@
...
+  const swapChecker = SwapOutChecker({
+    currentOnChainHotWalletBalance: onChainBalance,
+    minOnChainHotWalletBalanceConfig,
+    currentOutboundLiquidityBalance: outbound,
+    minOutboundLiquidityBalance,
+  })
+  const isOnChainWalletDepleted = swapChecker.isOnChainWalletDepleted()
+  const isOutboundLiquidityDepleted = swapChecker.isOutboundLiquidityDepleted()
+  logger.info(`SwapApp: isOnChainWalletDepleted: ${isOnChainWalletDepleted},
+    isOutboundLiquidityDepleted: ${isOutboundLiquidityDepleted}`)
```
nicolasburtey, line 39:
```suggestion
  logger.info({isOnChainWalletDepleted, isOutboundLiquidityDepleted }, "wallet depletion status")
```

nicolasburtey: our logger is a json logger. this make it easier to parse in the backend. see https://getpino.io/#/

src/app/swap/swap-out.ts
```diff
@@ -0,0 +1,69 @@
...
+  addAttributesToCurrentSpan({
+    "swap.isOutboundLiquidityDepleted": isOutboundLiquidityDepleted,
+    "swap.isOnChainWalletDepleted": isOnChainWalletDepleted,
+  })
+
+  if (isOnChainWalletDepleted && isOutboundLiquidityDepleted) {
+    const swapResult = await swapService.swapOut(toSats(amount))
+    return swapResult
+  } else {
+    return new SwapServiceError("No Swap Out needed")
```
nicolasburtey, line 50: is it an error? 

doesn't feel so because SwapOutChecker is part of this function. 

I think the function should return gracefully

src/app/swap/swap-out.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+
+async function getChannelLiquidityBalance(): Promise<Satoshis | Error> {
```
nicolasburtey, line 54: I'm not sure about the utility of this function (see comment above), but `getOffchainBalance` could be more descriptive.

nicolasburtey: also this function doesn't feel very tied to the swap service necessarily

src/domain/swap/index.ts
```diff
@@ -0,0 +1,45 @@
...
+  currentOutboundLiquidityBalance,
+  minOutboundLiquidityBalance,
+}) => {
+  const isOnChainWalletDepleted = () => {
+    if (currentOnChainHotWalletBalance < minOnChainHotWalletBalanceConfig) {
+      return true
+    } else {
+      return false
+    }
+  }
```
nicolasburtey, line 14:
```suggestion
  const isOnChainWalletDepleted = () => 
currentOnChainHotWalletBalance < minOnChainHotWalletBalanceConfig
```

src/domain/swap/index.ts
```diff
@@ -0,0 +1,45 @@
...
+  return {
+    isOnChainWalletDepleted,
+    isOutboundLiquidityDepleted,
+  }
+}
+
+export enum SwapProvider {
+  LOOP = "LOOP",
+  PEERSWAP = "PEERSWAP",
```
nicolasburtey, line 30:
```suggestion
```

src/servers/cron.ts
```diff
@@ -48,6 +49,13 @@ const main = async () => {
     if (result instanceof Error) throw result
   }
 
+  const swapOutJob = async () => {
+    const amount = toSats(getSwapConfig().swapOutAmount)
```
nicolasburtey, line 53: I think `getSwapConfig().swapOutAmount` should already returning the correct type

src/services/swap/providers/lightning-labs/loop-rest-client.ts
```diff
@@ -0,0 +1,134 @@
...
+
+import { getSwapConfig } from "@config"
+
+const loopMacaroon = process.env.LOOP_MACAROON
+  ? convertMacaroonToHexString(Buffer.from(process.env.LOOP_MACAROON, "base64"))
+  : ""
+
+const cert = process.env.LND1_TLS ? Buffer.from(process.env.LND1_TLS, "base64") : ""
+
+const key = process.env.LND1_TLS_KEY
```
nicolasburtey, line 14: value fetched from `process.env` should be in a `config` file (where the other `process.env` are)

src/services/swap/providers/lightning-labs/loop-rest-client.ts
```diff
@@ -0,0 +1,134 @@
...
+  loopOut,
+  swapStatus,
+  swapStatusAll,
+  liquidityParams,
+  loopOutTerms,
+  loopOutQuote,
+  convertMacaroonToHexString,
+}
+
+function loopClient() {
```
nicolasburtey, line 30: we typically don't use `function abc () {}` but `const abc = () => {}`

src/services/swap/providers/lightning-labs/loop-rest-client.ts
```diff
@@ -0,0 +1,134 @@
...
+async function swapStatus(swapId) {
+  let resp: AxiosResponse
+  try {
+    // https://lightning.engineering/loopapi/#v1-loop-swap
+    // The swap identifier which currently is the hash that locks the HTLCs.
+    // When using REST, this field must be encoded as URL safe base64.
+    const id = swapId.replaceAll("+", "-").replaceAll("/", "_")
+    resp = await loopClient().get(`${loopRestEndpoint}/v1/loop/swap/${id}`)
+  } catch (e) {
+    resp = e
```
nicolasburtey, line 86: we should follow the error pattern we have in the rest of the code, where we return a dedicated error object

src/services/swap/providers/lightning-labs/loop-rest-client.ts
```diff
@@ -0,0 +1,134 @@
...
+}
+
+async function liquidityParams() {
+  let resp: AxiosResponse
+  try {
+    resp = await loopClient().get(`${loopRestEndpoint}/v1/liquidity/params`)
+  } catch (e) {
+    resp = e
+  }
+  console.log(resp.data)
```
nicolasburtey, line 110: we don't use `console.log` but a (pino) `logger` instead

src/services/swap/providers/lightning-labs/loop-service.ts
```diff
@@ -0,0 +1,172 @@
...
+        } catch (e) {
+          state = SwapState.FAILED
+        }
+        let message
+        try {
+          const failureReason = data.getFailureReason()
+          message = Object.keys(FailureReason).find(
+            (key) => FailureReason[key] === failureReason,
+          )
+        } catch (e) {}
```
nicolasburtey, line 133: we should typically not silence errors

src/services/swap/swap-service.ts
```diff
@@ -0,0 +1,39 @@
...
+      return resp
+    } catch (e) {
+      return e as SwapServiceError
+    }
+  }
+
+  const swapListener = () => {
+    const listener = LoopService().swapListener()
+    return listener
+  }
```
nicolasburtey, line 29:
```suggestion
  const swapListener = LoopService().swapListener
```

nicolasburtey (commented):
did a first pass on the PR and provide some early comments

default.yaml
```diff
@@ -167,6 +167,13 @@ onChainWallet:
   scanDepthOutgoing: 2
   scanDepthChannelUpdate: 8
 
+swap:
+  minOutboundLiquidityBalance: 30000000
```
nicolasburtey, line 171: if there is a single parameter, shouldn't it be `minOnChainHotWalletBalance` ?

src/app/swap/swap-out.ts
```diff
@@ -0,0 +1,79 @@
...
+    const lnd = activeNode.lnd
+    const { channel_balance, inbound } = await getChannelBalance({ lnd })
+    let outbound = 0
+    const inboundBal = inbound ? inbound : 0
+    if (inbound) {
+      outbound = channel_balance - inboundBal
+    }
+    return toSats(outbound)
+  } catch (err) {
+    return new Error(err)
```
nicolasburtey, line 77: app should return a specialized error (extended), not a generic one.

default.yaml
ntheile: yeah, spoke to Justin on Friday to clarify what he meant by that. I have a lot of fixes coming on Monday.

src/app/swap/swap-out.ts
```diff
@@ -1,52 +1,48 @@
...
-import { getChannelBalance } from "lightning"
 import { addAttributesToCurrentSpan } from "@services/tracing"
+import { getOffChainChannelBalances } from "@app/lightning"
 
 const logger = baseLogger.child({ module: "swap" })
 
-export const swapOut = async (
-  amount: Satoshis,
-): Promise<SwapOutResult | SwapServiceError | null> => {
+export const swapOut = async (): Promise<SwapOutResult | SwapServiceError | null> => {
```
nicolasburtey, line 14: we are trying to avoid using null in our codename (there is still many of them from the time we didn't have this rules). we took that from Typescript codebase

ntheile: fixed with NoSwapAction type https://github.com/GaloyMoney/galoy/pull/1396/commits/203a2497f63b8df3f0e60b86c014182b54073831

default.yaml
ntheile: fixed with https://github.com/GaloyMoney/galoy/pull/1396/commits/8c464b23670f31ff6d4866e93a16276eaadd1b63#diff-94d4d5bb5c048793536625edb8adffb17e89b5f3821cec6d1c7c337578cda8e4R171

we also check to make to make sure we have enough outbound `hasEnoughOutboundLiquidity` src/domain/swap/swap-out-checker.ts

```typescript
const hasEnoughOutboundLiquidity = currentOutboundLiquidityBalance > swapOutAmount
```

test/integration/app/swap/swap-out.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+      const swapOutChecker = SwapOutChecker({
+        currentOnChainHotWalletBalance: onChain,
+        minOnChainHotWalletBalanceConfig,
+        currentOutboundLiquidityBalance: 50000,
+        minOutboundLiquidityBalance: 60000,
+      })
+
+      if (
+        swapOutChecker.isOnChainWalletDepleted() &&
+        swapOutChecker.isOutboundLiquidityDepleted()
```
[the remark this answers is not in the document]
ntheile: removed and replaced with 

```
const hasEnoughOutboundLiquidity = currentOutboundLiquidityBalance > swapOutAmount
```

src/app/lightning/get-balances.ts
```diff
@@ -41,6 +41,17 @@ export const getOffChainBalance = async (): Promise<Satoshis | ApplicationError>
     },
   })
 
+export const getOffChainChannelBalances = async () =>
+  cache.getOrSet({
```
nicolasburtey, line 45: I don't think this make sense to cache this value

nicolasburtey: if lnd is offline, the swap will fail anyway.

ntheile: cache removed

Makefile
```diff
@@ -108,3 +111,8 @@ create-tmp-env-ci:
 # 16 is exit code for critical https://classic.yarnpkg.com/lang/en/docs/cli/audit
 audit:
 	bash -c 'yarn audit --level critical; [[ $$? -ge 16 ]] && exit 1 || exit 0'
+
+mine-block:
+	container_id=$$(docker ps -q -f status=running -f name="bitcoind"); \
+	docker exec -it "$$container_id" /bin/sh -c 'ADDR=$$(bitcoin-cli getnewaddress "") && bitcoin-cli generatetoaddress 6 $$ADDR '
+
```
nicolasburtey, line 118: IDE issue. need `\n`

default.yaml
```diff
@@ -167,6 +167,16 @@ onChainWallet:
   scanDepthOutgoing: 2
   scanDepthChannelUpdate: 8
 
+swap:
+  minOnChainHotWalletBalance: 30000000
+  swapOutAmount:  25000000
```
nicolasburtey, line 172:
```suggestion
  minOnChainHotWalletBalance: 50000000
  swapOutAmount:  50000000
```

probably better default

dev/.gitignore
```diff
@@ -1 +1,6 @@
...
+lnd1-loop-tls.cert
```
nicolasburtey, line 3: shouldn't this be `dev/lnd/` instead of `dev/`?

docs/swap.md
```diff
@@ -0,0 +1,183 @@
...
+}
+```
+
+(2) Configure the loop params in `default.yaml`
+---------------------------------------
+
+```yaml
+swap:
+  minOnChainHotWalletBalance: 30000000
+  swapOutAmount:  25000000
```
nicolasburtey, line 53:
```suggestion
  minOnChainHotWalletBalance: 75000000
  swapOutAmount:  50000000
```

docs/swap.md
```diff
@@ -0,0 +1,183 @@
...
+```
+sudo touch /var/yaml/custom.yaml
+sudo chmod 755 /var/yaml/custom.yaml
+```
+
+Add the following config (you might need to 'retry as sudo' on mac+vscode):
+```
+swap:
+  minOnChainHotWalletBalance: 350000
+  swapOutAmount:  250000
```
nicolasburtey, line 73:
```suggestion
  minOnChainHotWalletBalance: 75000000
  swapOutAmount:  50000000
```

src/app/index.ts
nicolasburtey: ping @dolcalmi @vindard @bodymindarts

src/app/swap/swap-listener.ts
```diff
@@ -0,0 +1,61 @@
...
+      }
+      addAttributesToCurrentSpan({
+        "swap.success": JSON.stringify(swapFeeMetadata),
+      })
+      await recordSwapFeeToLedger(swapFeeMetadata)
+    }
+    if (state === SwapState.FAILED) {
+      addAttributesToCurrentSpan({
+        "swap.error": JSON.stringify({
+          swapiId: swapStatus.parsedSwapData.id,
```
nicolasburtey, line 38:
```suggestion
          swapId: swapStatus.parsedSwapData.id,
```

src/app/swap/swap-listener.ts
```diff
@@ -0,0 +1,61 @@
...
+          amt: swapStatus.parsedSwapData.amt.toString(),
+        }),
+      })
+    }
+  }
+  return true
+}
+
+// TODO handle
+export function handleSwapOutFailure(data) {
```
nicolasburtey, line 49: are we not already handling the failure here?

https://github.com/GaloyMoney/galoy/pull/1396/files#diff-0e844e1434164c814e3fb21ae1cb0649254d367a58697369661b66a9587dd5b6R35

nicolasburtey: one things that would be useful is to have an alert on loop out failure, but this would be outside this repo.

src/services/ledger/index.types.d.ts
```diff
@@ -57,6 +57,19 @@ type LnReceiveLedgerMetadata = NonIntraledgerLedgerMetadata & {
...
+  swapAmount: Satoshis
+  swapId: string
+  htlcAddress: string
+  onchainMinerFee: Satoshis
+  offchainRoutingFee: Satoshis
+  serviceProviderFee: Satoshis
+  serviceProvider: string
+  currency: WalletCurrency
+  type: LedgerTransactionType
+  pending: boolean
```
nicolasburtey, line 70: I think this will always be false?

```suggestion
  pending: false
```

src/services/swap/swap-service.ts
```diff
@@ -0,0 +1,39 @@
...
+    try {
+      const isServiceUp = await loopService.healthCheck()
+      if (isServiceUp) return true
+      return false
+    } catch (e) {
+      return false
+    }
+  }
+
+  const swapOut = async (amount) => {
```
nicolasburtey, line 19: imo this function should get both `{ amount, swapDestAddress }` as parameters, instead of solely relying on `amount`.

swapDestAddress should be injected at the app level.

this would remove the dependency from this service to the `OnChainService`

src/services/swap/swap-utils.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  const getLoopService = ({ macaroon, tlsCert, grpcEndpoint }) => {
+    return LoopService({ macaroon, tlsCert, grpcEndpoint })
+  }
+
+  // logic to choose the correct onChain address for the swap out destination
+  // active lnd node that has ["onChain"] wallet
+  const getSwapDestAddress = async () => {
+    const onChainService = OnChainService(TxDecoder(BTC_NETWORK))
```
nicolasburtey, line 47: `BTC_NETWORK` should be injected in `LoopUtils` when it's been created, so there there is no dependency on `@config`

(I'll let @vindard and @dolcalmi  chime in on this as well)

src/services/swap/swap-utils.ts
```diff
@@ -0,0 +1,58 @@
...
+      }
+    }
+    return loopService
+  }
+
+  const getLoopService = ({ macaroon, tlsCert, grpcEndpoint }) => {
+    return LoopService({ macaroon, tlsCert, grpcEndpoint })
+  }
+
+  // logic to choose the correct onChain address for the swap out destination
```
nicolasburtey, line 44: this logic doesn't belong to the swap service.

src/servers/trigger.ts
```diff
@@ -279,6 +283,25 @@ const listenerOffchain = ({ lnd, pubkey }: { lnd: AuthenticatedLnd; pubkey: Pubk
   })
 }
 
+const listenerSwapMonitor = async () => {
+  try {
+    const swapService = SwapService()
+    const isSwapServerUp = await swapService.healthCheck()
+    if (isSwapServerUp) {
```
nicolasburtey, line 290: we'll need a retry if server is offline or an error is thrown

src/services/lnd/index.ts
```diff
@@ -87,6 +87,27 @@ export const LndService = (
     }
   }
 
+  const getChannelBalances = async (pubkey?: Pubkey) => {
+    try {
+      const lnd = pubkey ? getLndFromPubkey({ pubkey }) : defaultLnd
+      if (lnd instanceof Error) return lnd
+      const { channel_balance, inbound } = await getChannelBalance({ lnd })
+      let outbound = 0
+      const inboundBal = inbound ? inbound : 0
```
nicolasburtey, line 96: this would do the same?

```suggestion
      const inboundBal = inbound ?? 0
```

src/services/lnd/index.ts
```diff
@@ -87,6 +87,27 @@ export const LndService = (
     }
   }
 
+  const getChannelBalances = async (pubkey?: Pubkey) => {
```
nicolasburtey, line 90: I think the function name is not accurate. we're getting aggregate balance here, not balance per channel. 

Also getBalance typically return a Satoshis amount, here we are returning  really inbound/output. so maybe `getInboundOutbalanceBalance`?

nicolasburtey:
@ntheile rebase needed

src/services/swap/swap-utils.ts
ntheile: does it belong under /app/swap ?

src/services/swap/providers/lightning-labs/loop-service.ts
```diff
@@ -0,0 +1,191 @@
...
+
+import * as grpc from "@grpc/grpc-js"
+import { getSwapConfig } from "@config"
+import { SwapClientNotResponding, SwapServiceError } from "@domain/swap/errors"
+
+import { SwapState as SwapStateType } from "@domain/swap/index"
+
+import { SwapType as DomainSwapType } from "@domain/swap"
+
+import { SwapClientClient } from "./protos/loop_grpc_pb"
```
ntheile, line 12: @nicolasburtey 

Is there a way to exclude these generated proto files from complying with the "no implicit any rules"?

nicolasburtey: @dolcalmi may know more. I think you need to generate the associated definition file: 

ie: [src/services/dealer-price/proto/services/price/v1/price_service_grpc_pb.d.ts](https://github.com/GaloyMoney/galoy/blob/main/src/services/dealer-price/proto/services/price/v1/price_service_grpc_pb.d.ts)

src/services/lnd/index.ts
ntheile: renamed to getInboundOutboundBalance

src/services/swap/swap-utils.ts
ntheile: refactored and removed swap-utils

docs/swap.md
ntheile: I updated these values in the default.yaml. In this readme instruction I have the user configure a smaller value for the local custom config in `/var/yaml/custom.yaml` so that the integration tests run without running out of regtest bitcoin

.envrc
```diff
@@ -14,6 +15,29 @@ export LNDONCHAIN_MACAROON=""
 export MACAROONOUTSIDE1=$(cat dev/lnd/$NETWORK/lnd-outside-1.admin.macaroon.base64)
 export MACAROONOUTSIDE2=$(cat dev/lnd/$NETWORK/lnd-outside-2.admin.macaroon.base64)
 
+fetch_loop_macaroon() {
```
nicolasburtey, line 18: could those macaroon and tls be injected like @openoms has done for lnd?

default.yaml
```diff
@@ -176,6 +176,16 @@ onChainWallet:
   scanDepthOutgoing: 2
   scanDepthChannelUpdate: 8
 
+swap:
+  minOnChainHotWalletBalance: 75000000
+  swapOutAmount:  50000000
```
nicolasburtey, line 181:
```suggestion
  swapOutAmount:  500000000
```

src/app/lightning/get-balances.ts
```diff
@@ -41,6 +41,12 @@ export const getOffChainBalance = async (): Promise<Satoshis | ApplicationError>
     },
   })
 
+export const getOffChainChannelBalances = async () => {
+  const offChainService = LndService()
+  if (offChainService instanceof Error) return offChainService
+  return offChainService.getInboundOutboundBalance()
+}
+
```
nicolasburtey, line 49: is this function of any use at all? why not directly using `getInboundOutboundBalance()`?

src/app/swap/get-active-loopd.ts
```diff
@@ -0,0 +1,33 @@
...
+    default: {
+      return LND1_LOOP_CONFIG
+    }
+  }
+}
+
+export const LND1_LOOP_CONFIG: LoopdConfig = {
+  btcNetwork: BTC_NETWORK,
+  grpcEndpoint: getSwapConfig().lnd1loopRpcEndpoint,
+  tlsCert: process.env.LND1_LOOP_TLS ?? "",
```
nicolasburtey, line 24: this should be injected from `src/config/process.ts`

src/app/swap/get-active-loopd.ts
```diff
@@ -0,0 +1,33 @@
...
+  macaroon: process.env.LND1_LOOP_MACAROON ?? "",
```
nicolasburtey, line 25: this should be injected from src/config/process.ts

src/app/swap/get-active-loopd.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export const LND2_LOOP_CONFIG: LoopdConfig = {
+  btcNetwork: BTC_NETWORK,
+  grpcEndpoint: getSwapConfig().lnd2loopRpcEndpoint,
+  tlsCert: process.env.LND2_LOOP_TLS ?? "",
+  macaroon: process.env.LND2_LOOP_MACAROON ?? "",
```
nicolasburtey, line 32: this should be injected from src/config/process.ts

src/app/swap/get-active-loopd.ts
```diff
@@ -0,0 +1,33 @@
...
+  if (activeNode instanceof Error) throw SwapErrorNoActiveLoopdNode
+  switch (activeNode.macaroon) {
+    case process.env.LND1_MACAROON: {
+      return LND1_LOOP_CONFIG
+    }
+    case process.env.LND2_MACAROON: {
+      return LND2_LOOP_CONFIG
+    }
+    default: {
+      return LND1_LOOP_CONFIG
```
nicolasburtey, line 16: I think we should default in that case?

src/app/swap/get-active-loopd.ts
```diff
@@ -0,0 +1,33 @@
+import { getActiveOnchainLnd } from "@services/lnd/utils"
+import { SwapErrorNoActiveLoopdNode } from "@domain/swap/errors"
+import { BTC_NETWORK, getSwapConfig } from "@config"
+
+export const getActiveLoopd = (): LoopdConfig => {
+  const activeNode = getActiveOnchainLnd()
+  if (activeNode instanceof Error) throw SwapErrorNoActiveLoopdNode
+  switch (activeNode.macaroon) {
+    case process.env.LND1_MACAROON: {
```
nicolasburtey, line 9: this should be injected from src/config/process.ts

src/app/swap/get-active-loopd.ts
```diff
@@ -0,0 +1,33 @@
...
+      return LND1_LOOP_CONFIG
+    }
+    case process.env.LND2_MACAROON: {
```
nicolasburtey, line 12: this should be injected from src/config/process.ts

src/app/swap/get-active-loopd.ts
```diff
@@ -0,0 +1,33 @@
+import { getActiveOnchainLnd } from "@services/lnd/utils"
+import { SwapErrorNoActiveLoopdNode } from "@domain/swap/errors"
+import { BTC_NETWORK, getSwapConfig } from "@config"
+
+export const getActiveLoopd = (): LoopdConfig => {
+  const activeNode = getActiveOnchainLnd()
+  if (activeNode instanceof Error) throw SwapErrorNoActiveLoopdNode
+  switch (activeNode.macaroon) {
```
nicolasburtey, line 8: it's surprising to do the switch on the macaroon. 

this is the admin macaroon from lnd1/lnd2? not loop?

src/services/loopd/loop-service.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  const healthCheck = async (): Promise<boolean> => {
+    try {
+      const request = new QuoteRequest()
+      request.setAmt(500000)
+      const resp = await clientHealthCheck(request)
+      const fee = resp.getSwapFeeSat()
+      if (fee) return true
+    } catch (error) {
+      console.log(error)
```
nicolasburtey, line 50: we don't use `console.log` to print error on the console. 

we use the logger with an error message, or more and more opentelemetry so we can have alert on them

src/services/loopd/loop-service.ts
```diff
@@ -0,0 +1,196 @@
...
+  const clientSwapOut = util.promisify<LoopOutRequest, SwapResponse>(
+    swapClient.loopOut.bind(swapClient),
+  )
+
+  const healthCheck = async (): Promise<boolean> => {
+    try {
+      const request = new QuoteRequest()
+      request.setAmt(500000)
+      const resp = await clientHealthCheck(request)
+      const fee = resp.getSwapFeeSat()
```
nicolasburtey, line 47: does this result in a request to the LL server? 

maybe in that a simpler request that doesn't require a query to a server outside our cluster would be preferable for a healthcheck?

src/app/swap/get-active-loopd.ts
ntheile: I basically need to know the name of the active onchain LND instance, specifically if its called LND1 or LND2. When I call `getActiveOnChainLnd()` it returns data about the macaroons and tls certs but it does not specifically tell me its name (LND1 or LND2) . This is what it return:

Screen Shot 2022-08-31 at 9 56 12 PM

After I know the name of the LND instance I can get its corresponding LOOP config based on the naming convention `LND1_LOOP_CONFIG` or `LND2_LOOP_CONFIG`. Any ideas on how to get the LND instance name?

ntheile: added name to LNDParams

src/app/swap/get-active-loopd.ts
nicolasburtey: I meant: I think we should error in the default case

ntheile: oic, will add the error

src/app/lightning/get-balances.ts
ntheile: remove this function

src/services/loopd/loop-service.ts
ntheile: yes, i updated the health check to use a simpler call to an internal token request service

src/services/loopd/loop-service.ts
```diff
@@ -0,0 +1,248 @@
...
+          state: state as SwapStateType,
+          message: message ? message : "",
+          swapType,
+        }
+        data.parsedSwapData = parsedSwapData
+        listener.resume()
+      })
+      return listener
+    } catch (error) {
+      throw new SwapServiceError(error)
```
nicolasburtey, line 166: I think this should be a `UnknownSwapServiceError` (@vindard ?)

src/services/loopd/loop-service.ts
```diff
@@ -0,0 +1,248 @@
...
+      return {
+        cltvDelta: resp.getCltvDelta(),
+        confTarget: resp.getConfTarget(),
+        htlcSweepFeeSat: resp.getHtlcSweepFeeSat(),
+        prepayAmtSat: resp.getPrepayAmtSat(),
+        swapFeeSat: resp.getSwapFeeSat(),
+        swapPaymentDest: resp.getSwapPaymentDest(),
+      } as SwapOutQuoteResult
+    } catch (error) {
+      return new SwapServiceError(error)
```
nicolasburtey, line 186: I think this should be a `UnknownSwapServiceError` (@vindard ?)

src/services/loopd/loop-service.ts
```diff
@@ -0,0 +1,248 @@
...
+      const request = new TermsRequest()
+      const resp = await clientSwapOutTerms(request)
+      return {
+        maxCltvDelta: resp.getMaxCltvDelta(),
+        maxSwapAmount: resp.getMaxSwapAmount(),
+        minCltvDelta: resp.getMinCltvDelta(),
+        minSwapAmount: resp.getMinSwapAmount(),
+      } as SwapOutTermsResult
+    } catch (error) {
+      return new SwapServiceError(error)
```
nicolasburtey, line 201: I think this should be a `UnknownSwapServiceError` (@vindard ?)

.envrc
ntheile: yep! done! https://github.com/GaloyMoney/galoy/pull/1396/commits/fabb350da08f693783af28701df02f526aba8e21

src/config/process.ts
```diff
@@ -101,3 +101,8 @@ export const LND_HEALTH_REFRESH_TIME_MS = parseInt(
   process.env.LND_HEALTH_REFRESH_TIME_MS || "20000",
   10,
 )
+
+export const LND1_LOOP_TLS = process.env.LND1_LOOP_TLS ?? ""
```
nicolasburtey, line 105: you should error if this is not present but necessary

something like:
```
const getLnd1LoopTls = () => {
   if (!process.env.LND1_LOOP_TLS) throw new Config Error
    return process.env.LND1_LOOP_TLS
}
```

ntheile: fixed in https://github.com/GaloyMoney/galoy/pull/1396/commits/146cb8791c7fc8ac731aad10585fb98eb3486ab1

.envrc
```diff
@@ -3,6 +3,7 @@ export NETWORK=regtest
 export NODE_ENV=development
 
 export LND1_TLS=$(cat dev/lnd/tls.cert.base64)
+export LND1_TLS_KEY=$(base64 dev/lnd/tls.key | tr -d '\n\r')
```
nicolasburtey, line 6: why is this needed?

nicolasburtey: note: seems below the value is already converted in `.base64` in the script so a simple `cat` is enough

dev/bin/start-loopd.sh
```diff
@@ -0,0 +1,54 @@
...
+sleep 5
+
+# test the lnd1-loop REST API with a quote request
+echo "loop1 results"
+LOOP1_MACAROON_HEXSTR=$(cat dev/lnd/loop/$NETWORK/loopd1-1.loop.macaroon | xxd -p |  awk '{print}' ORS='')
+# echo $LOOP1_MACAROON_HEXSTR
+# test loop rest api
+curl -k \
+    --request GET \
+    --url     https://localhost:8081/v1/loop/out/quote/500000 \
```
nicolasburtey, line 41: using the non https may make it that you don't need to use `-k`?

docker-compose.yml
```diff
@@ -156,6 +156,100 @@ services:
...
       cp /root/.lnd/admin.macaroon /root/.lnd/data/chain/bitcoin/regtest/admin.macaroon
       /bin/lnd
+  loopserver:
+    image: lightninglabs/loopserver:v0.9.52-beta
+    restart: unless-stopped
+    volumes:
+      - ${PWD}/dev/lnd/lnd.conf:/root/.lnd/lnd.conf
+      - ${PWD}/dev/lnd/tls.key:/root/.lnd/tls.key
+      - ${PWD}/dev/lnd/tls.cert:/root/.lnd/tls.cert
+      - ${PWD}/dev/lnd/regtest/lnd-outside-1.admin.macaroon:/root/.lnd/data/chain/bitcoin/regtest/admin.macaroon
```
nicolasburtey, line 166: can't you bind the directory instead?

docker-compose.yml
```diff
@@ -156,6 +156,100 @@ services:
...
+      - ${PWD}/dev/lnd/regtest/lnd-outside-1.admin.macaroon:/root/.lnd/data/chain/bitcoin/regtest/readonly.macaroon
+      - ${PWD}/dev/lnd/regtest/lnd-outside-1.admin.macaroon:/root/.lnd/data/chain/bitcoin/regtest/invoices.macaroon
+      - ${PWD}/dev/lnd/regtest/lnd-outside-1.admin.macaroon:/root/.lnd/data/chain/bitcoin/regtest/signer.macaroon
+      - ${PWD}/dev/lnd/regtest/lnd-outside-1.admin.macaroon:/root/.lnd/data/chain/bitcoin/regtest/chainnotifier.macaroon
+      - ${PWD}/dev/lnd/regtest/lnd-outside-1.admin.macaroon:/root/.lnd/data/chain/bitcoin/regtest/router.macaroon
+      - ${PWD}/dev/lnd/regtest/lnd-outside-1.admin.macaroon:/root/.lnd/data/chain/bitcoin/regtest/walletkit.macaroon
+    depends_on:
+      - lnd-outside-1
+    ports:
+      - "11009:11009"
```
nicolasburtey, line 176: for some reason ports are included in the docker-compose override files (not sure why that is, @bodymindarts ?)

src/app/swap/get-active-loopd.ts
```diff
@@ -0,0 +1,36 @@
...
+    case "LND2": {
+      return LND2_LOOP_CONFIG
+    }
+    default: {
+      throw SwapErrorNoActiveLoopdNode
+    }
+  }
+}
+
+export const LND1_LOOP_CONFIG: LoopdConfig = {
```
nicolasburtey, line 22: does this need to be exported?
```suggestion
const LND1_LOOP_CONFIG: LoopdConfig = {
```

src/app/swap/swap-listener.ts
```diff
@@ -0,0 +1,73 @@
...
+  if (isSwapServerUp instanceof Error === false) {
+    const listener = swapService.swapListener()
+    listener.on("data", (response) => {
+      // const swapData = response.parsedSwapData
+      // logger.info({ swapData }, "listenerSwapMonitor")
+      handleSwapOutCompleted(response)
+    })
+  } else {
+    return SwapErrorNoActiveLoopdNode
+  }
```
nicolasburtey, line 22: the logic with === false is convoluted IMO
```suggestion
  if (isSwapServerUp instanceof Error) {
      return SwapErrorNoActiveLoopdNode
    })
  } else {
     const listener = swapService.swapListener()
    listener.on("data", (response) => {
      // const swapData = response.parsedSwapData
      // logger.info({ swapData }, "listenerSwapMonitor")
      handleSwapOutCompleted(response)

   }
```

src/app/swap/swap-listener.ts
```diff
@@ -0,0 +1,73 @@
...
+import { SwapErrorNoActiveLoopdNode } from "@domain/swap/errors"
+import { sha256 } from "@domain/bitcoin/lightning"
+
+export const startSwapMonitor = async (swapService: ISwapService) => {
+  const isSwapServerUp = await swapService.healthCheck()
+  if (isSwapServerUp instanceof Error === false) {
+    const listener = swapService.swapListener()
+    listener.on("data", (response) => {
+      // const swapData = response.parsedSwapData
+      // logger.info({ swapData }, "listenerSwapMonitor")
```
nicolasburtey, line 17: code commented needs to be removed

src/app/swap/swap-out.ts
```diff
@@ -0,0 +1,72 @@
...
+  const swapOutAmount = swapChecker.getSwapOutAmount({
+    currentOnChainHotWalletBalance: onChainBalance,
+    currentOutboundLiquidityBalance: outbound,
+  })
+  if (swapOutAmount instanceof Error) return swapOutAmount
+
+  if (swapOutAmount > 0) {
+    logger.info(
+      { swapOutAmount, activeLoopdConfig },
+      `Initiating swapout for ${swapOutAmount} sats`,
```
nicolasburtey, line 48: remove the variable in the message (second arg). it make search through log easier.

src/app/swap/swap-out.ts
```diff
@@ -0,0 +1,72 @@
...
+    minOnChainHotWalletBalanceConfig,
+    swapOutAmount: toSats(getSwapConfig().swapOutAmount),
+  })
+  const swapOutAmount = swapChecker.getSwapOutAmount({
+    currentOnChainHotWalletBalance: onChainBalance,
+    currentOutboundLiquidityBalance: outbound,
+  })
+  if (swapOutAmount instanceof Error) return swapOutAmount
+
+  if (swapOutAmount > 0) {
```
nicolasburtey, line 45: I would start with the use case where there is no swap amount, so you can return early and don' t need to be in a if loop.

ie: 

```
if (swapOutAmount === 0) {
    return new NoSwapAction()
}

...
(rest of the code with no more if else clause)
```

src/domain/swap/index.ts
```diff
@@ -0,0 +1,23 @@
+export * from "./swap-out-checker"
+
+export enum SwapProvider {
```
nicolasburtey, line 3: IIRC we're avoiding enum. is that accurate @vindard? do you remember why?

nicolasburtey: https://github.com/GaloyMoney/galoy/blob/64d67269dbc343035d6f36075e5e2b699f694b45/CONTRIBUTING.md#enums-via-object-constants

src/services/loopd/loop-service.ts
```diff
@@ -0,0 +1,255 @@
...
+      }
+      return new SwapServiceError(error)
+    }
+  }
+
+  const swapListener = function (): SwapClientReadableStream<SwapListenerResponse> {
+    try {
+      const request = new MonitorRequest()
+      const listener = swapClient.monitor(request)
+      // listener.on("data" | "status" | "end")
```
nicolasburtey, line 115: remove comment

src/services/loopd/loop-service.ts
```diff
@@ -0,0 +1,255 @@
...
+        try {
+          state = Object.keys(SwapState).find((key) => {
+            // eslint-disable-next-line
+            // @ts-ignore
+            return SwapState[key] === stateVal
+          })
+        } catch (e) {
+          state = SwapState.FAILED
+        }
+        let message
```
nicolasburtey, line 130: implicit any? not catch by the compiler?

src/domain/swap/index.types.d.ts
```diff
@@ -0,0 +1,85 @@
...
+  serverMessage: string
+}
+
+type SwapListenerResponse =
+  | SwapStatusResultWrapper
+  | (SwapStatusResult & SwapStatusResultWrapper)
+  | SwapServiceError
+
+type SwapStatusResultWrapper = {
+  parsedSwapData?: SwapStatusResult
```
nicolasburtey, line 35: does it make sense for this to be optional?

src/services/loopd/loop-service.ts
```diff
@@ -0,0 +1,255 @@
...
+        const stateVal = data.getState()
+        let state
+        try {
+          state = Object.keys(SwapState).find((key) => {
+            // eslint-disable-next-line
+            // @ts-ignore
+            return SwapState[key] === stateVal
+          })
+        } catch (e) {
+          state = SwapState.FAILED
```
nicolasburtey, line 128: is there open telemetry set up for those errors?

src/domain/swap/index.types.d.ts
```diff
@@ -0,0 +1,85 @@
...
+  swapId: string
+  swapIdBytes: string
+  htlcAddress: string
+  serverMessage: string
+}
+
+type SwapListenerResponse =
+  | SwapStatusResultWrapper
+  | (SwapStatusResult & SwapStatusResultWrapper)
+  | SwapServiceError
```
nicolasburtey, line 32: we don't type the error within the type typically but at the function level.

nicolasburtey: ie 
```
fn(): SwapListenerResponse | SwapServiceError {
...
}
```

src/services/loopd/loop-service.ts
```diff
@@ -0,0 +1,255 @@
...
+  const swapListener = function (): SwapClientReadableStream<SwapListenerResponse> {
+    try {
+      const request = new MonitorRequest()
+      const listener = swapClient.monitor(request)
+      // listener.on("data" | "status" | "end")
+      listener.on("data", (data: SwapStatus & SwapStatusResultWrapper) => {
+        listener.pause()
+        // parse data to our interface
+        const stateVal = data.getState()
+        let state
```
nicolasburtey, line 120: implicit any?

src/services/loopd/loop-service.ts
```diff
@@ -0,0 +1,255 @@
...
+        try {
+          state = Object.keys(SwapState).find((key) => {
```
nicolasburtey, line 122: I'm having a hard time following the logic

src/services/loopd/loop-service.ts
```diff
@@ -0,0 +1,255 @@
...
+      request.setAmt(amt)
+      const resp = await clientSwapOutQuote(request)
+      return {
+        cltvDelta: resp.getCltvDelta(),
+        confTarget: resp.getConfTarget(),
+        htlcSweepFeeSat: resp.getHtlcSweepFeeSat(),
+        prepayAmtSat: resp.getPrepayAmtSat(),
+        swapFeeSat: resp.getSwapFeeSat(),
+        swapPaymentDest: resp.getSwapPaymentDest(),
+      } as SwapOutQuoteResult
```
nicolasburtey, line 191:
```suggestion
      return {
        cltvDelta: resp.getCltvDelta(),
        confTarget: resp.getConfTarget(),
        htlcSweepFeeSat: resp.getHtlcSweepFeeSat(),
        prepayAmtSat: resp.getPrepayAmtSat(),
        swapFeeSat: resp.getSwapFeeSat(),
        swapPaymentDest: resp.getSwapPaymentDest(),
      }: SwapOutQuoteResult
```

src/services/loopd/loop-service.ts
```diff
@@ -0,0 +1,255 @@
...
+  const swapOutTerms = async (): Promise<SwapOutTermsResult | SwapServiceError> => {
+    try {
+      const request = new TermsRequest()
+      const resp = await clientSwapOutTerms(request)
+      return {
+        maxCltvDelta: resp.getMaxCltvDelta(),
+        maxSwapAmount: resp.getMaxSwapAmount(),
+        minCltvDelta: resp.getMinCltvDelta(),
+        minSwapAmount: resp.getMinSwapAmount(),
+      } as SwapOutTermsResult
```
nicolasburtey, line 206: avoid the `as`. as highlighted above.

src/services/loopd/loop-service.ts
```diff
@@ -0,0 +1,255 @@
...
+    )
+    try {
+      const client: ServiceClient = new SwapClientClient(
+        grpcEndpoint,
+        credentials,
+        grpcOptions,
+      )
+      return client
+    } catch (e) {
+      throw SwapClientNotResponding
```
nicolasburtey, line 241: missing new?

test/integration/app/swap/swap-out.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+          )
+        }
+      }
+    }
+  })
+
+  it("Swap out without enough funds returns an error", async () => {
+    const isSwapServerUp = await swapService.healthCheck()
+    if (isSwapServerUp instanceof Error === false) {
+      const swapResult = await swapService.swapOut({ amount: toSats(5000000000) })
```
nicolasburtey, line 97:
```suggestion
      const swapResult = await swapService.swapOut({ amount: toSats(5_000_000_000) })
```

test/integration/app/swap/swap-out.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+      if (swapResult instanceof SwapClientNotResponding) {
+        return
+      }
+      expect(swapResult).toBeInstanceOf(Error)
+    }
+  })
+
+  it("Swap out if on chain wallet is depleted returns a swap result", async () => {
+    const isSwapServerUp = await swapService.healthCheck()
+    if (isSwapServerUp instanceof Error === false) {
```
nicolasburtey, line 107: remove `=== false`

test/integration/app/swap/swap-out.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+        )
+      } else {
+        expect("No swap Needed").toEqual("No swap Needed")
+      }
+    }
+  })
+
+  it("Swap out quote return quote result", async () => {
+    const isSwapServerUp = await swapService.healthCheck()
+    if (isSwapServerUp instanceof Error === false) {
```
nicolasburtey, line 142: remove `=== false`

test/integration/app/swap/swap-out.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+        }),
+      )
+    } else {
+      expect("No swap Needed").toEqual("No swap Needed")
+    }
+  })
+
+  it("Swap out terms return terms result", async () => {
+    const isSwapServerUp = await swapService.healthCheck()
+    if (isSwapServerUp instanceof Error === false) {
```
nicolasburtey, line 157: remove `=== false`

nicolasburtey (commented):
re-skimmed through the PR, added more comments

dev/bin/start-loopd.sh
ntheile: loopd only exposes the https server

docker-compose.yml
ntheile: yeah i will have to modify openoms script `dev/bin/save-lnd-data.sh` . right now the script does not export all the required macaroons for loopserver (readonly, invoices, signer, chainnotifier, router and walletkit).

docker-compose.yml
ntheile: yes, i will move them to override. maybe to make it easier to override if we get a port conflict

src/app/swap/get-active-loopd.ts
ntheile: i use it in the swap integration test to specifically target the lnd1-loop node

src/domain/swap/index.ts
ntheile: cool. i fixed them according to the documentation

src/services/loopd/loop-service.ts
ntheile: I moved the parsing logic out to separate functions. it's a bit confusing because the auto generated proto code uses numbers as the values and the human readable name as the key (see code below) . so you need to parse using `Object.keys` and then map to our Swap Domain model

```typescript
export interface FailureReasonMap {
  FAILURE_REASON_NONE: 0;
  FAILURE_REASON_OFFCHAIN: 1;
  FAILURE_REASON_TIMEOUT: 2;
  FAILURE_REASON_SWEEP_TIMEOUT: 3;
  FAILURE_REASON_INSUFFICIENT_VALUE: 4;
  FAILURE_REASON_TEMPORARY: 5;
  FAILURE_REASON_INCORRECT_AMOUNT: 6;
}
```

src/services/loopd/loop-service.ts
ntheile: Yeah returning SwapState.Failed will be picked up by the swap-listener and wrap it in a span see https://github.com/GaloyMoney/galoy/blob/37052f0cbe3cb6ca663dec8e2f7158d86edf2b1a/src/app/swap/swap-listener.ts#L47

src/app/index.ts
bodymindarts: In the context of this repository I think these are top level use cases that should be exposed via `src/app` like all other top level use cases.
Which `server` (main, daily-cron, trigger, <new-server>) ends up calling these use cases is another question.

I agree that this functionality should ideally be isolated from the 'core' backend - it belongs in the bitcoin layer - but due to coupling with the ledger (which is tightly integrated into this code base) I currently see no pragmatic alternative.

This PR is solving a short term acute pain point - so we did not set out to think about the ideal architecture for this functionality from first principles. With that background in mind I think this is correct - as it is consistent with how we are handling other similar functionalities (eg. hot -> cold wallet re-balance).

Considered in isolation with an idealised future state of our architecture in mind I would have probably advocated creating this as part of a new repository that starts bundling the bitcoin specific operational tasks - but that would be a much larger scope and create a lot more questions.

src/app/swap/get-active-loopd.ts
bodymindarts, line 22: Why are these exported?

src/app/swap/swap-listener.ts
```diff
@@ -0,0 +1,68 @@
...
+
+import { admin as LedgerAdmin } from "@services/ledger/admin"
+import { WalletCurrency } from "@domain/shared"
+import { LedgerTransactionType } from "@domain/ledger"
+import { SwapProvider, SwapState, SwapType } from "@domain/swap"
+import { addAttributesToCurrentSpan } from "@services/tracing"
+import { SwapErrorNoActiveLoopdNode } from "@domain/swap/errors"
+import { sha256 } from "@domain/bitcoin/lightning"
+
+export const startSwapMonitor = async (swapService: ISwapService) => {
```
bodymindarts, line 11: Wondering if this function belongs in a `server` file.

src/app/swap/swap-listener.ts
```diff
@@ -0,0 +1,68 @@
...
+  })
+}
+
+export const handleSwapOutCompleted = async (swapStatus: SwapStatusResultWrapper) => {
+  if (swapStatus.parsedSwapData) {
+    const type = swapStatus.parsedSwapData.swapType
+    const onchainMinerFee = toSats(swapStatus.parsedSwapData.onchainMinerFee)
+    const offchainRoutingFee = toSats(swapStatus.parsedSwapData.offchainRoutingFee)
+    const serviceProviderFee = toSats(swapStatus.parsedSwapData.serviceProviderFee)
+    const totalFees = onchainMinerFee + offchainRoutingFee + serviceProviderFee
```
bodymindarts, line 26: A few comments:
- type translations should be handled in the server (aka adapter). Ie `SwapStatusResultWrapper` should already have the correct types.
- you are using the 'old' `Satoshis` type - that type is deprecated. You should be using `BtcPaymentAmount` (right @vindard).

src/app/swap/swap-listener.ts
```diff
@@ -0,0 +1,68 @@
...
+    const state = swapStatus.parsedSwapData.state
+    if (type === SwapType.Swapout && state === SwapState.Success && totalFees > 0) {
+      const swapId = swapStatus.parsedSwapData.id as SwapId
+      const swapFeeMetadata: SwapTransactionMetadataUpdate = {
```
bodymindarts, line 30: Creating the XxxMetadata` types should be delegated to the `LedgerFacade` where they should be unit tested. See the functions [here](https://github.com/GaloyMoney/galoy/blob/main/src/services/ledger/tx-metadata.ts) exposed in the facade [here](https://github.com/GaloyMoney/galoy/blob/main/src/services/ledger/facade.ts#L19) and tested [here](https://github.com/GaloyMoney/galoy/blob/main/test/unit/services/ledger/domain/tx-metadata.spec.ts).

src/app/swap/swap-listener.ts
```diff
@@ -0,0 +1,68 @@
...
+      })
+    }
+  }
+}
+
+export async function recordSwapFeeToLedger(
+  swapFeeMetadata: SwapTransactionMetadataUpdate,
+) {
+  const description = `Swap out fee for swapId ${swapFeeMetadata.swapId}`
+  const journalResponse = await LedgerAdmin.addSwapFeeTxSend({
```
bodymindarts, line 63: Even though you are using `LedgerAdmin` instead of the facade I still think you should delegate creating the metadata to the ledger.

src/app/swap/swap-out.ts
```diff
@@ -0,0 +1,67 @@
...
+
+import { getActiveLoopd } from "./get-active-loopd"
+import { getSwapDestAddress } from "./get-swap-dest-address"
+
+const logger = baseLogger.child({ module: "swap" })
+
+export const swapOut = async (): Promise<
+  SwapOutResult | SwapServiceError | NoSwapAction
+> => {
+  logger.info("SwapApp: Started")
```
bodymindarts, line 20: Prefer a tracing event to logger output.

src/app/swap/swap-out.ts
```diff
@@ -0,0 +1,67 @@
...
+    minOnChainHotWalletBalanceConfig,
+    swapOutAmount: toSats(getSwapConfig().swapOutAmount),
+  })
+  const swapOutAmount = swapChecker.getSwapOutAmount({
+    currentOnChainHotWalletBalance: onChainBalance,
+    currentOutboundLiquidityBalance: outbound,
+  })
+
+  if (swapOutAmount instanceof Error) return swapOutAmount
+  if (swapOutAmount === 0) return new NoSwapAction()
```
bodymindarts, line 45: Not sure that `NoSwapAction` should be ist own class. First of all we are not using classes unless for errors (for specific reason). This is not an error but a happy path. I think this should rather be an instance of the `SwapOutResult` type that represents a NoOp.

src/domain/swap/index.ts
```diff
@@ -0,0 +1,24 @@
+export * from "./swap-out-checker"
+
+export const SwapProvider = {
+  Loop: "Loop",
+} as const
+
+export const SwapType = {
+  Swapout: "swapout",
```
bodymindarts, line 8: caml case?

src/domain/swap/index.ts
```diff
@@ -0,0 +1,24 @@
...
+  Unknown: "unknown",
```
bodymindarts, line 9: What is this for?

src/domain/swap/index.ts
```diff
@@ -0,0 +1,24 @@
+export * from "./swap-out-checker"
+
+export const SwapProvider = {
```
bodymindarts, line 3: I'm not sure that details of the swap provider implementation should be in domain.

src/domain/swap/index.ts
```diff
@@ -0,0 +1,24 @@
...
+export const SwapState = {
+  Initiated: "initiated",
+  PreimageRevealed: "preimageRevealed",
+  HtlcPublished: "htlcPublished",
+  Success: "success",
+  Failed: "failed",
+  InvoiceSettled: "invoiceSettled",
+} as const
+
+export const LoopdInstanceName = {
```
bodymindarts, line 21: Definitely don't want couple `domain` to the specifics of our infra setup (with the hard coded lnd1/2 names)... at least I'm doubtful that this is right.

src/domain/swap/swap-out-checker.ts
```diff
@@ -0,0 +1,40 @@
...
+  // if return an "amount" then swapout needed
+  // if return "error" then we don't have enough outbound
+  //   liquidity to perform a swap out.
+  const getSwapOutAmount = ({
+    currentOnChainHotWalletBalance,
+    currentOutboundLiquidityBalance,
+  }: {
+    currentOnChainHotWalletBalance: Satoshis
+    currentOutboundLiquidityBalance: Satoshis
+  }): Satoshis | SwapServiceError => {
```
bodymindarts, line 23: This function is in `domain` so it can not return `SwapServiceError` - as it is not a `service` function (ie. a function on `ISwapOutService`.

Also `Satoshis` is deprecated `BtcPaymentAmount` is the 'new' type for this.

src/services/ledger/admin.ts
```diff
@@ -97,4 +104,50 @@ export const admin = {
...
+          .credit(lndAccountingPath, totalSwapFee, {
+            currency: WalletCurrency.Btc,
+            pending: false,
+            type: LedgerTransactionType.Fee,
+          })
+          .debit(bankOwnerPath, totalSwapFee, {
+            currency: WalletCurrency.Btc,
+            pending: false,
+            type: LedgerTransactionType.Fee,
+          })
```
bodymindarts, line 135: This should use the `EntryBuilder`

bodymindarts (changes requested):
Hey good job overall - I think the general structure of the code is already very close to the style of the rest of the backend.

My comments aren't significant blockers - imo its already 'good enough' but it would avoid a bunch of rework if you use `BtcPaymentAmount` instead of `Satoshi` - create the `metadata` in a unit-tested ledger function and use the `EntryBuilder`. Those are the only significant 'boundary' issues I can see - the rest seems very good / just some nits.

dev/bin/save-loop-data.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+# script to generate the loop wallets for testing
+NETWORK=regtest
+mkdir ../lnd/loop/$NETWORK 2>/dev/null
+mkdir ../lnd/loop/server/$NETWORK 2>/dev/null
+
+fetch_loop_data() {
```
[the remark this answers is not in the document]
ntheile: yeah this script only needs to be ran once and the macaroons are saved. the script was already ran and all the loop macaroons are already saved. this script is kept around just in case we need to regenerate them in the future. i followed the same pattern openoms did with lnd macaroons in the `save-lnd-data.sh` script here https://github.com/GaloyMoney/galoy/blob/main/dev/bin/save-lnd-data.sh.

src/app/swap/get-active-loopd.ts
ntheile: i use them in the integration tests

src/app/swap/swap-listener.ts
ntheile: it gets called by the `trigger.ts` server

src/domain/swap/index.ts
ntheile: swapOut vs swapIn. We don't need swapIn's yet and the code is not written, so I did not yet include it in the SwapType

bodymindarts: hmm - then I guess this can be deleted. Don't want to keep dead code around.

ntheile: its actually used in the loop-service when parsing data from the loop grpc

src/services/ledger/admin.ts
ntheile: going to refactor this with the Ledger Cleanup work with Arvin in the next PR

src/app/swap/swap-listener.ts
ntheile: moved the metadata code to the ledger

src/domain/swap/index.ts
ntheile: removed this from the domain

src/app/swap/swap-out.ts
ntheile: added NoOp to SwapOutResult

ntheile (commented):
Made appropriate changes requested from Nicolas and Justin

bodymindarts (dismissed):

PR #1404, opened 2022-06-27
chore: type improvement 5b

nicolasburtey:
more type improvement to fix `no-implicit-any: false`

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,6 +111,7 @@
     "@jest/test-sequencer": "^28.1.1",
     "@types/babel__core": "^7.1.15",
     "@types/babel__preset-env": "^7.9.2",
+    "@types/bitcoinjs-lib": "^5.0.0",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -111,6 +111,7 @@
     "@jest/test-sequencer": "^28.1.1",
     "@types/babel__core": "^7.1.15",
     "@types/babel__preset-env": "^7.9.2",
+    "@types/bitcoinjs-lib": "^5.0.0",
```
dolcalmi, line 114: please remove, bitcoinjs-lib provides its own type definitions

dolcalmi (changes requested):
looks good, just a minor change

package.json
nicolasburtey: actually that works now. weird that may have been an IDE issue

dolcalmi (approved):

PR #1409, opened 2022-06-29
fix: float amounts validation in SatAmount type

dolcalmi:


nicolasburtey:
let's merge this one @dolcalmi. could be good to have an alert also

PR #1413, opened 2022-06-30
feat: proof of liabilities

UncleSamtoshi:


src/app/proof-of-liabilities/get-user-liabilities.ts
```diff
@@ -0,0 +1,19 @@
+import { LedgerService } from "@services/ledger"
+import { WalletsRepository } from "@services/mongoose"
+
+export const getUserLiabilities = async () => {
```
UncleSamtoshi, line 4: Wallet Currency should be a parameter to this function.

src/domain/wallets/index.types.d.ts
```diff
@@ -151,6 +151,7 @@ interface IWalletsRepository {
 
   findByAddress(address: OnChainAddress): Promise<Wallet | RepositoryError>
   listByAddresses(addresses: OnChainAddress[]): Promise<Wallet[] | RepositoryError>
+  listAll(): Promise<Wallet[] | RepositoryError>
```
UncleSamtoshi, line 154: Should take WalletCurrency as a parameter

src/app/proof-of-liabilities/get-user-liabilities.ts
```diff
@@ -0,0 +1,19 @@
...
+  const wallets = await WalletsRepository().listAll()
+  if (wallets instanceof Error) return wallets
+  const btcWallets = wallets.filter((wallet) => wallet.currency === "BTC")
+  const liabilities = btcWallets.map((wallet) => walletToLiability(wallet))
```
UncleSamtoshi, line 8: I believe you need to wrap this array with Promise.all, so it will be ```const liabilities = Promise.all(btcWallets.map(...))```

UncleSamtoshi (commented):
Looks really good! A couple feedback items.

src/app/proof-of-liabilities/get-user-liabilities.ts
```diff
@@ -0,0 +1,19 @@
+import { LedgerService } from "@services/ledger"
+import { WalletsRepository } from "@services/mongoose"
+
+export const getUserLiabilities = async () => {
+  const wallets = await WalletsRepository().listAll()
+  if (wallets instanceof Error) return wallets
+  const btcWallets = wallets.filter((wallet) => wallet.currency === "BTC")
```
dolcalmi, line 7: this filter should be at db level

UncleSamtoshi (commented):
After thinking about it more, I believe we should get rid of all references to liabilities in this repository. Instead we can create an endpoint to get all wallets, and then in the liabilities plugin we can convert the wallet id and balance into a liability object. This change would decouple the main galoy server from the liabilities plugin. It is currency implicitly coupled by creating a specific endpoint that uses the terminology ```liability```.

src/app/proof-of-liabilities/get-user-liabilities.ts
```diff
@@ -0,0 +1,18 @@
...
+import { WalletsRepository } from "@services/mongoose"
+
+export const getUserLiabilities = async () => {
+  const wallets = await WalletsRepository().listAll("BTC")
+  if (wallets instanceof Error) return wallets
+  const liabilities = wallets.map((wallet) => walletToLiability(wallet))
+  return liabilities
+}
+
+const walletToLiability = async (wallet: Wallet) => {
```
nicolasburtey, line 11: this is just a `getBalance` function. we should already have this as part of the API for wallet

src/app/proof-of-liabilities/get-user-liabilities.ts
```diff
@@ -0,0 +1,18 @@
+import { LedgerService } from "@services/ledger"
+import { WalletsRepository } from "@services/mongoose"
+
+export const getUserLiabilities = async () => {
+  const wallets = await WalletsRepository().listAll("BTC")
+  if (wallets instanceof Error) return wallets
+  const liabilities = wallets.map((wallet) => walletToLiability(wallet))
```
nicolasburtey, line 7: the use of Balance keyword here is probably more relevant (versus Liabilities).

this function doesn't need to be tied to proof of reserve. we're just returning a list of balance for wallets.

this should also be paginated, probably from the client level instead of the server level.

nicolasburtey (changes requested):
the admin API doesn't need proof of reserve specific vocabulary/API imo. 

a client calling the API can fetch the list of balance (with a paginated query). the logic for POR should then be at the client level.

nicolasburtey:
close per request from @thevaibhav-dixit

Issue #1414, opened 2022-07-01
settlementPrice should be 1 if Tx has the same currency of settlementAmount


dolcalmi:
I sent 100 usd to an empty usd wallet. The balance is ok but the tx shows a different value.

Initially I thought was an issue in mobile app but then checking settlementPrice realize that is an issue in display currency for usd txs (in this case probably in `updatePendingInvoice` method)

@sebastienverreault do we use settlementPrice in the dealer?

  ```json
    {
      "__typename": "Transaction",
      "id": "62be47...14ed40",
      "status": "SUCCESS",
      "direction": "RECEIVE",
      "memo": null,
      "createdAt": 1656637439,
      "settlementAmount": 10000,
      "settlementFee": 0,
      "settlementCurrency": "USD",
      "settlementPrice": {
        "base": 998699361100,
        "offset": 12,
        "currencyUnit": "USDCENT",
        "formattedAmount": "0.9986993611"
      },
      "initiationVia": {
        "__typename": "InitiationViaLn",
        "paymentHash": "9498ac1a31b4d...905b42ba3e8515ed"
      },
      "settlementVia": {
        "__typename": "SettlementViaLn",
        "paymentSecret": null
      }
    }
  ```

<img src="https://user-images.githubusercontent.com/2079600/176903594-3c1cd54a-0bfb-45dd-b673-b208cac92788.png" width="45%">  <img src="https://user-images.githubusercontent.com/2079600/176903630-9f1c5651-6507-46c2-9fcc-6862a1e13111.png" width="45%">

PR #1415, opened 2022-07-01
fix: update display amount to tx amount for usd txs

dolcalmi:
Fix issue https://github.com/GaloyMoney/galoy/issues/1414

### DIFF
--- a/src/app/wallets/update-pending-invoices.ts
+++ b/src/app/wallets/update-pending-invoices.ts
@@ -177,6 +177,11 @@ const updatePendingInvoice = async ({
       currency: DisplayCurrency.Usd,
     }
 
+    let amountDisplayCurrency = displayPaymentAmount.amount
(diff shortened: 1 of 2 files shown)


src/app/wallets/update-pending-invoices.ts
```diff
@@ -177,6 +177,11 @@ const updatePendingInvoice = async ({
       currency: DisplayCurrency.Usd,
     }
 
+    let amountDisplayCurrency = displayPaymentAmount.amount
```
UncleSamtoshi, line 180: Is price generated dynamically from the amountDisplayCurrency and settlement amount?

UncleSamtoshi (approved):

src/app/wallets/update-pending-invoices.ts
dolcalmi: in this case is from price service.. line 159

PR #1421, opened 2022-07-07
feat: add tx record type

vindard:
## Description

This PR adds a new `TransactionRecord` type, and adds type signatures to the `translateToLedgerTx` method.


src/services/ledger/schema.types.d.ts
```diff
@@ -1,3 +1,44 @@
+interface TransactionRecord {
```
dolcalmi, line 1: if this is is used to match medici we will need to add missing fields

https://github.com/GaloyMoney/galoy/blob/74ee147888b00c21825145c1097ace8aaf5e68d8/src/services/ledger/schema.types.d.ts

dolcalmi (changes requested):

src/services/ledger/schema.types.d.ts
vindard: Oh I didn't realise you all were making the same type. I swapped that new type in here instead.

dolcalmi (dismissed):

dolcalmi (approved):

PR #1424, opened 2022-07-10
feat: upgrade to graphqlws

### LINKED ISSUE #1365: switch to new graphql subscription protocol
We are using the old, deprecated and no longer maintained, protocol.

https://www.apollographql.com/docs/apollo-server/data/subscriptions/#switching-from-subscriptions-transport-ws

this requires updating both backend and frontend at the same time.

nicolasburtey:
fix #1365, plus some security consideration using the old library: https://github.com/enisdenjo/graphql-ws/issues/3

I ran into issues when I was importing both `subscriptions-transport-ws` and `graphql-ws` at the same time. instead of trying to understand the root of this, I decided to duplicated graphql-server and dynamically, at runtime, one or the other websocket library would be used.

by default the old library would be used - so that we can enable the use of the new library in another PR in the charts/deployment repo.

I recommend that we use a new url for BBW to do the migration from the mobile app. we could use api.bbw.sv to replace the current url. api.bbw.sv would have the `graphql-ws` library. mainnet.graphql.galoy.io would have the `subscriptions-transport-ws` library.


nicolasburtey:
legacy server (when no env variable is set) --> api.mainnet.galoy.io
"new" server (when NEW_SERVER is set) --> api.bbw.sv

nicolasburtey:
migrating in favor of https://github.com/GaloyMoney/galoy/pull/1424 (a squash to simplify the rebase)

PR #1427, opened 2022-07-11
feat: add metadata to ledger txns

vindard:
## Description

This PR adds metadata to ledger transactions via a 2nd call to the metadata collection for ledger methods that reference transactions specifically by `id` or `paymentHash`. It does not do the additional metadata collection lookup for ledger methods that list all transaction for a given user.

To support this, an additional `transactionById` gql query was added so that an api user can fetch the full transactions list normally and then drill down into the details of a given transaction using this additional query.

**This PR supersedes PR #1077**

### DIFF
--- a/src/domain/ledger/index.types.d.ts
+++ b/src/domain/ledger/index.types.d.ts
@@ -245,13 +245,13 @@ interface ILedgerService {
       | LnLedgerTransactionMetadataUpdate,
   ): Promise<true | LedgerServiceError>
 
-  getTransactionById(
+  getTransactionById<S extends WalletCurrency>(
     id: LedgerTransactionId,
-  ): Promise<LedgerTransaction<WalletCurrency> | LedgerServiceError>
+  ): Promise<LedgerTransactionWithMetadata<S> | LedgerServiceError>
 
-  getTransactionsByHash(
+  getTransactionsByHash<S extends WalletCurrency>(
     paymentHash: PaymentHash | OnChainTxHash,
-  ): Promise<LedgerTransaction<WalletCurrency>[] | LedgerServiceError>
+  ): Promise<LedgerTransactionWithMetadata<S>[] | LedgerServiceError>
 
   getTransactionsByWalletId(
     walletId: WalletId,
--- a/src/domain/ledger/index.types.d.ts
+++ b/src/domain/ledger/index.types.d.ts
@@ -24,8 +24,8 @@ type LedgerTransaction<S extends WalletCurrency> = {
   readonly id: LedgerTransactionId
   readonly walletId: WalletId | undefined // FIXME create a subclass so that this field is always set for liabilities wallets
   readonly type: LedgerTransactionType
-  readonly debit: S extends "BTC" ? Satoshis : UsdCents
-  readonly credit: S extends "BTC" ? Satoshis : UsdCents
+  readonly debit: Satoshis | UsdCents
+  readonly credit: Satoshis | UsdCents
   readonly fee: Satoshis
   readonly currency: S
   readonly timestamp: Date
--- a/src/graphql/types/object/transaction-with-metadata.ts
+++ b/src/graphql/types/object/transaction-with-metadata.ts
@@ -1,0 +1,83 @@
+import dedent from "dedent"
+
+import { GT } from "@graphql/index"
+
+import { SAT_PRICE_PRECISION_OFFSET } from "@config"
+
+import Memo from "../scalar/memo"
+
+import InitiationVia from "../abstract/initiation-via"
+import SettlementViaWithMetadata from "../abstract/settlement-via-with-metadata"
+import Timestamp from "../scalar/timestamp"
+import TxDirection, { txDirectionValues } from "../scalar/tx-direction"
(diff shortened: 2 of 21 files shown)


src/graphql/types/abstract/settlement-via.ts
```diff
@@ -31,13 +32,15 @@ const SettlementViaLn = GT.Object({
   fields: () => ({
     paymentSecret: {
       type: LnPaymentSecret,
-      resolve: (source: SettlementViaLn) => source.revealedPreImage,
+      resolve: () => undefined,
       deprecationReason:
-        "Shifting property to 'preImage' to improve granularity of the LnPaymentSecret type",
+        "Moving this property over to the TransactionWithMetadata object type.",
     },
     preImage: {
```
nicolasburtey, line 39: is this used at all, or could we scrap this code now?

src/graphql/types/object/transaction-with-metadata.ts
```diff
@@ -0,0 +1,83 @@
...
+import SettlementViaWithMetadata from "../abstract/settlement-via-with-metadata"
+import Timestamp from "../scalar/timestamp"
+import TxDirection, { txDirectionValues } from "../scalar/tx-direction"
+import TxStatus from "../scalar/tx-status"
+import SignedAmount from "../scalar/signed-amount"
+import WalletCurrency from "../scalar/wallet-currency"
+
+import Price from "./price"
+
+const TransactionWithMetadata = GT.Object<WalletTransactionWithMetadata>({
```
nicolasburtey, line 19: With this PR we seem to add quite a bit of complexity for the API consumer. 

why not do something like:

```
    fragment TransactionList on TransactionConnection {
    pageInfo {
      hasNextPage
      hasPreviousPage
      startCursor
      endCursor
    }
    edges {
      cursor
      node {
        __typename
        id
        status
        direction
        memo
        createdAt
        settlementAmount
        settlementFee
        settlementCurrency
        settlementPrice {
          base
          offset
          currencyUnit
          formattedAmount
        }
        initiationVia {
          __typename
          ... on InitiationViaIntraLedger {
            counterPartyWalletId
            counterPartyUsername
          }
          ... on InitiationViaLn {
            paymentHash
          }
          ... on InitiationViaOnChain {
            address
          }
        }
        settlementVia {
          __typename
          ... on SettlementViaIntraLedger {
            counterPartyWalletId
            counterPartyUsername
          }
          ... on SettlementViaLn {
            paymentSecret
          }
          ... on SettlementViaOnChain {
            transactionHash
          }
        }
      }
    }
  }
`

this for the main query, where we list all the transactions.

and for the individual query, something like:

```
    Transaction($transactionHash) {
      cursor
      node {
        __typename
        id
        status
        direction
        memo

        __preImage__
        __otherOptionalMetadata__

        createdAt
        settlementAmount
        settlementFee
        settlementCurrency
        settlementPrice {
          base
          offset
          currencyUnit
          formattedAmount
        }
        initiationVia {
          __typename
          ... on InitiationViaIntraLedger {
            counterPartyWalletId
            counterPartyUsername
          }
          ... on InitiationViaLn {
            paymentHash
          }
          ... on InitiationViaOnChain {
            address
          }
        }
        settlementVia {
          __typename
          ... on SettlementViaIntraLedger {
            counterPartyWalletId
            counterPartyUsername
          }
          ... on SettlementViaLn {
            paymentSecret
          }
          ... on SettlementViaOnChain {
            transactionHash
          }
        }
      }
    }
  }
`

the idea behind it is that we would not introduce a new type `WalletTransactionWithMetadata` but rather use the inherent property of Graphql to fetch, or not fetch, the individual metadata as needed.

for this to work, some of the logic for the additional/optional metadata would have to be in the graphql resolver. I'm not sure how this may interact with our domain logic yet (example here: https://stackoverflow.com/questions/61060046/graphql-only-fetch-if-certain-field-is-requested-in-query)

nicolasburtey (commented):
some early feedback on the design. I may not have full context on why we want to have another Transaction object with Metadata, but I feel we could get a better design by using the native property of Graphql and let the user choose what property he wants in query on the Transaction object without having to create multiple Transaction object.

src/domain/ledger/index.types.d.ts
```diff
@@ -24,8 +24,8 @@ type LedgerTransaction<S extends WalletCurrency> = {
   readonly id: LedgerTransactionId
   readonly walletId: WalletId | undefined // FIXME create a subclass so that this field is always set for liabilities wallets
   readonly type: LedgerTransactionType
-  readonly debit: S extends "BTC" ? Satoshis : UsdCents
-  readonly credit: S extends "BTC" ? Satoshis : UsdCents
+  readonly debit: Satoshis | UsdCents
```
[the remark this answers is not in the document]
vindard: I couldn't get the previous type to work properly when we typed the `translate` function more strongly

vindard:
Update: we decided that it would be a better approach to be more long-term with this solution and implement selective fetching at the query level on the same single `Transaction` type by using graphql's [`dataloader`](https://github.com/graphql/dataloader) package to manage data fetching.

PR #1432, opened 2022-07-12
feat: Decreased feecap percentage

daviroo:
@jireva recently did some data analysis and worked out that the feecap percentage could be dropped to 0.5% of payment amount without effecting 99% of payments.  This PR drops the feecap to the appropriate level.


nicolasburtey (approved):

nicolasburtey (dismissed):
actually not approving this as tests are failing now

daviroo:
> actually not approving this as tests are failing now

It was actually failing because in the unit tests the new feecap was dropping the fee below the `FEEMIN` value currently set to 10 sats.  Should we also adjust the min fee value since we are decreasing the feecap?

dolcalmi:
@vindard do we have 2 places to configure this? https://github.com/GaloyMoney/galoy/blob/main/src/domain/payments/ln-fees.ts#L9 would you mind to check?

test/unit/domain/shared/amount-calculator.spec.ts
```diff
@@ -150,4 +150,15 @@ describe("AmountCalculator", () => {
       })
     }
   })
+
+  describe("multiplies", () => {
+    it("correctly applies basis points to an amount", () => {
```
dolcalmi, line 155: please add a test where the amount should be rounded and other which return min value (with a small amount)

dolcalmi (approved):

src/domain/shared/calculator.ts
```diff
@@ -47,11 +47,18 @@ export const AmountCalculator = (): AmountCalculator => {
...
+  ) =>
+    divRound({ amount: amount.amount * basisPoints, currency: amount.currency }, 10_000n)
+
   return {
     sub,
     add,
     divRound,
     divFloor,
     divCeil,
+    applyBasisPoints,
```
nicolasburtey, line 62:
```suggestion
    mulBasisPoints,
```

nicolasburtey (commented):
one comment on naming

nicolasburtey (approved):

PR #1433, opened 2022-07-12
chore(deps): bump price-history proto

galoybot:


krtk6160 (approved):

dolcalmi (changes requested):
@krtk6160 price and history are different services if this is price history update what should be the name for price health proto? maybe we should change the names for health protos :thinking:

bodymindarts:
File needs renaming in price-history

PR #1438, opened 2022-07-13
chore: add log write in recordException

dolcalmi:
- This will be useful for first time setup (contributors without experience with tracing and honeycomb)
- Some edge cases requires review all the exceptions

Downsides
- We dont have redaction by default. Probably we will need to include as needed https://github.com/pinojs/pino/blob/master/docs/redaction.md

### DIFF
--- a/src/services/tracing.ts
+++ b/src/services/tracing.ts
@@ -30,6 +30,8 @@ import { ErrorLevel } from "@domain/shared"
 
 import type * as graphqlTypes from "graphql"
 
+import { baseLogger } from "./logger"
+
 propagation.setGlobalPropagator(new W3CTraceContextPropagator())
 
 // FYI this hook is executed BEFORE the `formatError` hook from apollo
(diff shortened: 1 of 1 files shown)


nicolasburtey:
where is the redaction done currently for opentelemetry? not in the galoy pod but in some opentelemetry dedicated server (the one responsible for collection of the data and sending it over honeycomb)?

bodymindarts:
> where is the redaction done currently for opentelemetry? not in the galoy pod but in some opentelemetry dedicated server (the one responsible for collection of the data and sending it over honeycomb)?

https://github.com/GaloyMoney/galoy-infra/blob/main/modules/services/opentelemetry-values.yml#L26-L54

nicolasburtey:
> > where is the redaction done currently for opentelemetry? not in the galoy pod but in some opentelemetry dedicated server (the one responsible for collection of the data and sending it over honeycomb)?
> 
> https://github.com/GaloyMoney/galoy-infra/blob/main/modules/services/opentelemetry-values.yml#L26-L54

thanks. 

I'm not sure we should merge this as is, as you said it introduce some new security flaw. 

maybe it's only with an option env variable flag that we output to the console.  otherwise I would think there are some open telemetry collector that could just print things in a console/stdout as well?

I think someone settings up a Galoy cluster would have to be familiar with open telemetry anyhow

PR #1446, opened 2022-07-17
chore: show in the log what is wrong

nicolasburtey:
when there is an error in the e2e test related to Graphql


nicolasburtey:
only work for query and mutation, not subscription. not sure how much this is useful

PR #1447, opened 2022-07-18
feat: add lightningAddressDomain to globals

samerbuna:


default.yaml
```diff
@@ -1,5 +1,7 @@
 name: "Galoy Banking"
 
+lightningAddressDomain: "bbw.sv"
```
nicolasburtey, line 3:
```suggestion
lightningAddressDomain: "pay.bbw.sv"
```

currnetly we use the pay subdomain - the logic is in the galoy-pay server. 

would be great to be at the top level thought, we would a redirection for that

samerbuna: Updated

samerbuna:
@daviroo if you can give this PR a review, I added the aliases field.

dolcalmi (approved):

PR #1460, opened 2022-07-21
fix: handle null/undefined entry description for medici 5

vindard:
## Description

This is one proposed fix for the issue where we see `"undefined"` string values for `memo` field in transactions instead of `undefined` object.

### Root cause

Medici 5 entry constructor handles the description argument different now:
- medici 5: `book.entry(undefined)` results in an entry with `{ memo: "undefined" }`
- medici 4: `book.entry(undefined)` results in an entry with `{ memo: undefined }`

### Proposed fix here
Inside the EntryBuilder, convert any `"undefined"` or `"null"` memo strings to an empty string (`""`) to enforce the falsiness we need.

### Alternative fix
Convert `undefined`/`false` inputs to the book.entry constructor throughout our application code so that an empty string always finally gets to the `book.entry` constructor. Also add a check in the entry builder constructor for `"undefined"`/`"null"` strings as the entry memo and then return an error if the check fails (entry builder does not have same error handling implemented that the payment flow builder has as yet).

### DIFF
--- a/src/services/ledger/domain/entry-builder.ts
+++ b/src/services/ledger/domain/entry-builder.ts
@@ -9,6 +9,10 @@ export const EntryBuilder = <M extends MediciEntry>({
   entry,
   metadata,
 }: EntryBuilderConfig<M>) => {
+  if (entry.journal.memo === "undefined" || entry.journal.memo === "null") {
+    entry.journal.memo = ""
+  }
+
   const withTotalAmount = ({
     usdWithFees,
     btcWithFees,
--- a/test/unit/services/ledger/domain/entry-builder.spec.ts
+++ b/test/unit/services/ledger/domain/entry-builder.spec.ts
@@ -109,6 +109,66 @@ describe("EntryBuilder", () => {
     more: "more",
   }
 
+  describe("truthy/falsy entry memo", () => {
+    // The mobile app uses a falsy memo to overwrite memos for certain types
+    // of transactions (e.g. usd/btc conversion)
+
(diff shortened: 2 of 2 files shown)


nicolasburtey:
> medici 5: book.entry(undefined) results in an entry with { memo: "undefined" }

Isn't this a bug of Medici 5 that could be fixed there?

bodymindarts:
I wouldn't fix it like this. It seems very hacky. Lets fix how the bad data is getting in first then we can run a migration to get rid of the old bad data.

I've noticed that there seems to be a typescript bug that may have contributed to this. The definition of the type does not match the definition... I'm surprised the typescript compiler didn't catch this:

compare the [`withoutInvoice`](https://github.com/GaloyMoney/galoy/blob/main/src/domain/payments/index.types.d.ts#L80-L85) type
```
  withoutInvoice({
    uncheckedAmount,
    description,
  }: {
    uncheckedAmount: number
    description: string | null
```
with [its implementation](https://github.com/GaloyMoney/galoy/blob/main/src/domain/payments/payment-flow-builder.ts#L80-L86).
```
  const withoutInvoice = ({
    uncheckedAmount,
    description,
  }: {
    uncheckedAmount: number
    description: string
  }): LPFBWithInvoice<S> | LPFBWithError => {
  ```

the `| null` is missing in the implementation and after that all the downstream code assumes a string for `description`.

@samerbuna do you know why the compiler doesn't catch this?

vindard:
Closing in favour of #1461

PR #1461, opened 2022-07-21
fix: inconsistency between type definition and implementation

bodymindarts:
For unknown reasons the typescript compiler didn't flag a difference
between type definition and implementation which allowed a `null` value
to be passed where none was assumed in the downstream code.

### DIFF
--- a/src/domain/payments/index.types.d.ts
+++ b/src/domain/payments/index.types.d.ts
@@ -82,7 +82,7 @@ type LightningPaymentFlowBuilder<S extends WalletCurrency> = {
     description,
   }: {
     uncheckedAmount: number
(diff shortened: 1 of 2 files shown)


src/domain/payments/index.types.d.ts
```diff
@@ -82,7 +82,7 @@ type LightningPaymentFlowBuilder<S extends WalletCurrency> = {
     description,
   }: {
     uncheckedAmount: number
-    description: string | null
+    description: string
```
samerbuna, line 85: @bodymindarts I think the problem here is that the type for `withoutInvoice` is defined in 2 places.

PR #1464, opened 2022-07-21
feat: add graphql endpoints to fetch all wallet id's and wallet

thevaibhav-dixit:
This PR adds two new endpoints in the admin Graphql query. An `allWalletIds` query that returns a list of all the wallet id's and a `wallet` query that returns an individual wallet based on the `walletId` passed.

### DIFF
--- a/src/app/wallets/index.ts
+++ b/src/app/wallets/index.ts
@@ -23,3 +23,8 @@ export const listWalletsByAccountId = async (
 ): Promise<Wallet[] | RepositoryError> => {
   return WalletsRepository().listByAccountId(accountId)
 }
+
+export const getAllWalletIds = async (): Promise<WalletId[] | RepositoryError> => {
+  const walletIds = await WalletsRepository().listAllWalletIds()
+  return walletIds
+}
--- a/src/app/wallets/index.ts
+++ b/src/app/wallets/index.ts
@@ -24,7 +24,8 @@ export const listWalletsByAccountId = async (
   return WalletsRepository().listByAccountId(accountId)
 }
 
-export const getAllWalletIds = async (): Promise<WalletId[] | RepositoryError> => {
-  const walletIds = await WalletsRepository().listAllWalletIds()
-  return walletIds
+export const getAllWalletIds = async (
+  walletCurrency: WalletCurrency,
+): Promise<WalletId[] | RepositoryError> => {
+  return WalletsRepository().listAllWalletIds(walletCurrency)
 }
--- a/src/app/wallets/index.ts
+++ b/src/app/wallets/index.ts
@@ -24,8 +24,10 @@ export const listWalletsByAccountId = async (
   return WalletsRepository().listByAccountId(accountId)
 }
 
-export const getAllWalletIds = async (
+export const listWalletIds = async (
   walletCurrency: WalletCurrency,
 ): Promise<WalletId[] | RepositoryError> => {
-  return WalletsRepository().listAllWalletIds(walletCurrency)
+  const wallets = await WalletsRepository().listByWalletCurrency(walletCurrency)
+  if (wallets instanceof Error) return wallets
+  return wallets.map((wallet) => wallet.id)
 }
--- a/src/domain/wallets/index.types.d.ts
+++ b/src/domain/wallets/index.types.d.ts
@@ -186,7 +186,7 @@ interface IWalletsRepository {
 
   findByAddress(address: OnChainAddress): Promise<Wallet | RepositoryError>
   listByAddresses(addresses: OnChainAddress[]): Promise<Wallet[] | RepositoryError>
-  listAllWalletIds(): Promise<WalletId[] | RepositoryError>
+  listAllWalletIds(walletCurrency: WalletCurrency): Promise<WalletId[] | RepositoryError>
 }
 
 type onChainDepositFeeArgs = {
--- a/src/domain/wallets/index.types.d.ts
+++ b/src/domain/wallets/index.types.d.ts
@@ -186,6 +186,7 @@ interface IWalletsRepository {
 
   findByAddress(address: OnChainAddress): Promise<Wallet | RepositoryError>
   listByAddresses(addresses: OnChainAddress[]): Promise<Wallet[] | RepositoryError>
+  listAllWalletIds(): Promise<WalletId[] | RepositoryError>
 }
 
 type onChainDepositFeeArgs = {
--- a/src/domain/wallets/index.types.d.ts
+++ b/src/domain/wallets/index.types.d.ts
@@ -186,7 +186,9 @@ interface IWalletsRepository {
 
   findByAddress(address: OnChainAddress): Promise<Wallet | RepositoryError>
   listByAddresses(addresses: OnChainAddress[]): Promise<Wallet[] | RepositoryError>
-  listAllWalletIds(walletCurrency: WalletCurrency): Promise<WalletId[] | RepositoryError>
+  listByWalletCurrency(
+    walletCurrency: WalletCurrency,
+  ): Promise<Wallet[] | RepositoryError>
 }
 
 type onChainDepositFeeArgs = {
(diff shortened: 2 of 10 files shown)


src/graphql/admin/root/query/all-walletids.ts
```diff
@@ -0,0 +1,13 @@
+import { GT } from "@graphql/index"
+import { Wallets } from "@app"
+import WalletId from "@graphql/types/scalar/wallet-id"
+
+const AllWalletIdsQuery = GT.Field({
+  type: GT.NonNullList(WalletId),
+  resolve: async () => {
+    const walletIds = await Wallets.getAllWalletIds()
```
UncleSamtoshi, line 8: should be ``return Wallets.getAllWalletIds()```

src/services/mongoose/wallets.ts
```diff
@@ -102,13 +103,26 @@ export const WalletsRepository = (): IWalletsRepository => {
       return new UnknownRepositoryError(err)
     }
   }
+  const listAllWalletIds = async (): Promise<WalletId[] | RepositoryError> => {
+    try {
+      const result: WalletRecord[] = await Wallet.find({})
+      if (!result || result.length === 0) {
```
UncleSamtoshi, line 109: Length of 0 is an acceptable reponse

UncleSamtoshi (commented):
Nice work!

src/app/wallets/index.ts
```diff
@@ -23,3 +23,9 @@ export const listWalletsByAccountId = async (
 ): Promise<Wallet[] | RepositoryError> => {
   return WalletsRepository().listByAccountId(accountId)
 }
+
+export const getAllWalletIds = async (
```
nicolasburtey, line 27: I would call it like this for consistency

```suggestion
export const listWalletIds = async (
```

src/app/wallets/index.ts
```diff
@@ -23,3 +23,9 @@ export const listWalletsByAccountId = async (
...
+  walletCurrency: WalletCurrency,
+): Promise<WalletId[] | RepositoryError> => {
+  return WalletsRepository().listAllWalletIds(walletCurrency)
+}
```
nicolasburtey, line 31: could even be reduced by the following:

```suggestion
): Promise<WalletId[] | RepositoryError> => 
 WalletsRepository().listAllWalletIds(walletCurrency)
```

src/domain/wallets/index.types.d.ts
```diff
@@ -186,6 +186,7 @@ interface IWalletsRepository {
 
   findByAddress(address: OnChainAddress): Promise<Wallet | RepositoryError>
   listByAddresses(addresses: OnChainAddress[]): Promise<Wallet[] | RepositoryError>
+  listAllWalletIds(walletCurrency: WalletCurrency): Promise<WalletId[] | RepositoryError>
```
nicolasburtey, line 189:
```suggestion
  listWalletIds(walletCurrency: WalletCurrency): Promise<WalletId[] | RepositoryError>
```

src/graphql/admin/queries.ts
```diff
@@ -18,6 +20,8 @@ const QueryType = GT.Object({
     transactionsByHash: TransactionsByHashQuery,
     lightningInvoice: LightningInvoiceQuery,
     lightningPayment: LightningPaymentQuery,
+    allWalletIds: AllWalletIdsQuery,
```
nicolasburtey, line 23:
```suggestion
    listWalletIds: ListWalletIdsQuery,
```

src/graphql/admin/root/query/all-walletids.ts
```diff
@@ -0,0 +1,16 @@
...
+import WalletId from "@graphql/types/scalar/wallet-id"
+
+const AllWalletIdsQuery = GT.Field({
+  type: GT.NonNullList(WalletId),
+  args: {
+    walletCurrency: { type: GT.NonNull(WalletCurrency) },
+  },
+  resolve: async (_, { walletCurrency }) => {
+    return Wallets.getAllWalletIds(walletCurrency)
+  },
```
nicolasburtey, line 13:
```suggestion
  resolve: (_, { walletCurrency }) => 
    Wallets.getAllWalletIds(walletCurrency)
```

src/graphql/admin/schema.graphql
```diff
@@ -300,10 +300,12 @@ type Query {
   accountDetailsByUserPhone(phone: Phone!): Account!
   accountDetailsByUsername(username: Username!): Account!
   allLevels: [AccountLevel!]!
+  allWalletIds(walletCurrency: WalletCurrency!): [WalletId!]!
```
nicolasburtey, line 303: I think we should have this (potentially array could be empty?): 

```suggestion
  allWalletIds(walletCurrency: WalletCurrency!): [WalletId]!
```

src/services/mongoose/wallets.ts
```diff
@@ -102,13 +103,28 @@ export const WalletsRepository = (): IWalletsRepository => {
       return new UnknownRepositoryError(err)
     }
   }
+  const listAllWalletIds = async (
+    walletCurrency: WalletCurrency,
+  ): Promise<WalletId[] | RepositoryError> => {
+    try {
+      const result: WalletRecord[] = await Wallet.find({ currency: walletCurrency })
```
nicolasburtey, line 110: this cast is not needed anymore since medici 5 afaik

```suggestion
      const result = await Wallet.find({ currency: walletCurrency })
```

src/services/mongoose/wallets.ts
```diff
@@ -102,13 +103,28 @@ export const WalletsRepository = (): IWalletsRepository => {
...
+      if (!result) {
+        return new CouldNotListWalletIdError()
+      }
+      const walletIds = result.map(({ id }) => id as WalletId)
```
nicolasburtey, line 114: `id` should already be WalletId. no cast should be needed

thevaibhav-dixit: I think without casting `id` will be `any` type

bodymindarts: Casting here is a bug - I think you want
```
 const walletIds = result.map(({ id }) => fromObjectId<WalletId>(id)
```
see src/services/mongoose/utils.ts

nicolasburtey: > I think without casting `id` will be `any` type

don't think it should be any: https://github.com/GaloyMoney/galoy/blob/7f17f41920d5d6fc2b05bc90526f9a6d78b43481/src/services/mongoose/wallets.ts#L116

nicolasburtey: oh ok you're not return the whole `Wallet` object but just `WalletId`

I would return all `Wallet` object, and do the filtering in the domain, so that the function is more generic at this level @thevaibhav-dixit 

thoughts @bodymindarts ?

bodymindarts: > I would return all Wallet object, and do the filtering in the domain, so that the function is more generic at this level @thevaibhav-dixit

Yes agree - `listByWalletCurrency(currency: WalletCurrency): AsyncGenerator<Wallet | RepositoryError>` would probably be the signature consistent with other similar use cases.

bodymindarts: Or perhaps `listByWalletCurrency(currency: WalletCurrency): Wallet[] | RepositoryError` - In some places we use `AsyncGenerator` which is preferred but I'm not sure if we use it in combination with `mongoose` or in other situations.

nicolasburtey (dismissed):

src/services/mongoose/wallets.ts
```diff
@@ -102,13 +103,27 @@ export const WalletsRepository = (): IWalletsRepository => {
       return new UnknownRepositoryError(err)
     }
   }
+  const listByWalletCurrency = async (
```
nicolasburtey, line 106:
```suggestion
  // TODO: future performance improvement might be needed
  // add pagination for instance which would have millions of wallets
  const listByWalletCurrency = async (
```

nicolasburtey (approved):

PR #1465, opened 2022-07-22
feat: add account custom fields

dolcalmi:
## How to test

Follow instructions here https://github.com/GaloyMoney/galoy/blob/0d80980d02ee79726773154c9e36c9da58b67230/docs/custom-account-fields.md

---

### PR Tasks

- [x] create account information config
- [x] add dynamic fields to graphql
- [x] choose naming (metadata, custom fields, dynamic fields, information, account extra info, ...) and graphql structure
- [x] add dynamic fields repository (query, update)
- [x] update app layer + graphql resolvers
- [x] add admin queries
- [x] add support to array fields
- [x] allow empty schema
- [x] add default values support
- [x] add list accounts by custom field


default.yaml
```diff
@@ -124,6 +124,19 @@ rateLimits:
 
 accounts:
   initialStatus: active
+  information:
```
dolcalmi, line 127: this is just test data  to be able to see the changes in the graphql schema file

src/graphql/helpers.ts
```diff
@@ -30,3 +34,30 @@ export const validateIsUsdWalletForMutation = async (
...
+    const type = getGT(field.type)
+    result[field.name] = {
+      type: field.required ? GT.NonNull(type) : type,
+    }
+  }
+  return result
+}
+
+const getGT = (type?: string): GraphQLScalarType => {
+  switch (type) {
```
dolcalmi, line 53: should we enable another type?

src/graphql/main/schema.graphql
```diff
@@ -27,6 +28,14 @@ interface Account {
...
 }
 
+type AccountInformation {
+  firstName: String!
+  lastName: String
+  otherField: Int
+  otherField2: Boolean
+  transactionsCallback: String
+}
+
```
samerbuna, line 38: Is this temp too?

The main GraphQL schema should not have any static values for account information. I'm not sure if there's a way to send back dynamic values though. Best I can think of is an array of arrays where everything is a string (but there's a `type` value in there somewhere:

`[ [name, type, value], [name, type, value], ...]`

dolcalmi: yep, if we remove default.yaml fields this will be removed too

samerbuna: 👍🏽 everything else look good.

bodymindarts:
I have doubts about exposing the fields individually over GQL I was thinking something much simpler like a `string` field that accepts a JSON blob. We should probably discuss this at standup / code review.

The downside I see is that a simple config change could mean a breaking API change - who is going to be responsible for this? Also what about the data already in the database - it could easily become inconsistent with the GQL definition.

default.yaml
nicolasburtey: I would call it `custom` or `customFields`

src/config/schema.ts
```diff
@@ -219,6 +219,24 @@ export const configSchema = {
...
+        information: {
+          type: "array",
+          items: {
+            type: "object",
+            properties: {
+              name: { type: "string" },
+              type: {
+                type: "string",
+                enum: ["string", "integer", "float", "boolean"],
+                default: "string",
```
nicolasburtey, line 231:
```suggestion
```

I would suggest not having a default here. ie: it has to be passed explicitly

src/config/schema.types.d.ts
```diff
@@ -16,6 +16,12 @@ type AccountLimitsConfig = {
   }
 }
 
+type DynamicField = {
```
nicolasburtey, line 19:
```suggestion
type CustomField = {
```

nicolasburtey: this would be my suggestion regarding name

src/graphql/main/schema.graphql
nicolasburtey: will the `schema.graphql` dynamically expand when passing a custom yaml @dolcalmi ?

nicolasburtey: (I think that would make sense and would be great, because it would make the custom fields discoverable from the API)

nicolasburtey:
> I have doubts about exposing the fields individually over GQL I was thinking something much simpler like a `string` field that accepts a JSON blob. We should probably discuss this at standup / code review.

Initially this is what I had in mind also, but I like the more stronger consistency of what @dolcalmi is suggesting. 
It make the API more self explanatory and really harness the power of Graphql here I think. 

> The downside I see is that a simple config change could mean a breaking API change - who is going to be responsible for this? Also what about the data already in the database - it could easily become inconsistent with the GQL definition.

This would put the burden on consistency on the operator of the instance/backend. But it gives guarantees to the API client. 

The JSON as a string opinion would put the burden on the client.

default.yaml
dolcalmi: updated https://github.com/GaloyMoney/galoy/pull/1465/commits/c536cd781df7aba2f2611bb8675e4ac38188f347

src/graphql/main/schema.graphql
dolcalmi: > will the `schema.graphql` dynamically expand when passing a custom yaml @dolcalmi ?

yes, it will

src/config/schema.ts
dolcalmi: done https://github.com/GaloyMoney/galoy/pull/1465/commits/c536cd781df7aba2f2611bb8675e4ac38188f347#diff-a25ae601baae9bca444c00d66e4934596768a26eafd5208c8fa077dcf31b4ae9L231

src/config/schema.types.d.ts
dolcalmi: done https://github.com/GaloyMoney/galoy/pull/1465/commits/c536cd781df7aba2f2611bb8675e4ac38188f347#diff-f10c7a06ab8f35c61100a7958d70ddfba331341e82be0017d9c7b955acdc230fR19

src/graphql/root/mutation/account-data-update.ts
```diff
@@ -0,0 +1,60 @@
...
+const customFieldsSchema = getAccountsConfig().customFields
+
+const AccountDataUpdateInput = GT.Input({
+  name: "AccountDataUpdateInput",
+  fields: () => {
+    const customFields = parseCustomFieldsSchema(customFieldsSchema)
+
+    return {
+      transactionsCallback: { type: GT.String },
+      ...customFields,
```
UncleSamtoshi, line 19: Since certain custom fields are required, will they always have to be passed to update any AccountData?

src/graphql/root/mutation/account-data-update.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const customFieldsSchema = getAccountsConfig().customFields
+
+const AccountDataUpdateInput = GT.Input({
+  name: "AccountDataUpdateInput",
+  fields: () => {
+    const customFields = parseCustomFieldsSchema(customFieldsSchema)
+
+    return {
+      transactionsCallback: { type: GT.String },
```
UncleSamtoshi, line 18: If ```transactioonsCallback``` is something we want present on all accounts, I think it should be separated from the customFields in how it is updated because it is not conceptually the same as customer specifc metadata and will impact behavior in the backend where as other metadata does not effect any behavior of the backend. Additionally, custom fields can override ```transactionsCallback``` since it is destructured after ```transactionsCallback``` is defined.

src/graphql/types/object/account-data.ts
```diff
@@ -0,0 +1,20 @@
+import { getAccountsConfig } from "@config"
+
+import { GT } from "@graphql/index"
+import { parseCustomFieldsSchema } from "@graphql/helpers"
+
+const customFieldsSchema = getAccountsConfig().customFields
+
+const AccountData = GT.Object({
```
UncleSamtoshi, line 8: What happens if a new required field is introduced? Will reading even non required fields fail because it is trying to resolve required fields? Or will it not try and resolve required fields unless they are specifically requested in the query?

src/graphql/root/mutation/account-data-update.ts
dolcalmi: yes (also an independent record is persisted in db every time the data is sent), if we want a different behavior we will need to add a mutation for `creation` and other to `update` 
@nicolasburtey any preference here?

src/graphql/root/mutation/account-data-update.ts
dolcalmi: done https://github.com/GaloyMoney/galoy/pull/1465/commits/c882fa2a11dbbb176ea1eff41e61cd8e6b702fbf

src/graphql/types/object/account-data.ts
dolcalmi: we will need to setup a default value. Added here https://github.com/GaloyMoney/galoy/pull/1465/commits/7b84e3ff269f84d560fa037334943463ec3144fd

nicolasburtey (commented):
some first questions. haven't done much of a review yet, will spend more time on it tomorrow.

src/app/accounts/update-account-custom-fields.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (account instanceof Error) return account
+
+  const accountCustomFields = await accountCustomFieldsRepo.findById(account.id)
+  const isEmpty = accountCustomFields instanceof CouldNotFindError
+  if (accountCustomFields instanceof Error && !isEmpty) return accountCustomFields
+
+  let data = customFields
+  if (!isEmpty) {
+    data = { ...accountCustomFields.customFields, ...customFields }
+  }
```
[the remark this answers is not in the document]
dolcalmi: `accountCustomFields` is an error if `isEmpty === true` so this suggestion will throw an exception

src/app/accounts/update-account-custom-fields.ts
```diff
@@ -0,0 +1,32 @@
...
+import { AccountCustomFieldsRepository, AccountsRepository } from "@services/mongoose"
+
+export const updateAccountCustomFields = async ({
+  accountId,
+  modifiedByUserId,
+  customFields,
+}: {
+  accountId: AccountId
+  modifiedByUserId: UserId
+  customFields: { [k: string]: string | number | boolean }
```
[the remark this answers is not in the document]
dolcalmi: uhmm we dont have bigint type in GT https://github.com/GaloyMoney/galoy/blob/main/src/graphql/index.ts#L33 @samerbuna any comment on this?

default.yaml
```diff
@@ -125,6 +125,30 @@ rateLimits:
 
 accounts:
   initialStatus: active
+  # customFields:
+  #   - name: "firstName"
```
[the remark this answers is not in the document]
dolcalmi: I added some comments https://github.com/GaloyMoney/galoy/pull/1465/commits/c5cdf416b447bbeb2dca0379ef00ce7e74d90ff6 do you think is enough? maybe we can update docs in another PR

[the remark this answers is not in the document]
dolcalmi: also there is an example in this gist https://gist.github.com/dolcalmi/42f1622f797b1734e356315c517084d5

default.yaml
```diff
@@ -125,6 +125,18 @@ rateLimits:
 
 accounts:
   initialStatus: active
+  # customFields:
+  #   - name: "customField"
+  #     type: <"string" | "integer" | "float" | "boolean">
+  #     defaultValue: <value> # optional
+  #     # default: false - true if field is an array of type
```
nicolasburtey, line 132: this line is not clear to me. do you mean `default: false` expect when `type: array`, in that case default is true?

default.yaml
```diff
@@ -125,6 +125,18 @@ rateLimits:
...
+  #     array: <true | false> # optional
+  #     # default: false - true if field can be updated by the user
```
nicolasburtey, line 134: does this applies to array? not sure to me that make sense as is

default.yaml
nicolasburtey: ok I think i get it: `#default` refers to the line below. not the line above. I think that could be make more clear by adding some line breaks.

src/app/accounts/get-account-custom-fields.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (val.defaultValue !== undefined) {
+    acc[val.name] = val.defaultValue
+  }
+  return acc
+}, {} as { [k: string]: AccountCustomFieldValues })
+
+export const getAccountCustomFields = async (
+  accountId: AccountId,
+): Promise<AccountCustomFields | ApplicationError> => {
+  if (!customFieldsSchema || customFieldsSchema.length <= 0)
```
nicolasburtey, line 19: why should it be an error if length is zero?

src/services/mongoose/schema.ts
```diff
@@ -351,3 +351,28 @@ const PhoneCodeSchema = new Schema({
 })
 
 export const PhoneCode = mongoose.model("PhoneCode", PhoneCodeSchema)
+
+const AccountCustomFieldsSchema = new Schema<AccountCustomFieldsRecord>({
+  accountId: Schema.Types.ObjectId,
+  customFields: Schema.Types.Mixed,
+  modifiedByUserId: Schema.Types.ObjectId,
```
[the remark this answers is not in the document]
nicolasburtey: otherwise we can rename this field `createdByUserId`?

default.yaml
```diff
@@ -125,6 +125,18 @@ rateLimits:
 
 accounts:
   initialStatus: active
+  # customFields:
```
nicolasburtey, line 128: I'm thinking we may need a `unique` property also?

(it almost like we're redefining a schema language here. wondering if there could be an approach where we just use a library? this would remove some burden on the maintenance)

src/app/accounts/update-account-custom-fields.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  let data = customFields
+  if (!isEmpty) {
+    data = { ...accountCustomFields.customFields, ...customFields }
+  }
+
+  return accountCustomFieldsRepo.persistNew({
+    accountId: account.id,
+    modifiedByUserId,
+    customFields: { ...defaultValues, ...data },
```
nicolasburtey, line 46: we're using defaultFields in two places. here (on write), but also https://github.com/GaloyMoney/galoy/pull/1465/files#diff-f5f42167109591eaca82e92d001bdd96f4db522ee0da89816e3ec9a56e7ff15bR25, on read. 

I'm wondering if this is a good behavior. 

I've never really design any database system like this, but my intuition tells me that default value would either be on read, or write, but not both.

nicolasburtey: one way to go with it is to not have any default settings. maybe that would be an option

at least for the current use case, I don't see a need to have default values

src/app/accounts/update-account-custom-fields.ts
```diff
@@ -0,0 +1,48 @@
...
+  const accountCustomFieldsRepo = AccountCustomFieldsRepository()
+
+  const account = await AccountsRepository().findById(accountId)
+  if (account instanceof Error) return account
+
+  const accountCustomFields = await accountCustomFieldsRepo.findById(account.id)
+  const isEmpty = accountCustomFields instanceof CouldNotFindError
+  if (accountCustomFields instanceof Error && !isEmpty) return accountCustomFields
+
+  let data = customFields
```
nicolasburtey, line 38: I guess the last object would have all the property for the accountId?

src/app/accounts/update-account-custom-fields.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (!isEmpty) {
+    data = { ...accountCustomFields.customFields, ...customFields }
```
nicolasburtey, line 40: I guess the last object would have all the property for the accountId?

nicolasburtey, line 40: one tricky part here is how to merge the field that are array. I guess here also the current behavior is that we would replace the array with the new value? (and not merge the array recursively?)

src/config/schema.types.d.ts
```diff
@@ -16,6 +16,16 @@ type AccountLimitsConfig = {
   }
 }
 
+type CustomField = {
+  name: string
+  type?: "string" | "integer" | "float" | "boolean"
```
nicolasburtey, line 21: I think this should be mandatory

default.yaml
nicolasburtey: maybe this is what `index` does. not sure to me.

src/config/schema.ts
```diff
@@ -219,6 +219,27 @@ export const configSchema = {
...
+                type: "string",
+                enum: ["string", "integer", "float", "boolean"],
+              },
+              defaultValue: { type: ["string", "number", "boolean", "array"] },
+              array: { type: "boolean", default: false },
+              required: { type: "boolean", default: false },
+              editable: { type: "boolean", default: false },
+              index: { type: "boolean", default: false },
+            },
+            required: ["name", "type"],
```
nicolasburtey, line 238: actually type is required here. ie: https://github.com/GaloyMoney/galoy/pull/1465/files#r944314491

src/config/yaml.ts
```diff
@@ -36,7 +36,7 @@ try {
 
 export const yamlConfigInit = merge(defaultConfig, customConfig)
 
-const ajv = new Ajv()
+const ajv = new Ajv({ allowUnionTypes: true })
```
nicolasburtey, line 39: why is that needed?

src/app/accounts/update-account-custom-fields.ts
nicolasburtey: note: this is the type of behavior that would be good to put in documentation.

src/domain/accounts/index.types.d.ts
```diff
@@ -46,6 +46,21 @@ type AccountStatusHistory = Array<{
...
+  | boolean
+  | string[]
+  | number[]
+  | boolean[]
+
+type AccountCustomFields = {
+  readonly accountId: AccountId
+  modifiedByUserId: UserId
+  customFields: { [k: string]: AccountCustomFieldValues }
+  createdAt: Date
```
nicolasburtey, line 61: shouldn't those be read only also?

for `createdAt` and `modifiedByUserId`, those values should be immutable also. 

for `customFields` it seems we are creating a new object when we're changing those property so it could be readonly also?

src/graphql/admin/mutations.ts
```diff
@@ -1,34 +1,70 @@
...
 
-    businessUpdateMapInfo: BusinessUpdateMapInfoMutation,
+  coldStorageRebalanceToHotWallet: ColdStorageRebalanceToHotWalletMutation,
+}
 
-    coldStorageRebalanceToHotWallet: ColdStorageRebalanceToHotWalletMutation,
-  }),
+const { customFields } = getAccountsConfig()
+if (customFields && customFields.length > 0) {
+  Object.assign(fields, { accountCustomFieldsUpdate: AccountCustomFieldsUpdateMutation })
```
nicolasburtey, line 40: When I was thinking about it, I had a different idea in mind in regards to the API. I guess this is related to the code review discussion we had Tuesday, that we may want to have specific use case for mutation.

instead of having a generic `AccountCustomFieldsUpdateMutation`,
we could create one API endpoint per customField

for instance, we could have:
```
AccountSetFirstname
AccountSetLastname
```
etc.

the result would be that the implementation of the custom field doesn't explicitly appear in the API, which I think is better. 

but this is just my thoughts. we should probably gain some consensus on this one.

src/graphql/admin/mutations.ts
```diff
@@ -1,34 +1,70 @@
...
+        },
+        // @ts-ignore-next-line no-implicit-any error
+        () => original(source, args, context, info),
+      )
+    }
+  }
+  return fields
+}
+
+const MutationType = GT.Object({
```
nicolasburtey, line 65: there was no such tracing in mutation before?

src/graphql/admin/queries.ts
```diff
@@ -7,22 +9,32 @@ import TransactionByIdQuery from "./root/query/transaction-by-id"
...
-    accountDetailsByUserPhone: AccountDetailsByUserPhoneQuery,
-    accountDetailsByUsername: AccountDetailsByUsernameQuery,
-    transactionById: TransactionByIdQuery,
-    transactionsByHash: TransactionsByHashQuery,
-    lightningInvoice: LightningInvoiceQuery,
-    lightningPayment: LightningPaymentQuery,
-    listWalletIds: ListWalletIdsQuery,
-    wallet: WalletQuery,
-  }),
+  fields: () => fields,
```
nicolasburtey, line 37: no adding tracing here? why is there an asymmetry with the mutation?

src/graphql/admin/root/query/accounts-details-by-custom-field.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { GT } from "@graphql/index"
+import { mapError } from "@graphql/error-map"
+import GraphQLAccount from "@graphql/admin/types/object/account"
+
+const { customFields: customFieldsSchema } = getAccountsConfig()
+
+const fields: Readonly<GraphQLEnumValueConfigMap> = customFieldsSchema.reduce(
+  (acc, val) => {
+    if (!val.array) {
```
nicolasburtey, line 15: why would you filter arrays?

src/graphql/types/object/business-account.ts
```diff
@@ -1,97 +1,129 @@
-import { CouldNotFindTransactionsForAccountError } from "@domain/errors"
-import { GT } from "@graphql/index"
-import { mapError } from "@graphql/error-map"
-import { connectionArgs, connectionFromArray } from "graphql-relay"
+import getUuidByString from "uuid-by-string"
```
nicolasburtey, line 1: maybe we can delete this business account until we have clear support for it. it would limit the code duplication/overhead on working on such PR in the meantime?

src/app/accounts/update-account-custom-fields.ts
bodymindarts: Do we want to support arrays? I don't think the use case we are aiming for here needs it and it adds a lot of complexity. Like you mention above we don't want to define our own schema definition language and merging semantics.

src/app/accounts/update-account-custom-fields.ts
bodymindarts: IMO defaults should be set on write (not read). But agree to keep the implementation to the minimum for the use case at hand.

src/domain/accounts/index.types.d.ts
```diff
@@ -113,15 +128,37 @@ type AccountValidator = {
...
+  customFields: { [k: string]: AccountCustomFieldValues }
+}
+
 interface IAccountsRepository {
   listUnlockedAccounts(): Promise<Account[] | RepositoryError>
   findById(accountId: AccountId): Promise<Account | RepositoryError>
   findByUserId(userId: UserId): Promise<Account | RepositoryError>
   findByUsername(username: Username): Promise<Account | RepositoryError>
   listBusinessesForMap(): Promise<BusinessMapMarker[] | RepositoryError>
+  listByIds(accountIds: AccountId[]): Promise<Account[] | RepositoryError>
```
bodymindarts, line 148: Wondering if we should have an explicit `IAdminAccountsRepository` or something like that.

bodymindarts:
Wonder if these tools can help: https://github.com/thenativeweb/get-graphql-from-jsonschema https://github.com/lifeomic/json-schema-to-graphql-types

nicolasburtey:
@bodymindarts highlighted that a benefits of the api that would be specific to the use case, as opposed to the more generic interface where one has to pass a key/value pair like in this PR, is that it will help to make it work/break at compile time, instead of run time.

src/services/mongoose/schema.ts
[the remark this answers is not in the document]
dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/1465/commits/542bbf3a35caa5348ca579e8d729604e323bc53c

default.yaml
dolcalmi: we create an index but not a unique index. Also this is a bit tricky to implement because need to have track of changes (duplicated records).

let me know if you want to include it in this PR

src/app/accounts/update-account-custom-fields.ts
dolcalmi: we need default values and also we need them on read because possible updates on required fields. Please check Sam's comment  https://github.com/GaloyMoney/galoy/pull/1465#discussion_r932725226

src/app/accounts/update-account-custom-fields.ts
dolcalmi: does not apply because we have optional fields (input) and we need to include current values in the last db record

dolcalmi: this was an additional request after first implementation. In this case full array is replaced

src/config/schema.types.d.ts
dolcalmi: type is not required, was an issue in schema validator https://github.com/GaloyMoney/galoy/pull/1465/commits/c77987b90c2efa222f03c8aa24b6e792af811ba2 also a unit test was added previously to validate default type https://github.com/GaloyMoney/galoy/pull/1465/files#diff-f2dc8f7e9dae5e09bcaebb4926c4fac76f9ec33907320097ae665330c04aa1afR27

src/config/schema.ts
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/commit/c77987b90c2efa222f03c8aa24b6e792af811ba2 also please check comment here https://github.com/GaloyMoney/galoy/pull/1465#discussion_r948646070

src/config/yaml.ts
dolcalmi: for default value in schema validator
```
defaultValue: { type: ["string", "number", "boolean", "array"] },
```

src/domain/accounts/index.types.d.ts
dolcalmi: Updated here https://github.com/GaloyMoney/galoy/pull/1465/commits/2d687e1d6ee24f6539fe767272ac77a138f8084a but customFields cant be readonly because need to be updated (default values use case)

src/graphql/admin/mutations.ts
dolcalmi: Changes according to code review https://github.com/GaloyMoney/galoy/pull/1465/commits/7ee35631e7b7da09da9382ced0a46dd02960d8a6

src/graphql/admin/mutations.ts
dolcalmi: No but missing it in admin queries https://github.com/GaloyMoney/galoy/pull/1465/commits/969d2247d4e9da3054663dd7a280800526daa6cd

src/graphql/admin/queries.ts
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/1465/commits/969d2247d4e9da3054663dd7a280800526daa6cd

src/graphql/admin/root/query/accounts-details-by-custom-field.ts
dolcalmi: Serialization (db and grapqhl), I tried different alternatives but was taking longer than expected. Any suggestion is welcome (input is a string and db aggregation use $toString to enable partial matches)

src/graphql/types/object/business-account.ts
dolcalmi: yep, you are right, this file is identical to consumer account and was not included in the schema https://github.com/GaloyMoney/galoy/pull/1465/commits/7a27835ff7d528473a4e67248f506f87f03228ec

src/domain/accounts/index.types.d.ts
dolcalmi: I thought the same but probably will end duplicating code.. any specific concerns about having this function in this repo?

nicolasburtey:
@dolcalmi can you add related documentation in `/doc`?

nicolasburtey:
> @dolcalmi can you add related documentation in `/doc`?

maybe something like that: https://github.com/GaloyMoney/galoy/blob/3744882b7cc03b34cb5081716c47158b597b812a/docs/swap.md

src/graphql/types/object/account-data.ts
nicolasburtey: I think doing an explicit db migration is better than having default value. 

it reduce the complexity in the code here, and the complexity to the migration, which I think is better:

those custom fields are not really meant to be dynamic and changed frequently.
when they changed, there might specific things to do on those changes, which I don't think is easy to manage from here, because each case will be different.

src/app/accounts/update-account-custom-fields.ts
nicolasburtey: I'm not sure I agree with Sam input, my response here: https://github.com/GaloyMoney/galoy/pull/1465#discussion_r949023827

src/app/accounts/update-account-custom-fields.ts
nicolasburtey: note: there is no need for Array with the current use case. they're ok with one response only instead of an array of response possible.

default.yaml
```diff
@@ -130,6 +130,19 @@ rateLimits:
...
+  #     type: <"string" | "integer" | "float" | "boolean"> # optional
+  #     defaultValue: <value> # optional
+  #     # default: false - true if field is an array of type
+  #     array: <true | false> # optional
+  #     # default: false - true if field can be updated by the user
+  #     editable: <true | false> # optional
+  #     # default: false - true if field is required by mutations
+  #     required: <true | false> # optional
+  #     # default: false - true if field should have an index in db
+  #     index: <true | false> # optional
```
nicolasburtey, line 145: no default for this one?

default.yaml
nicolasburtey: `unique` feels more important than `array` or `default`.

nicolasburtey:
note: in the admin API, there is a `key, value` pair query that we could change with a more specific query.

instead of 

```
type Query { 
  accountsDetailsByCustomField(field: CustomField, value: String!): [Account!]!
}
```

we could have 
```
type Query { 
  accountsDetailsByCustomFieldFirstName(value: String!): [Account!]!
  accountsDetailsByCustomFieldLastName(value: String!): [Account!]!
  accountsDetailsByCustomFieldID(value: String!): Account! # not need for an array if ID is unique
}
```

nicolasburtey:
> note: in the admin API, there is a `key, value` pair query that we could change with a more specific query.
> 
> instead of
> 
> ```
> type Query { 
>   accountsDetailsByCustomField(field: CustomField, value: String!): [Account!]!
> }
> ```
> 
> we could have
> 
> ```
> type Query { 
>   accountsDetailsByCustomFieldFirstName(value: String!): [Account!]!
>   accountsDetailsByCustomFieldLastName(value: String!): [Account!]!
>   accountsDetailsByCustomFieldID(value: String!): Account! # not need for an array if ID is unique
> }
> ```

Another reason to remove the `Array` usage from this PR: we can't pass an array in this query

Also, the value could be Int? or Boolean? here it's hardcoded as String

src/app/accounts/get-accounts-by-custom-fields.ts
```diff
@@ -0,0 +1,26 @@
...
+  field,
+  value,
+}: {
+  field: string
+  value: string
+}): Promise<Account[] | ApplicationError> => {
+  const fieldSchema = customFieldsSchema.find((s) => s.name === field)
+  if (!fieldSchema) return new InvalidCustomFieldError()
+
+  const accountIds = await AccountCustomFieldsRepository().listAccountIdsByCustomField({
```
bodymindarts, line 19: Up to now we have favoured the style of `listByCustomField` and then filtered `.map(|f| f.account_id)` in the domain.
This here is encoding use case into a repo which should purely be for data access.

src/app/accounts/update-account-custom-fields.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (!customFieldsSchema || customFieldsSchema.length <= 0)
+    return new NoAccountCustomFieldsError()
+
+  const accountCustomFieldsRepo = AccountCustomFieldsRepository()
+
+  const account = await AccountsRepository().findById(accountId)
+  if (account instanceof Error) return account
+
+  const accountCustomFields = await accountCustomFieldsRepo.findById(account.id)
+  const isEmpty = accountCustomFields instanceof CouldNotFindError
```
bodymindarts, line 35: We should early return on `isEmpty` or else we are coding an `upsert` use case (even though this is supposed to be an `update` use case).
I am NOT in favour of exposing `upsert`

src/app/accounts/update-account-custom-fields.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (accountCustomFields instanceof Error && !isEmpty) return accountCustomFields
+
+  let data = customFields
+  if (!isEmpty) {
+    data = { ...accountCustomFields.customFields, ...customFields }
+  }
+
+  return accountCustomFieldsRepo.persistNew({
```
bodymindarts, line 43: why are you using `repo.persistNew` in a use case called `update`?
Repos should expose `CRUD` methods - don't use `persistNew` (the `C` method) for an `update`.

src/config/schema.types.d.ts
bodymindarts: I don't understand - why is this optional?

src/domain/accounts/index.types.d.ts
```diff
@@ -113,15 +128,37 @@ type AccountValidator = {
...
   update(account: Account): Promise<Account | RepositoryError>
 }
 
+interface IAccountCustomFieldsRepository {
```
bodymindarts, line 152: Missing the `update` function.

src/domain/accounts/index.types.d.ts
```diff
@@ -113,15 +128,37 @@ type AccountValidator = {
...
+  findById(accountId: AccountId): Promise<AccountCustomFields | RepositoryError>
+  listAccountIdsByCustomField(
```
bodymindarts, line 154: Should be `listByCustomField`

src/config/schema.ts
```diff
@@ -219,6 +219,27 @@ export const configSchema = {
...
+          type: "array",
+          items: {
+            type: "object",
+            properties: {
+              name: { type: "string" },
+              type: {
+                type: "string",
+                enum: ["string", "integer", "float", "boolean"],
+              },
+              defaultValue: { type: ["string", "number", "boolean", "array"] },
```
bodymindarts, line 232: Why can default be an array?

src/config/schema.ts
```diff
@@ -219,6 +219,27 @@ export const configSchema = {
...
+              array: { type: "boolean", default: false },
```
bodymindarts, line 233: I thought we shall not support arrays initially?

src/config/schema.ts
dolcalmi: because we had array fields but now were removed so this was updated

src/config/schema.ts
dolcalmi: Removed

src/domain/accounts/index.types.d.ts
dolcalmi: could you please check the query? (it is an aggregate) really is only account ids query, there is no way to query the additional fields without do another query. do we want to do that?

src/app/accounts/update-account-custom-fields.ts
dolcalmi: in this moment it is not to do an upsert but we need the previous values to create the new record

src/domain/accounts/index.types.d.ts
dolcalmi: Updated with the new model https://github.com/GaloyMoney/galoy/pull/1465/commits/22c9f29082861a0a3ae06bfd1c0c82247e53368b

src/app/accounts/update-account-custom-fields.ts
dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/1465/commits/22c9f29082861a0a3ae06bfd1c0c82247e53368b

src/app/accounts/update-account-custom-fields.ts
dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/1465/commits/22c9f29082861a0a3ae06bfd1c0c82247e53368b

src/app/accounts/get-accounts-by-custom-fields.ts
dolcalmi: Updated with the new model https://github.com/GaloyMoney/galoy/pull/1465/commits/22c9f29082861a0a3ae06bfd1c0c82247e53368b

default.yaml
dolcalmi: updated in docs https://github.com/GaloyMoney/galoy/pull/1465/files#diff-3233f46d946532b99e17ec18cdf28ed0917b52aadcb201563cbd167491197709R14

src/app/accounts/update-account-custom-fields.ts
dolcalmi: array support was dropped

src/config/schema.types.d.ts
dolcalmi: default is string, only name is required

src/graphql/admin/root/query/accounts-details-by-custom-field.ts
dolcalmi: does not apply after remove array

default.yaml
dolcalmi: Unique added here https://github.com/GaloyMoney/galoy/pull/1465/commits/365d353f84a0c792aac7b84ddf202e2cc7c96f44

dolcalmi:
> > note: in the admin API, there is a `key, value` pair query that we could change with a more specific query.
> > instead of
> > ```
> > type Query { 
> >   accountsDetailsByCustomField(field: CustomField, value: String!): [Account!]!
> > }
> > ```
> > we could have
> > ```
> > type Query { 
> >   accountsDetailsByCustomFieldFirstName(value: String!): [Account!]!
> >   accountsDetailsByCustomFieldLastName(value: String!): [Account!]!
> >   accountsDetailsByCustomFieldID(value: String!): Account! # not need for an array if ID is unique
> > }
> > ```
> 
> Another reason to remove the `Array` usage from this PR: we can't pass an array in this query
> 
> Also, the value could be Int? or Boolean? here it's hardcoded as String

I know we already talked about this but for clarification:
- field param validation is enforced by grapqhl schema
- the suggested approach is ok for few fields but if we have 20 will be really difficult to implement in the frontend
- we search by string because we need to do partial searches. This is not possible for numbers by default (or other field types)

src/app/accounts/get-accounts-by-custom-fields.ts
```diff
@@ -0,0 +1,27 @@
+import { getAccountsConfig } from "@config"
+
+import { InvalidCustomFieldError } from "@domain/accounts"
+
+import { AccountCustomFieldsRepository, AccountsRepository } from "@services/mongoose"
+
+const { customFields: customFieldsSchema } = getAccountsConfig()
+
+export const getAccountsByCustomFields = async ({
```
bodymindarts, line 9: Small nit - should be `getAccountsByCustomField` (not `Fields`)

src/app/accounts/update-account-custom-fields.ts
```diff
@@ -0,0 +1,51 @@
...
+  accountId,
+  updatedByUserId,
+  customFields,
+}: {
+  accountId: AccountId
+  updatedByUserId: UserId
+  customFields: { [k: string]: AccountCustomFieldValues }
+}): Promise<AccountCustomFields | ApplicationError> => {
+  if (!customFieldsSchema || customFieldsSchema.length <= 0)
+    return new NoAccountCustomFieldsError()
```
bodymindarts, line 27: AFAICT this should never be called since the GQL that would access this is not included if they aren't configured. Correct?

src/domain/accounts/errors.ts
```diff
@@ -4,3 +4,9 @@ export class AccountError extends DomainError {}
 
 export class UsernameNotAvailableError extends AccountError {}
 export class UsernameIsImmutableError extends AccountError {}
+
+export class AccountCustomFieldsError extends DomainError {}
+
+export class InvalidCustomFieldError extends AccountCustomFieldsError {}
+export class NoAccountCustomFieldsError extends AccountCustomFieldsError {}
```
bodymindarts, line 11: This one should be `critical` as it should never happen.

src/graphql/admin/types/object/account.ts
```diff
@@ -1,59 +1,88 @@
...
+    }
+
+    if (customFields && customFields.length > 0) {
+      return Object.assign(fields, {
+        customFields: {
+          type: AccountCustomFields,
+          resolve: async (source: Account) => {
+            const accountCustomFields = await Accounts.getAccountCustomFields(source.id)
+            if (accountCustomFields instanceof CouldNotFindError) return null
+            if (accountCustomFields instanceof NoAccountCustomFieldsError) return null
```
bodymindarts, line 75: Shouldn't this one throw?

bodymindarts (approved):

docs/custom-account-fields.md
```diff
@@ -0,0 +1,182 @@
...
+This document shows you how to setup account custom fields for your galoy instance.
+
+## Configuration
+
+```
+customFields:
+  - name: "customFieldName"
+    type: <"string" | "integer" | "float" | "boolean"> # optional (default: "string")
+    defaultValue: <value> # optional
+    editable: <true | false> # optional (default: false) - true if field can be updated by the user
```
nicolasburtey, line 12: is it meant as: if false, only admin can edit?

docs/custom-account-fields.md
```diff
@@ -0,0 +1,182 @@
...
+    - name: "lastName"
+      type: "string"
+      editable: true
+    - name: "terms"
+      type: "boolean"
+      editable: true
+      defaultValue: false
+```
+
+If customFields config is empty then all related mutations and queries should not be visible by API clients.
```
nicolasburtey, line 41:
```suggestion
If customFields config is empty then all related mutations and queries will not be visible by API clients.
```

docs/custom-account-fields.md
```diff
@@ -0,0 +1,182 @@
...
+## Example
+
+Add the next configuration to `default.yaml`
+```
+accounts:
+  ...
+  customFields:
+    - name: "nationalId"
+      type: "integer"
+      defaultValue: 0
```
nicolasburtey, line 27: isn't it a use case where `index` and `unique` should be true, and not have a defaultValue?

docs/custom-account-fields.md
```diff
@@ -0,0 +1,182 @@
...
+      type: "string"
+      required: true
+      editable: true
+    - name: "lastName"
+      type: "string"
+      editable: true
+    - name: "terms"
+      type: "boolean"
+      editable: true
+      defaultValue: false
```
nicolasburtey, line 38: I think term should be more an integer. so that a user can approve any changes when terms are been updated?

0 may mean user has not accepted, but after that it's the term version number the customer has accepted.

docs/custom-account-fields.md
```diff
@@ -0,0 +1,182 @@
...
+        firstName
+        lastName
+        terms
+      }
+    }
+  }
+}
+```
+
+Note: please notice that `nationalId` is only visible as an editor (admin API)
```
nicolasburtey, line 89: I think the `editable` is conflating multiple use case. 
- it's possible that some fields could be write-once from the end user. Ie: the nationalId could be something submitted by the end user. 
-- similarly, the firstName is (in my mind) write-once by the user, but should not be editable.

- a field, even if not editable, could be still visible by the end user (I'm not sure about that use case, trying to think about it?)


The admin should be able to see/edit everything IMO, at least for now.

nicolasburtey: if we had a `visible` field, then this could be the flag on whether the mutation and query would be present or not for the end user API.

the `editable` field would then only be relevant if `visible: true` and if yes, means that the field is write only, if no, means that the fields can be write several time

wdyt?

docs/custom-account-fields.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Update values as an editor
```
nicolasburtey, line 91: I think the `editor` name is not very well defined. using `admin` for the `admin API` may a bit more clear this is a different endpoint.

```suggestion
### Update values as an admin
```

docs/custom-account-fields.md
```diff
@@ -0,0 +1,182 @@
...
+      firstName
+      lastName
+      terms
+    }
+    createdAt
+  }
+}
+```
+
+Input (updated user's phone)
```
nicolasburtey, line 132: why `updated`?

docs/custom-account-fields.md
```diff
@@ -0,0 +1,182 @@
...
+      firstName
+      lastName
+      terms
+    }
+    createdAt
+  }
+}
+```
+
+Input (updated user's username)
```
nicolasburtey, line 152: why `updated`?

we have not updated either phone number or username above?

docs/custom-account-fields.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+Input
+```
+{ "field": "nationalId", "value": "212"}
+or
+{ "field": "firstName", "value": "sato"}
+```
+
+Note: field is validated similar to the price query, only valid field names are valid values
```
nicolasburtey, line 182: `Note: field is validated similar to the price query,`: I'm not sure this will make sense to a new reader.

src/app/accounts/update-account-custom-fields.ts
```diff
@@ -0,0 +1,51 @@
...
+export const updateAccountCustomFields = async ({
+  accountId,
+  updatedByUserId,
+  customFields,
+}: {
+  accountId: AccountId
+  updatedByUserId: UserId
+  customFields: { [k: string]: AccountCustomFieldValues }
+}): Promise<AccountCustomFields | ApplicationError> => {
+  if (!customFieldsSchema || customFieldsSchema.length <= 0)
```
nicolasburtey, line 26:
```suggestion
  if (!customFieldsSchema || customFieldsSchema.length === 0)
```

docs/custom-account-fields.md
nicolasburtey: could also be `viewable` instead of `visible`

nicolasburtey (commented):
I've made some comments on the documentation. 

should we had another field `visible` or `viewable`?

still a big question for me whether we should have default value imo. (I think not, maybe we should sync up on this)

nicolasburtey:
> I've made some comments on the documentation.
> 
> should we had another field `visible` or `viewable`?
> 
> still a big question for me whether we should have default value imo. (I think not, maybe we should sync up on this)

And I know it might be late to think about this, but as we want to use Postgres, I'm wondering if this is not a use case that we should integrate directly into postgres? 

It would be a bit more work now, but ultimately, it would avoid a migration from this new collection to a table.

(not saying we should do this now, but just opening the topic)

docs/custom-account-fields.md
bodymindarts: if `visible` and `editable` are connected - perhaps there is a nested structure:
```
expose_to_user:
  visible: true
  editable: false
```

?

bodymindarts:
> > I've made some comments on the documentation.
> > should we had another field `visible` or `viewable`?
> > still a big question for me whether we should have default value imo. (I think not, maybe we should sync up on this)
> 
> And I know it might be late to think about this, but as we want to use Postgres, I'm wondering if this is not a use case that we should integrate directly into postgres?
> 
> It would be a bit more work now, but ultimately, it would avoid a migration from this new collection to a table.
> 
> (not saying we should do this now, but just opening the topic)

This is as good as any time - if we can handle the delay.
At the latest when I continue my work with casbin we would add it https://github.com/GaloyMoney/galoy/pull/1352

nicolasburtey:
I'm reading https://www.ory.sh/docs/kratos/manage-identities/overview and I'm wondering if the right data store for the custom field is actually not kratos? maybe we can discuss this in a call (still learning about kratos, but seeing how they managed identities seems like a potential good/better fit than mongodb?)

dolcalmi:
closed until we decide a better architecture

nicolasburtey:
re-opening @dolcalmi. we may still need this soon. now that I have a better understand of the data flow of kratos I can have another view on this.

PR #1470, opened 2022-07-22
why does this compile

bodymindarts:
`make check-code` fails without the type definitions in the method
```
(130) $ make check-code
yarn tsc-check-noimplicitany
yarn run v1.22.18
$ tsc --noEmit -p tsconfig.no-implicit-any.json --skipLibCheck
src/domain/payments/payment-flow-builder.ts:81:5 - error TS7031: Binding element 'uncheckedAmount' implicitly has an 'any' type.

81     uncheckedAmount,
       ~~~~~~~~~~~~~~~

src/domain/payments/payment-flow-builder.ts:82:5 - error TS7031: Binding element 'description' implicitly has an 'any' type.

82     description,
       ~~~~~~~~~~~


Found 2 errors in the same file, starting at: src/domain/payments/payment-flow-builder.ts:81

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
make: *** [check-implicit] Error 2
```

### DIFF
--- a/src/domain/payments/index.types.d.ts
+++ b/src/domain/payments/index.types.d.ts
@@ -82,7 +82,7 @@ type LightningPaymentFlowBuilder<S extends WalletCurrency> = {
     description,
   }: {
     uncheckedAmount: number
-    description: string | null
+    description: string
   }): LPFBWithInvoice<S> | LPFBWithError
 }
 
--- a/src/domain/payments/index.types.d.ts
+++ b/src/domain/payments/index.types.d.ts
@@ -82,7 +82,7 @@ type LightningPaymentFlowBuilder<S extends WalletCurrency> = {
     description,
   }: {
     uncheckedAmount: number
-    description: string
+    description: string | null
   }): LPFBWithInvoice<S> | LPFBWithError
 }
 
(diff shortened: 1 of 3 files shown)


src/domain/payments/index.types.d.ts
```diff
@@ -82,7 +82,7 @@ type LightningPaymentFlowBuilder<S extends WalletCurrency> = {
     description,
   }: {
     uncheckedAmount: number
-    description: string
+    description: string | null
```
samerbuna, line 85: Why did you add `| null` here?

bodymindarts: That was how it was before the fix. I thought that is what you wanted to see

samerbuna: oh, yeah. Then this was the problem, since this is the type TS is using. The implementation type was not being picked up.

I pushed an example of how I'd use the main type in the implementation (since without it, we have implicit anys).

dolcalmi:
@bodymindarts can we close this?

nicolasburtey:
closing this. we can reopen if we want to rediscuss it

PR #1484, opened 2022-07-26
feat: switch invoice generation to hodl invoices & external preimages

vindard:
_New changes diff: https://github.com/GaloyMoney/galoy/compare/3703545...hodl-add-single-subscribe_

## Discussion

This is a continuation from #1395 where we need to fix the trigger code for subscribing to hodl invoice `ACCEPT`/`CANCEL` events that don't come through the usual `subscribeInvoices` listener (see [here](https://github.com/lightningnetwork/lnd/issues/3120#issuecomment-1160732579)).

This PR will be to bring back in the #1395 changes and to add the subscribe-to-single-invoice code on top of it to get this feature working.

### Open questions
1. ~New implementation: is there a possibility that we end up with single-invoice subscriptions that don't close (if we somehow skip one of the `removeAllListeners` calls)?~

2. ~Previous implementation: how does the `subscribeInvoices` listener get restarted if it errors and `removeAllListeners` gets called?~

## Transition ([see comment](https://github.com/GaloyMoney/galoy/pull/1484#issuecomment-1205762776))

@dolcalmi pointed out that when we first roll this out our "updateInvoice" code will need to handle both hodl and non-hodl invoices. This was an easy fix for the "update" code path ([here](https://github.com/GaloyMoney/galoy/pull/1484/commits/4ae0917c559fe85fa9eb17c9949248fe7f2691c6)). For the "decline held invoices" code path though, I added a temporary fix that we should probably revert as soon as 24 hours passes and we are confident that no old invoices are in the system.

Does this sound right @nicolasburtey?

### DIFF
--- a/src/app/wallets/update-pending-invoices.ts
+++ b/src/app/wallets/update-pending-invoices.ts
@@ -12,55 +12,34 @@ import {
   WalletsRepository,
 } from "@services/mongoose"
 import { NotificationsService } from "@services/notifications"
-import { runInParallel } from "@utils"
+import { elapsedSinceTimestamp, runInParallel } from "@utils"
 import { WalletInvoiceReceiver } from "@domain/wallet-invoices/wallet-invoice-receiver"
 import * as LedgerFacade from "@services/ledger/facade"
 import { usdFromBtcMidPriceFn } from "@app/shared"
 
-export const updatePendingInvoices = async (logger: Logger): Promise<void> => {
+export const declineHeldInvoices = async (logger: Logger): Promise<void> => {
   const invoicesRepo = WalletInvoicesRepository()
 
-  const walletIdsWithPendingInvoices = invoicesRepo.listWalletIdsWithPendingInvoices()
+  const pendingInvoices = invoicesRepo.yieldPending()
 
-  if (walletIdsWithPendingInvoices instanceof Error) {
+  if (pendingInvoices instanceof Error) {
     logger.error(
-      { error: walletIdsWithPendingInvoices },
+      { error: pendingInvoices },
       "finish updating pending invoices with error",
     )
     return
   }
 
   await runInParallel({
-    iterator: walletIdsWithPendingInvoices,
+    iterator: pendingInvoices,
     logger,
-    processor: async (walletId: WalletId, index: number) => {
-      logger.trace(
-        "updating pending invoices for wallet %s in worker %d",
-        walletId,
-        index,
-      )
-      await updatePendingInvoicesByWalletId({ walletId, logger })
+    processor: async (walletInvoice: WalletInvoice, index: number) => {
+      logger.trace("updating pending invoices %s in worker %d", index)
+      await declineHeldInvoice({ walletInvoice, logger })
     },
   })
 
   logger.info("finish updating pending invoices")
-}
-
-export const updatePendingInvoicesByWalletId = async ({
-  walletId,
-  logger,
-}: {
-  walletId: WalletId
-  logger: Logger
-}) => {
-  const invoicesRepo = WalletInvoicesRepository()
-
-  const invoices = invoicesRepo.findPendingByWalletId(walletId)
-  if (invoices instanceof Error) return invoices
-
-  for await (const walletInvoice of invoices) {
-    await updatePendingInvoice({ walletInvoice, logger })
-  }
 }
 
 export const updatePendingInvoiceByPaymentHash = async ({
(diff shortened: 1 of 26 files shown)


vindard:
This isn't a fix, I was just curious to see where the memory limit is in GIthub Actions: https://github.com/GaloyMoney/galoy/pull/1484/commits/b534c67f1239aa2530ce1391d399f38b3959edf1

vindard:
## Transition

@dolcalmi pointed out that when we first roll this out our "updateInvoice" code will need to handle both hodl and non-hodl invoices. This was an easy fix for the "update" code path ([here](https://github.com/GaloyMoney/galoy/pull/1484/commits/4ae0917c559fe85fa9eb17c9949248fe7f2691c6)). For the "decline held invoices" code path though, I added a temporary fix that we should probably revert as soon as 24 hours passes and we are confident that no old invoices are in the system.

Does this sound right @nicolasburtey?

src/servers/trigger.ts
```diff
@@ -178,16 +180,16 @@ export const onchainBlockEventHandler = async (height: number) => {
   logger.info(`finish block ${height} handler with ${txNumber} transactions`)
 }
 
-export const invoiceUpdateEventHandler = async (invoice: GetInvoiceResult) => {
+export const invoiceUpdateEventHandler = async (
+  invoice: SubscribeToInvoiceInvoiceUpdatedEvent,
+): Promise<boolean | ApplicationError> => {
```
nicolasburtey, line 192: I think at some point (not necessarily with this PR) we should have `Success` instead of `boolean`. if if this is an alias to a boolean, this would give a better insights into what is being returned.

src/servers/trigger.ts
```diff
@@ -178,16 +180,16 @@ export const onchainBlockEventHandler = async (height: number) => {
...
-
-  const paymentHash = invoice.id as PaymentHash
-
-  await Wallets.updatePendingInvoiceByPaymentHash({ paymentHash, logger })
+  return invoice.is_held
+    ? Wallets.updatePendingInvoiceByPaymentHash({
+        paymentHash: invoice.id as PaymentHash,
+        logger,
+      })
+    : false
```
nicolasburtey, line 199: what is the scenario for which `invoice.is_held` will be false?

src/servers/trigger.ts
```diff
@@ -232,14 +234,40 @@ const listenerOnchain = (lnd: AuthenticatedLnd) => {
...
+    async (invoice: SubscribeToInvoiceInvoiceUpdatedEvent) => {
+      if (invoice.is_confirmed === true) {
+        subInvoice.removeAllListeners()
+      } else {
+        await invoiceUpdateHandler(invoice)
+      }
+    },
+  )
+  subInvoice.on("error", (err) => {
+    baseLogger.info({ err }, "error subChannels")
```
nicolasburtey, line 268: maybe we could have an opentelemetry event here as well?

src/servers/trigger.ts
nicolasburtey: I guess the first call of `listenerHoldInvoice`?

src/services/lnd/index.ts
```diff
@@ -404,6 +388,30 @@ export const LndService = (): ILightningService | LightningServiceError => {
...
+        } = after ? { token: after } : {}
+        const { invoices, next } = await getInvoices({ lnd, ...pagingArgs })
+        rawInvoices = [...rawInvoices, ...invoices]
+        after = (next as PagingContinueToken) || false
+      }
+      return rawInvoices.map(translateLnInvoiceLookup)
+    } catch (err) {
+      const errDetails = parseLndErrorDetails(err)
+      switch (errDetails) {
+        default:
```
nicolasburtey, line 409: are 3 lines are useful in any way?

src/app/wallets/update-pending-invoices.ts
```diff
@@ -35,7 +35,14 @@ export const declineHeldInvoices = async (logger: Logger): Promise<void> => {
...
       logger.trace("updating pending invoices %s in worker %d", index)
-      await declineHeldInvoice({ walletInvoice, logger })
+
+      walletInvoice.recipientWalletDescriptor.currency === WalletCurrency.Btc
+        ? await updatePendingInvoice({ walletInvoice, logger })
+        : await declineHeldInvoice({
+            pubkey: walletInvoice.pubkey,
+            paymentHash: walletInvoice.paymentHash,
+            logger,
+          })
```
nicolasburtey, line 45: the difference in signature between `updatePendingInvoice` and `declineHeldInvoice` feels inadequate. but maybe it's ok?

src/servers/trigger.ts
```diff
@@ -265,14 +267,36 @@ const listenerHodlInvoice = ({
...
 
   const invoices = await lndService.listInvoices(lnd)
   if (invoices instanceof Error) return invoices
 
   for (const lnInvoice of invoices) {
+    if (lnInvoice.isHeld) {
+      const invoicesRepo = WalletInvoicesRepository()
+      const walletInvoice = await invoicesRepo.findByPaymentHash(lnInvoice.paymentHash)
+      if (
+        walletInvoice instanceof Error ||
```
nicolasburtey, line 291: I guess there could be some transient error here (like mongodb is offline) but that's probably ok. we would anyway decline the invoice anyhow in this scenario if mongodb were to be offline because we would not process the invoice immediately and cancel the invoice on a later cronjob

nicolasburtey (commented):
some comments reviewing the PR but looks good on my end overall. would be great to merge it in the next day or so.

src/servers/trigger.ts
vindard: Yup when the invoice is first created and when it is subsequently settled

src/servers/trigger.ts
vindard: This one is more to allow us to not decline btc-only invoices (`walletInvoice.recipientWalletDescriptor.currency !== WalletCurrency.Btc`) if we receive the event here but the invoice has been held for some time.

It would decline non-BTC invoice, but can still settle btc-only invoices if for example the trigger were to restart and it finds held invoices in lnd.

src/services/lnd/index.ts
vindard: I did this as a consistency thing to signal to anybody else maintaining these. All the `catch` blocks in the lightning service have this pattern so that we can easily add known lnd error details as new cases and handle them in the switch as we discover them.

For some catch blocks, we haven't come across specific lnd errors we'd like to handle as yet, so their switches only have the default switch case.

src/app/wallets/update-pending-invoices.ts
nicolasburtey: yes but in one case we pass `walletInvoice`, in the other case we pass `walletInvoice.pubkey` and `walletInvoice.paymentHash`, in so both case we have walletInvoice ?

feels to me that the if we were doing the destructuring of the walletInvoice inside declineHeldInvoice instead of here, then the signature function would be the same?

maybe what you're doing is that there are other place we are calling `declineHeldInvoice`?

nicolasburtey (approved):

PR #1505, opened 2022-08-01
feat: skip fee probe for flagged destinations

vindard:
## Description

This PR fixes an issue where some services will always fail the fee probe whether there is a route or not (e.g. Muun). This is an issue because the attempt is logged in lnd's mission-control and it blocks any real payments attempts that are tried soon after the probe.

This PR allows us to configure flagged destinations that will fail and to skip the probe and simply return a max fee for these routes.

### DIFF
--- a/default.yaml
+++ b/default.yaml
@@ -188,3 +188,7 @@ kratosConfig:
 
 captcha:
   mandatory: false
+
+skipFeeProbe:
+  muun:
+    - "038f8f113c580048d847d6949371726653e02b928196bad310e3eda39ff61723f6"
--- a/src/app/payments/get-protocol-fee.ts
+++ b/src/app/payments/get-protocol-fee.ts
@@ -1,11 +1,19 @@
-import { decodeInvoice, defaultTimeToExpiryInSeconds } from "@domain/bitcoin/lightning"
+import { getPubkeysToSkipProbe, intersect } from "@config"
+
+import {
+  decodeInvoice,
+  defaultTimeToExpiryInSeconds,
+  parseFinalHopsFromInvoice,
+} from "@domain/bitcoin/lightning"
 import { checkedToWalletId } from "@domain/wallets"
 import {
   LnPaymentRequestNonZeroAmountRequiredError,
   LnPaymentRequestZeroAmountRequiredError,
+  SkipProbeForPubkeyError,
   PriceRatio,
 } from "@domain/payments"
 import { LndService } from "@services/lnd"
+
 import { PaymentFlowStateRepository } from "@services/payment-flow"
 import { WalletsRepository } from "@services/mongoose"
 import { NewDealerPriceService } from "@services/dealer-price"
@@ -128,10 +136,15 @@ const estimateLightningFee = async ({
       return PartialResult.err(lndService)
     }
 
-    const routeResult = await lndService.findRouteForInvoice({
-      invoice,
-      amount: btcPaymentAmount,
-    })
+    const skipProbe =
+      intersect(parseFinalHopsFromInvoice(invoice), getPubkeysToSkipProbe()).length > 0
+
+    const routeResult = skipProbe
+      ? new SkipProbeForPubkeyError()
+      : await lndService.findRouteForInvoice({
+          invoice,
+          amount: btcPaymentAmount,
+        })
     if (routeResult instanceof Error) {
       paymentFlow = await builder.withoutRoute()
       if (paymentFlow instanceof Error) {
--- a/src/app/payments/get-protocol-fee.ts
+++ b/src/app/payments/get-protocol-fee.ts
@@ -1,10 +1,4 @@
-import { getPubkeysToSkipProbe, intersect } from "@config"
-
-import {
-  decodeInvoice,
-  defaultTimeToExpiryInSeconds,
-  parseFinalHopsFromInvoice,
-} from "@domain/bitcoin/lightning"
+import { decodeInvoice, defaultTimeToExpiryInSeconds } from "@domain/bitcoin/lightning"
 import { checkedToWalletId } from "@domain/wallets"
 import {
   LnPaymentRequestNonZeroAmountRequiredError,
@@ -136,10 +130,7 @@ const estimateLightningFee = async ({
       return PartialResult.err(lndService)
     }
 
-    const skipProbe =
-      intersect(parseFinalHopsFromInvoice(invoice), getPubkeysToSkipProbe()).length > 0
-
-    const routeResult = skipProbe
+    const routeResult = (await builder.skipProbeForDestination())
       ? new SkipProbeForPubkeyError()
       : await lndService.findRouteForInvoice({
           invoice,
--- a/src/app/payments/helpers.ts
+++ b/src/app/payments/helpers.ts
@@ -3,6 +3,7 @@ import {
   getAccountLimits,
   MS_PER_DAY,
   MIN_SATS_FOR_PRICE_RATIO_PRECISION,
+  getPubkeysToSkipProbe,
 } from "@config"
 import { AccountLimitsChecker, TwoFALimitsChecker } from "@domain/accounts"
 import {
@@ -44,6 +45,7 @@ export const constructPaymentFlowBuilder = async <
   if (lndService instanceof Error) return lndService
   const paymentBuilder = LightningPaymentFlowBuilder({
     localNodeIds: lndService.listAllPubkeys(),
+    flaggedPubkeys: getPubkeysToSkipProbe(),
     usdFromBtcMidPriceFn,
     btcFromUsdMidPriceFn,
   })
--- a/src/config/schema.ts
+++ b/src/config/schema.ts
@@ -356,7 +356,11 @@ export const configSchema = {
     skipFeeProbe: {
       type: "object",
       patternProperties: {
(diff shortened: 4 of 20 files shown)


src/config/schema.ts
```diff
@@ -353,6 +353,13 @@ export const configSchema = {
       required: ["mandatory"],
       additionalProperties: false,
     },
+    skipFeeProbe: {
+      type: "object",
+      patternProperties: {
+        "^.*$": { type: "array", items: { type: "string" }, uniqueItems: true },
```
nicolasburtey, line 359: length could be fix to 66 (IIRC) characters?

default.yaml
```diff
@@ -188,3 +188,7 @@ kratosConfig:
 
 captcha:
   mandatory: false
+
+skipFeeProbe:
+  muun:
+    - "038f8f113c580048d847d6949371726653e02b928196bad310e3eda39ff61723f6"
```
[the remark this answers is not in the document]
vindard: Yup, it's captured in the code logic and in the test (test has an actual Muun invoice)

src/config/schema.ts
vindard: BOLT-11 says these are 33-bytes, so if it's hex-encoded then 66 sounds right 👍🏾

vindard: Done: https://github.com/GaloyMoney/galoy/pull/1505/commits/64efd9298363f055547d51669820da57537081ef

src/app/payments/get-protocol-fee.ts
```diff
@@ -128,10 +136,15 @@ const estimateLightningFee = async ({
       return PartialResult.err(lndService)
     }
 
-    const routeResult = await lndService.findRouteForInvoice({
-      invoice,
-      amount: btcPaymentAmount,
-    })
+    const skipProbe =
+      intersect(parseFinalHopsFromInvoice(invoice), getPubkeysToSkipProbe()).length > 0
```
bodymindarts, line 140: I think this logic belongs inside the builder - it can take the pub keys to skip as config:
```
type LightningPaymentFlowBuilderConfig = {
  localNodeIds: Pubkey[]
  skipProbeKeys: Pubkey[]
  usdFromBtcMidPriceFn: UsdFromBtcMidPriceFn
  btcFromUsdMidPriceFn: BtcFromUsdMidPriceFn
}
```
or something like that.

This step:
```
type LPFBWithConversion<S extends WalletCurrency, R extends WalletCurrency> = {
  withRoute({
    pubkey,
    rawRoute,
  }: {
    pubkey: Pubkey
    rawRoute: RawRoute
  }): Promise<PaymentFlow<S, R> | ValidationError | DealerPriceServiceError>
  withoutRoute(): Promise<PaymentFlow<S, R> | ValidationError | DealerPriceServiceError>

  btcPaymentAmount(): Promise<BtcPaymentAmount | DealerPriceServiceError>
  usdPaymentAmount(): Promise<UsdPaymentAmount | DealerPriceServiceError>

  isIntraLedger(): Promise<boolean | DealerPriceServiceError>
}
```
Already has some query methods - could also add another one `shouldSkipProbe`?

There is no way this logic should be integration tested.

src/config/utils.ts
```diff
@@ -2,3 +2,6 @@ import mergeWith from "lodash.mergewith"
 
 export const merge = (defaultConfig: unknown, customConfig: unknown) =>
   mergeWith(defaultConfig, customConfig, (a, b) => (Array.isArray(b) ? b : undefined))
+
+export const intersect = <T>(a: T[], b: T[]): T[] =>
```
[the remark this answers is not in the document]
vindard: I thought so as well, welcome to Javascript 🤷🏾‍♂️ 
https://exploringjs.com/impatient-js/ch_sets.html#missing-set-operations

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -855,6 +889,40 @@ describe("UserWallet - Lightning Pay", () => {
     expect(balanceAfter).toBe(0)
   })
 
+  it("skips fee probe for flagged pubkeys", async () => {
```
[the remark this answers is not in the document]
vindard: I added domain + unit test changes (https://github.com/GaloyMoney/galoy/pull/1505/commits/1ba3743f334fdc45148762abe3312d6334558703), but I think this integration test is still useful to confirm that the lnd service method isn't actually being called when we're using domain calculated states to pull together the use-case

src/app/payments/get-protocol-fee.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1505/commits/1ba3743f334fdc45148762abe3312d6334558703

src/config/utils.ts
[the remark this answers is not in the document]
vindard: There's still an open proposal to have them added 😅 
https://github.com/tc39/proposal-set-methods

default.yaml
```diff
@@ -188,3 +188,7 @@ kratosConfig:
 
 captcha:
   mandatory: false
+
+skipFeeProbe:
```
dolcalmi, line 192: why not just an array?

src/domain/payments/payment-flow-builder.ts
```diff
@@ -491,6 +500,13 @@ const LPFBWithConversion = <S extends WalletCurrency, R extends WalletCurrency>(
     return state.usdPaymentAmount
   }
 
+  const skipProbeForDestination = async () => {
+    const state = await statePromise
```
dolcalmi, line 504: why the other methods use `const state = await Promise.resolve(statePromise)` ?

src/graphql/error-map.ts
```diff
@@ -415,6 +415,7 @@ export const mapError = (error: ApplicationError): CustomApolloError => {
     case "SafeWrapperError":
     case "InvalidFeeProbeStateError":
     case "InvalidPubKeyError":
+    case "SkipProbeForPubkeyError":
```
dolcalmi, line 418: this should have a custom message. Sending `Unknown error occurred` is not good for this case

dolcalmi (changes requested):

default.yaml
vindard: I thought it would be useful to have some context if we had to add any other services.

src/domain/payments/payment-flow-builder.ts
vindard: Oh I thought I already standardized this. I believe the `Promise.resolve(statePromise)` is redundant. We can probably change all instances of this to be simpler.

src/graphql/error-map.ts
vindard: This is just to satisfy the type checker. This error never gets returned to the user right now. It's just used in a conditional to determine if to call `withoutRoute` or `withRoute` on the builder.

src/config/utils.ts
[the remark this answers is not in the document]
bodymindarts: Alright I guess that is to be expected from JS - still if we are augmenting the type it should still be `Set.intersect(Set) -> Set` and not return an `Array` or?

src/domain/payments/payment-flow-builder.ts
bodymindarts: yup I noticed this too lets make it consistent

vindard: Done: https://github.com/GaloyMoney/galoy/pull/1505/commits/11f8119a64597b1fc695e346e5c44b10502d7b9f

src/config/utils.ts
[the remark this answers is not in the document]
vindard: No reason not to: https://github.com/GaloyMoney/galoy/pull/1505/commits/a08ef9b70ca373c233aa18dbd68c24f9ffb15dd0

src/app/payments/helpers.ts
```diff
@@ -43,9 +44,12 @@ export const constructPaymentFlowBuilder = async <
...
   const paymentBuilder = LightningPaymentFlowBuilder({
-    localNodeIds: lndService.listAllPubkeys(),
-    usdFromBtcMidPriceFn,
-    btcFromUsdMidPriceFn,
+    config: {
+      localNodeIds: lndService.listAllPubkeys(),
+      usdFromBtcMidPriceFn,
+      btcFromUsdMidPriceFn,
+    },
+    flaggedPubkeys: getPubkeysToSkipProbe(),
```
bodymindarts, line 52: All looks good except I don't understand the extra `config` key... `flaggedPubkeys` is also config isn't it?

vindard: > All looks good except I don't understand the extra `config` key... `flaggedPubkeys` is also config isn't it?

Yup I noticed this literally after I pushed the last commit. Changed now.

bodymindarts (approved):

default.yaml
nicolasburtey: I would agree with @dolcalmi that just a series of keys would be enough.

you could have the name of the service in comments, or it's easy enough to fetch it otherwise.

ie: 
skipFeeProbe:
  - "038f8f113c580048d847d6949371726653e02b928196bad310e3eda39ff61723f6" # muun

PR #1508, opened 2022-08-02
chore: bump lightning version to get subscription fixes

vindard:
## Description

This supports changes in PR #1484

There is another issue here where one of the new dependencies is causing heap allocation errors in our CI at the 2048 bytes memory level. Options to fix are:
- bump the memory in CI tests (done here)
- to find the bad dependency version and roll it back one

---
### Logs

**Taken from [this test run](https://github.com/GaloyMoney/galoy/runs/7634630719?check_suite_focus=true):**

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,8 +88,11 @@ integration-in-ci:
 	make create-tmp-env-ci && \
 	TMP_ENV_CI=tmp.env.ci docker compose -f docker-compose.yml up integration-tests
 
+# NODE_OPTIONS line should be removed whenever we upgrade yarn.lock to see if
+# heap allocation issue has been resolved in dependencies (fails at 2048).
 execute-integration-from-within-container:
 	yarn install && \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -88,8 +88,11 @@ integration-in-ci:
 	make create-tmp-env-ci && \
 	TMP_ENV_CI=tmp.env.ci docker compose -f docker-compose.yml up integration-tests
 
+# NODE_OPTIONS line should be removed whenever we upgrade yarn.lock to see if
+# heap allocation issue has been resolved in dependencies (fails at 2048).
 execute-integration-from-within-container:
 	yarn install && \
+	NODE_OPTIONS="--max-old-space-size=3072" \
```
dolcalmi, line 95: @bodymindarts @krtk6160 could you please confirm if we have enough memory assigned in execution environment/pod ?

krtk6160: I think we should have enough memory

PR #1510, opened 2022-08-03
chore: Exposed node info in error msgs

daviroo:


src/services/tracing.ts
```diff
@@ -244,6 +244,7 @@ const recordException = (span: Span, exception: Exception, level?: ErrorLevel) =
   span.setAttribute("error.level", errorLevel)
   // @ts-ignore-next-line no-implicit-any error
   span.setAttribute("error.name", exception["name"])
+  span.setAttribute("error.message", exception["message"])
```
vindard, line 247: Can you confirm too that `exception["message"]` doesn't break if we don't include a message?

vindard (commented):
This looks good to me, just one comment

vindard (approved):

PR #1511, opened 2022-08-04
feat: add addresses arg to filter transactions selection for wallets

vindard:
## Description

This PR adds an additional optional arg `addresses` to the `transactions` selection under wallets. This arg can be used to fetch the transactions for a given address. If an empty array is passed the transaction fetches all onchain transactions for the wallet. Under the hood it also:
- ignores the cache and pings lnd for the latest incoming txns
- updates the cache with latest incoming data found

This mean that including the `addresses` arg can force a cache update with the latest incoming transactions state for pending transactions. This is useful if new transactions have been received but we are still within the redis cache's ttl window.


test/helpers/bitcoin-core.ts
```diff
@@ -33,6 +33,26 @@ export async function sendToAddressAndConfirm({
...
+export async function sendToAddress({
+  walletClient,
+  address,
+  amount,
+}: {
+  walletClient: BitcoindWalletClient
+  address: OnChainAddress
+  amount: number
+}) {
+  await walletClient.sendToAddress({ address, amount })
```
dolcalmi, line 45: just return, await is not needed

test/helpers/bitcoin-core.ts
```diff
@@ -33,6 +33,26 @@ export async function sendToAddressAndConfirm({
   await walletClient.generateToAddress({ nblocks: 6, address: RANDOM_ADDRESS })
 }
 
+export async function sendToAddress({
```
dolcalmi, line 36: helpers began previous to define some code rules but this should be
```
export const sendToAddress = ...
```

test/helpers/bitcoin-core.ts
```diff
@@ -33,6 +33,26 @@ export async function sendToAddressAndConfirm({
...
+  amount,
+}: {
+  walletClient: BitcoindWalletClient
+  address: OnChainAddress
+  amount: number
+}) {
+  await walletClient.sendToAddress({ address, amount })
+}
+
+export async function confirmSent({
```
dolcalmi, line 48: same here

test/helpers/bitcoin-core.ts
```diff
@@ -33,6 +33,26 @@ export async function sendToAddressAndConfirm({
...
+  walletClient,
+}: {
+  walletClient: BitcoindWalletClient
+}) {
+  await walletClient.generateToAddress({ nblocks: 6, address: RANDOM_ADDRESS })
```
dolcalmi, line 53: same here

src/app/wallets/get-transactions-by-addresses.ts
```diff
@@ -0,0 +1,91 @@
...
+  const onChainTxs = await onChain.listIncomingTransactions(ONCHAIN_MIN_CONFIRMATIONS)
+  if (onChainTxs instanceof Error) {
+    baseLogger.warn({ onChainTxs }, "impossible to get listIncomingTransactions")
+    return PartialResult.partial(confirmedHistory.transactions, onChainTxs)
+  }
+  redisCache.set({
+    key: CacheKeys.LastOnChainTransactions,
+    value: onChainTxs,
+    ttlSecs: SECS_PER_10_MINS,
+  })
```
dolcalmi, line 54: please use getorset like in `getTransactionsForWallets`
```
const onChainTxs = await redisCache.getOrSet({
    key: CacheKeys.LastOnChainTransactions,
    ttlSecs: SECS_PER_10_MINS,
    fn: () => onChain.listIncomingTransactions(ONCHAIN_MIN_CONFIRMATIONS),
  })
```
trigger server reset cache on new transactions

src/graphql/types/object/btc-wallet.ts
```diff
@@ -43,8 +46,26 @@ const BtcWallet = GT.Object<Wallet>({
...
     transactions: {
       type: TransactionConnection,
-      args: connectionArgs,
+      args: { ...connectionArgs, addresses: { type: GT.List(OnChainAddress) } },
       resolve: async (source, args) => {
+        let { addresses } = args
+        if (addresses) {
+          addresses = addresses.length ? addresses : source.onChainAddresses()
+          for (const address of addresses) {
+            if (address instanceof InputValidationError) throw address
```
dolcalmi, line 55: shouldn't be just `Error` ?

src/graphql/types/object/usd-wallet.ts
```diff
@@ -42,8 +45,26 @@ const UsdWallet = GT.Object<Wallet>({
...
     transactions: {
       type: TransactionConnection,
-      args: connectionArgs,
+      args: { ...connectionArgs, addresses: { type: GT.List(OnChainAddress) } },
       resolve: async (source, args) => {
+        let { addresses } = args
+        if (addresses) {
+          addresses = addresses.length ? addresses : source.onChainAddresses()
+          for (const address of addresses) {
+            if (address instanceof InputValidationError) throw address
```
dolcalmi, line 54: same here

dolcalmi (changes requested):
looks good, just minor changes

src/graphql/types/object/btc-wallet.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1511/commits/c7e25002ea2758ee3266f4133cf82eebb2202ffc

src/graphql/types/object/usd-wallet.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1511/commits/c7e25002ea2758ee3266f4133cf82eebb2202ffc

test/helpers/bitcoin-core.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1511/commits/b718f72b7ca6eebb857141bd31aa67dd11601b7a

test/helpers/bitcoin-core.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1511/commits/b718f72b7ca6eebb857141bd31aa67dd11601b7a

test/helpers/bitcoin-core.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1511/commits/423357013149bcde54e0310937d831695722ded4

test/helpers/bitcoin-core.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1511/commits/423357013149bcde54e0310937d831695722ded4

src/app/wallets/get-transactions-by-addresses.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1511/commits/1c46fe024f3f01f4cf0f76b1dc376e134de9ad88

dolcalmi (dismissed):

src/graphql/types/object/btc-wallet.ts
```diff
@@ -43,8 +44,26 @@ const BtcWallet = GT.Object<Wallet>({
...
+            if (address instanceof Error) throw address
+          }
+
+          const { result: transactions, error } =
+            await Wallets.getTransactionsForWalletsByAddresses({
+              wallets: [source],
+              addresses,
+            })
+          if (error instanceof Error || transactions === null) {
+            throw error
```
dolcalmi, line 62: does this needs to use mapError?

src/graphql/types/object/usd-wallet.ts
```diff
@@ -42,8 +43,26 @@ const UsdWallet = GT.Object<Wallet>({
...
+            if (address instanceof Error) throw address
+          }
+
+          const { result: transactions, error } =
+            await Wallets.getTransactionsForWalletsByAddresses({
+              wallets: [source],
+              addresses,
+            })
+          if (error instanceof Error || transactions === null) {
+            throw error
```
dolcalmi, line 61: same question here

src/graphql/types/object/btc-wallet.ts
vindard: Fixed: https://github.com/GaloyMoney/galoy/pull/1511/commits/5e625b2536da9d0e80b0064b3b505e80fb21c125

src/graphql/types/object/usd-wallet.ts
vindard: Fixed: https://github.com/GaloyMoney/galoy/pull/1511/commits/5e625b2536da9d0e80b0064b3b505e80fb21c125

dolcalmi (approved):

samerbuna:
I maybe overthinking it a bit but I'm not a fan of using nullability to mean "all transactions". If a client intention is to list transactions for addresses and accidentally passes nulls, the operation would not fail. It should.

Also, if there's a value for the new `addresses` argument, then the transaction list is all onchain, rendering the unions we have for `SettlementVia` and `initiationVia` invalid.

vindard:
> I maybe overthinking it a bit but I'm not a fan of using nullability to mean "all transactions". If a client intention is to list transactions for addresses and accidentally passes nulls, the operation would not fail. It should.

Yea thinking about it again now I think this makes sense. They can get onchain-only transactions anyway by using a `... on InitiationViaOnChain` filter in the request as well so this violates the "only one way to do a thing" idea too. I've fixed that behaviour here: https://github.com/GaloyMoney/galoy/pull/1511/commits/155776b2aba6e83196ec21b1bad31ce7b674d0c9

vindard:
> Also, if there's a value for the new `addresses` argument, then the transaction list is all onchain, rendering the unions we have for `SettlementVia` and `initiationVia` invalid.

Not sure I understand this point. Do you mean they break, or that they're just redundant because everything comes back as onchain type transactions? The intention here is to only surface transactions related to the addresses passed, so this will naturally only be onchain txns.

src/app/wallets/get-transactions-by-addresses.ts
```diff
@@ -0,0 +1,90 @@
...
+import { OnChainService } from "@services/lnd/onchain-service"
+
+const redisCache = RedisCacheService()
+
+export const getTransactionsForWalletsByAddresses = async ({
+  wallets,
+  addresses,
+}: {
+  wallets: Wallet[]
+  addresses: OnChainAddress[]
```
nicolasburtey, line 23: is there to check to verify `string[]` to `checkToOnChainAddress[]` is acceptable? ie: all onchain addresses should be valid addresses

vindard: Technically we did it in the graphql layer as well inside the `OnChainAddress` scaler, but I think we can do it again inside the method for type consistency: https://github.com/GaloyMoney/galoy/pull/1511/commits/9bde789bef1186e64beab233afc72a90e585a1d3

nicolasburtey: I actually think we don't need and could revert the last commit. I didn't know/remember we were already doing it on the graphql layer (I just look at the diff of this commit, didn't check the code outside of it). As we already do I think there is no need for double check imo.

samerbuna:
@vindard The validation on the `addresses` length is great, we should also make the type a list of non null addresses: `[address!]`

My point however was about the `addresses` argument itself being null to mean return all transactions. My thinking was maybe a different field, `transactionsForAddresses`, or maybe `onChainTransactions` and make the `addresses` argument there non nullable: `[address!]!`.

The second point is about how the object in this list is not exactly a `Transaction` (which could have LN/IntraLedger `initiationVia`/`settlementVia` values ). It's an `OnChainTransaction` (which we don't have). 

With the current implementation, using the new API endpoint is easier for a client (just pass `addresses` array, keep everything the same). However, it can be misused (with an unintentional null `addresses` value for example) and allowing the client to have inline fragments about LN/IntraLedger is not ideal.

vindard:
> My thinking was maybe a different field, `transactionsForAddresses`, or maybe `onChainTransactions` and make the `addresses` argument there non nullable: `[address!]!`.
> 
> The second point is about how the object in this list is not exactly a `Transaction` (which could have LN/IntraLedger `initiationVia`/`settlementVia` values ). It's an `OnChainTransaction` (which we don't have).
> 
> With the current implementation, using the new API endpoint is easier for a client (just pass `addresses` array, keep everything the same). However, it can be misused (with an unintentional null `addresses` value for example) and allowing the client to have inline fragments about LN/IntraLedger is not ideal.

Adding context: we discussed this with the team and decided that this was the right way to go. New selection now implemented as `filterTransactionsByAddress` with non-null `address` field now: https://github.com/GaloyMoney/galoy/pull/1511/commits/d9f7eb4c173bf08529dec19f06b8314335e1f0f1

src/graphql/types/object/btc-wallet.ts
```diff
@@ -48,13 +50,35 @@ const BtcWallet = GT.Object<Wallet>({
         const { result: transactions, error } = await Wallets.getTransactionsForWallets([
           source,
         ])
-        if (error instanceof Error || transactions === null) {
-          throw error
-        }
+        if (error instanceof Error) throw mapError(error)
+        if (transactions === null) throw error
```
dolcalmi, line 54: this should remains `if (error instanceof Error || transactions === null)`. We should not return an error if there are transactions

src/graphql/types/object/btc-wallet.ts
```diff
@@ -48,13 +50,35 @@ const BtcWallet = GT.Object<Wallet>({
...
+        const { address } = args
+        if (address instanceof Error) throw address
+
+        const { result: transactions, error } =
+          await Wallets.getTransactionsForWalletsByAddresses({
+            wallets: [source],
+            addresses: [address],
+          })
+        if (error instanceof Error) throw mapError(error)
+        if (transactions === null) throw error
```
dolcalmi, line 78: same here

src/graphql/types/object/usd-wallet.ts
```diff
@@ -47,9 +49,31 @@ const UsdWallet = GT.Object<Wallet>({
...
+        const { address } = args
+        if (address instanceof Error) throw address
+
+        const { result: transactions, error } =
+          await Wallets.getTransactionsForWalletsByAddresses({
+            wallets: [source],
+            addresses: [address],
+          })
+        if (error instanceof Error) throw mapError(error)
+        if (transactions === null) throw error
```
dolcalmi, line 76: same here

dolcalmi (changes requested):
looks good, just minor change

src/graphql/types/object/btc-wallet.ts
vindard: This was the original behaviour, I just decomposed it a bit to be able to map `error`.

Should we throw another error explicitly instead? Any ideas on this @samerbuna?

vindard: Ok so `transactions` can only be null if `error` is an instanceof Error, but it's tricky to get the type checker to understand this relationship so this conditional should never actually throw, but is just to signal to the type checker that `transactions` passed into the function just below is of type `WalletTransaction[]`

dolcalmi (approved):

src/graphql/types/object/btc-wallet.ts
vindard: Added a comment in case we forget again: https://github.com/GaloyMoney/galoy/pull/1511/commits/8ccde29578a77d7d469c0344f6337c82fdf11083

dolcalmi (dismissed):

vindard:
We originally named the new selection as `filterTransactionsByAddress` following this convention from "Production ready GraphQL" (see excerpt below)

We renamed it to `transactionsByAddress` (https://github.com/GaloyMoney/galoy/pull/1511/commits/0e2e6e6e851dcb05d0aa8fe5532e16f2dde6ba20) because of our own convention where we'd like to name `Query` queries as nouns and `Mutation` queries as verbs.

---
### Excerpt
image

dolcalmi (approved):

PR #1512, opened 2022-08-04
feat: add gql selection incoming pending onchain balance

vindard:
## Discussion

There is a subset of transactions that is currently not added to our ledger or accounted for in our balances. These are the incoming onchain transactions for a user that haven't met the required number of confirmations onchain as yet.

This PR is to calculate the sum of these pending transactions for a user and expose this balance via a new selection.

### Reasoning

If an API user would like to retrieve the pending incoming balance (unconfirmed onchain txns) for a wallet, their only option right now would be to use the `transactions` selection and:
- **paginate until their very 1st transaction**
   Because transactions can have different fees at different times, there's no guarantee that a confirmed transaction does not have prior unconfirmed transactions for that wallet still pending before it

- **filter by onchain & pending**
   There's no way (I think) otherwise to filter transactions at the graphql level, so all transactions must be fetched

- **sum the transaction amounts**

If we assume an API user wanting to retrieve their pending incoming onchain balance is a legitimate use-case, the rationale for having a selection that calculates this value for the user is similar to the one for exposing a `balance` selection instead of having the user also calculate this from `transactions` selection on the fly.

### DIFF
--- a/src/app/wallets/get-on-chain-txs.ts
+++ b/src/app/wallets/get-on-chain-txs.ts
@@ -1,0 +1,24 @@
+import { BTC_NETWORK, ONCHAIN_MIN_CONFIRMATIONS, SECS_PER_10_MINS } from "@config"
+
+import { OnChainError, TxDecoder } from "@domain/bitcoin/onchain"
+import { CacheKeys } from "@domain/cache"
+
+import { RedisCacheService } from "@services/cache"
+import { OnChainService } from "@services/lnd/onchain-service"
+import { baseLogger } from "@services/logger"
+
+// we are getting both the transactions in the mempool and the transaction that
+// have been mined by not yet credited because they haven't reached enough confirmations
+export default async () =>
+  RedisCacheService().getOrSet({
+    key: CacheKeys.LastOnChainTransactions,
+    ttlSecs: SECS_PER_10_MINS,
+    fn: async () => {
+      const onChain = OnChainService(TxDecoder(BTC_NETWORK))
+      if (onChain instanceof OnChainError) {
+        baseLogger.warn({ onChain }, "impossible to create OnChainService")
+        return onChain
+      }
+      return onChain.listIncomingTransactions(ONCHAIN_MIN_CONFIRMATIONS)
+    },
+  })
--- a/src/app/wallets/get-pending-onchain-balance-for-wallet.ts
+++ b/src/app/wallets/get-pending-onchain-balance-for-wallet.ts
@@ -1,0 +1,52 @@
+import { BTC_NETWORK, ONCHAIN_MIN_CONFIRMATIONS, SECS_PER_10_MINS } from "@config"
+
+import { CacheKeys } from "@domain/cache"
+import { OnChainError, TxDecoder, TxFilter } from "@domain/bitcoin/onchain"
+
+import { baseLogger } from "@services/logger"
+import { RedisCacheService } from "@services/cache"
+import { OnChainService } from "@services/lnd/onchain-service"
+import { IncomingOnChainTxHandler } from "@domain/bitcoin/onchain/incoming-tx-handler"
+
+const redisCache = RedisCacheService()
+
+export const getPendingOnChainBalanceForWallet = async (
+  wallets: Wallet[],
+): Promise<{ [key: WalletId]: CurrencyBaseAmount } | ApplicationError> => {
+  const onChain = OnChainService(TxDecoder(BTC_NETWORK))
+  if (onChain instanceof OnChainError) {
+    baseLogger.warn({ onChain }, "impossible to create OnChainService")
+    return onChain
+  }
+
+  // we are getting both the transactions in the mempool and the transaction that
+  // have been mined by not yet credited because they haven't reached enough confirmations
+  const onChainTxs = await redisCache.getOrSet({
+    key: CacheKeys.LastOnChainTransactions,
+    ttlSecs: SECS_PER_10_MINS,
+    fn: () => onChain.listIncomingTransactions(ONCHAIN_MIN_CONFIRMATIONS),
+  })
+  if (onChainTxs instanceof Error) {
+    baseLogger.warn({ onChainTxs }, "impossible to get listIncomingTransactions")
+    return onChainTxs
+  }
+
+  const filter = TxFilter({
+    confirmationsLessThan: ONCHAIN_MIN_CONFIRMATIONS,
+    addresses: wallets.flatMap((wallet) => wallet.onChainAddresses()),
+  })
+
+  const pendingIncoming = filter.apply(onChainTxs)
+
+  const balancesByWallet =
+    IncomingOnChainTxHandler(pendingIncoming).balanceByWallet(wallets)
+  if (balancesByWallet instanceof Error) return balancesByWallet
+
+  const normalizedBalances = {} as { [key: WalletId]: CurrencyBaseAmount }
+  for (const key of Object.keys(balancesByWallet)) {
+    const walletId = key as WalletId
+    const balance = Number(balancesByWallet[walletId]) as CurrencyBaseAmount
+    normalizedBalances[walletId] = balance
+  }
+  return normalizedBalances
+}
--- a/src/app/wallets/get-pending-onchain-balance-for-wallet.ts
+++ b/src/app/wallets/get-pending-onchain-balance-for-wallet.ts
@@ -9,7 +9,7 @@ import getOnChainTxs from "./get-on-chain-txs"
 
 export const getPendingOnChainBalanceForWallet = async (
   wallets: Wallet[],
-): Promise<{ [key: WalletId]: CurrencyBaseAmount } | ApplicationError> => {
+): Promise<{ [key: WalletId]: BtcPaymentAmount } | ApplicationError> => {
   const onChainTxs = await getOnChainTxs()
   if (onChainTxs instanceof Error) {
     baseLogger.warn({ onChainTxs }, "impossible to get listIncomingTransactions")
@@ -20,18 +20,7 @@ export const getPendingOnChainBalanceForWallet = async (
     confirmationsLessThan: ONCHAIN_MIN_CONFIRMATIONS,
     addresses: wallets.flatMap((wallet) => wallet.onChainAddresses()),
   })
-
   const pendingIncoming = filter.apply(onChainTxs)
 
-  const balancesByWallet =
-    IncomingOnChainTxHandler(pendingIncoming).balanceByWallet(wallets)
-  if (balancesByWallet instanceof Error) return balancesByWallet
-
-  const normalizedBalances = {} as { [key: WalletId]: CurrencyBaseAmount }
-  for (const key of Object.keys(balancesByWallet)) {
-    const walletId = key as WalletId
-    const balance = Number(balancesByWallet[walletId]) as CurrencyBaseAmount
-    normalizedBalances[walletId] = balance
-  }
-  return normalizedBalances
+  return IncomingOnChainTxHandler(pendingIncoming).balanceByWallet(wallets)
 }
--- a/src/app/wallets/get-pending-onchain-balance-for-wallet.ts
+++ b/src/app/wallets/get-pending-onchain-balance-for-wallet.ts
@@ -7,7 +7,7 @@ import { IncomingOnChainTxHandler } from "@domain/bitcoin/onchain/incoming-tx-ha
 
 import getOnChainTxs from "./get-on-chain-txs"
 
-export const getPendingOnChainBalanceForWallet = async (
+export const getPendingOnChainBalanceForWallets = async (
   wallets: Wallet[],
 ): Promise<{ [key: WalletId]: BtcPaymentAmount } | ApplicationError> => {
   const onChainTxs = await getOnChainTxs()
--- a/src/app/wallets/get-pending-onchain-balance-for-wallet.ts
+++ b/src/app/wallets/get-pending-onchain-balance-for-wallet.ts
@@ -13,10 +13,13 @@ const redisCache = RedisCacheService()
 export const getPendingOnChainBalanceForWallet = async (
   wallets: Wallet[],
 ): Promise<{ [key: WalletId]: CurrencyBaseAmount } | ApplicationError> => {
-  const onChain = OnChainService(TxDecoder(BTC_NETWORK))
-  if (onChain instanceof OnChainError) {
-    baseLogger.warn({ onChain }, "impossible to create OnChainService")
-    return onChain
+  const listIncomingToUpdateCache = async () => {
+    const onChain = OnChainService(TxDecoder(BTC_NETWORK))
+    if (onChain instanceof OnChainError) {
+      baseLogger.warn({ onChain }, "impossible to create OnChainService")
+      return onChain
+    }
+    return onChain.listIncomingTransactions(ONCHAIN_MIN_CONFIRMATIONS)
   }
 
   // we are getting both the transactions in the mempool and the transaction that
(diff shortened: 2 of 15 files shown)


nicolasburtey:
> ## Discussion
> There is a subset of transactions that is currently not added to our ledger or accounted for in our balances. These are the incoming onchain transactions for a user that haven't met the required number of confirmations onchain as yet.
> 
> This PR is to calculate the sum of these pending transactions for a user and expose this balance via a new selection.

could you clarify the use case for having the balance of unconfirmed transaction?

vindard:
> could you clarify the use case for having the balance of unconfirmed transaction?

From one of the ATM folks integrating our API:
> ".. it's literally just displayed for tracking purposes. That's how operators track the total value of the server wallet at any time, I don't think they export xpubs or anything like that, so they rely on that."

src/app/wallets/get-pending-onchain-balance-for-wallet.ts
```diff
@@ -0,0 +1,52 @@
...
+    baseLogger.warn({ onChain }, "impossible to create OnChainService")
+    return onChain
+  }
+
+  // we are getting both the transactions in the mempool and the transaction that
+  // have been mined by not yet credited because they haven't reached enough confirmations
+  const onChainTxs = await redisCache.getOrSet({
+    key: CacheKeys.LastOnChainTransactions,
+    ttlSecs: SECS_PER_10_MINS,
+    fn: () => onChain.listIncomingTransactions(ONCHAIN_MIN_CONFIRMATIONS),
```
nicolasburtey, line 27: here, and in 2 other places we have `redisCache.getOrSet({
    key: CacheKeys.LastOnChainTransactions })`, we are pre-fetching in advance the onchain Service, and failing in case the onchain service is down.

I believe this feels like an anti pattern. There are 2 reasons we are using redis:
- we want to reduce the load on lnd, which we achieve by using redis here
- we want to be resilient of error momentarily in case of a lnd restart

because we fail early if onchain service is down, we're removing many of the upside of using a cache here.

probably the fetching of the onchain service should only initiated if the value is not in cache.

```suggestion
    fn: () => {
      const onChain = OnChainService(TxDecoder(BTC_NETWORK))
  if (onChain instanceof OnChainError) {
    baseLogger.warn({ onChain }, "impossible to create OnChainService")
    return onChain
  }
    onChain.listIncomingTransactions(ONCHAIN_MIN_CONFIRMATIONS),
    }
```

src/app/wallets/get-pending-onchain-balance-for-wallet.ts
```diff
@@ -0,0 +1,52 @@
...
+import { OnChainError, TxDecoder, TxFilter } from "@domain/bitcoin/onchain"
+
+import { baseLogger } from "@services/logger"
+import { RedisCacheService } from "@services/cache"
+import { OnChainService } from "@services/lnd/onchain-service"
+import { IncomingOnChainTxHandler } from "@domain/bitcoin/onchain/incoming-tx-handler"
+
+const redisCache = RedisCacheService()
+
+export const getPendingOnChainBalanceForWallet = async (
```
nicolasburtey, line 13:
```suggestion
export const getPendingOnChainBalanceForWallets = async (
```
?

src/domain/bitcoin/onchain/incoming-tx-handler.ts
```diff
@@ -0,0 +1,66 @@
...
+        balancesByAddress[address] =
+          (balancesByAddress[address] || 0n) + balances[address]
+      }
+    }
+    return balancesByAddress
+  }
+
+  const balanceByWallet = (
+    wallets: Wallet[],
+  ): { [key: WalletId]: bigint } | ValidationError => {
```
nicolasburtey, line 23:
```suggestion
  ): { [key: WalletId]: Satoshis } | ValidationError => {
```
?

src/app/wallets/get-pending-onchain-balance-for-wallet.ts
```diff
@@ -0,0 +1,52 @@
...
+    addresses: wallets.flatMap((wallet) => wallet.onChainAddresses()),
+  })
+
+  const pendingIncoming = filter.apply(onChainTxs)
+
+  const balancesByWallet =
+    IncomingOnChainTxHandler(pendingIncoming).balanceByWallet(wallets)
+  if (balancesByWallet instanceof Error) return balancesByWallet
+
+  const normalizedBalances = {} as { [key: WalletId]: CurrencyBaseAmount }
```
nicolasburtey, line 45:
```suggestion
  const normalizedBalances: { [key: WalletId]: CurrencyBaseAmount } = {}
```

src/app/wallets/get-pending-onchain-balance-for-wallet.ts
```diff
@@ -0,0 +1,52 @@
...
+  for (const key of Object.keys(balancesByWallet)) {
+    const walletId = key as WalletId
```
nicolasburtey, line 47:
```suggestion
  for (const walletId of Object.keys(balancesByWallet) as Array<
    keyof typeof balancesByWallet
  >) {
```

an alternative (takes one more line so not sure that a good one)

src/domain/bitcoin/onchain/incoming-tx-handler.ts
nicolasburtey: seems Satoshis is still not a bigint?

src/app/wallets/get-pending-onchain-balance-for-wallet.ts
nicolasburtey, line 45: maybe `typedAmount` might be more descriptive? versus normalized?

when I think about normalized I think like we are bounding the values to `[0, 1]`

src/domain/bitcoin/onchain/incoming-tx-handler.ts
```diff
@@ -0,0 +1,66 @@
+import { paymentAmountFromNumber, WalletCurrency } from "@domain/shared"
+
+export const IncomingOnChainTxHandler = (
+  txns: IncomingOnChainTransaction[],
+): IncomingOnChainTxHandler => {
+  const balanceByAddress = (): { [key: OnChainAddress]: bigint } | ValidationError => {
+    const pendingBalances = txns.map(balanceFromIncomingTx)
+
+    const balancesByAddress = {} as { [key: OnChainAddress]: bigint }
```
nicolasburtey, line 9:
```suggestion
    const balancesByAddress: { [key: OnChainAddress]: bigint } = {}
```

src/domain/bitcoin/onchain/incoming-tx-handler.ts
```diff
@@ -0,0 +1,66 @@
...
+    for (const out of outs) {
+      if (!out.address) continue
+      balanceByAddress[out.address] = balanceByAddress[out.address] || 0n
+      const outAmount = paymentAmountFromNumber({
+        amount: out.sats,
+        currency: WalletCurrency.Btc,
+      })
+      if (outAmount instanceof Error) return outAmount
+
+      balanceByAddress[out.address] = balanceByAddress[out.address] + outAmount.amount
```
nicolasburtey, line 57:
```suggestion
      if (!out.address) continue
      const outAmount = paymentAmountFromNumber({
        amount: out.sats,
        currency: WalletCurrency.Btc,
      })
      if (outAmount instanceof Error) return outAmount

      balanceByAddress[out.address] =
        (balanceByAddress[out.address] || 0n) + outAmount.amount
    }
```

nicolasburtey: we could just set the default initial value at the end with the same instruction

src/graphql/types/object/btc-wallet.ts
```diff
@@ -43,6 +43,15 @@ const BtcWallet = GT.Object<Wallet>({
         return balanceSats
       },
     },
+    pendingIncomingBalance: {
+      type: GT.NonNull(SignedAmount),
+      description: "An unconfirmed incoming onchain balance stored in BTC.",
```
nicolasburtey, line 48:
```suggestion
      description: "An unconfirmed incoming onchain balance.",
```

`stored in BTC` feels redundant? we are in BTC wallet.

Plus I believe the balance in in sats so the `stored in BTC` can be confusing.

src/graphql/types/object/usd-wallet.ts
```diff
@@ -42,6 +42,15 @@ const UsdWallet = GT.Object<Wallet>({
         return Math.floor(balanceCents)
       },
     },
+    pendingIncomingBalance: {
```
nicolasburtey, line 45: we can't receive USD onchain for now

src/app/wallets/get-pending-onchain-balance-for-wallet.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1512/commits/eea915b7e389d52d9ee712c81b733a04f31e96e8 & https://github.com/GaloyMoney/galoy/pull/1512/commits/2c7b1344df4196582af7e748de047e67308dcda8

src/app/wallets/get-pending-onchain-balance-for-wallet.ts
vindard: I guess in a data context you're right. I was aiming for a context (meaning) similar to how JS's `String.prototype.normalize()` standardizes inputs to a certain format.

`typed...` doesn't quite feel like it. Maybe `standardized...`? Either way, this code block is now removed and we've fallen back to an existing method with similar name.

src/domain/bitcoin/onchain/incoming-tx-handler.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1512/commits/ab147ab0c6a8fdf749f8f815273c2ec93034cc6a

src/domain/bitcoin/onchain/incoming-tx-handler.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1512/commits/2bae66a20bf0a3540d960a9fd3cdc5f37b8d021c

src/app/wallets/get-pending-onchain-balance-for-wallet.ts
vindard: Fixed: https://github.com/GaloyMoney/galoy/pull/1512/commits/1fa7470bd5bfc2d36e22bab00ef85b3a2f2717e2

src/app/wallets/get-on-chain-txs.ts
```diff
@@ -0,0 +1,24 @@
...
+import { OnChainError, TxDecoder } from "@domain/bitcoin/onchain"
+import { CacheKeys } from "@domain/cache"
+
+import { RedisCacheService } from "@services/cache"
+import { OnChainService } from "@services/lnd/onchain-service"
+import { baseLogger } from "@services/logger"
+
+// we are getting both the transactions in the mempool and the transaction that
+// have been mined by not yet credited because they haven't reached enough confirmations
+export default async () =>
```
nicolasburtey, line 12: this will be a big improvement and may help to not have any more errors when lnd1 is restarting

src/domain/bitcoin/onchain/incoming-tx-handler.ts
```diff
@@ -0,0 +1,82 @@
...
+  ): { [key: WalletId]: BtcPaymentAmount } | ValidationError => {
+    const balancesByAddress = balanceByAddress()
+    if (balancesByAddress instanceof Error) return balancesByAddress
+
+    const balancesByWallet: { [key: WalletId]: BtcPaymentAmount } = {}
+    for (const wallet of wallets) {
+      balancesByWallet[wallet.id] = ZERO_SATS
+      for (const key of Object.keys(balancesByAddress)) {
+        const address = key as OnChainAddress
+        if (wallet.onChainAddresses().includes(address as OnChainAddress)) {
```
nicolasburtey, line 43:
```suggestion
        if (wallet.onChainAddresses().includes(address)) {
```

src/domain/bitcoin/onchain/index.types.d.ts
```diff
@@ -64,6 +64,13 @@ type PayToAddressArgs = {
   targetConfirmations: TargetConfirmations
 }
 
+type IncomingOnChainTxHandler = {
+  balanceByAddress(): { [key: OnChainAddress]: BtcPaymentAmount } | ValidationError
```
nicolasburtey, line 68: shouldn't this function take an arg? ie:

```suggestion
  balanceByAddress(address: OnchainAddress): { [key: OnChainAddress]: BtcPaymentAmount } | ValidationError
```

nicolasburtey: otherwise maybe it should be `balanceByAddresses`?

src/graphql/types/object/usd-wallet.ts
```diff
@@ -42,6 +43,11 @@ const UsdWallet = GT.Object<Wallet>({
...
+      type: GT.NonNull(SignedAmount),
+      description: "An unconfirmed incoming onchain balance.",
+      resolve: () => notBtcWalletForQueryError,
```
nicolasburtey, line 49: I'm wondering if we need this error? or if the same code that we have for BTC wallet would just work?

assuming we can't generate an address currently on usd wallet for onchain, this function would basically always return 0 which is also a valid result?

another option would be to simply not expose `pendingIncomingBalance`? (maybe we can't because there is some inheritance?)

nicolasburtey (commented):
some nits but looks good to me otherwise.

src/domain/bitcoin/onchain/index.types.d.ts
vindard: It uses the `txns` array internal to the `IncomingOnChainTxHandler`. Changed the name here: https://github.com/GaloyMoney/galoy/pull/1512/commits/7267097055d3f68eb8ebb24bb0e32ae4b178bc41

src/app/wallets/get-on-chain-txs.ts
```diff
@@ -0,0 +1,24 @@
+import { BTC_NETWORK, ONCHAIN_MIN_CONFIRMATIONS, SECS_PER_10_MINS } from "@config"
```
bodymindarts, line 1: I've been thinking about our file structure. I think its a bit confusing that we have both these 'helper' files on the same level as the 'use-cases'. This file for instance should be considered 'module private' (only visible from `app/wallets`). Perhaps we should start adding a subfolder for these to highlight the difference in the folder structure.

I'm thinking of a subfolder called `private` or `internal` or something like that. I don't like `helpers` or `utils` so much because - even though somewhat accurate those are overloaded words that will probably not serve the correct purpose.

bodymindarts (approved):

src/domain/bitcoin/onchain/index.types.d.ts
nicolasburtey: I actually prefer `balanceByAddresses` than `balanceByTxnsAddresses`. the thing that was "bothering" me in the previous commit is the use of singular where a list were returned.

nicolasburtey (approved):

src/domain/bitcoin/onchain/index.types.d.ts
vindard: I can change in a follow-up PR 👍🏾

src/app/wallets/get-on-chain-txs.ts
vindard: Makes sense, can address in a follow-up PR

src/domain/bitcoin/onchain/index.types.d.ts
vindard: Addressed: #1630

src/app/wallets/get-on-chain-txs.ts
vindard: Addressed: #1630

PR #1517, opened 2022-08-08
chore: reconnect to nodes on every cronjob

nicolasburtey:
grafana shows that there are many inactives nodes. running reconnect automatically should help.

only downside of bringing balanceofsatoshis as a dependancy is that it brings a lot more dependancies, but I think most of those dependancies are small library from alex. 

also I would envision that this "issue" of no auto reconnection would be resolved at some point by lnd. 

thoughts on this PR?

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "apollo-server-errors": "^3.3.1",
     "apollo-server-express": "^3.8.2",
     "axios": "^0.27.2",
+    "balanceofsatoshis": "^12.18.0",
     "bip32": "^3.0.1",
     "bitcoin-core-ts": "^3.0.3",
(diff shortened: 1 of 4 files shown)


openoms:
That is indeed a lot of new dependencies, but for what it's worth I am running `bos reconnect` connected to my lnd nodes also. 
Bos might even be used for other purposes later on.

nicolasburtey:
> That is indeed a lot of new dependencies, but for what it's worth I am running `bos reconnect` connected to my lnd nodes also. Bos might even be used for other purposes later on.

I would assume we would not keep `bos` forever as this is something that would be fixed on lnd at some point.

I would also use `bos send` to rebalance between the different internal node.

openoms (approved):

PR #1519, opened 2022-08-09
refactor: fix exposed error messages via API

vindard:
## Description

This PR cleans up our error handling at the boundary between the application service and the graphql api (adapter). It was prompted by [this error message](https://chat.galoy.io/galoy/pl/huh38waxjfga5gb8fknsy8g7uy) below that was leaked through to the frontend.

We also standardise the error message that gets added to `UnknownLightningServiceError` in this PR.

---
### Screenshot
New Project

### DIFF
--- a/src/graphql/error-map.ts
+++ b/src/graphql/error-map.ts
@@ -27,7 +27,7 @@ const assertUnreachable = (x: never): never => {
 
 export const mapError = (error: ApplicationError): CustomApolloError => {
   const errorName = error.name as ApplicationErrorKey
-  let message: string = errorName || ""
+  let message = ""
   switch (errorName) {
     case "WithdrawalLimitsExceededError":
       message = error.message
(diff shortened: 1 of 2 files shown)


src/graphql/error-map.ts
```diff
@@ -27,7 +27,7 @@ const assertUnreachable = (x: never): never => {
 
 export const mapError = (error: ApplicationError): CustomApolloError => {
   const errorName = error.name as ApplicationErrorKey
-  let message = error.message || errorName || ""
+  let message: string = errorName || ""
```
dolcalmi, line 30: we need to handle properly the error but the message needs to be passed (we are using it in a lot of errors). IMO error message needs to remain here

dolcalmi (changes requested):

src/graphql/error-map.ts
vindard: Yea I was trying to figure out where we use these for the unhandled ones (this only counts for unhandled errors in the map). What's an example of where we use the message? Maybe we can figure out something that makes sense without exposing internal details accidentally.

dolcalmi (approved):

Issue #1516, opened 2022-08-08
Anchor channels support


openoms:
Changing to channels with anchor commitments would improve the security of the node in case of a high fee environment.

Currently the accounting doesn't support the changes in the commitment fees

When removing 
```
protocol.no-anchors=true 
```
from
https://github.com/GaloyMoney/galoy/blob/8d468c4151a0d4b90edbb17370b2f7ea246d1ece/dev/lnd/lnd.conf#L41

The tests fail:

```
  Bitcoind
    ✓ check no wallet (7 ms)
    ✓ create cold wallet (316 ms)
    ✓ create outside wallet (1076 ms)
    ✓ should be funded mining 10 blocks (259 ms)
    ✓ funds outside lnd node (3431 ms)
    ✓ funds lnd1 node (918 ms)

WARN [1659981612148] (140548 on debian): not balanced
    module: "test"
    assetsLiabilitiesDifference: 0
    bookingVersusRealWorldAssets: -330
    assets: -99992363
    liabilities: 99992363
    bankOwnerBalance: 292363
    lnd: 99988893
    lightning: -99989223
    bitcoind: 0
    bitcoin: 0
WARN [1659981613963] (140548 on debian): not balanced
    module: "test"
    assetsLiabilitiesDifference: 0
    bookingVersusRealWorldAssets: -660
    assets: -99984726
    liabilities: 99984726
    bankOwnerBalance: 284726
    lnd: 99977786
    lightning: -99978446
    bitcoind: 0
    bitcoin: 0
WARN [1659981615722] (140548 on debian): not balanced
    module: "test"
    assetsLiabilitiesDifference: 0
    bookingVersusRealWorldAssets: -660
    assets: -99984726
    liabilities: 99984726
    bankOwnerBalance: 284726
    lnd: 99977786
    lightning: -99978446
    bitcoind: 0
    bitcoin: 0
WARN [1659981617569] (140548 on debian): not balanced
    module: "test"
    assetsLiabilitiesDifference: 0
    bookingVersusRealWorldAssets: -990
    assets: -99977089
    liabilities: 99977089
    bankOwnerBalance: 277089
    lnd: 99966679
    lightning: -99967669
    bitcoind: 0
    bitcoin: 0
FAIL test/integration/01-setup/01-lightning-channels.spec.ts (9.735 s)
  Lightning channels
    ✕ opens channel from lnd1 to lnd2 (1699 ms)
[...]

PR #1520, opened 2022-08-09
feat: support anchor commitments in tests

openoms:
fixes: #1516

### DIFF
--- a/src/services/lnd/utils.ts
+++ b/src/services/lnd/utils.ts
@@ -120,10 +120,8 @@ export const lndBalances = async (lnd: AuthenticatedLnd) => {
   const selfInitiatedChannels = channels.filter(
     ({ is_partner_initiated }) => is_partner_initiated === false,
   )
-  let anchorBalance = 0
-  for (let selfInitiatedChannel of selfInitiatedChannels) {
-    anchorBalance += 330
-  }
+  const satsAnchorOutput = 330
+  const anchorBalance = selfInitiatedChannels.length * satsAnchorOutput
 
   const total =
     chain_balance +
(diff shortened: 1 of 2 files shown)


src/services/lnd/utils.ts
```diff
@@ -115,8 +115,19 @@ export const lndBalances = async (lnd: AuthenticatedLnd) => {
...
 
+  // adds 330 sats for every selfInitiatedChannel even if not all channels use anchor
+  const { channels } = await getChannels({ lnd })
+  const selfInitiatedChannels = channels.filter(
+    ({ is_partner_initiated }) => is_partner_initiated === false,
+  )
+  let anchorBalance = 0
+  for (let selfInitiatedChannel of selfInitiatedChannels) {
+    anchorBalance += 330
+  }
```
nicolasburtey, line 126: Const satsAnchorOutput = 330
Let anchorBalance = selfInitiatedChannel * satsAnchorOutput

nicolasburtey (changes requested):
You don't need a loop here

nicolasburtey:
Also check code is failing. Make sure to run 'make check-code' and debug any issue that would be shown

src/services/lnd/utils.ts
openoms: I have used a loop because selfInitiatedChannels is an array.

nicolasburtey: Ok then you can just take the length of the array

openoms: thanks for the suggestion, made the change in [be1f13c](https://github.com/GaloyMoney/galoy/pull/1520/commits/be1f13c8748a814e458375c2bde6117898cf140b) and tests are passing locally

nicolasburtey (approved):

PR #1525, opened 2022-08-10
feat: auto balance internal channels

nicolasburtey:


nicolasburtey:
@vindard @dolcalmi how do you feel about testing for this PR?
I'm using the balanceofsatoshis which in itself is already tested, so not sure whether I need to do in depth testing, but let me know if you think this would be important to do a before/after rebalancing and check channel balance.

vindard:
> before/after rebalancing and check channel balance

I think a simple before/after check could be useful still. It would test the `rebalancingInternalChannels` function generally so that even if we switch up the internal implementation we'd still have the expected behaviour encoded.

nicolasburtey:
note: tests may be flaky. actively trying to improve it.

.gitignore
```diff
@@ -26,3 +26,7 @@ public/*
 coverage
 lib
 artifacts
+
+tls.cert
```
bodymindarts, line 30: Why are these needed? Are they artefacts resulting from a 'normal' test run or specific to your dev workflow? If they were just a 1 off artifact I wouldn't add them.

dev/lnd/lnd.conf
```diff
@@ -1,32 +1,31 @@
 [Application Options]
 
 accept-keysend=1
+; keysend-hold-time=2s
```
bodymindarts, line 4: what does the `;` signify? is it commented out? If so shouldn't we delete it?

package.json
```diff
@@ -6,16 +6,16 @@
...
-    "cron": "yarn build && source scripts/export-local.sh && node src/servers/cron.js",
+    "cron": ". ./.envrc && yarn build && node lib/servers/cron.js",
     "exporter": "yarn build && node lib/servers/exporter.js",
-    "lnd-connection": "source scripts/export-local.sh && echo $MACAROON | base64 -d > ./macaroon && echo $TLS | base64 -d > ./tls.cert && lncli --network regtest --macaroonpath=./macaroon --tlscertpath=./tls.cert --rpcserver=${LNDDNS}:${LNDRPCPORT} getinfo",
-    "lncli": "source scripts/export-local.sh && echo $MACAROON | base64 -d > ./macaroon && echo $TLS | base64 -d > ./tls.cert && lncli --network regtest --macaroonpath=./macaroon --tlscertpath=./tls.cert --rpcserver=${LNDDNS}:${LNDRPCPORT}",
-    "bitcoin-cli": "source ./scripts/export-local.sh && kubectl exec bitcoind-0 -- bitcoin-cli -rpcuser=rpcuser -rpcpassword=$BITCOINDRPCPASS -regtest",
+    "lncli1": "echo $LND1_MACAROON | base64 -d > ./macaroon1 && echo $LND1_TLS | base64 -d > ./tls.cert && lncli --network regtest --macaroonpath=./macaroon1 --tlscertpath=./tls.cert --rpcserver=${LND1_DNS}:${LND1_RPCPORT}",
```
bodymindarts, line 16: can't we put these under bin/ instead of driving them via yarn? They have no nodejs dependencies.

test/helpers/lightning.ts
```diff
@@ -211,11 +215,22 @@ export const resetLnds = async () => {
 }
 
 export const closeAllChannels = async ({ lnd }) => {
+  let channels
+  try {
+    ;({ channels } = await getChannels({ lnd }))
```
bodymindarts, line 220: what does the `;` do?

src/services/lnd/utils-bos.ts
```diff
@@ -0,0 +1,91 @@
...
+
+import { LndService } from "."
+
+export const reconnectNodes = async () => {
+  const lndService = LndService()
+  if (lndService instanceof Error) throw lndService
+
+  const lnds = getLnds({ type: "offchain", active: true })
+
+  for (const lnd of lnds) {
```
vindard, line 21: Why did we change these lines away from:
```ts
const lndsParamsAuth = getLnds({ type: "offchain" })
for (const { lnd } of lndsParamsAuth) {
  await reconnect({ lnd })
}
```

It looks like `reconnect` takes a `{ lnd: <Authenticated LND gRPC API Object> }` argument.

vindard (changes requested):
Looks mostly ok to me, just had one question that might be a bug.

src/services/lnd/utils-bos.ts
nicolasburtey: yup. you've catched a bug! maybe that's why reconnect was flacky 😒

test/helpers/lightning.ts
nicolasburtey: it's part of the auto formatting we're doing - because we choose some settings a year ago for lint.

I think it's needed whenever we do a deconstruct that have an await. 

not sure why exactly, this is the common pattern in our codebase.

package.json
nicolasburtey: revert the changes. we can update this back once we have the deterministic lnd on regtest

package.json
```diff
@@ -6,16 +6,16 @@
...
-    "cron": "yarn build && source scripts/export-local.sh && node src/servers/cron.js",
+    "cron": ". ./.envrc && yarn build && node lib/servers/cron.js",
     "exporter": "yarn build && node lib/servers/exporter.js",
-    "lnd-connection": "source scripts/export-local.sh && echo $MACAROON | base64 -d > ./macaroon && echo $TLS | base64 -d > ./tls.cert && lncli --network regtest --macaroonpath=./macaroon --tlscertpath=./tls.cert --rpcserver=${LNDDNS}:${LNDRPCPORT} getinfo",
-    "lncli": "source scripts/export-local.sh && echo $MACAROON | base64 -d > ./macaroon && echo $TLS | base64 -d > ./tls.cert && lncli --network regtest --macaroonpath=./macaroon --tlscertpath=./tls.cert --rpcserver=${LNDDNS}:${LNDRPCPORT}",
-    "bitcoin-cli": "source ./scripts/export-local.sh && kubectl exec bitcoind-0 -- bitcoin-cli -rpcuser=rpcuser -rpcpassword=$BITCOINDRPCPASS -regtest",
+    "lncli1": "lncli --network regtest --macaroonpath=./dev/lnd/lnd1.macaroon --tlscertpath=./dev/lnd/tls.cert --rpcserver=${LND1_DNS}:${LND1_RPCPORT}",
+    "lncli2": "lncli --network regtest --macaroonpath=./dev/lnd/lnd2.macaroon --tlscertpath=./dev/lnd/tls.cert --rpcserver=${LND2_DNS}:${LND2_RPCPORT}",
```
bodymindarts, line 17: I thought I left a comment here about moving these scripts to `bin` since they have no dependency on nodejs so running them via `yarn` has disadvantages.
But it perhaps I forgot to submit it.

bodymindarts (approved):

package.json
nicolasburtey: you did. I just responded that we're going to change the script anyway with the deterministic lnd and so we can update this later.

PR #1526, opened 2022-08-11
feat: add spans to update-invoices sub-functions

vindard:
## Description

This PR is to wrap the handle-invoice sub functions in their own spans so that we can collect function-level attributes for each invoice processed.

### `wrapAsyncToRunInSpan`
We are now switching away from `asyncRunInSpan` to `wrapAsyncToRunInSpan` to be able to standardise how we name/namespace things more easily and to be able to flag if a span is a `root` span (doesn't get nested under other spans).

### Auto wrapping
If we import a module like:
-  `import * as Wallets from "@app/wallets"` there is no wrapping
-  `import { Wallets } from "@app"` there is auto wrapping

For this PR, I mixed the two by importing Wallets twice and renaming the one with wrapping at the import level. This was to not doubly wrap `declineHeldInvoices` but also to apply missing wrapping to the other `Wallets` methods in `trigger.ts`.

### DIFF
--- a/src/app/wallets/update-pending-invoices.ts
+++ b/src/app/wallets/update-pending-invoices.ts
@@ -17,7 +17,11 @@ import {
 } from "@services/mongoose"
 import { NotificationsService } from "@services/notifications"
 import * as LedgerFacade from "@services/ledger/facade"
-import { addAttributesToCurrentSpan } from "@services/tracing"
+import {
+  addAttributesToCurrentSpan,
+  asyncRunInSpan,
+  SemanticAttributes,
+} from "@services/tracing"
 
 import { elapsedSinceTimestamp, runInParallel } from "@utils"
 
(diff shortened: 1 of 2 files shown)


src/app/wallets/update-pending-invoices.ts
```diff
@@ -239,65 +255,79 @@ export const declineHeldInvoice = async ({
...
-  // TODO: REMOVE THIS after hodl invoices has been deployed for 24 hours.
-  if (lnInvoiceLookup.isSettled) {
-    addAttributesToCurrentSpan({ ["isHodlInvoice"]: false })
-    const walletInvoice = await WalletInvoicesRepository().findByPaymentHash(paymentHash)
-    if (walletInvoice instanceof Error) return walletInvoice
-    return updatePendingInvoice({ walletInvoice, logger })
-  }
-  addAttributesToCurrentSpan({ ["isHodlInvoice"]: true })
+}): Promise<boolean | ApplicationError> =>
+  asyncRunInSpan(
```
dolcalmi, line 259: `declineHeldInvoice` is already wrapped please only add missing attributes

src/app/wallets/update-pending-invoices.ts
```diff
@@ -78,158 +82,170 @@ const updatePendingInvoice = async ({
...
-      recipientWalletDescriptor: walletInvoiceReceiver.recipientWalletDescriptor,
-      amountToCreditReceiver: {
-        usd: walletInvoiceReceiver.usdToCreditReceiver,
-        btc: walletInvoiceReceiver.btcToCreditReceiver,
-      },
-      bankFee: {
-        usd: walletInvoiceReceiver.usdBankFee,
-        btc: walletInvoiceReceiver.btcBankFee,
+}): Promise<boolean | ApplicationError> =>
+  asyncRunInSpan(
```
dolcalmi, line 86: please use `wrapAsyncToRunInSpan` and then `addAttributesToCurrentSpan`

dolcalmi (changes requested):

src/app/wallets/update-pending-invoices.ts
vindard: Are you sure? The prompt for this PR was the `isHodlInvoice` attribute being assigned to the parent `handleHeldInvoices` span resulting in us losing the granularity from individual runs of this function -> https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/eaYcthFH7Uj

src/app/wallets/update-pending-invoices.ts
vindard: Ah are we doing this now instead of using `asyncRunInSpan`?

src/app/wallets/update-pending-invoices.ts
vindard: Made some selective-wrapping changes in trigger from here: https://github.com/GaloyMoney/galoy/pull/1526/commits/c215222764e2f1b83f309853366bbebb89021e02

src/app/wallets/update-pending-invoices.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1526/commits/50c2399e9e6f0f7d095eb88a1f2d289bbba7ec05

dolcalmi (dismissed):

dolcalmi (dismissed):

dolcalmi (approved):

PR #1527, opened 2022-08-11
fix: populate testAccount role in create-user

bodymindarts:


src/app/users/create-user.ts
```diff
@@ -38,7 +39,11 @@ export const createUser = async ({
   const phoneNumberValid = checkedToPhoneNumber(phone)
   if (phoneNumberValid instanceof Error) return phoneNumberValid
 
-  const userRaw: NewUserInfo = { phone: phoneNumberValid, phoneMetadata }
+  const userRaw: NewUserInfo = {
+    phone: phoneNumberValid,
+    phoneMetadata,
+    role: getTestAccounts().find(({ phone }) => phone === phoneNumberValid)?.role,
+  }
```
samerbuna, line 46: We don't need this change for the test accounts we create locally (through tests) but I'm guessing we need it in production?

Looking up test-accounts every time we create a user is not ideal. Can we think of an alternative?

bodymindarts: Agree not perfect solution - hack is fine though IMO until we implement proper authorization logic. (ie casbin)

daviroo (approved):
Had a discussion on a call with @bodymindarts.  We tripped over the issue in staging where the test accounts weren't created with the correct permissions.  I think it's ok to have this hack in for now until we do something better.

PR #1531, opened 2022-08-14
test: add e2e usd invoice test

bodymindarts:


dev/stablesats.yml
```diff
@@ -0,0 +1,19 @@
+pubsub:
+  host: redis
+
+price_server:
+  enabled: true
+  server:
+    listen_port: 3325
+  fees:
+    base_fee_rate: 0.005
```
nicolasburtey, line 9: this represents 50 bps? I think the dealer v1 had more like 5 bps as the default value

src/services/dealer-price/proto/services/price/v1/price_service.proto
```diff
@@ -0,0 +1,83 @@
+syntax = "proto3";
```
nicolasburtey, line 1: could we merge this one first? 

https://github.com/GaloyMoney/galoy/pull/1429/files

to get a smaller diff

bodymindarts: done

test/helpers/redis.ts
```diff
@@ -19,3 +23,43 @@ export const clearLimitersWithExclusions = async (exclusions: string[]) => {
...
+let ioRedis
+let pricePublishInterval
+export const publishOkexPrice = () => {
+  ioRedis = new Redis(6379, process.env.REDIS_0_INTERNAL_IP || "localhost")
+  pricePublishInterval = setInterval(() => {
+    const timestamp = Math.floor(new Date().getTime() / 1000)
+    const message = {
+      meta: {
+        publishedAt: timestamp,
+        correlationId: crypto.randomUUID(),
```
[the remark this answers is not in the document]
bodymindarts: `CorrelationId` is a pattern by which you can link together discrete process steps that don't necessarily have a parent / child relationship (like when using `trace_id` / `span_id`).

In the Stablesats example it makes it possible to tie together the receiving of the price update from okex with the call to the price server.

see this query for instance https://ui.honeycomb.io/galoy/datasets/jc-stablesats-dev/result/dScQBZGeetV

dev/stablesats.yml
bodymindarts: You're right there is a `0` missing https://github.com/GaloyMoney/dealer/blob/e22cad5c61909cf532c3f4eb549fcb0673f3d176/dealer/test/integration/DealerPriceService.spec.ts#L16-L18

test/helpers/redis.ts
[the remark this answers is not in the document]
nicolasburtey: seems the [opentelemetry convention](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/messaging/) for that is `conversation_id`, but I find `message_id` more meaningful. 

anyway feels free to keep the name if you prefer it like that

test/e2e/servers/graphql-main-server/with-auth-requests.spec.ts
```diff
@@ -336,6 +353,36 @@ describe("graphql", () => {
...
+      const input = {
+        walletId: usdWalletId,
+        amount: 1000,
+        memo: "This is a lightning invoice",
+      }
+      const result = await apolloClient.mutate({ mutation, variables: { input } })
+      const { invoice, errors } = result.data.lnUsdInvoiceCreate
+      expect(errors).toHaveLength(0)
+      expect(invoice).toHaveProperty("paymentRequest")
+      expect(invoice.paymentRequest.startsWith("lnbcrt4")).toBeTruthy()
```
nicolasburtey, line 369: shouldn't we check that the amount in sats is the correct one?

test/e2e/servers/graphql-main-server/with-auth-requests.spec.ts
```diff
@@ -336,6 +353,36 @@ describe("graphql", () => {
...
+      const input = { walletId: usdWalletId, amount: 1000 }
+      const result = await apolloClient.mutate({ mutation, variables: { input } })
+      const { invoice, errors } = result.data.lnUsdInvoiceCreate
+      expect(errors).toHaveLength(0)
+      expect(invoice).toHaveProperty("paymentRequest")
+      expect(invoice.paymentRequest.startsWith("lnbcrt4")).toBeTruthy()
+      expect(invoice).toHaveProperty("paymentHash")
+      expect(invoice).toHaveProperty("paymentSecret")
+    })
+  })
```
nicolasburtey, line 384: this test doesn't seems to bring more value than the prior one, as we're testing the dealer integration here and the only difference should be in price/amount

test/e2e/servers/graphql-main-server/with-auth-requests.spec.ts
bodymindarts: That would couple this test to the business logic of the dealer. The purpose of e2e tests is to assert that all cross service boundary communication driven by the public interface are working.

The dealer business logic is being tested in https://github.com/GaloyMoney/stablesats-rs/blob/698cb07c56625ecb3e2924f2ec27feea5f105ff6/price-server/tests/price_app.rs#L20. If it changes we don’t want to break this test.

bodymindarts: In general we are using these e2e tests for too much. Imo e we could probably delete quite a few completely or move the assertions into integration and unit tests. 

That would reduce our testing runtime while keeping the same correctness guarantees

test/e2e/servers/graphql-main-server/with-auth-requests.spec.ts
bodymindarts: True. I just copy pasted the 2 tests from the non usd structure. Will delete it

test/e2e/servers/graphql-main-server/with-auth-requests.spec.ts
nicolasburtey: integration tests are currently taking longer than e2e tests

nicolasburtey: > The dealer business logic is being tested in https://github.com/GaloyMoney/stablesats-rs/blob/698cb07c56625ecb3e2924f2ec27feea5f105ff6/price-server/tests/price_app.rs#L20. If it changes we don’t want to break this test.

that makes sense

nicolasburtey (approved):

test/helpers/redis.ts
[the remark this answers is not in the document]
bodymindarts: Thanks for the link - very useful article!

A few things to note:
- this is a field in the message body (ie managed by our own code) and not metadata automatically added by the pubsub system. 
- `message_id` in the convention doc is defined as `A value used by the messaging system as an identifier for the message, represented as a string.` Redis (unlike their examples of Kafka and RabbitMQ) doesn't have such a message id.
- If this were purely for message passing I agree that `conversation_id` might fit well. As they point out: `conversation ID (sometimes called correlation ID). ` I prefer correlation here because it is more general  as we are using it to tie a link between a received message and a grpc response.

test/e2e/servers/graphql-main-server/with-auth-requests.spec.ts
nicolasburtey: btw the `4` in `lnbcrt4` still implies some sats amount been returned (just the first digit)

bodymindarts: > integration tests are currently taking longer than e2e tests
=> unit tests FTW!
> btw the 4 in lnbcrt4 still implies some sats amount been returned (just the first digit)
Okay I used the same number of chars as for the other tests. I doubt the first digit would change - but if it does and the test breaks we can remove it.

vindard: Based on the fix from #1558 this test should've failed here (see [BOLT-11 spec](https://github.com/lightning/bolts/blob/master/11-payment-encoding.md#human-readable-part)). Why did it pass?

PR #1535, opened 2022-08-17
refactor: reimburse sats on failed USD payment

vindard:
## Description

This PR is to add a condition where we reimburse the original sats sent instead of reverting the original usd transaction for a failed USD LN payment.

The original USD transaction entries are left untouched, and the new code simply adds a `recordReceive` call which adds 2 new BTC transactions to the ledger (a credit to the user, and a corresponding debit to lnd).

I explored optionally recreating the part of the medici [`void`](https://github.com/flash-oss/medici/blob/4d83496c158b4907393689d1c4c40a97c63a8340/src/Book.ts#L215) method that only updates the original journal/transactions but skips the [`addReversedTransactions`](https://github.com/flash-oss/medici/blob/4d83496c158b4907393689d1c4c40a97c63a8340/src/Book.ts#L254) call, but doing this would significantly expand the scope of this PR (introduce a Journal schema, replicate consistency guarantees and add tests etc.). So given this, the original USD transactions remain untouched.

### DIFF
--- a/src/app/payments/update-pending-payments.ts
+++ b/src/app/payments/update-pending-payments.ts
@@ -103,13 +103,16 @@ const updatePendingPayment = async ({
     paymentHash,
   })
   if (lnPaymentLookup instanceof Error) {
-    const lightningLogger = logger.child({
-      topic: "payment",
-      protocol: "lightning",
-      transactionType: "payment",
-      onUs: false,
-    })
-    lightningLogger.error({ err: lnPaymentLookup }, "issue fetching payment")
+    logger.error(
+      {
+        err: lnPaymentLookup,
+        topic: "payment",
+        protocol: "lightning",
+        transactionType: "payment",
+        onUs: false,
+      },
+      "issue fetching payment",
+    )
     return lnPaymentLookup
   }
 
@@ -120,8 +123,8 @@ const updatePendingPayment = async ({
   }
 
   if (status === PaymentStatus.Settled || status === PaymentStatus.Failed) {
-    const ledgerService = LedgerService()
     return LockService().lockPaymentHash(paymentHash, async () => {
+      const ledgerService = LedgerService()
       const recorded = await ledgerService.isLnTxRecorded(paymentHash)
       if (recorded instanceof Error) {
         paymentLogger.error({ error: recorded }, "we couldn't query pending transaction")
--- a/src/app/payments/update-pending-payments.ts
+++ b/src/app/payments/update-pending-payments.ts
@@ -7,13 +7,14 @@ import {
   LedgerTransactionType,
   UnknownLedgerError,
 } from "@domain/ledger"
-import { WalletCurrency } from "@domain/shared"
+import { ErrorLevel, WalletCurrency } from "@domain/shared"
 
 import { LedgerService, getDealerWalletIds } from "@services/ledger"
 import { LndService } from "@services/lnd"
 import { LockService } from "@services/lock"
 import { WalletsRepository } from "@services/mongoose"
 import { PaymentFlowStateRepository } from "@services/payment-flow"
+import { recordExceptionInCurrentSpan } from "@services/tracing"
 
 import { Wallets } from "@app"
 import { runInParallel } from "@utils"
@@ -199,6 +200,7 @@ const updatePendingPayment = async ({
           if (voided instanceof Error) {
             const error = `error voiding payment entry`
             logger.fatal({ success: false, result: lnPaymentLookup }, error)
+            recordExceptionInCurrentSpan({ error: voided, level: ErrorLevel.Critical })
             return voided
           }
         } else {
@@ -210,6 +212,10 @@ const updatePendingPayment = async ({
           if (reimbursed instanceof Error) {
             const error = `error reimbursing usd payment entry`
             logger.fatal({ success: false, result: lnPaymentLookup }, error)
+            recordExceptionInCurrentSpan({
+              error: reimbursed,
+              level: ErrorLevel.Critical,
+            })
             return reimbursed
           }
         }
--- a/src/app/payments/update-pending-payments.ts
+++ b/src/app/payments/update-pending-payments.ts
@@ -9,7 +9,7 @@ import {
 } from "@domain/ledger"
 import { WalletCurrency } from "@domain/shared"
 
-import { LedgerService } from "@services/ledger"
+import { LedgerService, getDealerWalletIds } from "@services/ledger"
 import { LndService } from "@services/lnd"
 import { LockService } from "@services/lock"
 import { WalletsRepository } from "@services/mongoose"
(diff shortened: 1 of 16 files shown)


nicolasburtey:
> I explored optionally recreating the part of the medici [void](https://github.com/flash-oss/medici/blob/4d83496c158b4907393689d1c4c40a97c63a8340/src/Book.ts#L215) method that only updates the original journal/transactions but skips the [addReversedTransactions](https://github.com/flash-oss/medici/blob/4d83496c158b4907393689d1c4c40a97c63a8340/src/Book.ts#L254) call, but doing this would significantly expand the scope of this PR (introduce a Journal schema, replicate consistency guarantees and add tests etc.). So given this, the original USD transactions remain untouched.

Your thinking on this was to update prior ledger transaction in place? 

if this is the case: typically we don't want to change any written transaction. a ledger should be thoughts as an append only database.

src/app/wallets/reimburse-failed-usd.ts
```diff
@@ -0,0 +1,80 @@
...
+  const paymentHash = paymentFlow.paymentHashForFlow()
+  if (paymentHash instanceof Error) return paymentHash
+
+  const metadata: FailedPaymentLedgerMetadata = {
+    hash: paymentHash,
+    type: LedgerTransactionType.Payment,
+    pending: false,
+    related_journal: journalId,
+
+    usd: Number(
```
nicolasburtey, line 36: somewhat a duplicate from `centsAmount`?

src/app/wallets/reimburse-failed-usd.ts
```diff
@@ -0,0 +1,80 @@
...
+  paymentFlow: PaymentFlow<S, R>
+  journalId: LedgerJournalId
+  accountId: AccountId
+}): Promise<true | ApplicationError> => {
+  const paymentHash = paymentFlow.paymentHashForFlow()
+  if (paymentHash instanceof Error) return paymentHash
+
+  const metadata: FailedPaymentLedgerMetadata = {
+    hash: paymentHash,
+    type: LedgerTransactionType.Payment,
```
nicolasburtey, line 32: shouldn't we have a more precise type for this?

src/app/payments/update-pending-payments.ts
```diff
@@ -187,15 +188,27 @@ const updatePendingPayment = async ({
...
+          }
+        } else {
+          const reimbursed = await Wallets.reimburseFailedUsdPayment({
+            journalId: pendingPayment.journalId,
+            paymentFlow,
+            accountId: wallet.accountId,
+          })
+          if (reimbursed instanceof Error) {
+            const error = `error reimbursing usd payment entry`
+            logger.fatal({ success: false, result: lnPaymentLookup }, error)
```
nicolasburtey, line 209: as we rely less and less on logs and more and more on opentelemetry for alert, I'm wondering if we should not have an event here in addition, or instead?

the current `Error` being returned here would not triggered back an alert on honeycomb.

ie: `InvalidAccountForWalletIdError`, `CouldNotFindWalletForWalletCurrencyError`, `CouldNotListWalletsFromAccountIdError` are not `ErrorLevel.Critical` (nor they should be in other contexts)

src/app/wallets/reimburse-failed-usd.ts
```diff
@@ -0,0 +1,80 @@
...
+  )
+  if (btcWallet === undefined) return new CouldNotFindWalletForWalletCurrencyError()
+  const btcWalletDescriptor = { id: btcWallet.id, currency: btcWallet.currency }
+
+  const result = await LedgerFacade.recordReceive({
+    description: "usd failed payment reimburse",
+    recipientWalletDescriptor: btcWalletDescriptor,
+    amountToCreditReceiver: paymentFlow.totalAmountsForPayment(),
+    metadata,
+    txMetadata,
```
nicolasburtey, line 75: it's probably not the point of this PR to change this, but the `metadata` and txMetadata` property is pretty confusing.

src/domain/ledger/index.ts
```diff
@@ -70,6 +73,7 @@ export const checkedToLedgerTransactionId = (
 export const inputAmountFromLedgerTransaction = (
   txn: LedgerTransaction<WalletCurrency>,
 ) => {
-  const fee = txn.currency == WalletCurrency.Usd ? txn.feeUsd : txn.fee
+  const fee = txn.currency == WalletCurrency.Usd ? txn.centsFee : txn.satsFee
```
nicolasburtey, line 76:
```suggestion
  const fee = txn.currency === WalletCurrency.Usd ? txn.centsFee : txn.satsFee
```

nicolasburtey (commented):
A few comments but looks good to me overall.

Some additional commits to this PR that you can merge if you want: https://github.com/GaloyMoney/galoy/pull/1537

src/app/wallets/reimburse-failed-usd.ts
vindard: Yup, but we need this duplicate in for now for consistency with the `LedgerTransaction` type across other non-`"payment"` tx types. Will be addressed as part of the cleanup [here](https://github.com/GaloyMoney/galoy/projects/9#card-84279899).

src/app/wallets/reimburse-failed-usd.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1535/commits/f44d455388e9d38ac8496e30e19c17dc78c04b65

src/app/payments/update-pending-payments.ts
vindard: Good point, done: https://github.com/GaloyMoney/galoy/pull/1535/commits/750bdbeaac5f6afa95c41013e310c4f4e5e920b9

src/domain/ledger/index.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1535/commits/97d377e1ea62175118eb7e3da86b4069587165a7

src/app/wallets/reimburse-failed-usd.ts
vindard: Yea I think we'd need to clean this up at some point when we sort out how we'll be reading from the `transaction_metadata` table (gql [dataloader](https://github.com/GaloyMoney/galoy/pull/1427#issuecomment-1187422982) planned piece of work).

nicolasburtey:
@vindard there is a conflict with this PR

src/app/wallets/reimburse-failed-usd.ts
nicolasburtey: maybe I'm missing what is being implemented, but I had in mind having something else than `LedgerTransactionType.Payment`. 

maybe something like `LedgerTransactionType.PaymentReimbursement.`

nicolasburtey (approved):

PR #1536, opened 2022-08-18
feat: add callback for invoice updates

dolcalmi:


nicolasburtey:
@dolcalmi I think for the callback design, we should use another process to actually process the callback itself, and use redis or postgres to create a queue and pass the messages between the galoy backend and the "callback-trigger" Pod.

the main reason is that callback really needs a retry with a backoff mechanism if companies start to rely on it. if we don't have any retry, and if their services is down, even only momentarily, they would not miss the event and have to do a full reconciliation/rescan to see if there is any discrepancies.

event are important because they are used to synchronize accounting. it's just not an informal notification that, if it's lost, is not really an issue. 

(note: assuming the "callback-trigger" would be responsible to just do callback from redis messages in the form of queue, this is the type of Pod that could be done in Rust? this would take much less memory than having to load nodejs)

the behavior on the callback could be:
- before doing a callback, consume the message on the queue from redis
- if a callback returned an error,  place back a message on the queue with a "execution time" and "retry count". there should be an exponential backoff. ie: 5s, 15s, 2min, 10min, 1h, 4h. then skip the messages.

src/graphql/main/schema.graphql
```diff
@@ -322,6 +327,11 @@ input LnInvoiceCreateInput {
   """
   amount: SatAmount!
 
+  """
+  Optional callback for the lightning invoice updates.
```
nicolasburtey, line 331: I don't think the callback should be at the invoice level. I think it should be at the account level. 

there are some reasons for that:
- the first one is that if the invoice is set at the invoice level, it will be easier for an attacker to have many urls as callback and trigger all of them with 1 sats invoices. 
- if we implement the callback at the account level, we can rate limit that, so that the callback url can only be changed once a day or something like that

The company/dev I talked to who wants to have callback functionality, they would want to have all the invoices coming back as a callback. so having to add callback for invoices one by one seems unecessary. 

the callback itself should have the information necessary for them to process it in the payload, instead of relying on a different callback url for each invoice.

nicolasburtey:
another attack vector to have in mind is the injecting value to discover environement variable. ie: 

callback being something along like: `https://my-server/callback?${process.env}` (this might already been filtered out with the current regex, I didn't checked.

one benefits of splitting the "`trigger-callback`" to it own pod is that fact it would not get access to all the environment variable the main pod has

nicolasburtey:
I think I like the suggestion of Justin that if you want to add a callback to the invoice when you use the "onBehalfOf", you have to be authenticated.

Like this, it could be possible to associate a callback domain with an account if we see potential abuse/attack from it

I think this would well for the platform that want to use this feature. They would not mind to authenticate

PR #1539, opened 2022-08-18
fix: add missing midPriceRatio return from dealer

vindard:
## Description

Fixes bug where we missed returning the mid-price-ratio dealer from dealer and were always falling back to previous non-dealer price server.

### DIFF
--- a/src/app/shared/mid-price.ts
+++ b/src/app/shared/mid-price.ts
@@ -12,7 +12,6 @@ import {
 } from "@services/tracing"
 
 const usdHedgeEnabled = getDealerConfig().usd.hedgingEnabled
-
 const dealer = NewDealerPriceService()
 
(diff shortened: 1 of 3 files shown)


bodymindarts (approved):
Lgtm. Why use the mock again instead of integrating (in the integration test)?

vindard:
> Lgtm. Why use the mock again instead of integrating (in the integration test)?

Was keeping it simple in this one for now. There's potential edge cases with live services where both services return the same price, or the price changes between calls for the values we're comparing.

bodymindarts:
You can babe the stabkesats price response deterministic by calling https://github.com/GaloyMoney/galoy/blob/2ca9f7630db38b87fe7809431f22d4278e1bbaa5/test/helpers/redis.ts#L29 in a beforeAll

PR #1551, opened 2022-08-20
feat: add invoice payment status query

dolcalmi:
Add `lnInvoicePaymentStatus` public query (same behavior and name as its subscription). This will avoid the use of fee probe for status validation (-_-) 

### How to test

```graphql
query LnInvoicePaymentStatus($input: LnInvoicePaymentStatusInput!) {
  status: lnInvoicePaymentStatus(input: $input)
}
```

```grapqhl
{ "input": { "paymentRequest": "lnbcrt500u...k8n2p" } }
```

### DIFF
--- a/src/graphql/root/query/ln-invoice-payment-status.ts
+++ b/src/graphql/root/query/ln-invoice-payment-status.ts
@@ -1,0 +1,29 @@
+import { Lightning } from "@app"
+
+import { GT } from "@graphql/index"
+import { mapError } from "@graphql/error-map"
+import InvoicePaymentStatus from "@graphql/types/scalar/invoice-payment-status"
+import LnInvoicePaymentStatusInput from "@graphql/types/object/ln-invoice-payment-status-input"
+
+const LnInvoicePaymentStatusQuery = GT.Field({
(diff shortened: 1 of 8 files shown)


vindard (approved):

src/graphql/root/query/ln-invoice-payment-status.ts
```diff
@@ -0,0 +1,29 @@
...
+    const { paymentRequest } = args.input
+    if (paymentRequest instanceof Error) throw paymentRequest
+
+    const paymentStatusChecker = await Lightning.PaymentStatusChecker(paymentRequest)
+    if (paymentStatusChecker instanceof Error) throw mapError(paymentStatusChecker)
+
+    const paid = await paymentStatusChecker.invoiceIsPaid()
+    if (paid instanceof Error) throw mapError(paid)
+
+    if (paid) return "PAID"
```
nicolasburtey, line 23: instead of using "PAID" and "PENDING" as String, we should have an enum or something similar?

dolcalmi: I will include it in another PR (I just copy code from subscription to this new query).

PR #1574, opened 2022-08-24
fix(tracing): only update span with same/higher ranked errors

vindard:
## Description

This PR checks the current span first before updating its error attributes when recording a new exception in span. It only updates attributes if new error is at same or higher error level. This is to mitigate the issue that we first discovered in #1567 where `critical` errors were being surpressed.

The new errors are recorded as events either way, and this change will only affect the attributes being written to the span.

### DIFF
--- a/src/services/tracing.ts
+++ b/src/services/tracing.ts
@@ -254,12 +254,25 @@ const updateErrorForSpan = ({
 const recordException = (span: Span, exception: Exception, level?: ErrorLevel) => {
   // @ts-ignore-next-line no-implicit-any error
   const errorLevel = level || exception["level"] || ErrorLevel.Warn
+
+  // Write error attributes if update checks pass
   if (updateErrorForSpan({ span, errorLevel })) {
     span.setAttribute("error.level", errorLevel)
     // @ts-ignore-next-line no-implicit-any error
     span.setAttribute("error.name", exception["name"])
     span.setAttribute("error.message", exception["message"])
   }
+
+  // Append error with next index
+  let nextIdx = 0
+  while (span.attributes[`error.${nextIdx}.level`] !== undefined) {
+    nextIdx++
+  }
+  span.setAttribute(`error.${nextIdx}.level`, errorLevel)
+  // @ts-ignore-next-line no-implicit-any error
+  span.setAttribute(`error.${nextIdx}.name`, exception["name"])
+  span.setAttribute(`error.${nextIdx}.message`, exception["message"])
+
   span.recordException(exception)
   span.setStatus({ code: SpanStatusCode.ERROR })
 }


bodymindarts:
Why move it out of errors file?

vindard:
> Why move it out of errors file?

It was being included in the exhaustive check in the error-map ([here](https://github.com/GaloyMoney/galoy/pull/1574/files#diff-ff969139182e57d8fc79cdb1b5a83000fb99932e9c509ce84337168cd1b0cfbdL402)).

bodymindarts:
> > Why move it out of errors file?
> 
> It was being included in the exhaustive check in the error-map ([here](https://github.com/GaloyMoney/galoy/pull/1574/files#diff-ff969139182e57d8fc79cdb1b5a83000fb99932e9c509ce84337168cd1b0cfbdL402)).

I’d prefer that than the circular dependency you introduced

vindard:
> I’d prefer that than the circular dependency you introduced

Fixed https://github.com/GaloyMoney/galoy/pull/1574/commits/a9a419262a647185b7f4a2aa887efb12d941bd73

src/services/tracing.ts
```diff
@@ -238,13 +238,28 @@ export const recordExceptionInCurrentSpan = ({
...
+  span,
+  errorLevel,
+}: {
+  span: Span
+  errorLevel: ErrorLevel
+}): boolean => {
+  const spanErrorRank = RankedErrorLevel.indexOf(span.attributes["error.level"])
+  const errorRank = RankedErrorLevel.indexOf(errorLevel)
+
+  return errorRank >= spanErrorRank
```
nicolasburtey, line 251: if they're equal we're just overriding another error from the same level? is that useful?

src/services/tracing.ts
```diff
@@ -238,13 +238,28 @@ export const recordExceptionInCurrentSpan = ({
...
   // @ts-ignore-next-line no-implicit-any error
   const errorLevel = level || exception["level"] || ErrorLevel.Warn
-  span.setAttribute("error.level", errorLevel)
-  // @ts-ignore-next-line no-implicit-any error
-  span.setAttribute("error.name", exception["name"])
-  span.setAttribute("error.message", exception["message"])
+  if (updateErrorForSpan({ span, errorLevel })) {
+    span.setAttribute("error.level", errorLevel)
+    // @ts-ignore-next-line no-implicit-any error
+    span.setAttribute("error.name", exception["name"])
```
nicolasburtey, line 260: assuming there are multiple errors, maybe we should add them here?

ie: 
```
span.setAttribute("error.name.1", exception["name"])`
span.setAttribute("error.message.1", exception["message"])
```

then:
```
span.setAttribute("error.name.2", exception["name"])`
span.setAttribute("error.message.2", exception["message"])
```


for the new error (assuming we know how many errors we are adding)

src/services/tracing.ts
vindard: At this point it would be down to if we favor exposing earlier or later errors in the call stack. The important thing here though would be that we're not overwriting a higher level with a lower one for our alerting.

We'd be able to get details of all the actual errors still from the `Events` section in the span, and can query with attributes like `exception.message` & `exception.type` (e.g. [here](https://ui.honeycomb.io/galoy/datasets/arvin-dev/result/tsieg7hWCGz/trace/3D9Ja83hgKA?fields%5B%5D=c_name&fields%5B%5D=c_service.name&span=954e8400b489147b) where we got one instance of `InvalidContactAlias` from an explicit `recordException` and another from the function return)

src/services/tracing.ts
vindard: I'm not sure if this would be useful since we also already have all the error details for multiple errors in the `Events` sections (e.g. [trace](https://ui.honeycomb.io/galoy/datasets/arvin-dev/result/tsieg7hWCGz/trace/3D9Ja83hgKA?fields%5B%5D=c_name&fields%5B%5D=c_service.name&span=954e8400b489147b))

vindard: Done: https://github.com/GaloyMoney/galoy/pull/1574/commits/7df2168e978e9300b71a4c0d7d20e79da6d2f76e

bodymindarts (approved):

PR #1576, opened 2022-08-25
feat: add max amounts validation for sats/cents inputs

vindard:
## Description

This PR adds checks in validation methods to make sure values are no greater than a defined max amount. The max amount used was `100,000` BTC since this is the max value that lnd would let you create an invoice for.

These changes were prompted by 2 error message types observed from lnd for the `createHodlInvoice` method with the following amount inputs ([examples traces](https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/6xBE7yPk9wp)):
- (1) `10_000_000_000_001`
  `'invoice amount 100000.00000001 BTC is too large, max is 100000 BTC'`

- (2) `1_064_138_827_560_000_000`
  `'payments of negative value are not allowed, value is -5772328715153993728'`

- (2) `476_057_064_138_827_560_000_000`
  `'payments of negative value are not allowed, value is -1000'`

### DIFF
--- a/src/domain/shared/primitives.ts
+++ b/src/domain/shared/primitives.ts
@@ -26,6 +26,18 @@ export const ZERO_BANK_FEE = {
   btcBankFee: ZERO_SATS,
 }
 
+// 100_000 BTC is the max amount you can make a hodl invoice for in lnd
+export const MAX_SATS = {
+  currency: WalletCurrency.Btc,
+  amount: 10_000_000_000_000n,
+} as const
+
+// Assumes optimistic price under sat-cent parity
+export const MAX_CENTS = {
+  currency: WalletCurrency.Usd,
+  amount: 10_000_000_000_000n,
+}
+
 export const BtcWalletDescriptor = (walletId: WalletId) => {
   return {
     id: walletId,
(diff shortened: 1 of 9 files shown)


src/domain/shared/primitives.ts
```diff
@@ -26,6 +26,18 @@ export const ZERO_BANK_FEE = {
...
+// 100_000 BTC is the max amount you can make a hodl invoice for in lnd
+export const MAX_SATS = {
+  currency: WalletCurrency.Btc,
+  amount: 10_000_000_000_000n,
+} as const
+
+// Assumes optimistic price under sat-cent parity
+export const MAX_CENTS = {
+  currency: WalletCurrency.Usd,
+  amount: 10_000_000_000_000n,
```
vindard, line 38: I just realized, this should probably be based on a pessimistic price to function properly as a max that doesn't trigger lnd errors when it gets converted to sats. Maybe something like `10_000_000_000` which allows you to do less effective sats after conversion, but works as expected for prices down to $1k/BTC.

Does it make sense to have this max here, or should I add checks individually in certain places (like `registerInvoice` method) to check amounts after conversion from USD?

nicolasburtey: a nit: for cents you may want to put the `_` after two `00`? ie: `const hundredk = 100_000_00`

vindard: Oh yes, it's 100B actually. I fixed the format. I left the number as is to be closer to the value we're defining for `MAX_SATS` since this is more for data type validation than business logic https://github.com/GaloyMoney/galoy/pull/1576/commits/e00ef5e1710f0aa5d9973b7162cd20802ef06edd

PR #1583, opened 2022-08-26
debug: add script to clean pending payments for dead lnd nodes

vindard:
## Description

Script to clean pending payments leftover from nodes that have been permanently removed for a given environment. Will fix traces like these: https://ui.honeycomb.io/galoy/datasets/galoy-hack/result/defJvEY2iuA

### DIFF
--- a/src/debug/clean-pending-payments-for-dead-node.ts
+++ b/src/debug/clean-pending-payments-for-dead-node.ts
@@ -1,0 +1,85 @@
+/**
+ * how to run:
+ *	. ./.envrc && yarn ts-node \
+ *		--files \
+ *			-r tsconfig-paths/register \
+ *			-r src/services/tracing.ts \
+ *		src/debug/clean-pending-payments-for-dead-node.ts
+ */
+
+import { LedgerTransactionType, UnknownLedgerError } from "@domain/ledger"
+
+import { isUp } from "@services/lnd/health"
+import { params as unauthParams } from "@services/lnd/unauth"
+
+import { setupMongoConnection } from "@services/mongodb"
+import { LedgerService, translateToLedgerTx } from "@services/ledger"
+import { LockService } from "@services/lock"
+import { baseLogger } from "@services/logger"
+
+import { MainBook } from "@services/ledger/books"
+
+const DEAD_LND_PUBKEYS = [
+  "02ae3c066e67cf3951bb5230cf4e13aee053bc06465cd25f870988b691469140b0", // dead node in hack environment
+]
+
+const listAllPendingPayments = async (): Promise<
+  LedgerTransaction<WalletCurrency>[] | LedgerError
+> => {
+  try {
+    const { results } = await MainBook.ledger({
+      type: LedgerTransactionType.Payment,
+      pending: true,
+    })
+
+    return results.filter((tx) => tx.debit > 0).map((tx) => translateToLedgerTx(tx))
+  } catch (err) {
+    return new UnknownLedgerError(err)
+  }
+}
+
+const main = async (): Promise<true | ApplicationError> => {
+  const pendingPayments = await listAllPendingPayments()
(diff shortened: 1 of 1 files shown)


src/debug/clean-pending-payments-for-dead-node.ts
```diff
@@ -0,0 +1,85 @@
...
+
+import { setupMongoConnection } from "@services/mongodb"
+import { LedgerService, translateToLedgerTx } from "@services/ledger"
+import { LockService } from "@services/lock"
+import { baseLogger } from "@services/logger"
+
+import { MainBook } from "@services/ledger/books"
+
+const DEAD_LND_PUBKEYS = [
+  "02ae3c066e67cf3951bb5230cf4e13aee053bc06465cd25f870988b691469140b0", // dead node in hack environment
```
nicolasburtey, line 23: those should be passed from the cli?

nicolasburtey: we may not need an array. I think it's ok to run the cli once for every node that may needs some clean up (currently running 2 would means we have removed both node which would itself be unlikely)

src/debug/clean-pending-payments-for-dead-node.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    if (DEAD_LND_PUBKEYS.includes(pubkey)) {
+      await LockService().lockPaymentHash(
+        paymentHash,
+        async (): Promise<true | LedgerServiceError> => {
+          const ledgerService = LedgerService()
+          const settled = await ledgerService.settlePendingLnPayment(paymentHash)
+          if (settled instanceof Error) {
+            baseLogger.error({ error: settled }, "no transaction to update")
+            return settled
```
nicolasburtey, line 58: we may want to `continue` here instead instead of `return`?

maybe we should also delete the record from the ledgerService?

vindard: The `return`s inside here are for the `lockPaymentHash` internal function. They wouldn't exit the outer `for` loop I think (based on my testing so far locally).

src/debug/clean-pending-payments-for-dead-node.ts
vindard: Changed: https://github.com/GaloyMoney/galoy/pull/1583/commits/1c39f876c162ab2a5c726bc76bf6ea85d2527a3d

src/debug/clean-pending-payments-for-dead-node.ts
nicolasburtey: ohh yeah ok didn't pay attention to the nest anonymous function!

src/debug/clean-pending-payments-for-dead-node.ts
```diff
@@ -0,0 +1,85 @@
...
+import { params as unauthParams } from "@services/lnd/unauth"
+
+import { setupMongoConnection } from "@services/mongodb"
+import { LedgerService, translateToLedgerTx } from "@services/ledger"
+import { LockService } from "@services/lock"
+import { baseLogger } from "@services/logger"
+
+import { MainBook } from "@services/ledger/books"
+
+const PUBKEY = process.env.PUBKEY
```
nicolasburtey, line 23: I think it would be more typically to pass this as an argument to the function, but it doesn't really matter.

nicolasburtey (dismissed):

src/debug/clean-pending-payments-for-dead-node.ts
vindard: Thinking about this again, it's probably cleaner as a script arg because we don't know if we'd be overwriting some env variable with the env approach. Changed: https://github.com/GaloyMoney/galoy/pull/1583/commits/de7b1b16a86a2ee79c0ab57da00f1ca4f8e465b0

nicolasburtey (approved):

PR #1591, opened 2022-08-27
chore(tracing): add check to catch undefined gql 'source'

vindard:
## Description

This PR is an attempt to find somewhere in the graphql layer where `errors` property is being read from an `undefined` object. Looking at the code, these changes seem like a good first attempt at finding this issue.

Relevant traces: https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/xj4WGXgcjpj

_Found from this query for errors with missing error codes that could become an alert: [query](https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/pvvg6LLveDD)_

### DIFF
--- a/src/graphql/root/subscription/price.ts
+++ b/src/graphql/root/subscription/price.ts
@@ -6,9 +6,12 @@ import { GT } from "@graphql/index"
 import PricePayload from "@graphql/types/payload/price"
 import SatAmount from "@graphql/types/scalar/sat-amount"
 import ExchangeCurrencyUnit from "@graphql/types/scalar/exchange-currency-unit"
+import { UnknownClientError } from "@graphql/error"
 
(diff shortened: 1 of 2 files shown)


src/graphql/root/subscription/price.ts
```diff
@@ -47,6 +50,14 @@ const PriceSubscription = {
     source: { errors: IError[]; satUsdCentPrice?: number },
     args: PriceResolveArgs,
   ) => {
+    if (source === undefined) {
```
dolcalmi, line 53: but the trace in comment is only from myUpdates... do you have a trace for this?

vindard: No trace for this one, but I was thinking behaviour could maybe be the same at some point?

PR #1600, opened 2022-08-30
fix: handle zero ledger entries for zero-amount transactions

vindard:
## Description

This fixes the `UnknownLedgerError` that popped up in [this trace](https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/z9Kv45P5m1t/trace/qyNq5XZqQyK?fields[]=c_name&fields[]=c_service.name&span=60c2b6e7eee28dff) where someone paid an invoice which `995 msats` which got rounded down to `0` in our code and resulted in an empty `entry` that threw the error.

An interesting insight from this error is that amounts that come in on a paid zero-amount invoice can maybe be considered user input if our lnd channel policies don't match our domain expectations.

_Note: There is additional work to improve Facade tracing that I'll likely add in another PR (see [`add-facade-tracing`](https://github.com/GaloyMoney/galoy/tree/add-facade-tracing))_


dev/lnd/lnd.conf
```diff
@@ -25,6 +25,8 @@ bitcoin.active=true
 bitcoin.defaultchanconfs=0
 bitcoin.node=bitcoind
 bitcoin.regtest=true
+; bitcoin.minhtlc=1
+bitcoin.minhtlcout=1
```
vindard, line 29: Got help from Alex to eventually figure this setting out to be able to do the `995 msats` integration test

test/integration/02-user-wallet/02-receive-lightning.spec.ts
```diff
@@ -546,6 +548,54 @@ describe("UserWallet - Lightning", () => {
...
+    ).toBeInstanceOf(NoTransactionsForEntryError)
+    // should be idempotent (not return error when called again)
+    expect(
+      await Wallets.updatePendingInvoiceByPaymentHash({
+        paymentHash: hash as PaymentHash,
+        logger: baseLogger,
+      }),
+    ).not.toBeInstanceOf(Error)
+
+    // Check that new txns are added
```
dolcalmi, line 580: no new txs are added :)

test/integration/02-user-wallet/02-receive-lightning.spec.ts
```diff
@@ -546,6 +548,54 @@ describe("UserWallet - Lightning", () => {
...
+    if (walletInvoice instanceof Error) throw walletInvoice
+    expect(walletInvoice.paid).toBeTruthy()
+
+    // Check that wallet balance is unchanged
+    const finalBalance = await getBalanceHelper(walletIdB)
+    expect(finalBalance).toBe(initBalanceB)
+
+    // Return msats to balance overall accounting
+    const { request } = await createInvoice({ lnd: lndOutside1 })
+    await pay({ lnd: lnd1, request, mtokens })
```
dolcalmi, line 596: nice hack

src/services/ledger/helpers.ts
```diff
@@ -1,4 +1,5 @@
-import { UnknownLedgerError } from "./domain/errors"
+import { UnknownLedgerError } from "@domain/ledger"
```
dolcalmi, line 1: I was not part of ledger refactor but I guess this was on purpose @UncleSamtoshi @bodymindarts could you please check this?

dolcalmi (dismissed):

test/integration/02-user-wallet/02-receive-lightning.spec.ts
vindard: Thank fixed 😅

src/services/ledger/helpers.ts
vindard: Relevant commit here for @UncleSamtoshi / @bodymindarts
https://github.com/GaloyMoney/galoy/pull/1600/commits/52b248d351e7ad140d985fb720c4a735fdd92425

UncleSamtoshi: I dont recall the reason for it and think it would be fine to change. Perhaps @bodymindarts has an opinion on it.

bodymindarts: We want to isolate the ledger internally. It should not depend on @domain. Rather use its own domain types. Exception is dependencies on shared

bodymindarts (commented):
IMO the entry builder should not know about `ValidationError`. Validation is a concern at the boundary of a use case not deep in the business logic. I sense there should be an early return higher in the call stack.

dev/lnd/lnd.conf
nicolasburtey: we could also set the minhtlc to 1000 msats?

I'm not sure whether we need to support sub sats payment use cases.

vindard:
Isolated builder commits to remove in latest [`force-push`](https://github.com/GaloyMoney/galoy/compare/2b31af764e2493f3342e71094a49f23d00da1a7a..3a60eff8a5d9a583722cd929d0e66885c3d5490a)

src/app/wallets/update-pending-invoices.ts
```diff
@@ -113,10 +121,22 @@ const updatePendingInvoiceBeforeFinally = async ({
 
   const {
     lnInvoice: { description },
-    roundedDownReceived,
+    roundedDownReceived: uncheckedRoundedDownReceived,
   } = lnInvoiceLookup
 
   // TODO: validate roundedDownReceived as user input
+  const roundedDownReceived = checkedToSats(uncheckedRoundedDownReceived)
```
vindard, line 128: Is this the correct boundary for this validation?

I opted to do it here instead of inside `lndService.lookupInvoice` because:
- it feels like `roundedDownReceived == 0` is invalid in this context, but could be valid in other contexts
- if we can successfully fetch an invoiceLookup then we can call `declineHeldInvoice` on it instead of simply "not canceling" the invoice and just deleting the db record for it ([here](https://github.com/GaloyMoney/galoy/pull/1600/commits/f253ac1d6204c50ae66d84fce67438a6bb647123#diff-f02445547836eff187be3a4b54e41d92b4223dd7c304847b688ef3016eec317fR113))

bodymindarts (approved):

bodymindarts (approved):

PR #1618, opened 2022-09-01
chore: graphql ws2

nicolasburtey:
fix https://github.com/GaloyMoney/galoy/issues/1365, plus some security consideration using the old library: https://github.com/enisdenjo/graphql-ws/issues/3

I ran into issues when I was importing both subscriptions-transport-ws and graphql-ws at the same time. instead of trying to understand the root of this, I decided to duplicated graphql-server and dynamically, at runtime, one or the other websocket library would be used.

by default the old library would be used - so that we can enable the use of the new library in another PR in the charts/deployment repo.

I recommend that we use a new url for BBW to do the migration from the mobile app. we could use api.bbw.sv to replace the current url. api.bbw.sv would have the graphql-ws library. mainnet.graphql.galoy.io would have the subscriptions-transport-ws library.

legacy server (when no env variable is set) --> api.mainnet.galoy.io
"new" server (when NEW_SERVER is set) --> api.bbw.sv

note: I didn't see there was a way to do backward compability with both subscription services

[ws](https://github.com/websockets/ws) server usage with [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws) backwards compatibility

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@ory/client": "^0.0.1-alpha.123",
     "@ory/kratos-client": "^0.10.1",
     "ajv": "^8.10.0",
-    "apollo-server-core": "^3.10.1",
+    "apollo-server-core": "^3.10.2",
     "apollo-server-errors": "^3.3.1",
     "apollo-server-express": "^3.10.2",
     "axios": "^0.27.2",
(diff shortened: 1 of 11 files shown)


nicolasburtey:
may need to replaced with https://github.com/enisdenjo/graphql-ws 

look at code example with #[ws](https://github.com/websockets/ws) server usage with [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws) backwards compatibility

PR #1627, opened 2022-09-02
chore: bump lnd docker images to v0.15.1-beta

krtk6160:


bodymindarts:
Integration test failed - we should consider rolling back production until we figure this out!

nicolasburtey:
maybe you can look into this @openoms you debug a similar issue recently. it seems lnd 15.1 has new default that makes channel opening different in size than with lnd 15.0

openoms:
LND v0.15.1 switched to use Taproot addresses by default which does affect the transaction sizes. Will need to work out the new numbers. Looking into it.

bodymindarts (approved):

Issue #1638, opened 2022-09-06
TransactionList cursors can yield duplicate transactions


bodymindarts:
because the ordering is newest to oldest a cursor like `echo "arrayconnection:1" | base64` could very well return the same transactions multiple times if new transactions have occurred which would invalidate the state of the last cursor.

We must either reverse the default order of transactions or change the way we represent cursor.

How is this currently handled in the mobile app @daviroo ?

daviroo:
I didn't write this code but this is my current understanding.  We keep a ref to the cursor we see at the end of the returned edges.  Then on the next call we compare the cursor we see at the last edge to the ref we stored previously, if they differ we know it's a new page of data and add it to an array to be rendered.

Here is an example: https://github.com/GaloyMoney/galoy-mobile/blob/6911c5ede595019b64ae0b579f412a939bc0f191/app/screens/transaction-screen/transaction-screen.tsx#L96

nicolasburtey:
> We must either reverse the default order 

This would not fit the use case from mobile/web wallet where we show the most recent tx

bodymindarts:
Let me be more clear the following pseudo code:
```
call GQL to get list
-> get cursor from response
-> call GQL again passing cursor
```
can yield the same transactions multiple times. Perhaps for the mobile use case this is okay but in general it is unexpected and afaik not how cursors usually work.

or @samerbuna ?

daviroo:
> Let me be more clear the following pseudo code:
> 
> ```
> 
> call GQL to get list
> 
> -> get cursor from response
> 
> -> call GQL again passing cursor
> 
> ```
> 
> can yield the same transactions multiple times. Perhaps for the mobile use case this is okay but in general it is unexpected and afaik not how cursors usually work.
> 

This would result in duplicate transactions being rendered on the screen (a behaviour we haven't seen in production) however it is a behaviour I saw when I tried to implement relayStylePagination caching a while back so maybe the problem you are referring to is why I couldn't get the caching to work.

bodymindarts:
> This would result in duplicate transactions being rendered on the screen (a behaviour we haven't seen in production) however it is a behaviour I saw when I tried to implement relayStylePagination caching a while back so maybe the problem you are referring to is why I couldn't get the caching to work.

It would occur if new transactions took place between the time of the first and second call. In that case the cursor index would no longer point to the same transaction - just the same position in the list.

daviroo:
[...]

PR #1639, opened 2022-09-06
fix: use internal definition of connectionFromArray

bodymindarts:
fixes https://github.com/GaloyMoney/galoy/issues/1638

Looks like @samerbuna [was already aware of the issue](https://github.com/GaloyMoney/galoy/blob/main/src/graphql/connections.ts#L10-L18) but forgot to use the work around everywhere.


daviroo (approved):
Approved assuming the checks pass

PR #1641, opened 2022-09-07
feat: add option to enable usd wallets globally

vindard:
## Description

This PR adds a config option that allows an instance owner to toggle whether USD wallets are also added for each new user created.

### DIFF
--- a/default.yaml
+++ b/default.yaml
@@ -130,9 +130,8 @@ rateLimits:
 
 accounts:
   initialStatus: active
-  wallets:
-    enabledCurrencies:
-      - "BTC"
+  initialWallets:
+    - "BTC"
 
 accountLimits:
   withdrawal:
--- a/default.yaml
+++ b/default.yaml
@@ -133,7 +133,6 @@ accounts:
   wallets:
     enabledCurrencies:
       - "BTC"
-    defaultCurrency: "BTC"
 
 accountLimits:
   withdrawal:
--- a/default.yaml
+++ b/default.yaml
@@ -130,6 +130,7 @@ rateLimits:
 
 accounts:
   initialStatus: active
+  hasUsdWallet: false
 
 accountLimits:
   withdrawal:
--- a/default.yaml
+++ b/default.yaml
@@ -130,7 +130,10 @@ rateLimits:
 
 accounts:
   initialStatus: active
-  hasUsdWallet: false
+  wallets:
+    enabledCurrencies:
+      - "BTC"
+    defaultCurrency: "BTC"
 
 accountLimits:
   withdrawal:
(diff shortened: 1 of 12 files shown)


nicolasburtey (commented):
One suggestion on how to "architecture" this

default.yaml
```diff
@@ -130,6 +130,7 @@ rateLimits:
 
 accounts:
   initialStatus: active
+  hasUsdWallet: false
```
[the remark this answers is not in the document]
vindard: Oh yea I like this. I've already had folks asking if we could run our stack with USD-only.

src/app/users/create-user.ts
```diff
@@ -10,21 +9,45 @@ import {
...
+    enabledWallets[currency] = wallet
+  }
 
-  account.defaultWalletId = wallet.id
+  // Set default wallet explicitly, or implicitly as 1st element in
+  // walletsEnabledConfig array.
+  const defaultWalletId =
+    enabledWallets[defaultWalletConfig]?.id || enabledWallets[walletsEnabledConfig[0]]?.id
+
+  if (defaultWalletId === undefined) {
```
nicolasburtey, line 47: I think that is an acceptable use case to not have a defaultWalletId ?

like for API/business users (when we eventually get there), you create your account, but you don't necessarily have to have a default wallet at the start.

that said, maybe it's possible our backend doesn't handle yet this use case (?). In that case I would recommand still making it optional to not have a default wallet from the config view point, and we can dive more into the backend if there are some edges cases to fix when the need arise

src/app/users/login.ts
```diff
@@ -60,7 +61,10 @@ export const login = async ({
   if (user instanceof CouldNotFindUserFromPhoneError) {
     subLogger.info({ phone }, "new user signup")
     const userRaw: NewUserInfo = { phone }
-    user = await createUserForPhoneSchema(userRaw)
+    user = await createUserForPhoneSchema({
+      newUserInfo: userRaw,
+      config: getAccountsConfig(),
```
nicolasburtey, line 66: I would rename the function more into something like `getDefaultAccountsConfig`.

the idea is that account will be able to evolve over time. ie: as a enterprise user, I may want to add new wallets to my account.

src/config/schema.ts
```diff
@@ -219,8 +220,26 @@ export const configSchema = {
       type: "object",
       properties: {
         initialStatus: { type: "string", enum: Object.values(AccountStatus) },
+        wallets: {
```
nicolasburtey, line 223: suggestion: defaultWallets or initialWallets

nicolasburtey (commented):
some suggestions on the wording, and also the behavior when the initial wallet set is empty, but otherwise looks good.

vindard:
> some suggestions on the wording, and also the behavior when the initial wallet set is empty, but otherwise looks good. 

Made some changes based on suggestions.

nicolasburtey (approved):

PR #1650, opened 2022-09-10
chore: moving User To Account

nicolasburtey:


src/app/users/create-user.ts
```diff
@@ -42,7 +42,6 @@ export const createUser = async ({
   const userRaw: NewUserInfo = {
     phone: phoneNumberValid,
     phoneMetadata,
-    role: getTestAccounts().find(({ phone }) => phone === phoneNumberValid)?.role,
```
bodymindarts, line 45: I cannot find an equivalent of this line moved into code where accounts are being created. Without this creating a new environment requires manually editing db records before its in a workable state.

nicolasburtey: code was moved here: https://github.com/GaloyMoney/galoy/pull/1650/files#diff-99cd32ee09e06e1567ef911c157dcd78f2b91d64e1d1dd537ebce47bac79bfb2R173

I thought this was used for test only.

Let me think about it to see if I can find of a better solution

bodymindarts (approved):

PR #1653, opened 2022-09-12
feat!: add oathkeeper to auth requests

nicolasburtey:
Note: familiarity with [Oathkeeper](https://www.ory.sh/docs/oathkeeper) and specially the mutators and authenticator will be useful to understand the PR.

Here is the goal of the PR: add Oathkeeper as a mandatory path to the login flow. 

Oathkeeper will be configured in production with keys settings from the following file: dev/ory/oathkeeper.yml

```
authenticators:
  bearer_token:
    enabled: true
    config:
      check_session_url: http://e2e-tests:4002/auth/validatejwt

  anonymous:
    enabled: true
    config:
      subject: anon
```

the request will be authenticated in 2 different ways: 

either it will contain a "Authorization" Header, and will go to the bearer_token flow, or it doesn't and will be consider an anonymous flow.

for the bearer_token flow, oathkeeper will reach out to the galoy backend under /auth/validatejwt, and validate the current symmetric JWT, and return the userId if this is valid. if the JWT is not valid, the request will fail.

next step: mutators.

```
mutators:
  id_token:
    enabled: true
    config: 
      jwks_url: file:///home/ory/jwks.json
      issuer_url: "galoy.io"
```

here the requests will go through the id_token flow.

Oathkeeper will add an __asymmetric__ short lived JWT keys for __every requests__ (whether they comes from the bearer_token flow or anonymous flow) which will be verify from the backend using the public key

Note: this is the model of zero trust networking.


src/servers/middlewares/auth-router.ts
```diff
@@ -54,4 +56,36 @@ authRouter.post("/browser", async (req, res) => {
...
+      const rawToken = authz.slice(7) as string
+
+      tokenPayload = jwt.verify(rawToken, JWT_SECRET, {
+        algorithms: jwtAlgorithms,
+      })
+    } catch (err) {
+      console.log("can't decode token")
+    }
+
+    console.log({ tokenPayload }, "tokenPayload")
```
dolcalmi, line 78: please remove

src/servers/middlewares/auth-router.ts
```diff
@@ -54,4 +56,36 @@ authRouter.post("/browser", async (req, res) => {
...
+  const authz = headers.authorization || headers.Authorization
+  if (authz) {
+    try {
+      const rawToken = authz.slice(7) as string
+
+      tokenPayload = jwt.verify(rawToken, JWT_SECRET, {
+        algorithms: jwtAlgorithms,
+      })
+    } catch (err) {
+      console.log("can't decode token")
```
dolcalmi, line 75: please remove or return a proper error

src/servers/middlewares/auth-router.ts
```diff
@@ -54,4 +56,36 @@ authRouter.post("/browser", async (req, res) => {
...
+
+    console.log({ tokenPayload }, "tokenPayload")
+
+    if (typeof tokenPayload === "string") {
+      throw new Error("tokenPayload should be an object")
+    }
+  }
+  if (tokenPayload) {
+    res.json({ subject: "", extra: { uid: tokenPayload.uid } })
+  } else {
```
dolcalmi, line 86: else is not necessary

src/servers/middlewares/auth-router.ts
```diff
@@ -1,12 +1,14 @@
...
 
 import { Users } from "@app"
-import { baseLogger } from "@services/logger"
-import { isDev, getKratosConfig } from "@config"
+import { getKratosConfig, isDev, JWT_SECRET } from "@config"
 import { parseIps } from "@domain/users-ips"
 import { mapError } from "@graphql/error-map"
+import { baseLogger } from "@services/logger"
+
+import * as jwt from "jsonwebtoken"
```
dolcalmi, line 11: please move to the beginning of the file

src/servers/graphql-server.ts
```diff
@@ -82,21 +96,20 @@ const sessionContext = ({
...
         if (loggedInUser instanceof Error)
           throw new ApolloError("Invalid user authentication", "INVALID_AUTHENTICATION", {
             reason: loggedInUser,
           })
         domainUser = loggedInUser
 
-        const loggedInDomainAccount = await Accounts.getAccount(
-          domainUser.defaultAccountId,
-        )
+        const loggedInDomainAccount = await Accounts.getAccount(uid)
```
dolcalmi, line 112: please keep previous code, we have not migrated accounts/users so this works now but will create a problem in the future

src/servers/graphql-server.ts
```diff
@@ -224,9 +236,11 @@ export const startApolloServer = async ({
     }),
   )
 
+  const secret = jwksRsa.expressJwtSecret(getJwksArgs()) as GetVerificationKey // https://github.com/auth0/express-jwt/issues/288#issuecomment-1122524366
```
dolcalmi, line 239: will this invalidate current jwt tokens?

src/config/yaml.ts
```diff
@@ -301,3 +301,14 @@ export const getRewardsConfig = () => {
...
   initialStatus: config.accounts.initialStatus as AccountStatus,
 })
+
+export const getJwksArgs = (config = yamlConfig) => {
+  const urlJkws = config.oathkeeperConfig.urlJkws
+
+  return {
+    cache: true,
+    rateLimit: true,
+    jwksRequestsPerMinute: 5,
```
dolcalmi, line 311: should this be part of oathkeeperConfig ?

dev/ory/oathkeeper.yml
```diff
@@ -0,0 +1,63 @@
...
+      enabled: true
+      config:
+        verbose: true
+
+serve:
+  proxy:
+    cors:
+      enabled: true
+      # allowed_origins:
+      #   - http://localhost:5050
```
dolcalmi, line 47: please remove

Makefile
```diff
@@ -84,7 +87,6 @@ reset-integration: reset-deps integration
 reset-e2e: reset-deps e2e
 
 integration-in-ci:
-	docker compose -f docker-compose.yml up integration-deps -d && \
```
dolcalmi, line 87: are you sure about this? probably will break in concourse

Makefile
```diff
@@ -61,12 +61,15 @@ watch-unit:
...
+oathkeeper:
+	oathkeeper serve -c dev/ory/oathkeeper.yml --sqa-opt-out
+
 e2e:
+	# need oathkeeper running in another terminal
 	yarn build && \
 	yarn test:e2e
 
 e2e-in-ci:
-	docker compose -f docker-compose.yml up integration-deps -d && \
```
dolcalmi, line 69: are you sure about this? probably will break in concourse

Makefile
```diff
@@ -61,12 +61,15 @@ watch-unit:
 watch-compile:
 	$(BIN_DIR)/tsc --watch  --noEmit --skipLibCheck
 
+oathkeeper:
+	oathkeeper serve -c dev/ory/oathkeeper.yml --sqa-opt-out
+
 e2e:
+	# need oathkeeper running in another terminal
```
dolcalmi, line 68: if it is need it why not execute the cmd here?

dolcalmi (changes requested):

Makefile
nicolasburtey: it's a server so it doesn't return. it has to stay on until the test complete.

it would be possible to do a `authkeeper & yarn build && yarn test:e2e` but if authkeeper error you would not see it.

any suggestion how to do better?

Makefile
nicolasburtey: we don't need this anymore I think because we now have hardcoded variable environment. we no longer need to run the script to populate macaroon and tls.

as you still think that can break concourse (?) I will make a separate PR out of it

nicolasburtey: relevant commit: https://github.com/GaloyMoney/galoy/commit/040eaafbd80a9a392ad9a34eb66b9d22712c9ccb

src/servers/middlewares/auth-router.ts
nicolasburtey: any clue why this is not enforced by our linter?

src/servers/graphql-server.ts
nicolasburtey: not in the short term, but in the long term yes (the idea is to no longer use long term symmetric JWT)

let me update the PR to give more context on the approach to update the auth system

Makefile
nicolasburtey: https://github.com/GaloyMoney/galoy/pull/1668

src/config/yaml.ts
nicolasburtey: do you mean exposing this to the yaml?

src/servers/graphql-server.ts
nicolasburtey: I mixed uid/aid above. that doesn't have an impact for now. UserId/User Collection (as in: in mongodb) will be removed shortly.

nicolasburtey: let me know if that respond to your concern: 3207ed17bee981bf20b88d9651565707b2033637

nicolasburtey:
this PR needs more work on the websocket side, putting it back as draft

src/config/yaml.ts
dolcalmi: yes

nicolasburtey:
ready for review again @dolcalmi

src/services/oathkeeper/index.ts
```diff
@@ -0,0 +1,46 @@
...
+    const res = await axios({
+      url: requestUrl,
+      method: "POST",
+      headers,
+    })
+    const jwt = res.headers.authorization
+    return jwt.slice(7) as JwtToken
+  } catch (err) {
+    if (err.response?.status === 401) {
+      return new OathkeeperUnauthorizedServiceError(err)
```
dolcalmi, line 37: err in general is an object so it will try to serialize.. maybe is better to use `new OathkeeperUnauthorizedServiceError(err.message || err)`

src/services/oathkeeper/index.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    if (err.response?.status === 403) {
+      return new OathkeeperForbiddenServiceError(err)
```
dolcalmi, line 41: same here

src/services/oathkeeper/index.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    return new UnknownOathkeeperServiceError(err)
```
dolcalmi, line 44: and here

src/servers/graphql-server.ts
```diff
@@ -69,43 +78,43 @@ const sessionContext = ({
...
         domainAccount = loggedInDomainAccount
-      }
 
-      addAttributesToCurrentSpan({ [ACCOUNT_USERNAME]: domainAccount?.username })
+        addAttributesToCurrentSpan({ [ACCOUNT_USERNAME]: domainAccount?.username })
+      }
 
       return {
         logger,
-        uid: userId,
```
dolcalmi, line 108: I cant see modifications in queries and mutations. We dont have TS in graphql layer so you need to search `uid` and update all queries and mutations that use `uid`.

dolcalmi (dismissed):

src/servers/graphql-server.ts
nicolasburtey: there are only 2 and they are in the admin api. I think they are passed explicitly as an argument as opposed to be deduced from the context. can you double check?

dolcalmi: yep, you are right. they are passed as parameters.

src/services/oathkeeper/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  OathkeeperError,
+  OathkeeperForbiddenServiceError,
+  UnknownOathkeeperServiceError,
+} from "@domain/oathkeeper/errors"
+import axios from "axios"
+
+// import oathkeeper from "@ory/oathkeeper-client"
+
+// note: decisions is the admin API
+const decisionsPrefix = "http://oathkeeper:4456/decisions/"
```
nicolasburtey, line 12: note for me and @bodymindarts 

this will need to be set to the correct domain name

dev/ory/jwks.json
```diff
@@ -0,0 +1,18 @@
+{
+  "keys": [
```
nicolasburtey, line 2: note @bodymindarts a private key should be dynamically generated here. 

this can be done with `  oathkeeper credentials generate --alg RS256`

bodymindarts: Will have to research in the chart how to pre-provision it and generate in terraform

bodymindarts: DO we want the id_token https://github.com/ory/k8s/blob/e3c6951bdb22f7216b9f55fea751ca58df86a6c3/docs/helm/oathkeeper.md#json-web-key-set-for-mutator-id_token or the authenticator jwks https://github.com/ory/k8s/blob/e3c6951bdb22f7216b9f55fea751ca58df86a6c3/docs/helm/oathkeeper.md#json-web-key-set-for-authenticator?
I think the id_token right.

nicolasburtey: id_token for the mutator

nicolasburtey: https://github.com/GaloyMoney/galoy/pull/1653/files#diff-343ce5007f35549b95e3d0db237fce76fbce3130d06916a33423288f98fdd3b1R27

src/services/oathkeeper/index.ts
```diff
@@ -0,0 +1,49 @@
...
+import {
+  OathkeeperUnauthorizedServiceError,
+  OathkeeperError,
+  OathkeeperForbiddenServiceError,
+  UnknownOathkeeperServiceError,
+  OathkeeperMissingAuthorizationHeaderError,
+} from "@domain/oathkeeper/errors"
+import axios from "axios"
+
+export const sendOathkeeperRequest = async (
```
bodymindarts, line 11: I thought the call graph is oath keeper -> galoy not vica versa. This introduces a circular dependency...

bodymindarts (approved):

PR #1654, opened 2022-09-12
chore: remove use of uid and (galoy) UserId

nicolasburtey:


bodymindarts (approved):

src/servers/middlewares/wallet-id.ts
```diff
@@ -16,12 +16,12 @@ const validateWalletId = async (
...
 
-  if (!context.domainUser) {
-    return new Error("Invalid wallet")
+  if (!context.domainAccount) {
+    return new Error("Invalid Account")
   }
 
   const hasPermissions = await Accounts.hasPermissions(
-    context.domainUser.id,
+    context.domainAccount.id,
```
bodymindarts, line 24: since this is a top level use case we could also pass the whole `domainAccount` instead of `id`. I know out of scope for this PR - just thinking.

PR #1662, opened 2022-09-15
chore: add temporary 'UncaughtTypescriptError' for debugging

vindard:
## Description

Proposal for figuring out an open issue.

There's a `TypeError` coming in from the `txs.filter` call that seems to only happen if `txs` is an instance of `Error` or a non-array object. We'd get a different error message if `txs` was `undefined` or `[]`. According to the type checker this should not be possible though, and I can't seem to replicate locally. This PR is to add visibility to what could possibly be happening in prod.

Once we figure this out, we should be able to cleanly revert these changes.


vindard:
Closing since we were able to eventually debug this locally (#1686) and fix (#1693)

PR #1687, opened 2022-09-21
perf: avoid to add contact for funder intraledger transactions

dolcalmi:
Due to the size of contacts in funder's account the performance of the contacts update is really slow.

This PR does an early return on `intraledgerPaymentSendWalletId` to avoid adding contacts to funder's account

### DIFF
--- a/src/app/payments/send-intraledger.ts
+++ b/src/app/payments/send-intraledger.ts
@@ -26,6 +26,7 @@ import {
 import { LockService } from "@services/lock"
 import { LedgerService } from "@services/ledger"
 import * as LedgerFacade from "@services/ledger/facade"
+import { getFunderWalletId } from "@services/ledger/caching"
 import { NotificationsService } from "@services/notifications"
 
 import { ResourceExpiredLockServiceError } from "@domain/lock"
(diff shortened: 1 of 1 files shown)


vindard (approved):
Do we also need to remove the existing contacts from funder somehow?

nicolasburtey:
Yup was thinking the same thing.

Maybe that's a better fix

nicolasburtey:
it's actually a performance hit for every non quizz transaction. could we do it a way where:
1/ we clean the funder account
2/ make sure we don't add new record in the quizz script?

I think if we remove the username for the funder this have the achieved effect without incurring the penalty for every non quizz payment. 

it's also the idea that the complexity of the quizz should not be in the "main codepath" but only in the (future) "addon pod" (when we have federation)

nicolasburtey:
maybe the performance hit is negligible because getFunderWalletId is cached.

I think the design consideration are still important

dolcalmi:
> I think if we remove the username for the funder this have the achieved effect without incurring the penalty for every non quizz payment.

the issue is not adding funder as a contact for the user. It is adding the user as contact for the funder

dolcalmi:
> Yup was thinking the same thing.
> 
> Maybe that's a better fix

it is ok to clean the funder's contacts but is not a fix is only delay the issue again

nicolasburtey:
>  It is adding the user as contact for the funder

ok I think maybe a good solution for that would be to have account that doens't have contact attached to it because it doesn't really make sense to do so.

I think admin accounts are falling into this category.

maybe we could have some flag in the account that have "contactEnabled"? and if no, we basically do not needed the attached contact object or we always keep it as default. 

do you think that could be a better solution?

dolcalmi:
closed in favor of https://github.com/GaloyMoney/galoy/pull/1691

PR #1690, opened 2022-09-21
chore: move kratosUserId creation to Account

nicolasburtey:
This PR move kratosUserId creation to Account and move some function calls from User to Account.

The idea is to continue deprecating current User domain as we're moving it to Kratos. This PR is in preparation of the kratos integration.

### DIFF
--- a/src/app/accounts/create-account.ts
+++ b/src/app/accounts/create-account.ts
@@ -1,6 +1,7 @@
 import { ConfigError, getTestAccounts } from "@config"
+import { checkedToKratosUserId } from "@domain/accounts"
 import { WalletCurrency } from "@domain/shared"
-import { checkedToKratosUserId, checkedToPhoneNumber } from "@domain/users"
+import { checkedToPhoneNumber } from "@domain/users"
 import { WalletType } from "@domain/wallets"
 import { baseLogger } from "@services/logger"
 import {
@@ -11,17 +12,14 @@ import {
 import { TwilioClient } from "@services/twilio"
 
 const setupAccount = async ({
-  userId,
+  account,
   config,
   phoneNumberValid,
 }: {
-  userId: UserId
+  account: Account
   config: AccountsConfig
   phoneNumberValid?: PhoneNumber
 }): Promise<Account | ApplicationError> => {
-  const account = await AccountsRepository().findByUserId(userId)
-  if (account instanceof Error) return account
-
   const newWallet = (currency: WalletCurrency) =>
     WalletsRepository().persistNew({
       accountId: account.id,
@@ -60,13 +58,13 @@ const setupAccount = async ({
 }
 
 // no kratos user is been added (currently with PhoneSchema)
-export const createUserForPhoneSchema = async ({
+export const createAccountForPhoneSchema = async ({
   newUserInfo: { phone, phoneMetadata },
   config,
 }: {
   newUserInfo: NewUserInfo
   config: AccountsConfig
-}) => {
+}): Promise<Account | RepositoryError> => {
   const phoneNumberValid = checkedToPhoneNumber(phone)
   if (phoneNumberValid instanceof Error) return phoneNumberValid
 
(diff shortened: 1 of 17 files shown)


src/app/users/login.ts
```diff
@@ -55,25 +59,30 @@ export const login = async ({
   await rewardFailedLoginAttemptPerIpLimits(ip)
   await rewardFailedLoginAttemptPerPhoneLimits(phone)
 
-  const userRepo = UsersRepository()
```
vindard, line 58: Nit, adding a suggestion that could be more intuitive as we separate the `User` and `Account` concerns in persistence layer.

```ts
  const user = await UsersRepository().findByPhone(phone)
  if (user instanceof Error && !(user instanceof CouldNotFindUserFromPhoneError)) {
    return user
  }

  if (user instanceof CouldNotFindUserFromPhoneError) {
    subLogger.info({ phone }, "new user signup")
    const userRaw: NewUserInfo = { phone }
    const account = await createAccountForPhoneSchema({
      newUserInfo: userRaw,
      config: getDefaultAccountsConfig(),
    })

    return createToken({ uid: account.id, network })
  }

  subLogger.info({ phone }, "user login")
  const account = await AccountsRepository().findByUserId(user.id)

  return createToken({ uid: account.id, network })
}

```

src/domain/accounts/index.types.d.ts
```diff
@@ -119,6 +119,10 @@ interface IAccountsRepository {
   listUnlockedAccounts(): Promise<Account[] | RepositoryError>
   findById(accountId: AccountId): Promise<Account | RepositoryError>
   findByUserId(userId: UserId): Promise<Account | RepositoryError>
+
+  findByKratosUserId(kratosUserId: KratosUserId): Promise<Account | RepositoryError>
+  persistNewKratosUser(kratosUserId: KratosUserId): Promise<Account | RepositoryError>
```
vindard, line 124:
```suggestion
  persistNewByKratosUserId(kratosUserId: KratosUserId): Promise<Account | RepositoryError>
```

src/app/accounts/create-account.ts
```diff
@@ -89,30 +87,31 @@ export const createUserForPhoneSchema = async ({
...
-}) => {
+}): Promise<Account | RepositoryError> => {
   const kratosUserIdValid = checkedToKratosUserId(kratosUserId)
   if (kratosUserIdValid instanceof Error) return kratosUserIdValid
 
-  const user = await UsersRepository().persistNewKratosUser({
-    kratosUserId: kratosUserIdValid,
-  })
-  if (user instanceof Error) return user
+  let account = await AccountsRepository().persistNewKratosUser(kratosUserIdValid)
```
vindard, line 110: We're probably approaching the point where we'd need to separate Accounts from Users at the persistence layer as well.

```suggestion
  let account = await AccountsRepository().persistNewByKratosUserId(kratosUserIdValid)
```

src/services/mongoose/accounts.ts
```diff
@@ -137,7 +138,38 @@ export const AccountsRepository = (): IAccountsRepository => {
     }
   }
 
+  const persistNewKratosUser = async (
```
vindard, line 141:
```suggestion
  const persistNewByKratosUserId = async (
```

src/app/accounts/create-account.ts
```diff
@@ -89,30 +87,31 @@ export const createUserForPhoneSchema = async ({
   const user = await UsersRepository().persistNew(userRaw)
   if (user instanceof Error) return user
 
-  const account = await setupAccount({ userId: user.id, config, phoneNumberValid })
+  let account = await AccountsRepository().findByUserId(user.id)
```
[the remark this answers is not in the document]
nicolasburtey: this is the intention. I'll make that in the next PR if that's fine with you to avoid rebasing work.

src/domain/accounts/index.types.d.ts
nicolasburtey: this method will be removed in a future PR and will only be called `persistNew`, where the kratosUserId will be a mandatory parameters (alongside other parameters currently passed for phone authentication)

vindard (approved):
Approving since comments will be addressed in follow-up PRs

PR #1693, opened 2022-09-23
fix: skip writing errors as serialized objects to cache

vindard:
## Description

We recently had an issue (see #1662) where an unexpected object was somehow being read out of redis caching. On diving in further into this, I found that the root cause is because the redis `setCache` method can actually cache an `Error` object by just caching its readable properties, and when this is read back out from cache it comes out as a non-Error type object that looks like:
```ts
{ name: "OnChainServiceError", level: "critical" }
```

The fix here is to include a new `NonError` type as part of the `set` method's args signature, and to fix places where this lint breaks.

### DIFF
--- a/src/domain/cache/index.types.d.ts
+++ b/src/domain/cache/index.types.d.ts
@@ -5,7 +5,7 @@ type CacheKeys =
 
 type LocalCacheSetArgs<T> = {
   key: CacheKeys | string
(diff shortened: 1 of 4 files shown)


nicolasburtey (approved):

bodymindarts:
Awesome!

PR #1694, opened 2022-09-23
chore: move quiz from domain user to account

nicolasburtey:
- for now. this is one of the last field that will not be migrated to kratos.

it should eventually be it's own collection but this can be done when a migration out of the main codebase is done with an apollo federation

### DIFF
--- a/src/domain/accounts/index.types.d.ts
+++ b/src/domain/accounts/index.types.d.ts
@@ -62,6 +62,19 @@ type Account = {
   readonly contacts: AccountContact[]
   readonly isEditor: boolean
   role?: string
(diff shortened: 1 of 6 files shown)


nicolasburtey:
@vindard here is an example of the graphql API that need to be updated. currently Quiz is attached to the me query, which is in turn attached to the User.

But with User moving to kratos and doing only authentication, this query doesn't make much sense. I'm continue to drain the User domain from its property for now just to introduce kratos that will replace the User domain, but there will be quite a bit of cleaning up needed afterwards

bodymindarts (approved):

PR #1695, opened 2022-09-23
chore: expose adminApi of kratos

nicolasburtey:


src/servers/auth-router.ts
```diff
@@ -16,14 +16,15 @@ const graphqlLogger = baseLogger.child({
   module: "graphql",
 })
 
-const { serverURL, corsAllowedOrigins } = getKratosConfig()
-
 const authRouter = express.Router({ caseSensitive: true })
 
+// TODO: why is cors origin policies mapped to kratos config?
```
bodymindarts, line 21: Because its cors for the endpoint that exchanges the cookie from the web wallet to a backend signed jwt. Its literally cors only for the kratos related endpoint called from the web-wallet.

bodymindarts, line 21:
```suggestion
```

bodymindarts (approved):

bodymindarts (approved):

PR #1691, opened 2022-09-22
feat: add contact enabled account setting

dolcalmi:
Add contact enabled as discussed https://github.com/GaloyMoney/galoy/pull/1687#issuecomment-1253842233

Replace PR https://github.com/GaloyMoney/galoy/pull/1687

PR #1696, opened 2022-09-23
feat: add contact enabled setting

dolcalmi:
Split PR according to https://github.com/GaloyMoney/galoy/pull/1691#issuecomment-1254949049

### DIFF
--- a/src/app/accounts/create-account.ts
+++ b/src/app/accounts/create-account.ts
@@ -49,8 +49,8 @@ const setupAccount = async ({
 
   // FIXME: to remove when Casbin is been introduced
   const role = getTestAccounts().find(({ phone }) => phone === phoneNumberValid)?.role
-  account.role = role
-  account.contactEnabled = !account.role || account.role === "user"
+  account.role = role || "user"
+  account.contactEnabled = account.role === "user" || account.role === "editor"
 
   const updatedAccount = await AccountsRepository().update(account)
   if (updatedAccount instanceof Error) return updatedAccount
--- a/src/app/accounts/create-account.ts
+++ b/src/app/accounts/create-account.ts
@@ -50,6 +50,7 @@ const setupAccount = async ({
   // FIXME: to remove when Casbin is been introduced
   const role = getTestAccounts().find(({ phone }) => phone === phoneNumberValid)?.role
   account.role = role
+  account.contactEnabled = !account.role || account.role === "user"
 
   const updatedAccount = await AccountsRepository().update(account)
   if (updatedAccount instanceof Error) return updatedAccount
--- a/src/app/payments/send-intraledger.ts
+++ b/src/app/payments/send-intraledger.ts
@@ -54,7 +54,7 @@ export const intraledgerPaymentSendWalletId = async ({
   const { senderWallet, recipientWallet, recipientAccount } = validatedPaymentInputs
 
   const { id: recipientWalletId, currency: recipientWalletCurrency } = recipientWallet
-  const { id: recipientAccountId, username: recipientUsername } = recipientAccount
+  const { username: recipientUsername } = recipientAccount
 
   const paymentBuilder = LightningPaymentFlowBuilder({
     localNodeIds: [],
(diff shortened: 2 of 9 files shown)


src/app/accounts/create-account.ts
```diff
@@ -50,6 +50,7 @@ const setupAccount = async ({
   // FIXME: to remove when Casbin is been introduced
   const role = getTestAccounts().find(({ phone }) => phone === phoneNumberValid)?.role
   account.role = role
+  account.contactEnabled = !account.role || account.role === "user"
```
nicolasburtey, line 53: I'm not sure this is very explicit on the intent: `!account.role`

src/app/payments/send-intraledger.ts
```diff
@@ -281,28 +279,24 @@ const executePaymentViaIntraledger = async ({
...
 }: {
-  senderAccountId: AccountId
-  senderUsername: Username | undefined
-  recipientAccountId: AccountId
-  recipientUsername: Username | undefined
+  senderAccount: Account
+  recipientAccount: Account
 }): Promise<true | ApplicationError> => {
-  if (recipientUsername) {
+  if (senderAccount.contactEnabled && recipientAccount.username) {
```
nicolasburtey, line 288: I'm wondering it the use is more something like:

```
if (!(senderAccount.contactEnabled && recipientAccount.contactEnabled)) {
   return
}

continue...
```

ie: only add contact if both contact have contactEnabled

in the case of the quiz, I think it doens't make sense to have the funder add contacts, but also doesn't make sense to have the end user have the contact of the funder. 

wdyt?

src/migrations/20220922030802-update-contact-enabled.ts
```diff
@@ -0,0 +1,18 @@
+module.exports = {
+  async up(db) {
+    try {
+      const collection = db.collection("users")
+      const result = await collection.updateMany(
+        { role: { $ne: "user" } },
```
nicolasburtey, line 6: maybe editor could be excluded also.

src/app/accounts/create-account.ts
vindard: Based on [this comment](https://github.com/GaloyMoney/galoy/pull/1696/files#r979055348) should `editor` be included here too?

test/helpers/user.ts
```diff
@@ -191,7 +191,11 @@ export const createUserAndWallet = async (entry: TestEntry) => {
   }
 
   if (entry.role) {
-    await User.findOneAndUpdate({ _id: toObjectId<UserId>(uid) }, { role: entry.role })
+    const contactEnabled = entry.role === "user"
```
vindard, line 194: Same comment here

vindard (commented):
Just one small thing based on Nicolas's comments

test/helpers/user.ts
dolcalmi: fixed https://github.com/GaloyMoney/galoy/pull/1696/commits/ab3bd473794af580a5739e38e2862bf15a03b806

src/app/accounts/create-account.ts
dolcalmi: fixed https://github.com/GaloyMoney/galoy/pull/1696/commits/ab3bd473794af580a5739e38e2862bf15a03b806

src/app/payments/send-intraledger.ts
dolcalmi: makes sense.. updated here https://github.com/GaloyMoney/galoy/pull/1696/commits/b35293a215a997ad35b6583d001bbbd8b5bfa8c5

src/migrations/20220922030802-update-contact-enabled.ts
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/1696/commits/ab3bd473794af580a5739e38e2862bf15a03b806

vindard (commented):
Looks good to me, just one small question

src/migrations/20220922030802-update-contact-enabled.ts
```diff
@@ -0,0 +1,18 @@
+module.exports = {
+  async up(db) {
+    try {
+      const collection = db.collection("users")
+      const result = await collection.updateMany(
+        { role: { $in: ["dealer", "funder", "bankowner"] } },
+        { $set: { contactEnabled: false, contacts: [] } },
```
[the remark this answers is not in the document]
dolcalmi: schema default value is true so there is no need to do the update

vindard (approved):

PR #1697, opened 2022-09-23
fix: reduce ledger error recording for pending invoices

dolcalmi:


vindard (approved):
Makes sense. Right now we `call invoiceUsers -> call external service -> call medici_transactions`, and this change will pull the external servoce call up higher and then group the rest of mongo calls together like `call external service -> call invoiceUsers -> call medici_transactions`

PR #1698, opened 2022-09-23
feat: adding kratos stack + service test

nicolasburtey:
introducing the service for kratos. 
this is a no-op in term of code execution for the public/admin API.


nicolasburtey:
@bodymindarts do you have an input on this one: https://github.com/GaloyMoney/galoy/pull/1698/commits/628e1a1124808134f2aa3f8f420ef0475c552a09

if I set localhost, then it works for local development (ie: `make integration`), but not for ci, because ci needs to access the service through the kratos domain name. 
the reason is `e2e-in-ci` starts the main server in a sub process. this subprocess (of a container) have access to the dns of the docker network, but not localhost.

bodymindarts:
> @bodymindarts do you have an input on this one: [628e1a1](https://github.com/GaloyMoney/galoy/commit/628e1a1124808134f2aa3f8f420ef0475c552a09)
> 
> if I set localhost, then it works for local development (ie: `make integration`), but not for ci, because ci needs to access the service through the kratos domain name. the reason is `e2e-in-ci` starts the main server in a sub process. this subprocess (of a container) have access to the dns of the docker network, but not localhost.

We are using environment variables for situations like this. In Stablesats they are set directly in [docker-compose](https://github.com/GaloyMoney/stablesats-rs/blob/main/docker-compose.yml#L27-L33)

In this repo they are set in https://github.com/GaloyMoney/galoy/blob/main/.env.ci and interpolated via https://github.com/GaloyMoney/galoy/blob/main/Makefile#L107-L109 when the tests startup.
Now that we no longer need dynamic lookup of lnd secrets this could be simplified to be like in Stablesats or similar as (afaik) no more interpolation after startup should be required.

nicolasburtey:
> 

yeah that could be a good idea but currently the yaml file are not overridable from the the environment variable.

probably the yaml and env variable should be merged together to allow for this use case.

right now I guess I'm going to hard code the fact that if an environement variable exist it will take precedent on the yaml for the public/private kratos endpoint

PR #1702, opened 2022-09-23
feat: adding kratos stack + service test

nicolasburtey:
this PR implements some service call for kratos + some tests. this is a no-op for what is run in prod.


src/domain/users/errors.ts
```diff
@@ -0,0 +1,14 @@
+import { DomainError, ErrorLevel } from "@domain/shared"
```
vindard, line 1: This file needs to be added to `src/app/errors.ts` so that the linter catches missing error types in `src/graphql/error-map.ts`

dev/ory/phone_v0.identity.schema.json
```diff
@@ -0,0 +1,30 @@
...
+  "type": "object",
+  "properties": {
+    "traits": {
+      "type": "object",
+      "properties": {
+        "phone": {
+          "type": "string",
+          "format": "string",
+          "title": "phone",
+          "minLength": 3,
```
bodymindarts, line 14: this seems off for a phone number or?

dev/ory/phone_with_password_v0.identity.schema.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$id": "https://schemas.ory.sh/presets/kratos/quickstart/phone-no-password/identity.schema.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "PhoneNoPassword",
```
bodymindarts, line 4: Identical to the file above.

docker-compose.yml
```diff
@@ -3,19 +3,20 @@ services:
...
-    - redis
-    - mongodb
-    - postgres-price-history
-    - bitcoind
-    - lnd1
-    - lnd2
-    - lnd-outside-1
-    - lnd-outside-2
-    - otel-agent
+      - price
```
bodymindarts, line 6: This diff is noise. Why change the indentation?

docker-compose.yml
```diff
@@ -166,6 +167,9 @@ services:
       - ${TMP_ENV_CI:-.env.ci}
     volumes:
       - ./:/repo
+    environment:
+      - KRATOS_PUBLIC_API=http://kratos:4433
```
bodymindarts, line 171: While I like having these settings here (and I believe in theory we could now get rid of `.env.ci` I think for the time being I would add these values to `.env.ci` to have all the environment variables colocated.

docker-compose.yml
```diff
@@ -176,6 +180,9 @@ services:
...
```
bodymindarts, line 184: as with e2e tests

src/config/yaml.ts
```diff
@@ -276,7 +276,21 @@ export const getTestAccounts = (config = yamlConfig): TestAccount[] =>
   }))
 
 export const getCronConfig = (config = yamlConfig): CronConfig => config.cronConfig
-export const getKratosConfig = (config = yamlConfig): KratosConfig => config.kratosConfig
+
+export const getKratosConfig = (config = yamlConfig): KratosConfig => {
+  let { publicApi, adminApi } = config.kratosConfig
+  if (process.env.KRATOS_PUBLIC_API) {
```
bodymindarts, line 282: Why do we need both `yaml` and `env` config options here?

src/services/kratos/index.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  try {
+    result = await kratosPublic.submitSelfServiceRegistrationFlow(flow.data.id, {
+      traits,
+      method,
+      password,
+    })
+  } catch (err) {
+    if (err.message === "Request failed with status code 400") {
+      return new LikelyUserAlreadyExistError(err)
```
bodymindarts, line 85: Why `Likely`?

src/services/kratos/index.ts
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  // TODO active: true
+
+  return session.identity.id as KratosUserId
+}
+
+export const activateUser = async (
+  kratosUserId: KratosUserId,
+): Promise<void | KratosError> => {
```
bodymindarts, line 146: I think we have just been returning `true | XXX` in situations like this.

src/services/kratos/index.ts
```diff
@@ -0,0 +1,194 @@
...
+  Configuration,
+  Identity,
+  Session,
+  SuccessfulSelfServiceLoginWithoutBrowser,
+  SuccessfulSelfServiceRegistrationWithoutBrowser,
+  V0alpha2Api,
+  V0alpha2ApiInterface,
+} from "@ory/client"
+
+import { AxiosResponse } from "node_modules/@ory/client/node_modules/axios/index"
```
bodymindarts, line 21: will this hard coded path also work when embedded in the image?

src/services/kratos/index.ts
```diff
@@ -0,0 +1,194 @@
+import { getKratosConfig, getKratosMasterPhonePassword } from "@config"
+import {
+  AuthenticationKratosError,
+  IncompatibleSchemaUpgradeError,
+  KratosError,
+  LikelyNoUserWithThisPhoneExistError,
+  LikelyUserAlreadyExistError,
+  UnknownKratosError,
+} from "@domain/users/errors"
```
bodymindarts, line 9: I'm thinking these errors should probably be in their own `@domain/authentication` module not under `user`. The definition of the:
```
interface KratosService { .. }
```
If you create one should live there too.

bodymindarts (commented):
General structure looks good. Can't really comment on the boundaries of the service definition without seeing its usage but I don't see any obvious flaws.

src/services/kratos/index.ts
nicolasburtey: I'm using the public API here. this is the only way to get a session token. so the api doesn't return much context on the error by design - because it's public. 

so the error can be that it's exist, but something else also. the only way to know for sure is to look at the logs.

src/services/kratos/index.ts
nicolasburtey: yup. but I think @dolcalmi said there is a better way to do it with resolutions? not sure, but at least this works.

src/config/yaml.ts
nicolasburtey: oh I think I did it before I was overriding value in `.env.ci` or `docker-compose`. I can probably remove it.

src/services/kratos/index.ts
bodymindarts: We discussed exactly this pattern and decided on `true | ` - but we can bring it up again.

nicolasburtey: there are a lot of `<void | Error>` as well

image

PR #1707, opened 2022-09-26
chore: parse & map repository errors

vindard:
## Description

This is to map known connection issues to take the place of this class of `UnknownRepositoryError` errors.


bodymindarts (dismissed):

src/services/shared/repository.ts
```diff
@@ -0,0 +1,11 @@
+import { CantConnectToMongoDbError, UnknownRepositoryError } from "@domain/errors"
```
bodymindarts, line 1: Shouldn't this be scoped under `mongoose`? its not a service in itself.

vindard: Oh yea, the vast majority of our repository implementations are already in there anyway.

There's a few places outside of `mongoose` where we call `import { parseRepositoryError } from "@services/mongoose/utils"`. Should we expose those `utils` at the package level, or is it ok to reach in like this?

src/domain/errors.ts
```diff
@@ -13,6 +13,9 @@ export class DuplicateError extends RepositoryError {}
 
 export class BadInputsForFindError extends RepositoryError {}
 export class CouldNotFindError extends RepositoryError {}
+export class CantConnectToMongoDbError extends RepositoryError {
```
bodymindarts, line 16: shouldn't couple domain to `Mongo` implementation. I also don't like leaving out `'` in names like this. `CanNotConnectToDatabaseError` I guess - but its preference not a big deal.

src/services/payment-flow/index.ts
```diff
@@ -8,6 +8,7 @@ import {
 import { InvalidLightningPaymentFlowStateError, PaymentFlow } from "@domain/payments"
 import { paymentAmountFromNumber, WalletCurrency } from "@domain/shared"
 import { safeBigInt } from "@domain/shared/safe"
+import { parseRepositoryError } from "@services/mongoose/utils"
```
bodymindarts, line 11: this file (ie `src/services/payment-flow`) should be under `mongoose` like all the other repositories.

src/services/redis/routes.ts
```diff
@@ -16,7 +17,7 @@ export const RoutesCache = (): IRoutesCache => {
       await redis.set(key, value, "EX", defaultTimeToExpiryInSeconds)
       return routeToCache
     } catch (err) {
-      return new UnknownRepositoryError(err)
+      return parseRepositoryError(err)
```
bodymindarts, line 20: The `redis` service should not be calling anything under `mongoose/utils` I don't see why the error messages should match anyway. `redis` service needs its own parse function.

src/services/ledger/services/transactions-metadata.ts
```diff
@@ -1,6 +1,6 @@
-import { NoTransactionToUpdateError, UnknownRepositoryError } from "@domain/errors"
+import { NoTransactionToUpdateError } from "@domain/errors"
 import { CouldNotFindTransactionMetadataError } from "@domain/ledger"
-import { fromObjectId, toObjectId } from "@services/mongoose/utils"
+import { fromObjectId, toObjectId, parseRepositoryError } from "@services/mongoose/utils"
```
bodymindarts, line 3: This isn't great but as this is also a repository implemented with `mongoose` its okay as an exception.

src/services/redis/routes.ts
vindard: Fixed

src/services/payment-flow/index.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1707/commits/a10d5f494d8e8f4ba12e260937ce5e9e261a03ca

src/domain/errors.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1707/commits/bd06ee7f110f2a85d1afdca9c34c64bda333b761

bodymindarts (approved):

PR #1716, opened 2022-09-28
chore: handle "mixed case" ln decode error

vindard:
## Description

Ln invoices are case insensitive so we can ideally lower-case all invoices before decoding them. This PR changes how we take in payment requests so that they are validated/lower-cased/decoded in a single step from a `string` instead of from `EncodedPaymentRequest` type cast.

Traces: https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/ihXEPwA4KsQ

### DIFF
--- a/src/app/lightning/payment-status-checker.ts
+++ b/src/app/lightning/payment-status-checker.ts
@@ -2,8 +2,8 @@ import { RepositoryError } from "@domain/errors"
 import { decodeInvoice } from "@domain/bitcoin/lightning"
 import { LedgerService } from "@services/ledger"
 
-export const PaymentStatusChecker = async (paymentRequest: string) => {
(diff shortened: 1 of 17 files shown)


nicolasburtey:
> This is just to remove all the casting right?

QR code encode them in uppercase because it's a more efficient encoding scheme for QR code, but lnd only accept lowercase invoice. bc1 encoding format assumes lowercase

vindard:
> This is just to remove all the casting right?

Partially. We could have a `toEncodedPaymentRequest` method separately, but I've done this implicitly inside `decodeInvoice` now since we don't ever actually need the string outside of that method's return.

It more accurately represents that we haven't validated the input coming in to our use-cases.

PR #1719, opened 2022-09-29
fix: update MONGODB_ADVERTISED_HOSTNAME for ci env

dolcalmi:


daviroo (approved):
@dolcalmi it works for me on m1 mac but I wondered if removing the overrides from the `test-migrate` run would have other effects?  I'll trust your judgement if you say it doesn't.

PR #1721, opened 2022-09-29
feat(api): extend gql error type with 'code' selection

vindard:
## Description

This is to map the `code` property across to graphql error objects returned in the API. This has been an open `TODO` for a while, and the prompt now is that we need this property here to be able to pick it up in tracing to add to traces (see #1722).

### Details 

The result of this will be the following additions to our API:
- remove existing (but unused) PaymentError and InputError (`IError` implementations)
- new implementation of the `IError` interface called `GraphQLApplicationError`
- add `code` selection to `IError` interface

This is all to expose the `code` property to be picked up by `gqlResponseHook` to be added in tracing as `graphql.error.code`/`graphql.data.error.code` attributes. We will then change the "unknown message" honeycomb alert (https://github.com/GaloyMoney/galoy-deployments/pull/2246) to be based on error codes instead of error messages (see #1722 for new error code to be used).


vindard:
Reverted the API change here (https://github.com/GaloyMoney/galoy/pull/1721/commits/fb0bc34075a273deeccf06a82edd46a939397d10) to see if the `code` property will make it to `gqlResponseHook`. It didn't, so we'd need to return the new `code` property as a selection on `Error` somehow in the API.

src/graphql/main/schema.graphql
```diff
@@ -33,7 +33,45 @@ input AccountUpdateDefaultWalletIdInput {
 
 type AccountUpdateDefaultWalletIdPayload {
   account: ConsumerAccount
-  errors: [Error!]!
+  errors: [AppError!]!
```
vindard, line 36: Is this type name change a breaking change?

vindard:
I tried out changing the `ApiError` to a fragment of types out of curiosity. Interestingly it works, but not viable because it breaks the API: https://github.com/GaloyMoney/galoy/commit/e2d54e03757983e761ee41ceba8c0989e602139b

src/graphql/main/schema.graphql
vindard: We resolved this by removing the unused error enums so that `code` could be set as a `GT.String` in the interface

vindard:
Final solution was to remove all unused `IError` implementations, add a new `GraphqlQLApplicationError` implementation with `isTypeOf` set to `() => true` so that all interface calls get passed to this object, and then to simply add the `code` field to `IError` as a `GT.String`.

src/graphql/main/schema.graphql
```diff
@@ -230,6 +231,12 @@ type Globals {
   nodesIds: [String!]!
 }
 
+type GraphQLApplicationError implements Error {
```
bodymindarts, line 234: It all LGTM but I'm not sure about the semantics of `GraphQLApplicationError` replacing `InputError` and `PaymentError`. What exactly are we signifying? `ApplicationError` is very broad. Is removing the other 2 errors breaking?

vindard: Samer and I went through this to confirm. Both errors were included as interfaces in the schema but were not actually being used anywhere. We actually pass the `IError` interface into places that need to return an error in the schema, and this was being auto-resolved to `InputError` only via `isTypeOf: () => true && ...` in the `InputError` object.

`InputError` isn't ideal for this resolution though since it has a very limited enum for the `code` field that isn't relevant. This change is to have that interface to resolve a new `GraphQLApplicationError` that has broader `code` type. I'm not sure about the name either, but the intention is that this error is for all `CustomApolloError` codes that can come through.

I had also originally considered doing a broader enum on those custom apollo errors for the `code` field ([see here](https://github.com/GaloyMoney/galoy/commit/53ea1f6205c5b6cae1bd9e359e8708087811b24c#diff-5309dec6f007762be2e567aaf4b2f973cf0a7764ca0833c4763dbdf96d524931R45)), but that broke when things like input validation return errors since those don't have codes as yet, so I reverted to simply using `GT.String` for `code`.

vindard: p.s. should we consider adding something like this to our actions to detect breaking api changes?
https://www.the-guild.dev/graphql/inspector/docs/products/action

bodymindarts (approved):

PR #1727, opened 2022-09-30
fix: swap alerting with swap-checker span

ntheile:


bodymindarts:
I think the error is five it just shouldn’t be critical perhaps

nicolasburtey:
> I think the error is fine it just shouldn’t be critical perhaps

agree

docs/swap.md
```diff
@@ -166,6 +166,9 @@ In Honeycomb you can query the following:
 swap.error exists
 swap.submitted exists
 swap.success exists
+swap.checker.message exists
```
nicolasburtey, line 169: I think you should query in a honeycomb board instead of having some documentation here

src/domain/swap/errors.ts
```diff
@@ -1,6 +1,9 @@
 import { DomainError, ErrorLevel } from "@domain/shared"
 
 export class SwapError extends DomainError {}
+export class SwapErrorNonCritical extends DomainError {
+  level = ErrorLevel.Info
```
nicolasburtey, line 5: this is the default afaik. no need to set it explicitly

src/domain/swap/errors.ts
```diff
@@ -1,6 +1,9 @@
...
+}
 export class SwapTriggerError extends SwapError {}
 export class SwapServiceError extends SwapError {}
```
nicolasburtey, line 8: swap service error is already non critical

src/domain/swap/errors.ts
nicolasburtey: skimming through it, I think your issue is that you're throwing on `cron.ts` if an error is being returned. this is what is also trigger pager duty

nicolasburtey (changes requested):
the use of `NonCritical` will not changed anything on the alerting

src/app/swap/swap-out.ts
```diff
@@ -48,16 +48,17 @@ export const swapOut = async (): Promise<SwapOutResult | SwapServiceError> => {
...
 
-  if (swapOutAmount.amount === 0n)
-    return {
-      noOp: true,
-      htlcAddress: "" as OnChainAddress,
-      serverMessage: "",
-      swapId: "" as SwapId,
-      swapIdBytes: "",
-    }
+  if (swapOutAmount instanceof SwapErrorNonCritical) return swapNoOp
```
ntheile, line 59: @nicolasburtey  this is where I handle a non-critical error and instead of returning an instance of an Error (which throws from the caller...i.e cron) i instead return a NoOp SwapOutResult

src/domain/swap/errors.ts
ntheile: yep, i was following the other cron jobs as a template. should I not throw?

src/servers/trigger.ts
```diff
@@ -404,6 +407,15 @@ const listenerOffchain = ({ lnd, pubkey }: { lnd: AuthenticatedLnd; pubkey: Pubk
   })
 }
 
+const startSwapMonitor = async (swapService: ISwapService) => {
+  const isSwapServerUp = await swapService.healthCheck()
+  if (isSwapServerUp instanceof Error) return new SwapErrorNoActiveLoopdNode()
```
vindard, line 412: Why not return `isSwapServerUp`?

src/app/swap/swap-out.ts
vindard: In places like this in other use-cases we would also simply return something like `false`. Having errors carry context like this could be tricky.

ntheile: I made it a NoOp swap result based on this feedback from Justin https://github.com/GaloyMoney/galoy/pull/1396#discussion_r973624924

PR #1731, opened 2022-10-03
fix: inject dealer fns depending on senderWallet.currency

bodymindarts:


vindard:
@dolcalmi and I were able to replicate the bad conversion locally, apply this fix, and confirm that in the specific case where it's a BTC sender and USD recipient that this works (1 sat converts to 0 cents after fix, 1 cent before fix)

vindard (approved):

vindard (approved):

PR #1732, opened 2022-10-03
chore: remove updatePendingPaymentsByWalletId from getBalance

nicolasburtey:
this is handled both by trigger and in cron as a fallback.

eventually, this could be in the hourly cron (instead of daily)

cc @ntheile


bodymindarts (approved):

nicolasburtey:
note: to not merge until we have related event in trigger

nicolasburtey:
@dolcalmi it's a good task I think if you want to pick it up (adding the event on trigger so we can merge this one)

PR #1682, opened 2022-09-20
feat: add self trade tx type, volume & limits checks

vindard:
## Description

This PR is to separate self-trade transaction types from other intraledger transactions. The goal here is to have separate limits for self-trades vs. other intraledger transfers.

It also separates out a new "send-lightning-limits" test file and adds a comprehensive set of tests around limits (now possible because of #1518).

I also did some cleanup and added new methods to our integrations tests to generate wallets/accounts on the fly to make each limits test a bit more self-contained. Before now, we would use wallets that were previously instantiated and used in other tests which created dependencies across different testing functions.

PR #1750, opened 2022-10-05
chore: re-intrduce "feat: add self trade tx type, volume & limits checks (#1682)"

vindard:
## Description

Re-introduces changes from #1682, reverted in #1745


vindard:
Tested limits in staging here, all looks well: https://ui.honeycomb.io/galoy/datasets/galoy-staging/result/aveVWWMTgvM

PR #1084, opened 2022-02-14
Chore  make memo also compatible with USD wallets

nicolasburtey:

PR #1767, opened 2022-10-06
fix: add usd wallets memo filter

dolcalmi:
updated code from https://github.com/GaloyMoney/galoy/pull/1084

### DIFF
--- a/galoy.yaml
+++ b/galoy.yaml
@@ -129,6 +129,7 @@ accountLimits:
       "2": 20000000
 spamLimits:
   memoSharingSatsThreshold: 1000
+  memoSharingCentsThreshold: 50
 twoFALimits:
   threshold: 10000
 ipRecording:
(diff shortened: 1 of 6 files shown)


galoy.yaml
```diff
@@ -129,6 +129,7 @@ accountLimits:
       "2": 20000000
 spamLimits:
   memoSharingSatsThreshold: 1000
+  memoSharingCentsThreshold: 50
```
vindard, line 132: This seems high, any particular reason for this?

vindard (commented):
This looks ok for now. I think eventually we get `satsAmount`/`centsAmount` into all transaction types we can pass thru a price ratio from the `LedgerTransaction` and get this back down to one config setting

galoy.yaml
dolcalmi: was the value in the previous PR.. nothing in particular.. what value do you suggest?

vindard (approved):

PR #1776, opened 2022-10-07
chore: instrumenting validationtoken

nicolasburtey:
note: this `/validatetoken` function will only exist during the transition to kratos. once we stopped accepting the `legacyJwt`, this function will disappear. which is why I think the current hacky approach is better suited than adding express opentelemetry collector. the later option would otherwise generated many spans.

### DIFF
--- a/src/servers/middlewares/auth-router.ts
+++ b/src/servers/middlewares/auth-router.ts
@@ -64,7 +64,7 @@ const jwtAlgorithms: jwt.Algorithm[] = ["HS256"]
 authRouter.post(
   "/validatetoken",
   wrapAsyncToRunInSpan({
-    namespace: "daily-balance-notification",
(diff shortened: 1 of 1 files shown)


src/servers/middlewares/auth-router.ts
```diff
@@ -60,35 +61,45 @@ const jwtAlgorithms: jwt.Algorithm[] = ["HS256"]
...
-  }
-
-  // the sub (subject) sent to oathkeeper as a response is the uid from the original token
-  // which is the AccountId
-  res.json({ sub: tokenPayload.uid })
-})
+authRouter.post(
+  "/validatetoken",
+  wrapAsyncToRunInSpan({
+    namespace: "daily-balance-notification",
```
bodymindarts, line 67: Copy paste error

PR #1779, opened 2022-10-08
feat: add better swap tracing

ntheile:
This PR adds better Honeycomb tracing for swaps. It also adds logic to not start the swapListener if swaps are not enabled in the default.yaml

### DIFF
--- a/docs/swap.md
+++ b/docs/swap.md
@@ -160,13 +160,7 @@ make start-cron
 
 Tracing
 =======
-In Honeycomb you can query the following:
-
-```
-swap.error exists
-swap.submitted exists
-swap.success exists
-```
+In Honeycomb you can query for `swap`
 
 Generate GRPC Types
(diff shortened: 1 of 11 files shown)


docs/swap.md
```diff
@@ -160,13 +160,7 @@ make start-cron
...
 Tracing
 =======
-In Honeycomb you can query the following:
-
-```
-swap.error exists
-swap.submitted exists
-swap.success exists
-```
+In Honeycomb you can query for `swap`
```
nicolasburtey, line 163: do you mean `name: swap`?

ntheile: yeah actually `code.namespace = services.swap`. i think i remember Justin mentioning to use`addAttributesToCurrentSpan` instead of creating a new one, due to cost.

nicolasburtey (approved):

PR #1789, opened 2022-10-10
feat: add kratos interface

nicolasburtey:


src/domain/authentication/index.types.d.ts
```diff
@@ -0,0 +1,53 @@
+type KratosAuthResponse = {
```
bodymindarts, line 1: Up to now we have been careful not to couple to external implementations in `domain`. Is this a `AuthenticatedUserSession` or something like that.

src/domain/authentication/index.types.d.ts
```diff
@@ -0,0 +1,53 @@
...
+type LoginWithPhoneNoPasswordSchemaResponse = KratosAuthResponse
+
+type CreateKratosUserForPhoneNoPasswordSchemaResponse = KratosAuthResponse
+
+type KratosUserId = string & { readonly brand: unique symbol }
+type KratosSessionToken = string & { readonly brand: unique symbol }
+
+type KratosError = import("./errors").KratosError
+
+type IdentityKratos = import("@ory/client").Identity
```
bodymindarts, line 15: Should not couple to external dependency in `domain`. The only file / module that should reference `@ory/client` should be `services/kratos/*`.
If we need a type like:
```
type AuthenticatedUserIdentity {
}
```
we should declare it here explicitly and add a translation / adapter in the `kratos` service implementation. This protects `domain` code and anything that depends on it from breaking changes in `@ory/client` (which has happened since we've been using it). Then there is only 1 place in the entire code base (the adapter) that needs changing on breaking updates.

src/services/kratos/auth-phone-no-password.ts
```diff
@@ -0,0 +1,143 @@
...
+} from "@ory/client"
+
+import { AxiosResponse } from "node_modules/@ory/client/node_modules/axios/index"
+
+import { kratosAdmin, kratosPublic } from "."
+
+// login with phone
+const password = getKratosMasterPhonePassword()
+
+export const AuthWithPhoneNoPassword = (): AuthWithPhoneNoPasswordSchema => {
```
bodymindarts, line 25: should this be `KratosAuthenticationService = (): IAuthenticationService`?

src/services/kratos/index.ts
```diff
@@ -0,0 +1,296 @@
...
+import { AxiosResponse } from "node_modules/@ory/client/node_modules/axios/index"
+
+export * from "./auth-phone-no-password"
+
+const { publicApi, adminApi } = getKratosConfig()
+
+const KratosSdk: (kratosEndpoint: string) => V0alpha2ApiInterface = (kratosEndpoint) =>
+  new V0alpha2Api(new Configuration({ basePath: kratosEndpoint }))
+
+export const kratosPublic = KratosSdk(publicApi)
```
bodymindarts, line 27: this must be module private. If we are leaking the `kratos` client outside of this module we are not implementing the `service` correctly.

src/services/kratos/index.ts
```diff
@@ -0,0 +1,296 @@
...
+export const kratosAdmin = KratosSdk(adminApi)
+
+export const LoginWithPhoneAndPasswordSchema = async ({
```
bodymindarts, line 30: Why are these functions implemented in isolation and not part of a `XxxService` container?

dev/ory/email_and_phone_with_password_v0.identity.schema.json
```diff
@@ -0,0 +1,39 @@
...
+          }
+        },
+        "phone": {
+          "title": "Phone",
+          "type": "string",
+          "format": "string",
+          "ory.sh/kratos": {
+            "credentials": {
+              "password": {
+                "identifier": true
```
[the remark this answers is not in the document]
nicolasburtey: you can log with either field + password.

I did a test, and from my test the same password is valid for both field/traits.

docker-compose.yml
```diff
@@ -272,3 +273,24 @@ services:
...
+    command: '"kratos migrate sql up -y -e && kratos serve -c /home/ory/kratos.yml --sqa-opt-out"'
+    environment:
+      DSN: postgres://dbuser:secret@pg:5432/default?sslmode=disable
+    links:
+      - pg:pg
+    volumes:
+      - type: bind
+        source: dev/ory
+        target: /home/ory
+  pg:
```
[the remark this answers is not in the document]
nicolasburtey: yup, renamed `pg` to `kratos-pg`. not sure renaming the other part is useful?

src/domain/authentication/index.types.d.ts
```diff
@@ -0,0 +1,53 @@
...
+
+type KratosPassword = string & { readonly brand: unique symbol }
+type KratosEmail = string & { readonly brand: unique symbol }
+
+type ValidateKratosTokenResult = {
+  kratosUserId: KratosUserId
+  session: Session
+}
+
+interface AuthWithPhoneNoPasswordSchema {
```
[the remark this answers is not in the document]
nicolasburtey: yes it's an interface, but it's not that generic

login via phone will have different interface than email. most likely only phone will really be in our stack. email will be fully handled by the kratos server.

src/services/kratos/index.ts
nicolasburtey: for tests and learning about kratos. new filename should help clarify this

src/config/process.ts
```diff
@@ -116,3 +116,10 @@ export const getLoopConfig = () => {
     lnd2LoopMacaroon: process.env.LND2_LOOP_MACAROON as Macaroon,
   }
 }
+
+export const getKratosMasterPhonePassword = () => {
```
bodymindarts, line 120: I understand the reason for this - just wondering wether we shouldn't be generating a unique PW per user instead of using the same one everywhere. If this gets leaked what are the repercussions? Will this go away once the symmetric JWT is gone?
We can have self-serviced on boarding from mobile talking directly to kratos (like in web wallet) shouldn't we be doing that?

src/domain/authentication/index.types.d.ts
```diff
@@ -0,0 +1,37 @@
...
+
+type IdentityPassword = string & { readonly brand: unique symbol }
+
+type KratosUserId = string & { readonly brand: unique symbol }
+type SessionToken = string & { readonly brand: unique symbol }
+
+type Identity = {
+  id: KratosUserId
+  schemaId: SchemaId
+  traits: {
```
bodymindarts, line 14: I would flatten the traits here - no need to keep the structure of the `ory/client` type.

src/domain/authentication/index.types.d.ts
```diff
@@ -0,0 +1,37 @@
...
+    phone?: PhoneNumber
```
bodymindarts, line 15: Why is this `optional?`? This is often an indication that we are combining use cases. Under what circumstances is `phone` going to be present and when is it not? Are these distinct code paths that can use 2 different types?

src/domain/authentication/index.types.d.ts
```diff
@@ -0,0 +1,37 @@
...
+type Session = {
+  identity: Identity
+  id: SessionId
+  expiresAt: Date
+}
+
+interface IAuthWithPhoneNoPassword {
+  login(
+    phone: PhoneNumber,
+  ): Promise<LoginWithPhoneNoPasswordSchemaResponse | AuthenticationError>
```
bodymindarts, line 28: The response types are no longer declared in this file.

src/domain/authentication/errors.ts
```diff
@@ -0,0 +1,3 @@
+import { DomainError } from "@domain/shared"
+
+export class AuthenticationError extends DomainError {}
```
bodymindarts, line 3: To follow with the patterns of the other subdomains I think we need an `AuthenticationServiceError`.

src/domain/authentication/index.types.d.ts
```diff
@@ -0,0 +1,37 @@
...
+type AuthenticationError = import("./errors").AuthenticationError
+
+type IdentityPassword = string & { readonly brand: unique symbol }
+
+type KratosUserId = string & { readonly brand: unique symbol }
+type SessionToken = string & { readonly brand: unique symbol }
+
+type Identity = {
+  id: KratosUserId
+  schemaId: SchemaId
```
bodymindarts, line 13: Why is `schemaId` required in the app use cases? Isn't this an internal detail?


Do we have logic that switches on this field? What could a domain based representation look like? Perhaps 2 types `EmailIdentity` and `PasswordIdentity`? Or if they must be unified then perhaps `IdentityType` that is an explicitly declared enum.

src/services/kratos/auth-phone-no-password.ts
```diff
@@ -0,0 +1,143 @@
...
+} from "@ory/client"
+
+import { AxiosResponse } from "node_modules/@ory/client/node_modules/axios/index"
+
+import { kratosAdmin, kratosPublic, toDomainIdentity } from "./private"
+
+// login with phone
+const password = getKratosMasterPhonePassword()
+
+export const AuthWithPhoneNoPassword = (): IAuthWithPhoneNoPassword => {
```
bodymindarts, line 25: Still not sure about this name. Is it perhaps a `PaswordlessAuthenticationService`? Having something like `XxxService` is just to stay consistent with all the rest of our `ExternalService` implementations.

src/services/kratos/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./auth-phone-no-password"
+export * from "./tests-but-not-prod"
```
bodymindarts, line 2: This should not be exported as they are not (yet) part of the public interface (exposed via `index`). I suggest the tests that need them to depend directly on the sub-module `@services/kratos/tests-but-not-prod`.

bodymindarts (commented):
Okay this is certainly a step in the right direction. There are a lot fewer definitions in `domain` and they are more explicit.

Next step is to think about the usage patterns - I think we can improve the domain types in particular exposing `schema` seems like a smell to me.
I could see us having 2 domain types (`EmailIdentity`, `PhoneIdentity`) or perhaps an `IdentityType` enum if they must be unified - but not expose `schema` that is a concept that should remain internal to the `kratos` service IMO. It is after all simply a config detail.

src/config/process.ts
nicolasburtey: > I understand the reason for this - just wondering wether we shouldn't be generating a unique PW per user instead of using the same one everywhere.

there is no reason to. it's more state to manage with no clear benefits

> If this gets leaked what are the repercussions?

it should be kept in the backend. it's like having a private key, or the prior secret for signing the JWT. yes if this get leaked this means a breach happen but in that case an attacker would not just take this password but more likely private keys. 

> We can have self-serviced on boarding from mobile talking directly to kratos (like in web wallet) shouldn't we be doing that?

we can't for phone. it will always (or at least for the time being) be behind some logic in our api. the main reasons:
1/ there is no phone support in kratos
2/ there is no captcha support in kratos (and this one will likely never be resolved if there is no open source catcha solution)

unlike with email, phone __needs__ to be protected behind a captcha, otherwise there will "DDOS text" attacks

src/domain/authentication/index.types.d.ts
nicolasburtey: yup good point

src/services/kratos/private.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const toDomainSession = (session: KratosSession) => {
+  // is throw ok? this should not happen I (nb) believe but the type say it can
+  // this may probably be a type issue in kratos SDK
+  if (!session.expires_at) throw new MissingExpiredAtKratosError()
+
+  return {
+    id: session.id as SessionId,
+    identity: toDomainIdentityPhone(session.identity),
+    expiresAt: new Date(session.expires_at),
```
bodymindarts, line 22: `expiresAt` is no longer declared on the domain `Session` type.

src/services/kratos/private.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const { publicApi, adminApi } = getKratosConfig()
+
+const KratosSdk: (kratosEndpoint: string) => V0alpha2ApiInterface = (kratosEndpoint) =>
+  new V0alpha2Api(new Configuration({ basePath: kratosEndpoint }))
+
+export const kratosPublic = KratosSdk(publicApi)
+export const kratosAdmin = KratosSdk(adminApi)
+
+export const toDomainSession = (session: KratosSession) => {
```
bodymindarts, line 14: Definitely need to declare the return type here IMO - inference could cause issues.

src/services/kratos/private.ts
```diff
@@ -0,0 +1,29 @@
...
+  // is throw ok? this should not happen I (nb) believe but the type say it can
+  // this may probably be a type issue in kratos SDK
+  if (!session.expires_at) throw new MissingExpiredAtKratosError()
```
bodymindarts, line 17: If it's truly something that should never happen throw is okay IMO.
OTOH protecting our code from potentially faulty dependencies is what this layer is about - I'd leave it like this and see if it comes up in prod - then reassess.

bodymindarts (approved):
Okay I think the domain types are much cleaner now - I don't see any point in continuing to think about them without example usage. Perhaps on the next PR there will be some additional information that may inform some changes.

Other than that you need to explicitly type the return of `toDomainSession` and it should be all good IMO.

PR #1798, opened 2022-10-11
feat!: activate kratos

nicolasburtey:


Makefile
```diff
@@ -80,34 +80,53 @@ e2e-in-ci:
...
+	make create-tmp-env-ci && \
+	TMP_ENV_CI=tmp.env.ci docker compose -f docker-compose.yml run --name e2e-tests e2e-tests make start-main-ci || \
+	docker rm `docker ps -q -a -f status=exited`
+
+main-in-ci-cached-no-restart:
+	make create-tmp-env-ci && \
+	TMP_ENV_CI=tmp.env.ci docker compose run --name e2e-tests e2e-tests make start-main-ci || \
+	docker rm `docker ps -q -a -f status=exited`
+
+del-containers:
```
bodymindarts, line 117: don't want to add this as it is destructive to more than just the containers that get created in this repo.

Is the command you're looking for `docker compose rm -sfv`?

dev/ory/oathkeeper.yml
```diff
@@ -13,6 +13,15 @@ authenticators:
     config:
       subject: anon
 
+  # cookie_session:
```
bodymindarts, line 16: Shouldn't keep this in here - if its prep for a next step then at least add a comment to indicate when it will become relevant.

dev/ory/oathkeeper_rules.yaml
```diff
@@ -27,6 +27,36 @@
...
     handler: allow
   mutators:
+    - handler: header
+      config:
+        headers:
+          orgauthorization: '{{ .MatchContext.Header.Get "Authorization" }}'
     - handler: id_token
       config:
         claims: '{"sub": "{{ print .Subject }}"}'
+#  {
```
bodymindarts, line 37: Don't want commented config

src/app/users/login.ts
```diff
@@ -64,33 +65,66 @@ export const login = async ({
...
 
-  if (user instanceof CouldNotFindUserFromPhoneError) {
-    subLogger.info({ phone }, "new user signup")
+  const authService = AuthWithPhonePasswordlessService()
+
+  let kratosResult = await authService.login(phone)
+  if (kratosResult instanceof LikelyNoUserWithThisPhoneExistError) {
+    // user has not migrated to kratos or it's a new user
+
+    kratosResult = await authService.createWithSession(phone)
```
bodymindarts, line 76: reading it here in usage I think this should be `createIdentityWithSession` or something like that. Its not obvious what is being `created`.

src/app/users/login.ts
```diff
@@ -64,33 +65,66 @@ export const login = async ({
...
+      subLogger.info({ phone }, "new user signup")
+
+      // TODO: look at where is phone metadata stored
+      const accountRaw: NewAccountInfo = { phone, kratosUserId }
+      const account_ = await createAccountForPhoneSchema({
+        newAccountInfo: accountRaw,
+        config: getDefaultAccountsConfig(),
+      })
+
+      if (account_ instanceof Error) return account_
```
bodymindarts, line 95: what is the suffix `account_` for? Typo? I don't think we have used it anywhere up to now. I have only seen `_account` for variables that aren't going to be used (but it doesn't apply here as it is being returned).

src/app/users/login.ts
```diff
@@ -64,33 +65,66 @@ export const login = async ({
...
+    const kratosUserId = kratosResult.kratosUserId
+
+    const user = await UsersRepository().findByPhone(phone)
+    if (user instanceof CouldNotFindUserFromPhoneError) {
+      // brand new user
+      subLogger.info({ phone }, "new user signup")
+
+      // TODO: look at where is phone metadata stored
+      const accountRaw: NewAccountInfo = { phone, kratosUserId }
+      const account_ = await createAccountForPhoneSchema({
```
bodymindarts, line 90: I think we removed the word `schema` from the domain types but its still here as part of the app use case. Its not terrible but can we find something more descriptive?
`createAccountWithPhoneIdentifier` perhaps?

src/app/users/login.ts
```diff
@@ -64,33 +65,66 @@ export const login = async ({
...
+        newAccountInfo: accountRaw,
+        config: getDefaultAccountsConfig(),
+      })
+
+      if (account_ instanceof Error) return account_
+    } else {
+      if (user instanceof Error) return user
+      else {
```
bodymindarts, line 98: this else is redundant because of the `return` - you can remove the indentation.

src/app/users/login.ts
```diff
@@ -64,33 +65,66 @@ export const login = async ({
...
+        // account exist but doesn't have a kratos user yet
+
+        let account = await AccountsRepository().findByUserId(user.id)
+        if (account instanceof Error) return account
+
+        account = await AccountsRepository().attachKratosUser({
```
bodymindarts, line 104: This is not really how a repository should be used as it is putting behaviour in the repo.

Typically repositories only expose `crud` operations.

The cleaner version would be:
```

File: src/app/users/login.ts [hunk 9]
```diff
@@ -64,33 +65,66 @@ export const login = async ({
...
+          id: account.id,
+          kratosUserId,
+        })
+
+        if (account instanceof Error) {
+          recordExceptionInCurrentSpan({
```
bodymindarts, line 110: Shouldn't this be called in `AccountsRepository`?

bodymindarts (commented):
Only started but already publishing findings up to here

src/domain/accounts/index.types.d.ts
```diff
@@ -136,13 +137,23 @@ type AccountValidator = {
   validateWalletForAccount(wallet: Wallet): true | ValidationError
 }
 
+type NewAccountInfo = {
+  kratosUserId: KratosUserId
+  phone?: PhoneNumber
```
bodymindarts, line 142: how is phone optional? Later when we support email perhaps - but now?

src/servers/middlewares/auth-router.ts
```diff
@@ -11,23 +10,22 @@ import { mapError } from "@graphql/error-map"
 import { baseLogger } from "@services/logger"
 import { wrapAsyncToRunInSpan } from "@services/tracing"
 
-export const KratosSdk: (kratosEndpoint?: string) => V0alpha2ApiInterface = (
-  kratosEndpoint,
-) => new V0alpha2Api(new Configuration({ basePath: kratosEndpoint }))
+import { AccountsRepository } from "@services/mongoose"
+import { kratosPublic } from "@services/kratos/private"
```
bodymindarts, line 14: should not import anything under `private` (as the name implies).

src/servers/middlewares/auth-router.ts
```diff
@@ -38,9 +36,9 @@ authRouter.post("/browser", async (req, res) => {
   const logger = graphqlLogger.child({ ip, body: req.body })
 
   try {
-    const { data } = await kratos.toSession(undefined, req.header("Cookie"))
+    const { data } = await kratosPublic.toSession(undefined, req.header("Cookie"))
```
bodymindarts, line 39: This `toSession` method should be exposed via a `Service` - we shouldn't call the client directly. Even if its just a thin wrapper - catching and translating errors already has value.

src/servers/middlewares/auth-router.ts
```diff
@@ -38,9 +36,9 @@ authRouter.post("/browser", async (req, res) => {
...
 
-    const kratosLoginResp = await Users.loginWithKratos({
+    const kratosLoginResp = await Users.loginWithEmail({
```
bodymindarts, line 41: I'm wondering if the boundary is still correct here. Perhaps this should actually be `loginWithCookie` and the use case calls the `AuthService` to resolve the cookie etc.

src/servers/middlewares/auth-router.ts
```diff
@@ -11,23 +10,22 @@ import { mapError } from "@graphql/error-map"
 import { baseLogger } from "@services/logger"
 import { wrapAsyncToRunInSpan } from "@services/tracing"
 
-export const KratosSdk: (kratosEndpoint?: string) => V0alpha2ApiInterface = (
-  kratosEndpoint,
-) => new V0alpha2Api(new Configuration({ basePath: kratosEndpoint }))
+import { AccountsRepository } from "@services/mongoose"
+import { kratosPublic } from "@services/kratos/private"
+import { validateKratosToken } from "@services/kratos/tests-but-not-prod"
```
bodymindarts, line 15: Another import from a private sub module - we're missing a container here.

src/servers/middlewares/auth-router.ts
```diff
@@ -59,35 +57,54 @@ authRouter.post("/browser", async (req, res) => {
...
+
+      if (!authz) {
+        res.status(401).send({ error: "Missing token" })
+        return
+      }
+
+      const rawToken = authz.slice(7) as string
+
+      // new flow
+      if (rawToken.length === 32) {
```
bodymindarts, line 80: The logic here regarding which token we have and how to validate / verify it should be its own use case - don't want this logic in the middleware IMO.

src/servers/middlewares/auth-router.ts
```diff
@@ -102,4 +119,10 @@ authRouter.post(
   }),
 )
 
+authRouter.post("/after_settings_hooks", (req, res) => {
```
bodymindarts, line 122: Can we delete this if its `// not currently use`?

src/servers/middlewares/update-token.ts
```diff
@@ -0,0 +1,132 @@
...
+    next()
+    return
+  }
+
+  if (!tokenPayload) {
+    next()
+    return
+  }
+
+  const uid = tokenPayload.uid
```
bodymindarts, line 43: is this `uid` now always the `kratosUserId`?

src/servers/middlewares/update-token.ts
```diff
@@ -0,0 +1,132 @@
...
+  const twoMonths = (60 * 60 * 24 * 30) as Seconds
+
+  await RedisCacheService().set<SessionToken>({
+    key: rawToken,
+    value: kratosToken,
+    ttlSecs: twoMonths,
+  })
+
+  return
+}
```
bodymindarts, line 132: This whole `update-token` file is a bit confusing to me. We are interleaving application logic with express details. For small trivial things I guess its fine but these are 132 lines and I would prefer a more explicit separation of layers. There is a missing function that handles the use case in a way decoupled from the middleware implementation details. Don't know exactly what its called or where it lives but I think we need it.

todo.txt
```diff
@@ -0,0 +1,4 @@
+make it work with cookie
```
bodymindarts, line 1: guessing this file is an accidental commit.

src/servers/middlewares/auth-router.ts
nicolasburtey: it's meant to be just temporary in the transition stage. do you still think I should extract it?

src/servers/middlewares/auth-router.ts
nicolasburtey: I didn't really touch this code (expect a function rename). This code is meant to be deleted on Cookie is activated and webwallet updated. (it was a "hack"/proxy for when there was no Oathkeeper that knew how to manage cookie)

src/servers/middlewares/auth-router.ts
nicolasburtey: I didn't really touch this code (expect a function rename). This code is meant to be deleted on Cookie is activated and webwallet updated. (it was a "hack"/proxy for when there was no Oathkeeper that knew how to manage cookie)

src/servers/middlewares/auth-router.ts
bodymindarts: It would be cleaner but its not a must have if the conditional is temporary. Can leave it I guess.

src/servers/middlewares/auth-router.ts
nicolasburtey: I didn't really touch this code (expect a function rename). This code is meant to be deleted on Cookie is activated and webwallet updated. (it was a "hack"/proxy for when there was no Oathkeeper that knew how to manage cookie)

src/domain/accounts/index.types.d.ts
bodymindarts: Then shouldn't these be 2 different types?

src/servers/middlewares/update-token.ts
bodymindarts: Re floating this - if you are looking for something to cleanup I would extract this into a use case - session handling is application logic.

bodymindarts (commented):
It looks okay to me - I guess most issues will go away post migration. Re-tagged a few earlier comments that you perhaps missed.

Makefile
```diff
@@ -80,34 +80,64 @@ e2e-in-ci:
...
+	docker stop `docker ps -q -a` && \
+	docker rm `docker ps -q -a`
```
dolcalmi, line 119: this will affect other containers running in dev machine

Makefile
```diff
@@ -80,34 +80,64 @@ e2e-in-ci:
...
+main-in-ci-build:
+	yarn build && \
+	make create-tmp-env-ci && \
+	TMP_ENV_CI=tmp.env.ci docker compose -f docker-compose.yml run --name e2e-tests e2e-tests make start-main-ci || \
+	docker rm `docker ps -q -a -f status=exited`
+
+main-in-ci-cached:
+	make create-tmp-env-ci && \
+	TMP_ENV_CI=tmp.env.ci docker compose -f docker-compose.yml run --name e2e-tests e2e-tests make start-main-ci || \
+	docker rm `docker ps -q -a -f status=exited`
```
dolcalmi, line 110: we should try to be specific, this will remove all stopped containers (including not related to galoy)

Makefile
```diff
@@ -80,34 +80,64 @@ e2e-in-ci:
...
+delete-e2e:
+	docker container kill `docker ps -f name="e2e-tests" -q` && \
+	sleep 1 && \
+	docker container rm `docker ps -f name="e2e-tests" -q`
+
+main-in-ci-build:
+	yarn build && \
+	make create-tmp-env-ci && \
+	TMP_ENV_CI=tmp.env.ci docker compose -f docker-compose.yml run --name e2e-tests e2e-tests make start-main-ci || \
+	docker rm `docker ps -q -a -f status=exited`
```
dolcalmi, line 105: same here

Makefile
```diff
@@ -80,34 +80,64 @@ e2e-in-ci:
...
+
+main-in-ci-cached:
+	make create-tmp-env-ci && \
+	TMP_ENV_CI=tmp.env.ci docker compose -f docker-compose.yml run --name e2e-tests e2e-tests make start-main-ci || \
+	docker rm `docker ps -q -a -f status=exited`
+
+main-in-ci-cached-no-restart:
+	make create-tmp-env-ci && \
+	TMP_ENV_CI=tmp.env.ci docker compose run --name e2e-tests e2e-tests make start-main-ci || \
+	docker rm `docker ps -q -a -f status=exited`
```
dolcalmi, line 115: and here

src/services/mongoose/accounts.ts
```diff
@@ -171,7 +185,7 @@ export const AccountsRepository = (): IAccountsRepository => {
   }
 
   return {
-    persistNewKratosUser,
+    persistNew,
```
dolcalmi, line 188: uhmm so now user = account?

src/services/kratos/tests-but-not-prod.ts
```diff
@@ -1,22 +1,17 @@
+import { SuccessfulSelfServiceLoginWithoutBrowser } from "@ory/client"
```
dolcalmi, line 1: is this files needed? `src/services/kratos/tests-but-not-prod.ts`

src/services/kratos/index.ts
```diff
@@ -1 +1,50 @@
+import { AuthenticationKratosError, UnknownKratosError } from "./errors"
```
dolcalmi, line 1: why is this file not using an interface?

src/services/kratos/cron.ts
```diff
@@ -0,0 +1,34 @@
+import { kratosAdmin, listSessionsInternal } from "./private"
+
+import { listIdentities } from "."
```
dolcalmi, line 3: possible circular reference, please check

src/services/kratos/cron.ts
```diff
@@ -0,0 +1,34 @@
+import { kratosAdmin, listSessionsInternal } from "./private"
```
dolcalmi, line 1: this should be also in service with a normal structure...

src/servers/cron.ts
```diff
@@ -15,6 +15,7 @@ import { ColdStorage, Lightning, Wallets, Payments, Swap } from "@app"
 import { getCronConfig, TWO_MONTHS_IN_MS } from "@config"
 
 import { rebalancingInternalChannels, reconnectNodes } from "@services/lnd/utils-bos"
+import { extendSessions } from "@services/kratos"
```
dolcalmi, line 18: Servers should try to only access app layer

src/domain/accounts/index.types.d.ts
```diff
@@ -136,13 +137,33 @@ type AccountValidator = {
   validateWalletForAccount(wallet: Wallet): true | ValidationError
 }
 
+type NewAccountWithEmailIdentifier = {
+  kratosUserId: KratosUserId
+  phone?: undefined
+  phoneMetadata?: undefined
```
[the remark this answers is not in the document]
nicolasburtey: good catch. I've been struggling a little bit with the type here. 

it's because I have a `NewAccount` type below, and typescript would not be happy if there was no `phone` or `phoneMetadata` properties

src/servers/graphql-server.ts
```diff
@@ -88,29 +91,33 @@ const sessionContext = ({
...
     },
     async () => {
-      if (!(maybeAid instanceof InvalidAccountIdError)) {
-        const userId = maybeAid as string as UserId // FIXME: fix until User is attached to kratos
-        const loggedInUser = await Users.getUserForLogin({ userId, ip, logger })
+      // note: value should match (ie: "anon") if not an accountId
+      // settings from dev/ory/oathkeeper.yml/authenticator/anonymous/config/subjet
+      const maybeKratosUserId = checkedToKratosUserId(tokenPayload.sub || "")
+      if (!(maybeKratosUserId instanceof ValidationError)) {
+        const userId = maybeKratosUserId
```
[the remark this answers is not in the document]
nicolasburtey: not strictly needed but the aim is to show that `maybeKratosUserId` could be `anon` also.

src/servers/graphql-server.ts
```diff
@@ -88,29 +91,33 @@ const sessionContext = ({
...
+
+        const account = await Accounts.getAccountFromKratosUserId(userId)
+        if (account instanceof Error) throw Error
+        domainAccount = account
+
+        const loggedInUser = await Users.getUserForLogin({
+          userId: account.id as string as UserId,
```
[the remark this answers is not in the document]
nicolasburtey: `userId` collection from the mongoose collection is been removed. may untimely be back with `KratosUserId` but I'm not mixing the type for now.

src/services/mongoose/accounts.ts
nicolasburtey: `User` from mongoose = `Account` domain object (mongoose collection to be renamed in follow up PR)

src/services/kratos/tests-but-not-prod.ts
nicolasburtey: yes, used in tests for now. 

more function will be moved in production in follow up PRs.

src/services/kratos/cron.ts
nicolasburtey: will check. is there a way to automatically detect those references?

src/services/kratos/cron.ts
nicolasburtey: any example to follow?

src/servers/cron.ts
nicolasburtey: fixed in 664a70632ec7c5cd00c54fafe559477bdf004c80

src/services/kratos/cron.ts
nicolasburtey: fixed in 664a70632ec7c5cd00c54fafe559477bdf004c80

src/services/kratos/index.ts
nicolasburtey: I'm still working on the interface as I learn about how kratos works

src/domain/accounts/index.types.d.ts
```diff
@@ -136,13 +137,33 @@ type AccountValidator = {
...
   listUnlockedAccounts(): Promise<Account[] | RepositoryError>
   findById(accountId: AccountId): Promise<Account | RepositoryError>
   findByUserId(userId: UserId): Promise<Account | RepositoryError>
 
   findByKratosUserId(kratosUserId: KratosUserId): Promise<Account | RepositoryError>
-  persistNewKratosUser(kratosUserId: KratosUserId): Promise<Account | RepositoryError>
+  persistNew({ kratosUserId }: NewAccount): Promise<Account | RepositoryError>
+
+  persistNew({
+    kratosUserId,
```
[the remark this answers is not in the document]
bodymindarts: Yes we have been keeping `domain` clean of any external dependencies. Samer and I originally introduced `kratosUserId` as a quick way to get web-wallet working and Nicolas kept that.
At the time reasons were:
- quick way to move forward
- clear differentiation with `userId`
- low cognitive overhead in needing to map a more abstract name

At this point as we are taking more time to think about modelling we could
- introduce a better name to use in the long trem
- keep it as is until the 'legacy' `userId` is out of the code base and change it then (when there is less chance for confusion)
- keep it like this as we are hard coupled to kratos and can accept this

Alternative names I would suggest:
- `identityId`
- `externalUserId`
- ... other?

Either way as long as the auth code is in transition I don't think it's critical - we will probably come up with a better name / concept post transition.

src/domain/accounts/index.types.d.ts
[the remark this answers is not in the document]
bodymindarts: This ties into my previous comment that `NewAccount` is probably 2 types. `NewAccountWithPhoneIdenitty`, `NewAccountWithEmailIdentity` or something like that.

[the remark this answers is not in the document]
bodymindarts: In general `optional?` fields are almost always indications / symptoms of different use-cases being unified. Should be avoided / questioned when we encounter them.

Makefile
bodymindarts: Yeah I already had a comment here (looks like it was removed).

I think you want `docker compose down -sfv`

src/app/auth/index.ts
```diff
@@ -0,0 +1,18 @@
+import { extendSession, listIdentities } from "@services/kratos"
+
+// TODO: interface/type should be reworled so that it doesn't have to come from private
+import { listSessionsInternal } from "@services/kratos/private"
```
bodymindarts, line 4: This is what `private` is meant for. To be used internally to the `auth` module and not `export`ed... looks fine to me.

nicolasburtey: it's exported to the use case here. so outside the service. does that still looks good for you?

src/domain/accounts/index.types.d.ts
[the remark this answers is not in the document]
dolcalmi: probably is better to use `Omit` https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys to create the new type

krtk6160 (approved):

src/domain/accounts/index.types.d.ts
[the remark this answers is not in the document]
nicolasburtey: thanks for your input @dolcalmi. this will go away in the next PR so not spending much time on this right now.

PR #1803, opened 2022-10-11
chore: migrate remaining intraledger types to self trade

vindard:
## Discussion

This PR is to follow #1750. It executes the migration for intraledger transactions that were self trades but did not get tagged with this new type before we implemented the self-trade code.

### Rollout process

We should do a check of which transactions are currently on the old type and should be the new type pre-migration, and then confirm that these types change as expected post-migration. I've added a rollback migration step as well for in case anything goes wrong from this check.

I can work with Jose maybe to execute this check before we roll out.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -162,6 +162,28 @@ When developing migrations the best way to test them on a clean database is:
 make test-migrate
 ```
 
+#### Create a new migration
+
+Create the migration file
+
+```bash
+npx migrate-mongo create <migration-name> \
+  -f src/migrations/migrate-mongo-config.js
+```
+
+Write the migration in the newly created migration file and then test/run with the following:
+```bash
+# Migrate
+npx migrate-mongo up \
+  -f src/migrations/migrate-mongo-config.js
+
+# Rollback
+npx migrate-mongo down \
+  -f src/migrations/migrate-mongo-config.js
+```
+
+When testing, to isolate just the current migration being worked on in local dev you can temporarily move the other migrations to another dir.
+
 ### Known issues
 
 * **Test suite timeouts**: increase jest timeout value. Example:
--- a/src/migrations/20221010162913-add-self-trade-type.ts
+++ b/src/migrations/20221010162913-add-self-trade-type.ts
@@ -1,6 +1,15 @@
 /* eslint @typescript-eslint/no-var-requires: "off" */
 const { Types } = require("mongoose")
 
+const LedgerTransactionType = {
+  IntraLedger: "on_us",
+  LnIntraLedger: "ln_on_us",
+  OnchainIntraLedger: "onchain_on_us",
+  WalletIdTradeIntraAccount: "self_trade",
+  LnTradeIntraAccount: "ln_self_trade",
+  OnChainTradeIntraAccount: "onchain_self_trade",
+}
+
 const toWalletId = (walletIdPath) => {
   const path = walletIdPath.split(":")
   return Array.isArray(path) && path.length === 2 && path[0] === "Liabilities" && path[1]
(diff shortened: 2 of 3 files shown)


nicolasburtey:
@vindard tests are not passing on this one

DEV.md
```diff
@@ -160,6 +160,28 @@ When developing migrations the best way to test them on a clean database is:
...
+
+```bash
+node_modules/.bin/migrate-mongo create <migration-name> \
+  -f src/migrations/migrate-mongo-config.js
+```
+
+Write the migration in the newly created migration file and then test/run with the following:
+```bash
+# Migrate
+node_modules/.bin/migrate-mongo up \
```
dolcalmi, line 175: instead of this use `npx migrate-mongo up ...` (applies for all cmds)

src/migrations/20221010162913-add-self-trade-type.ts
```diff
@@ -0,0 +1,186 @@
...
+
+const getNonUserWalletIds = async ({ usersCollection, walletsCollection }) => {
+  const accountIds = []
+  const usersCursor = usersCollection.find({
+    role: { $in: ["funder", "dealer", "bankowner"] },
+  })
+  for await (const user of usersCursor) {
+    const accountId = user._id.toString()
+    accountIds.push(accountId)
+  }
```
dolcalmi, line 31: you dont need to use cursor for 3 accounts but is not an issue

src/migrations/20221010162913-add-self-trade-type.ts
```diff
@@ -0,0 +1,186 @@
...
+
+    const accountIdsByWalletId = {}
+    const getSelfTradeType = async (txnAgg) => {
+      // BUILD 'txns' FOR JOURNALID
+      // =====
+      const txns = []
+      const txnsCursor = txnsCollection.find({ _journal: txnAgg._id })
+      for await (const txn of txnsCursor) {
+        // Filter internal wallet IDs
+        if (nonUserWalletIds.includes(toWalletId(txn.accounts))) continue
```
dolcalmi, line 63: not a blocker but you can also use `account_path` (it is an array so you dont have to split)

src/migrations/20221010162913-add-self-trade-type.ts
```diff
@@ -0,0 +1,186 @@
...
+        default:
+          return false
+      }
+    }
+
+    // FETCH INTRALEDGER TXNS AND UPDATE TYPE FOR SELF-TRADES
+    // =====
+    const journalCursor = txnsCollection.aggregate([
+      {
+        $match: { type: { $regex: "on_us" }, currency: "USD" },
```
dolcalmi, line 120: is this always `on_us`? cant be `ln_on_us` or `onchain_on_us`?

dolcalmi (changes requested):

DEV.md
bodymindarts: Don't think we have `npx` as a mandatory dev dependency... should probably move to nix based dev envs at some point.

src/migrations/20221010162913-add-self-trade-type.ts
vindard: Regex pattern catches all of the above

DEV.md
dolcalmi: `Since npm version 5.2.0 npx is pre-bundled with npm`

vindard: Changed it to `npx`... `yarn` had some issue with node versions when I tried to run with it https://github.com/GaloyMoney/galoy/compare/1da5b8504fd7c5f01955a13eb91cec00841cbf01..0d7142e1bf1e3d38fd75c5463c7041faeb5af44b

dolcalmi (approved):

PR #1811, opened 2022-10-12
chore: remove apollo usage reporting

nicolasburtey:
I don't think we use it anymore?


bodymindarts (approved):

bodymindarts:
Isn't some config injection missing from the deletions?

PR #1812, opened 2022-10-12
refactor: move conversion fn selection logic from app layer to domain

vindard:
## Description

We've discovered a new dimension of conversion method selection where the decision to use a `buy` or `sell` method is dependent on the sender wallet currency. Our initial fix was to do this switching in the app layer. This PR is to move this selection logic to the LPFB and to unit test these new pathways.

### DIFF
--- a/src/app/payments/get-protocol-fee.ts
+++ b/src/app/payments/get-protocol-fee.ts
@@ -85,14 +85,14 @@ const estimateLightningFee = async ({
     senderWallet,
     invoice,
     uncheckedAmount,
-    usdFromBtc:
-      senderWallet.currency === WalletCurrency.Btc
-        ? dealer.getCentsFromSatsForFutureBuy
-        : dealer.getCentsFromSatsForFutureSell,
-    btcFromUsd:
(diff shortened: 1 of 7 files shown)


src/app/payments/get-protocol-fee.ts
```diff
@@ -85,14 +85,14 @@ const estimateLightningFee = async ({
...
     uncheckedAmount,
-    usdFromBtc:
-      senderWallet.currency === WalletCurrency.Btc
-        ? dealer.getCentsFromSatsForFutureBuy
-        : dealer.getCentsFromSatsForFutureSell,
-    btcFromUsd:
-      senderWallet.currency === WalletCurrency.Btc
-        ? dealer.getSatsFromCentsForFutureBuy
-        : dealer.getSatsFromCentsForFutureSell,
+    hedgeBuy: {
```
bodymindarts, line 88: can we rename to `hedgeBuyUsd` to make it clear what we're buying

bodymindarts (commented):
Yeah this is fine - just a small suggestion for naming improvement.

bodymindarts (approved):

PR #1813, opened 2022-10-12
fix: update oathkeeper hosts with docker host ip

dolcalmi:
How to test
```bash
make reset-deps
make start-main-fast
# open http://localhost:4002/graphql and query price (should not get unauthorized error)

make reset-e2e 
make clean-deps
make e2e-in-ci
```


test/helpers/integration-server.ts
```diff
@@ -13,7 +13,6 @@ export const startServer = async (command: string): Promise<PID> => {
     })
     const serverPid = serverProcess.pid as PID
     serverProcess.stdout.on("data", (data) => {
-      console.log(data.toString())
```
nicolasburtey, line 16: this is helpful (necessary imo) otherwise the stdout of the subprocess are just eaten

dolcalmi: fixed

nicolasburtey (approved):

PR #1830, opened 2022-10-16
chore: kratos cleanup

nicolasburtey:
this PR:
- remove the user domain that relied on mongoose. 
- store in kratos the phone, phoneMetadata, language, token for firebase.
- prepare the fact that other schemas can be used for kratos, such as email+password


bodymindarts:
It would be easier to review this if it would merge into the `activate-kratos` branch

nicolasburtey:
> It would be easier to review this if it would merge into the `activate-kratos` branch

updated. I used `main` because other the tests don't run.

bodymindarts:
Could you make a PR just for deleting 2FA - we could merge it right away (no dependency on kratos) and it would make this diff significantly smaller.

PR #1839, opened 2022-10-18
chore: Fixed integration tests on m1 mac

daviroo:
On my local machine both stablesats and the integration-deps runner was starting up before mongodb had actually started.  Allowing them to restart on failure gives enough time for mongodb to start.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,6 +18,7 @@ services:
       - lnd-outside-2
       - otel-agent
       - oathkeeper
+    restart: on-failure:10
   otel-agent:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -18,6 +18,7 @@ services:
       - lnd-outside-2
       - otel-agent
       - oathkeeper
+    restart: on-failure:10
```
nicolasburtey, line 21: what does the `:10` means?

daviroo: It will retry a maximum of 10 times.

nicolasburtey (approved):

PR #1846, opened 2022-10-19
test: add cases for hedge-price usage

vindard:
## Discussion

This PR is to start adding a set of test cases to ensure our use-cases are configured properly with the dealer conversion functions they use for different pathways.

### DIFF
--- a/test/helpers/user.ts
+++ b/test/helpers/user.ts
@@ -1,5 +1,9 @@
 import { getDefaultAccountsConfig, yamlConfig } from "@config"
+
 import { CouldNotFindUserFromPhoneError } from "@domain/errors"
+import { WalletCurrency } from "@domain/shared"
+import { WalletType } from "@domain/wallets"
+import { adminUsers } from "@domain/admin-users"
 import {
   AuthenticationError,
   LikelyNoUserWithThisPhoneExistError,
@@ -12,14 +16,21 @@ import {
 } from "@services/mongoose"
 import { User } from "@services/mongoose/schema"
 import { toObjectId } from "@services/mongoose/utils"
-import { addWallet, addWalletIfNonexistent } from "@app/accounts/add-wallet"
-import { WalletCurrency } from "@domain/shared"
-import { WalletType } from "@domain/wallets"
-import { adminUsers } from "@domain/admin-users"
 import { UsersIpRepository } from "@services/mongoose/users-ips"
-import { createAccountWithPhoneIdentifier } from "@app/accounts"
 
 import { AuthWithPhonePasswordlessService } from "@services/kratos"
+import { baseLogger } from "@services/logger"
+
+import { Wallets } from "@app"
+import {
+  addWallet,
+  addWalletIfNonexistent,
+  createAccountWithPhoneIdentifier,
+} from "@app/accounts"
+
+import { sleep } from "@utils"
+
+import { lndOutside1, safePay } from "./lightning"
 
 const users = UsersRepository()
 
@@ -286,6 +297,55 @@ export const createNewWalletFromPhone = async ({
     { _id: toObjectId<UserId>(user.id) },
     { deviceToken: ["test-token"] },
   )
+
+  return wallet
+}
+
+export const createAndFundNewWalletForPhone = async <S extends WalletCurrency>({
+  phone,
+  balanceAmount,
+}: {
+  phone: PhoneNumber
+  balanceAmount: PaymentAmount<S>
+}) => {
+  // Create new wallet
+  const wallet = await createNewWalletFromPhone({
+    phone,
+    currency: balanceAmount.currency,
+  })
+  if (wallet instanceof Error) throw wallet
+
+  // Fund new wallet if a non-zero balance is passed
+  if (balanceAmount.amount === 0n) return wallet
+
+  const lnInvoice = await Wallets.addInvoiceForSelf({
+    walletId: wallet.id,
+    amount: Number(balanceAmount.amount),
+    memo: `Fund new wallet ${wallet.id}`,
+  })
+  if (lnInvoice instanceof Error) throw lnInvoice
+  const { paymentRequest: invoice, paymentHash } = lnInvoice
+
+  const updateInvoice = () =>
+    Wallets.updatePendingInvoiceByPaymentHash({
+      paymentHash,
+      logger: baseLogger,
+    })
+
+  const promises = Promise.all([
+    safePay({ lnd: lndOutside1, request: invoice }),
+    (async () => {
+      // TODO: we could use event instead of a sleep to lower test latency
+      await sleep(500)
+      return updateInvoice()
+    })(),
+  ])
+
+  {
+    // first arg is the outsideLndpayResult
+    const [, result] = await promises
+    expect(result).not.toBeInstanceOf(Error)
+  }
 
   return wallet
 }
--- a/test/integration/services/dealer/hedge-price.spec.ts
+++ b/test/integration/services/dealer/hedge-price.spec.ts
@@ -366,8 +366,8 @@ describe("arbitrage strategies", () => {
     // an invoice up to 78 sats for $0.01 from a USD wallet and then convert back
     // to $0.01 for 53 sats.
 
-    describe("pay max btc-amount invoice from usd wallet, replenish $0.01 back to usd wallet", () => {
-      describe("replenish with $0.01 pull from usd wallet", () => {
+    describe("pay max btc-amount invoice from usd wallet", () => {
+      describe("replenish with $0.01 pull back to usd wallet", () => {
         it("via usd-denominated invoice", async () => {
           const accountAndWallets = await newAccountAndWallets()
           const { newBtcWallet, newUsdWallet, newAccount } = accountAndWallets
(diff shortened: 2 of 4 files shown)


test/integration/services/dealer/hedge-price.spec.ts
```diff
@@ -0,0 +1,1892 @@
...
+    amount: Number(btcPaymentAmount.amount),
+    ...sendArgs,
+  })
+  if (result instanceof Error) {
+    if (!(result instanceof ZeroAmountForUsdRecipientError)) throw result
+    return result
+  }
+  const afterBtc = await getBalanceHelper(newBtcWallet.id)
+  const diff = (beforeBtc - afterBtc) as CurrencyBaseAmount
+  return diff
```
dolcalmi, line 110: if variable will not be used you can return the difference directly

test/integration/services/dealer/hedge-price.spec.ts
```diff
@@ -0,0 +1,1892 @@
...
+  }
+
+  const beforeBtc = await getBalanceHelper(newBtcWallet.id)
+
+  const result = await Payments.intraledgerPaymentSendWalletId({
+    amount: Number(btcPaymentAmount.amount),
+    ...sendArgs,
+  })
+  if (result instanceof Error) {
+    if (!(result instanceof ZeroAmountForUsdRecipientError)) throw result
```
dolcalmi, line 105: why is only this error being thrown?

test/integration/services/dealer/hedge-price.spec.ts
```diff
@@ -0,0 +1,1892 @@
...
+    balanceAmount: usdFundingAmount,
+  })
+
+  const newAccount = await AccountsRepository().findById(newBtcWallet.accountId)
+  if (newAccount instanceof Error) throw newAccount
+
+  return { newBtcWallet, newUsdWallet, newAccount }
+}
+
+const getBtcForUsdEquivalentIntraledger = async ({
```
dolcalmi, line 83: not sure about this name, you are using this as a diff function but this indicates something else

test/integration/services/dealer/hedge-price.spec.ts
```diff
@@ -0,0 +1,1892 @@
...
+  if (result instanceof Error) {
+    if (!(result instanceof ZeroAmountForUsdRecipientError)) throw result
+    return result
+  }
+  const afterBtc = await getBalanceHelper(newBtcWallet.id)
+  const diff = (beforeBtc - afterBtc) as CurrencyBaseAmount
+  return diff
+}
+
+const getUsdForBtcEquivalentWithAmountInvoice = async ({
```
dolcalmi, line 113: same here

test/integration/services/dealer/hedge-price.spec.ts
```diff
@@ -0,0 +1,1892 @@
...
+  const result = await Payments.payInvoiceByWalletId({
+    uncheckedPaymentRequest: lnInvoice.paymentRequest,
+    memo: null,
+    senderWalletId: newUsdWallet.id,
+    senderAccount: newAccount,
+  })
+  if (result instanceof Error) throw result
+  const afterUsd = await getBalanceHelper(newUsdWallet.id)
+  const diff = (beforeUsd - afterUsd) as CurrencyBaseAmount
+  return diff
```
dolcalmi, line 137: same here

test/integration/services/dealer/hedge-price.spec.ts
```diff
@@ -0,0 +1,1892 @@
...
+}
+
+const getUsdForBtcEquivalentWithAmountInvoiceAndProbe = async ({
```
dolcalmi, line 140: same `name` comment here

test/integration/services/dealer/hedge-price.spec.ts
```diff
@@ -0,0 +1,1892 @@
...
+  const payResult = await Payments.payInvoiceByWalletId({
+    uncheckedPaymentRequest: lnInvoice.paymentRequest,
+    memo: null,
+    senderWalletId: newUsdWallet.id,
+    senderAccount: newAccount,
+  })
+  if (payResult instanceof Error) throw payResult
+  const afterUsd = await getBalanceHelper(newUsdWallet.id)
+  const diff = (beforeUsd - afterUsd) as CurrencyBaseAmount
+  return diff
```
dolcalmi, line 170: same here

test/integration/services/dealer/hedge-price.spec.ts
```diff
@@ -0,0 +1,1892 @@
...
+    amount: Number(btcPaymentAmount.amount),
+    uncheckedPaymentRequest: lnInvoice.paymentRequest,
+    memo: null,
+    senderWalletId: newBtcWallet.id,
+    senderAccount: newAccount,
+  })
+  if (result instanceof Error) {
+    if (!(result instanceof ZeroAmountForUsdRecipientError)) throw result
+    return result
+  }
```
dolcalmi, line 240: apply to all diff methods, if return zero amount error is required then probably is cleaner do something like this:
```
if (result instanceof ZeroAmountForUsdRecipientError) return result
```

test/integration/services/dealer/hedge-price.spec.ts
```diff
@@ -0,0 +1,1892 @@
...
+  const afterBtc = await getBalanceHelper(newBtcWallet.id)
+  const diff = (beforeBtc - afterBtc) as CurrencyBaseAmount
+  return diff
+}
+
+const getMaxBtcAmountToEarn = async ({
+  startingBtcAmount,
+  accountAndWallets,
+  diffFn,
```
dolcalmi, line 249: I guess this is not just diff, it's a no change comment and I know we are using this in src but this introduces an overhead in logic and is really hard to track, especially if the fn name doesn't include what the function actually does

dolcalmi (changes requested):
not sure if worth it but we can add a helper file for diff and amount to earn/spend functions

test/helpers/user.ts
```diff
@@ -286,3 +295,52 @@ export const createNewWalletFromPhone = async ({
...
+      paymentHash,
+      logger: baseLogger,
+    })
+
+  const promises = Promise.all([
+    safePay({ lnd: lndOutside1, request: invoice }),
+    (async () => {
+      // TODO: we could use event instead of a sleep to lower test latency
+      await sleep(500)
+      return updateInvoice()
```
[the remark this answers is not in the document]
vindard: I just tried this and the test timed out because I assume it doesn't get past the `await safePay` line to execute the update.

I also tried calling `safePay` without the `await` and the test still timed out.

test/integration/services/dealer/hedge-price.spec.ts
vindard: Only the `ZeroAmountForUsdRecipientError` is not being thrown because we'd like to use that value outside this function. All other errors are thrown here like we usually do in tests to fail the test.

test/integration/services/dealer/hedge-price.spec.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1846/commits/4ebd8ddd7b4fc84653a2c1d96c36044d73429e30

test/integration/services/dealer/hedge-price.spec.ts
vindard: Fixed: https://github.com/GaloyMoney/galoy/pull/1846/commits/4ebd8ddd7b4fc84653a2c1d96c36044d73429e30

vindard:
> not sure if worth it but we can add a helper file for diff and amount to earn/spend functions

I would usually only do this if the functions were going to be used in more than 1 file. Do you think it makes sense here since we're using only in these tests?

test/integration/services/dealer/hedge-price.spec.ts
vindard: With the new names does this function make more sense now? I tried to explain what the function does with comments throughout the function as well. It executes a 3 step strategy to find the max BTC amount one can use to pay an invoice or recipient wallet.

test/integration/services/dealer/hedge-price.spec.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1846/commits/399649b05f40db5975be04c1869c36fc2cd6e1d9

test/integration/services/dealer/hedge-price.spec.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1846/commits/399649b05f40db5975be04c1869c36fc2cd6e1d9

test/integration/services/dealer/hedge-price.spec.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1846/commits/72daf78c0ff9c3500e4ce70d277bff077080a864

test/integration/services/dealer/hedge-price.spec.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1846/commits/72daf78c0ff9c3500e4ce70d277bff077080a864

test/integration/services/dealer/hedge-price.spec.ts
vindard: Fixed: https://github.com/GaloyMoney/galoy/pull/1846/commits/72daf78c0ff9c3500e4ce70d277bff077080a864

dolcalmi (approved):

PR #1861, opened 2022-10-22
chore: kratos cleanup 3

nicolasburtey:
main goal of the PR: removing `phone`, `phoneMetadata`, `deviceTokens`, `language` property from `Account` and putting it in a new User collection

plus some minors refactoring/cleanup


src/app/users/add-device-token.ts
```diff
@@ -0,0 +1,28 @@
+import { checkedToDeviceToken } from "@domain/users"
+import { UsersRepository } from "@services/mongoose"
+
+type AddDeviceTokenArgs = {
```
bodymindarts, line 4: Don't think we have been putting type definitions like this in `.ts` files.

src/app/users/update-language.ts
```diff
@@ -1,17 +1,22 @@
 import { checkedToLanguage } from "@domain/users"
 import { UsersRepository } from "@services/mongoose"
 
+type UpdateLanguageArgs = {
```
bodymindarts, line 4: Why not put the type in an `index.types.d.ts`?

src/app/wallets/update-pending-invoices.ts
```diff
@@ -15,9 +15,9 @@ import { NewDealerPriceService } from "@services/dealer-price"
 import { LndService } from "@services/lnd"
 import {
   AccountsRepository,
-  UsersRepository,
```
bodymindarts, line 18: Why the noise? What is the point of switching line 18 and 20 other than that it creates noise in the PR?

src/domain/accounts/index.types.d.ts
```diff
@@ -136,33 +137,13 @@ type AccountValidator = {
...
-type NewAccountWithPhoneIdentifier = {
-  kratosUserId: KratosUserId
-  phone: PhoneNumber
-  phoneMetadata?: PhoneMetadata
-}
-
-type NewAccount = NewAccountWithEmailIdentifier | NewAccountWithPhoneIdentifier
-
 interface IAccountsRepository {
+  list(): Promise<Account[] | RepositoryError>
```
bodymindarts, line 141: This seems like a potential performance issue. RU sure we want to expose being able to list _all_ accounts? Without a filter criteria or pagination (ie. iterator pattern)?

src/domain/authentication/index.types.d.ts
```diff
@@ -39,3 +40,9 @@ interface IAuthWithPhonePasswordlessService {
     password: IdentityPassword
   }): Promise<IdentityPhone | AuthenticationError> // TODO: should be IdentityPhoneWithPassword
 }
+
+interface IIdentityRepository {
+  getIdentity(id: KratosUserId): Promise<IdentityPhone | KratosError>
```
bodymindarts, line 45: A bit inconsistent naming here. If we are using `IdentityRepository` as then it should be `IdentityId` and `IdentityError` - not `Kratos`...

src/graphql/root/mutation/user-update-username.ts
```diff
@@ -35,7 +34,9 @@ const UserUpdateUsernameMutation = GT.Field({
     if (result instanceof Error) {
       return {
         errors: [mapAndParseErrorForGqlResponse(result)],
-        ...(result instanceof UsernameIsImmutableError ? { user: domainUser } : {}),
+
+        // FIXME: what is this return for?
+        // ...(result instanceof UsernameIsImmutableError ? { user: domainUser } : {}),
```
bodymindarts, line 39: Commenting this out seems like a breaking change in API behaviour. Need to validate.

src/services/mongoose/schema.ts
```diff
@@ -315,16 +280,90 @@ const UserSchema = new Schema<UserRecord>(
...
+      enum: [...Languages, ""],
+    },
+    deviceTokens: {
+      type: [String],
+      default: [],
+    },
+    phoneMetadata: {
+      type: {
+        carrier: {
+          error_code: String, // check this is the right syntax
```
bodymindarts, line 343: Does this need checking pre-merge?

src/services/mongoose/schema.types.d.ts
```diff
@@ -110,8 +106,23 @@ interface UserRecord {
   title?: string
   coordinates?: CoordinateObjectForUser
 
+  // TODO: delete
```
bodymindarts, line 109: Pretty unspescific comment... when does it get deleted?

src/services/mongoose/users.ts
```diff
@@ -1,60 +1,86 @@
...
 
-      return userFromRaw(result)
+      return translateToUser(result)
+    } catch (err) {
+      //  else
+      return parseRepositoryError(err)
+    }
+  }
+
+  const list = async (): Promise<User[] | RepositoryError> => {
```
bodymindarts, line 50: Think this is probably not a good idea to expose a way to get all users. At least we should return an iterater so as to not blow up memory.

todo.txt
```diff
@@ -0,0 +1,4 @@
+migration:
```
bodymindarts, line 1: I think this file is added by accident or? Should probably be in a ticket or PR.

bodymindarts (commented):
looks good overall...

mostly its renaming which is fine. Though I would introduce `IdentityError` and `IdentityId` rather than coupling to `KratosUserId` - also I think there is a circular dependency from domain -> service -> domain as the errors are defined in `service/kratos` rather than `domain/authentication`. Though I'm not sure I think we are generally defining the errors in the domain layer up to now.

src/domain/accounts/index.types.d.ts
nicolasburtey: I don't think this is actually been used, only the function below `listUnlockedAccounts` is being used based on a quick search. I have removed `list` 

`listUnlockedAccounts` should indeed probably get some pagination but it's currently only used once a day for cronjob so I think this can wait until it's needed

src/domain/authentication/index.types.d.ts
nicolasburtey: yeah I've been switching back and forth on the fact it's should be generic (`Identity`) or dedicated to Kratos. unclear to me how we should handle this, but I've started with Kratos and moved to Identity

src/graphql/root/mutation/user-update-username.ts
nicolasburtey: @samerbuna do you know what this line is for?

At least on the happy path (username is correctly set) commenting this line doesn't change anything

src/services/mongoose/schema.ts
nicolasburtey: I don't think so. this code was copy/paste: https://github.com/GaloyMoney/galoy/pull/1861/files/33b17223db559cb9b1dc8d5e49df8d7b53cd5ec3#diff-3e0c24fa2039e07c9e45083c6d45ecebbbc4d31d9133de31e215ecb31e52996cL202

There are new element that twilio recently added, so probably a good thing to dive into while reviewing this TODO: https://www.twilio.com/blog/detect-sim-swap-sms-otp-2fa

src/services/mongoose/schema.types.d.ts
nicolasburtey: when this second migration has been run: https://github.com/GaloyMoney/galoy/pull/1861/files/33b17223db559cb9b1dc8d5e49df8d7b53cd5ec3

src/graphql/admin/types/object/account.ts
```diff
@@ -36,7 +38,7 @@ const Account: GraphQLObjectType<Account> = GT.Object<Account>({
 
       type: GT.NonNull(GraphQLUser),
       resolve: async (source) => {
-        const user = await Users.getUser(source.ownerId)
+        const user = await UsersRepository().findById(source.kratosUserId)
```
nicolasburtey, line 41: @dolcalmi does this have any implication for open telemetry?

Makefile
```diff
@@ -127,7 +127,7 @@ execute-e2e-from-within-container:
 	NODE_ENV=test LOGLEVEL=error $(BIN_DIR)/jest --config ./test/jest-e2e.config.js --bail --runInBand --ci --reporters=default --reporters=jest-junit
```
vindard, line 127: Small nit, this might be possible

```suggestion
	make execute-e2e-from-within-container-cached
```

galoy.yaml
```diff
@@ -130,8 +130,6 @@ accountLimits:
 spamLimits:
   memoSharingSatsThreshold: 1000
   memoSharingCentsThreshold: 50
-twoFALimits:
```
vindard, line 133: This still exists in the `YamlSchema` type. Is the plan to remove from there as well?

src/services/mongoose/users.ts
```diff
@@ -1,60 +1,87 @@
...
   }
 
   const findByPhone = async (phone: PhoneNumber): Promise<User | RepositoryError> => {
     try {
-      const result = await User.findOne({ phone }, projection)
-      if (!result) {
-        return new CouldNotFindUserFromPhoneError(phone)
-      }
+      const result = await User.findOne({ phone })
+      if (!result) return new CouldNotFindError()
```
vindard, line 48: Why switch away from `CouldNotFindUserFromPhoneError` here?

vindard (commented):
Added some initial comments

src/services/kratos/identity.ts
```diff
@@ -21,34 +23,78 @@ export const getNextPage = (link: string): number | undefined => {
...
     }
 
-    // FIXME(nb) function above return duplicated query for the first 2 calls, so removing them here
-    const uniqueIdentities = identities.filter(
-      (value, index, self) => index === self.findIndex((t) => t.id === value.id),
-    )
-    assert(totalCount == uniqueIdentities.length)
+    const identity = identities.find((identity) => identity.traits.phone === phone)
+
+    if (!identity) return new PhoneIdentityInexistentError(phone)
```
vindard, line 90: Optional nit, `PhoneIdentityNonExistentError` or `PhoneIdentityDoesNotExistError` might be better here

src/services/kratos/auth-phone-no-password.ts
```diff
@@ -144,7 +145,7 @@ export const AuthWithPhonePasswordlessService = (): IAuthWithPhonePasswordlessSe
     const adminIdentity: AdminUpdateIdentityBody = {
       ...identity,
       credentials: { password: { config: { password } } },
-      state: "active",
+      state: identity.state as IdentityState, // FIXME? type bug from ory library?
```
vindard, line 148: Or if it's not a bug and `undefined` is possible, we can check for it just above this and throw an error if undefined. This may be safer if we can't confirm whether `identity.state` absolutely can't be `undefined` as the type suggests.

src/servers/graphql-server.ts
```diff
@@ -135,32 +137,32 @@ const sessionContext = ({
...
-      if (!(maybeKratosUserId instanceof ValidationError)) {
-        const userId = maybeKratosUserId
+      const maybeUserId = checkedToUserId(tokenPayload?.sub || "")
+      if (!(maybeUserId instanceof ValidationError)) {
+        const userId = maybeUserId
 
-        const account = await Accounts.getAccountFromKratosUserId(userId)
-        if (account instanceof Error) throw Error
+        const account = await Accounts.getAccountFromUserId(userId)
+        if (account instanceof Error) throw new Error(account.name)
```
vindard, line 145: Why not `throw mapError(account)` like we normally do?

src/servers/graphql-server.ts
```diff
@@ -135,32 +137,32 @@ const sessionContext = ({
...
           logger,
         })
-        if (loggedInUser instanceof Error)
-          throw new ApolloError("Invalid user authentication", "INVALID_AUTHENTICATION", {
-            reason: loggedInUser,
-          })
-        domainUser = loggedInUser
+
+        const userRes = await UsersRepository().findById(account.kratosUserId)
+        if (userRes instanceof Error) throw new Error(userRes.name)
```
vindard, line 156: We should only be throwing `CustomApolloError` from here
```suggestion
        if (userRes instanceof Error) throw mapError(userRes)
```

src/graphql/root/mutation/user-update-username.ts
vindard: It would return the user still in the response if the error type is `UsernameIsImmutableError`. Commenting here means that the `user` selection will always be undefined if `result` is an error.

There may be use-cases that rely on getting that user back even if the username value passed is invalid?

src/services/kratos/errors.ts
```diff
@@ -0,0 +1,25 @@
+import { AuthenticationError } from "@domain/authentication/errors"
+import { ErrorLevel } from "@domain/shared"
+
+export class KratosError extends AuthenticationError {}
+export class LikelyNoUserWithThisPhoneExistError extends KratosError {}
+export class LikelyUserAlreadyExistError extends KratosError {}
+export class AuthenticationKratosError extends KratosError {}
+
+export class PhoneIdentityInexistentError extends KratosError {}
```
vindard, line 9: All other errors already exist in `src/domain/authentication/errors.ts`. The only new error here is `PhoneIdentityInexistentError` and it should likely be added to that file and this file deleted

vindard (changes requested):
Some more comments, and a requested change around Kratis error definitions

src/graphql/admin/types/object/account.ts
vindard: Yes, this removes the span wrapping we do in `src/app/index.ts`. It's probably ok though, because any inner tracing on this repository method should be caught by the parent span here I think.

src/app/wallets/update-on-chain-receipt.ts
```diff
@@ -129,10 +129,10 @@ const processTxForWallet = async (
       return recorded
     }
 
-    const account = await AccountsRepository().findById(wallet.accountId)
-    if (account instanceof Error) return account
-
     if (!recorded) {
+      const account = await AccountsRepository().findById(wallet.accountId)
```
vindard, line 133: @dolcalmi did we want this account existence check for some reason outside the `if (!recorded)` block?

test/integration/notifications/notification.spec.ts
```diff
@@ -49,24 +49,30 @@ describe("notification", () => {
       expect(sendNotification.mock.calls.length).toBeGreaterThan(0)
 
       let usersWithDeviceTokens = 0
-      for (const { ownerId } of activeAccounts) {
-        const user = await UsersRepository().findById(ownerId)
-        if (user instanceof Error) throw user
+      for (const { kratosUserId } of activeAccounts) {
+        const user = await UsersRepository().findById(kratosUserId)
+        if (user instanceof Error) return user
```
vindard, line 54: Why return here? We usually throw from tests to stop execution when an unexpected error happens

galoy.yaml
nicolasburtey: good point. removed.

yes the goal is to re-introduce 2fa in a different way. instead of attaching 2fa to payment call, it should be a separate `validate2fa` graphql call, that will give you an elevated session for the next x minutes. 

in the UI, then you would do the 2fa, and move on the validation screen next if success. then the call in the backend for the payment would be same whether it needs 2fa or not. 

I think it's a much better design than what we have attempted so far (having 2fa as part of the payments API)

src/services/mongoose/users.ts
nicolasburtey: no reason, probably too many refactoring.

src/app/accounts/add-earn.ts
```diff
@@ -27,6 +25,8 @@ export const addEarn = async ({
   quizQuestionId: QuizQuestionId
   accountId: AccountId /* AccountId: aid validation */
 }): Promise<QuizQuestion | ApplicationError> => {
+  const rewardsConfig = getRewardsConfig()
```
dolcalmi, line 28: any reason to do this change?

dolcalmi (changes requested):
initial pass

test/integration/notifications/notification.spec.ts
```diff
@@ -49,24 +49,30 @@ describe("notification", () => {
...
 
       let usersWithDeviceTokens = 0
-      for (const { ownerId } of activeAccounts) {
-        const user = await UsersRepository().findById(ownerId)
+      for (const { kratosUserId } of activeAccounts) {
+        const user = await UsersRepository().findById(kratosUserId)
         if (user instanceof Error) throw user
 
-        if (user.deviceTokens && user.deviceTokens.length > 0) usersWithDeviceTokens++
+        if (user.deviceTokens.length > 0) usersWithDeviceTokens++
```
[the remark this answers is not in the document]
nicolasburtey: yes, should be enforced by mongoose

src/domain/authentication/index.types.d.ts
```diff
@@ -33,9 +34,15 @@ interface IAuthWithPhonePasswordlessService {
...
-    kratosUserId: KratosUserId
+    kratosUserId: UserId
     password: IdentityPassword
   }): Promise<IdentityPhone | AuthenticationError> // TODO: should be IdentityPhoneWithPassword
 }
+
+interface IIdentityRepository {
+  getIdentity(id: UserId): Promise<IdentityPhone | KratosError>
+  listIdentities(): Promise<IdentityPhone[] | KratosError>
+  slowFindByPhone(phone: PhoneNumber): Promise<IdentityPhone | KratosError>
```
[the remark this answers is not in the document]
nicolasburtey: It's currently really slow and should not be used expect in debug/cron, so I think putting that in the name for now is relevant

src/debug/migrate-user-metadata-1.ts
```diff
@@ -0,0 +1,53 @@
+/**
```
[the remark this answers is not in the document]
nicolasburtey: you are right. I did that because initially the first PR had kratos involved to store the informations. now hat we don't store anything on kratos anymore I can actually have a migration only script.

src/app/accounts/add-earn.ts
nicolasburtey: not sure anymore. 

slight performance optimization?

if we have the config loaded in the function, instead of been executed at startup time, it may speed up the launch of the app

I think the drawback is on memory as a new variable will be created everytime the function is been called.

src/graphql/admin/types/object/account.ts
dolcalmi: also graphql layer should only access app layer

src/graphql/root/mutation/user-update-username.ts
```diff
@@ -35,7 +35,9 @@ const UserUpdateUsernameMutation = GT.Field({
     if (result instanceof Error) {
       return {
         errors: [mapAndParseErrorForGqlResponse(result)],
-        ...(result instanceof UsernameIsImmutableError ? { user: domainUser } : {}),
+
+        // FIXME: what is this return for?
```
vindard, line 39: Do we still need this `FIXME` comment?

vindard (commented):
Changes look ok so far. Some checks in GitHub Actions are failing so far though

src/graphql/root/mutation/user-update-username.ts
nicolasburtey: yeah I think the line above is not used/useful, but it's probably better to delete in another segregated PR

vindard (dismissed):

vindard (approved):

src/graphql/root/mutation/user-update-username.ts
vindard: Related: https://github.com/GaloyMoney/galoy/pull/1991

PR #1871, opened 2022-10-25
fix: re-order onchain payment logic

dolcalmi:


src/app/wallets/send-on-chain.ts
```diff
@@ -397,6 +382,50 @@ const executePaymentViaOnChain = async ({
...
+    if (
+      txHash instanceof InsufficientOnChainFundsError ||
+      txHash instanceof CPFPAncestorLimitReachedError
+    ) {
+      const reverted = await ledgerService.revertOnChainPayment({
+        journalId: journal.journalId,
+      })
+      return reverted || txHash
+    }
+    if (txHash instanceof Error) {
```
vindard, line 400: Why aren't we reverting for all errors? In lightning we revert for all send errors except `LnPaymentPendingError` I think.

src/services/ledger/send.ts
```diff
@@ -114,6 +135,24 @@ export const send = {
       return new UnknownLedgerError(err)
     }
   },
+
+  revertOnChainPayment: async ({
+    journalId,
+    description = "Protocol error",
+  }: RevertOnChainPaymentArgs): Promise<void | LedgerServiceError> => {
```
vindard, line 142: Is there anything we can return here except `void` (maybe `true` or something similar). I know at one point we were trying to avoid implicit returns of void.

vindard (commented):
A couple initial comments

src/services/ledger/send.ts
dolcalmi: I used `revertLightningPayment` as template... should I change that method too?

dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/1871/commits/e23a8038437318b7475ab5a2ca00e0d5d2320ed2

src/app/wallets/send-on-chain.ts
```diff
@@ -397,6 +382,51 @@ const executePaymentViaOnChain = async ({
...
+      })
+      if (reverted instanceof Error) return reverted
+      return txHash
+    }
+    if (txHash instanceof Error) {
+      logger.error(
+        { err: txHash, address, tokens: amountToSend, success: false },
+        "Impossible to sendToChainAddress",
+      )
+      return txHash
```
vindard, line 406: Thinking about this again, I wonder if it makes sense to record a new critical exception here alongside the return. I'm thinking that if a critical alert gets thrown here it may not be immediately apparent that we should also check if the onchain payment actually failed and a manual reversion is needed.

vindard (commented):
Added a follow-up question

src/app/wallets/send-on-chain.ts
vindard: I'm just thinking also, maybe we can catch this in a new `updatePendingOnChainPayments` cronjob. We could:
- record the `txid` as a separate metadata field for successful transactions
- create a new `updatePendingOnChainPayments` method that cycles through pending onchain payments and uses their journalIds to look for onchain txns in our lnd that:
  - we can settle against if found
  - revert if not found

dolcalmi: `updatePendingOnChainPayments` is a pending task and will be added in another PR

vindard (approved):
Looks good now

PR #1885, opened 2022-10-26
chore: get loop config when it's needed instead of doing it at startup

nicolasburtey:


openoms (approved):

openoms:
I think this doesn't stop
https://github.com/GaloyMoney/galoy/blob/8dc41f0dbfeef1a257439747d26b30f2ad7c9156/src/config/process.ts#L106-L111 throwing errors.

ntheile:
this looks good. let my pull down this branch and test locally!

ntheile (approved):

PR #1735, opened 2022-10-04
chore: upgrade to note18

nicolasburtey:
need to be merged before: https://github.com/GaloyMoney/concourse-shared/pull/14

PR #1886, opened 2022-10-26
docs: update DEV.md node version

joshr4:
https://github.com/GaloyMoney/galoy/pull/1735 Updated `package.json` to require node 18, resulting in an out of date version in the dev readme. This PR bumps the dev readme version to `18.12.0`.

I'm using asdf locally to manage project specific node versions, it added a `.tool-version` file to track this. Hope thats ok to add also. https://asdf-vm.com/guide/getting-started.html#local


bodymindarts:
Thank you for your contribution. In generally we don't want to enforce a specific local toolchain like `nvm` or `asdf` and we cannot support them all either. If you want to use `.tool-version` I recommend you add it to your global `.gitignore` so that it doesn't interfere with repos.

If we did decide to support something to manage the local dev setup it would probably be something deterministic like `nix`.

bodymindarts (approved):

PR #1733, opened 2022-10-03
fix: avoid errors if Loop is not enabled

openoms:
Currently Loop is  not running on signet https://github.com/lightninglabs/loop/issues/522 so will needed to keep it off in https://github.com/GaloyMoney/charts/pull/1125. 

Fixing the error when deploying on signet with Loop disabled:
```
{"level":30,"time":1666863947270,"pid":1,"hostname":"api-97c4b8f8d-b4j5h","msg":"loading custom.yaml"}
/app/lib/config/process.js:88
        throw new error_1.ConfigError("Missing LND1_LOOP_TLS config");
        ^

ConfigError: Missing LND1_LOOP_TLS config
    at getLoopConfig (/app/lib/config/process.js:88:15)
    at Object.<anonymous> (/app/lib/app/swap/get-active-loopd.js:27:42)
    at Module._compile (node:internal/modules/cjs/loader:1159:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1213:10)
    at Module.load (node:internal/modules/cjs/loader:1037:32)
    at Module._load (node:internal/modules/cjs/loader:878:12)
    at Module.require (node:internal/modules/cjs/loader:1061:19)
    at Hook._require.Module.require (/app/node_modules/require-in-the-middle/index.js:80:39)
    at Hook._require.Module.require (/app/node_modules/require-in-the-middle/index.js:80:39)
    at Hook._require.Module.require (/app/node_modules/require-in-the-middle/index.js:80:39) {
  data: undefined
}

Node.js v18.11.0
k3d@z420:~/charts/dev$ k -n galoy-sig-galoy   logs  api-97c4b8f8d-b4j5h -f
Defaulted container "api" out of: api, wait-for-mongodb-migrate (init)
{"level":30,"time":1666863947270,"pid":1,"hostname":"api-97c4b8f8d-b4j5h","msg":"loading custom.yaml"}
/app/lib/config/process.js:88
        throw new error_1.ConfigError("Missing LND1_LOOP_TLS config");
        ^

ConfigError: Missing LND1_LOOP_TLS config
    at getLoopConfig (/app/lib/config/process.js:88:15)
    at Object.<anonymous> (/app/lib/app/swap/get-active-loopd.js:27:42)
    at Module._compile (node:internal/modules/cjs/loader:1159:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1213:10)
    at Module.load (node:internal/modules/cjs/loader:1037:32)
[...]

PR #1891, opened 2022-10-27
fix: allow loop creds to be empty if not enabled

openoms:
moved from https://github.com/GaloyMoney/galoy/pull/1733

### DIFF
--- a/src/config/process.ts
+++ b/src/config/process.ts
@@ -112,13 +112,14 @@ export const getLoopConfig = () => {
       throw new ConfigError("Missing LND1_LOOP_MACAROON config")
     if (!process.env.LND2_LOOP_MACAROON)
       throw new ConfigError("Missing LND2_LOOP_MACAROON config")
+    return {
+      lnd1LoopTls: process.env.LND1_LOOP_TLS,
+      lnd1LoopMacaroon: process.env.LND1_LOOP_MACAROON as Macaroon,
+      lnd2LoopTls: process.env.LND2_LOOP_TLS,
+      lnd2LoopMacaroon: process.env.LND2_LOOP_MACAROON as Macaroon,
+    }
   }
-  return {
-    lnd1LoopTls: process.env.LND1_LOOP_TLS,
-    lnd1LoopMacaroon: process.env.LND1_LOOP_MACAROON as Macaroon,
-    lnd2LoopTls: process.env.LND2_LOOP_TLS,
-    lnd2LoopMacaroon: process.env.LND2_LOOP_MACAROON as Macaroon,
-  }
+  throw new ConfigError("getLoopConfig() was called though swapEnabled is false")
 }
 
 export const getKratosMasterPhonePassword = () => {
--- a/src/config/process.ts
+++ b/src/config/process.ts
@@ -1,3 +1,5 @@
+import { getCronConfig } from "@config"
+
 import { ConfigError } from "./error"
 
 type TwilioConfig = {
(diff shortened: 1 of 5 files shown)


src/config/process.ts
```diff
@@ -103,12 +105,14 @@ export const LND_HEALTH_REFRESH_TIME_MS = parseInt(
 )
 
 export const getLoopConfig = () => {
-  if (!process.env.LND1_LOOP_TLS) throw new ConfigError("Missing LND1_LOOP_TLS config")
-  if (!process.env.LND2_LOOP_TLS) throw new ConfigError("Missing LND2_LOOP_TLS config")
-  if (!process.env.LND1_LOOP_MACAROON)
-    throw new ConfigError("Missing LND1_LOOP_MACAROON config")
-  if (!process.env.LND2_LOOP_MACAROON)
-    throw new ConfigError("Missing LND2_LOOP_MACAROON config")
+  if (getCronConfig().swapEnabled) {
```
openoms, line 108: > this should be part of service validation
copied from https://github.com/GaloyMoney/galoy/pull/1733#discussion_r1007103937

src/services/loopd/loop-service.ts
```diff
@@ -41,6 +43,9 @@ export const LoopService = ({
   lndInstanceName,
 }: LoopdConfig): ISwapService => {
   const mac = Buffer.from(macaroon, "base64").toString("hex") as Macaroon
+  if (!tlsCert) {
```
openoms, line 46: >probably here should be the validation, also service should not throw exceptions. Please check onchain service as example

copied from https://github.com/GaloyMoney/galoy/pull/1733#discussion_r1007112288

src/config/process.ts
ntheile: i did not think we were allowed to call process.env from any other file other than `process.ts`?

src/services/loopd/loop-service.ts
bodymindarts: This should not ever happen - either config is set and `tlsCert` is never null or it is not set and this code should never be called see https://github.com/GaloyMoney/galoy/pull/1891/commits/eab6cd41a50e4af67934fbd277726dff945fad85

bodymindarts:
I have added a commit https://github.com/GaloyMoney/galoy/pull/1891/commits/eab6cd41a50e4af67934fbd277726dff945fad85 that throws in the config if we try to access the loop config in the event that `enabledSwap = false`... it should never be called so we can flag the error right there.

ntheile:
> 

Yeah this commit looks good and makes sense. 

I am still having trouble getting tests to run locally when testing this change, similar error to the e2e test ` TypeError: Cannot read properties of undefined (reading 'tracingServiceName')`

openoms:
> I am still having trouble getting tests to run locally when testing this change, similar error to the e2e test ` TypeError: Cannot read properties of undefined (reading 'tracingServiceName')`

Not sure why the `tracingConfig` came in the picture here, but [ce26a65](https://github.com/GaloyMoney/galoy/pull/1891/commits/ce26a65a66c7ba9b8294f09611294aa5f368ce8e) lets the tests pass.

ntheile:
i tested it and it works locally now. i would defer to @bodymindarts about the tracing config

bodymindarts (approved):

ntheile:
I created a custom docker image of this commit and it works with signet on my local dev charts node! Here is the config I set:

dev/galoy/galoy-signet-values.yml
```
galoy:
  network: signet
  images:
    ## Custom Test Galoy Application Image details
    app:
      repository: ntheile/galoy-test:0.0.9
      digest: "sha256:622984b81ad44d730666f1b95379978bf24156c540ec39e60ca1f9361db5dc8e"
      git_ref: "622984b"

  config:
    test_accounts:
      - ref: A
        phone: "INSERT_PHONE"
        code: "INSERT_CODE"
        role: "bankowner"
        username: "bankowner"
        ref: "bankowner"
        needUsdWallet: true
    apollo:
      playground: true
    cronConfig:
      swapEnabled: false

     ....
```

ntheile (approved):

src/config/process.ts
openoms: I think we are beck to the original problem with https://github.com/GaloyMoney/galoy/commit/eab6cd41a50e4af67934fbd277726dff945fad85
since the process.ts is called every time and will throw the error when `getLoopConfig` is called.

PR #1893, opened 2022-10-27
chore: add global attributes to spans

UncleSamtoshi:
- moves session context logic out of ApolloServer and into middleware so that all graphql spans will be descendants that inherit session context attributes
- add session attributes to top level subscription fields because they do not execute as children of the express middleware spans


dolcalmi (approved):
Approving but there is a code scanning alert. please check before merge

PR #1894, opened 2022-10-28
chore: add function arg redaction in spans

UncleSamtoshi:
- enable redaction of function arguments when creating span wrapped functions

### DIFF
--- a/src/services/tracing.ts
+++ b/src/services/tracing.ts
@@ -328,12 +328,14 @@ const resolveFunctionSpanOptions = ({
   functionArgs,
   spanAttributes,
   root,
+  redactFnArgs,
 }: {
   namespace: string
   functionName: string
   functionArgs: Array<unknown>
   spanAttributes?: Attributes
   root?: boolean
+  redactFnArgs?: boolean
 }): SpanOptions => {
   const attributes = {
     [SemanticAttributes.CODE_FUNCTION]: functionName,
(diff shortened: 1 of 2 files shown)


src/services/tracing.ts
```diff
@@ -346,7 +348,9 @@ const resolveFunctionSpanOptions = ({
     for (const key in params) {
       // @ts-ignore-next-line no-implicit-any error
       const value = params[key]
-      attributes[`${SemanticAttributes.CODE_FUNCTION}.params.${key}`] = value
+      attributes[`${SemanticAttributes.CODE_FUNCTION}.params.${key}`] = redactFnArgs
```
dolcalmi, line 351: I dont think this should be the solution... there is a config for redacted params.. also this is not flexible enough to have the non-sensitive params

dolcalmi (changes requested):

src/services/cache/local-cache.ts
```diff
@@ -63,5 +63,6 @@ export const LocalCacheService = (): ICacheService => {
   return wrapAsyncFunctionsToRunInSpan({
     namespace: "services.cache.local",
     fns: { set, get, getOrSet, clear },
+    redactFnArgs: true,
```
dolcalmi, line 66: this is not good... why redact all feature that use cache?

dolcalmi (changes requested):

src/services/cache/local-cache.ts
UncleSamtoshi: Using our previous method of redaction, we would have to redact `code.function.param.key` as well as `code.function.param.value` for all features that use the cache anyway.

UncleSamtoshi: In addition to changing the api of the cache service to change single parameter functions into object parameters. For example changing `get(key)` to `get({key})`.

dolcalmi: cache interface probably was implemented previous the the params rule so make sense change the interface to use named params. Also since this is for the temp hack is easier add/remove redacted to config files

UncleSamtoshi:
Moved to pr #1895

PR #1907, opened 2022-11-01
fix: avoid error because undefinied tokenPayload

dolcalmi:


UncleSamtoshi (dismissed):

daviroo (dismissed):

src/servers/graphql-server.ts
```diff
@@ -84,7 +84,6 @@ const setGqlContext = async (
   res: Response,
   next: NextFunction,
 ): Promise<void> => {
-  console.log(req)
```
UncleSamtoshi, line 87: Add a line here `if (!tokenPayload) next()`

UncleSamtoshi (approved):

src/servers/graphql-server.ts
dolcalmi: this will not record ip in span and also could not pass  logger and other values to public queries/muttions

PR #1916, opened 2022-11-06
refactor: trigger onchain tx event handler

dolcalmi:
Update onchain receipt logic was larger than expected so this is the first split https://github.com/GaloyMoney/galoy/pull/1871#discussion_r1007301963

This PR includes:

- Refactor trigger onchainTransactionEventHandler to use only app layer
- Fix notifications to include amounts for each UTXO (pending to do because was sending tx amount)
- Fix issue with pending incoming txs (didn't include fee discount)
- Small refactor in onchain service to include named params + description (label)
- Fix tests.


src/app/wallets/update-on-chain-payments.ts
```diff
@@ -0,0 +1,113 @@
...
+  if (senderWallet instanceof Error) return senderWallet
+
+  const senderAccount = await AccountsRepository().findById(senderWallet.accountId)
+  if (senderAccount instanceof Error) return senderAccount
+
+  const senderUser = await UsersRepository().findById(senderAccount.ownerId)
+  if (senderUser instanceof Error) return senderUser
+
+  const walletTx = translateLedgerTxnToWalletTxn(payment)
+  const paymentAmount = paymentAmountFromNumber({
```
vindard, line 84: Just making a note, we'll eventually get these directly from new `satsAmount` & `displayAmount` fields in PR #1408

src/services/lnd/onchain-service.ts
```diff
@@ -136,13 +140,39 @@ export const OnChainService = (
     txHash,
     scanDepth,
   }: LookupOnChainFeeArgs): Promise<Satoshis | OnChainServiceError> => {
-    const onChainTxs = await listOutgoingTransactions(scanDepth)
+    const onChainTxs = await listOutgoingTransactions({ scanDepth })
```
vindard, line 143: Can we call `lookupOnChainPayment` here instead and simply pass forward `args` to be more direct?

src/services/lnd/onchain-service.ts
```diff
@@ -208,7 +238,10 @@ export const OnChainService = (
       getBalance,
       getPendingBalance,
       listIncomingTransactions,
+      listOutgoingTransactions,
```
vindard, line 241: Why add this new export?

src/app/wallets/notify-on-chain-receipt-unconfirmed.ts
```diff
@@ -0,0 +1,105 @@
...
+}): Promise<void | ApplicationError> => {
+  const displayCurrencyPerSat = await getCurrentPrice()
+  if (displayCurrencyPerSat instanceof Error) return displayCurrencyPerSat
+
+  const converter = DisplayCurrencyConverter(displayCurrencyPerSat)
+  const account = await AccountsRepository().findById(wallet.accountId)
+  if (account instanceof Error) return account
+
+  const walletAddresses = wallet.onChainAddresses()
+  for (const { sats, address } of tx.rawTx.outs) {
```
vindard, line 75: Just noting that a hypothetical spam side-effect here is recipient could have something happen like receiving multiple notifications for a single transaction if someone for e.g. made a txn with 100 utxos all with some small sat amount paid to the same address.

It's not too likely because there is a cost attached, but it's still good to be aware of.

src/domain/bitcoin/onchain/index.types.d.ts
```diff
@@ -79,22 +86,47 @@ type IncomingOnChainTxHandler = {
...
-  listIncomingTransactions(
-    scanDepth: ScanDepth,
-  ): Promise<IncomingOnChainTransaction[] | OnChainServiceError>
+  listIncomingTransactions({
+    scanDepth,
+  }: ListIncomingTransactionsArgs): Promise<
+    IncomingOnChainTransaction[] | OnChainServiceError
+  >
+
+  listOutgoingTransactions({
```
vindard, line 109: Same comment, why new export (didn't find it used anywhere outside package)

src/app/wallets/update-on-chain-payments.ts
```diff
@@ -0,0 +1,113 @@
...
+  txHash: OnChainTxHash
+  scanDepth?: ScanDepth
+  logger: Logger
+}): Promise<true | ApplicationError> => {
+  const onChainService = OnChainService(TxDecoder(BTC_NETWORK))
+  if (onChainService instanceof Error) return onChainService
+
+  const tx = await onChainService.lookupOnChainPayment({ txHash, scanDepth })
+  if (tx instanceof Error) return tx
+  if (tx.confirmations === 0) return new UnconfirmedOnChainTxError()
```
vindard, line 33: Should we leave this comment in here somewhere still, or signal explicitly in the code somewhere that we don't know the txid? Is it still a problem that trigger can happen before we know the txid?
```
      // we have to return here because we will not know whose user the the txid belong to
      // this is because of limitation for lnd onchain wallet. we only know the txid after the
      // transaction has been sent. and this events is trigger before

vindard (commented):
Added some initial comments. Just have a few more files to review still.

vindard (commented):
Ok I finished the reviewing, rest looks good. So it's just those original comments

src/services/lnd/onchain-service.ts
dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/1916/commits/ccf62bb393a1e6e0541e9fdb7860869cd3063813

src/services/lnd/onchain-service.ts
dolcalmi: this is part of the second PR, the reason is because this is a split

src/app/wallets/notify-on-chain-receipt-unconfirmed.ts
dolcalmi: uhmm yep but still we need to do it in this way for final settlement but... you are right, there is a cost to do that so I think is not a problem

src/domain/bitcoin/onchain/index.types.d.ts
dolcalmi: this is part of the second PR, wil be used by cron

src/app/wallets/update-on-chain-payments.ts
dolcalmi: Added https://github.com/GaloyMoney/galoy/pull/1916/commits/b7e61a2c458d234872e57443603ea7bbedef45d8

vindard (approved):

PR #1917, opened 2022-11-07
chore: extend all sessions

nicolasburtey:


bodymindarts:
I think we can close this - the issue was misconfiguration in charts.

PR #1927, opened 2022-11-09
wip: dataloader

samerbuna:


nicolasburtey:
@samerbuna will the data loader solve the pagination issue for transactions/stablesats?

samerbuna:
> @samerbuna will the data loader solve the pagination issue for transactions/stablesats?

No. But the solution for that will be different with dataloader, and a bit easier to implement.

nicolasburtey:
> > @samerbuna will the data loader solve the pagination issue for transactions/stablesats?
> 
> No. But the solution for that will be different with dataloader, and a bit easier to implement.

I think pagination might be a much more urgent issue than dataloader. 

I don't think dataloader solves much problem in production right now. the only use-case I see for it is for the pre-image because it's store in a separate collection than medici, but I also think this is solvable at the ledger level eventually

dolcalmi:
@vindard can this be closed in favor of https://github.com/GaloyMoney/galoy/pull/2252 ?

vindard:
> @vindard can this be closed in favor of https://github.com/GaloyMoney/galoy/pull/2252 ?

Yea I believe this was just a reference PR (can always reopen if we need it again)

PR #1942, opened 2022-11-25
refactor: map missing gql errors

vindard:
## Description

This PR is an exercise in reviewing all our thrown/returned errors in our graphql layer and ensuring they are properly mapped using our `mapAndParseErrorForGqlResponse` method. This method is important because it adds the `code` field to the gql response that we now use in our honeycomb alerting to filter things.

The results of this PR are to:
- avoid unnecessary `missing-error-code` alerts (false positives from unmapped errors)
- for mutations, return the correct error message in the selection field consistently from our `error-map` switch to the end user
- for queries, throw the appropriate error at the top level of the response

**This PR was prompted by the alert that led to the fix in #1935**

### DIFF
--- a/src/domain/accounts/errors.ts
+++ b/src/domain/accounts/errors.ts
@@ -5,4 +5,6 @@ export class AccountError extends DomainError {}
 export class UsernameNotAvailableError extends AccountError {}
 export class UsernameIsImmutableError extends AccountError {}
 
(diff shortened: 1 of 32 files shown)


nicolasburtey:
would be a good code review topic

nicolasburtey (dismissed):

nicolasburtey (approved):

Issue #1611, opened 2022-08-31
expose current limit on the API


nicolasburtey:
it's currently not possible to know how much limit is being left for withdraw or trade. there should be an API for that, so that frontend apps could display this values, and even have some validation and warning before attempting a payment that is over limits.

dolcalmi:
Added in accounts https://github.com/GaloyMoney/galoy/blob/main/src/graphql/main/schema.graphql#L6

PR #1953, opened 2022-11-28
feat: add transactions limit/volumes api endpoints

vindard:
## Description

This PR exposes new endpoints for user limits across the following limits:
- intraledger
- withdrawal
- conversion

_**Note: this was implemented so that each new limit selection has its own separate resolver. Each `totalLimit` and `remainingLimit` selection is separately executed.**_

Sample query:
```gql
query me {
  me {
    defaultAccount {
      limits {
        withdrawal {
          ...on OneDayAccountLimit {
            __typename
            totalLimit
            remainingLimit
            interval
          }
        }
        internalSend {
          __typename
          totalLimit 
          remainingLimit
          interval
        }
        convert {
          __typename
          totalLimit
          remainingLimit
          interval
        }
      }
    }
  }
}
```

Result from sample query:
```json
{
  "data": {
    "me": {
      "defaultAccount": {
        "limits": {
          "withdrawal": [
            {
              "__typename": "OneDayAccountLimitRolling",
              "totalLimit": 100000,
              "remainingLimit": 100000,
              "interval": 86400
            }
          ],
          "internalSend": [
            {
              "__typename": "OneDayAccountLimitRolling",
              "totalLimit": 200000,
              "remainingLimit": 200000,
              "interval": 86400
            }
          ],
          "convert": [
            {
              "__typename": "OneDayAccountLimitRolling",
              "totalLimit": 5000000,
              "remainingLimit": 5000000,
              "interval": 86400
            }
          ]
        }
      }
    }
  }
}
```

#### Open `TODO`s
- [x] get `schema.gql` to reflect updated schema
- [x] Test properly (initial manual tests on local API were failing with `PriceNotAvailableError` error)
- [x] Fix volumes to check all wallets for a given account (fixed https://github.com/GaloyMoney/galoy/pull/1973 and rebased in)
- [x] Try to figure out how to pass `IAccountLimit` interface instead of Object instances (solution: include objects in `ALL_INTERFACE_TYPES`)
- [x] Finalize new selection names chosen (implemented `option1` from [this repl](https://github.com/GaloyMoney/galoy/pull/1973))
- [x] Rename limit types from `DailyWithdrawal...` to `OneDayWithdrawal...` (see https://github.com/GaloyMoney/galoy/pull/1953#issuecomment-1357689726, done: https://github.com/GaloyMoney/galoy/pull/1953/commits/5f338dd379a53242df20b7f45038740b312b7a91)
- [x] Add an add'l  nesting level for `rolling` limits (see https://github.com/GaloyMoney/galoy/pull/1953#issuecomment-1357689726, done: https://github.com/GaloyMoney/galoy/pull/1953/commits/cb4f1789a8f6ad45f0b2d8453a42a214316194c6)

---

## `TODO` in follow-up PRs
- `sort` optional param

---

## Notes

Complete in a follow-up PR, cleaning up the `limitName` args in volume/check functions like this https://github.com/GaloyMoney/galoy/compare/a3583194056d0ac9151b0e33e74c8e397e31e472...0deae6bbe2d86c7ce10d1d3d21fa3617e83f984f.

This failed on the first attempt because for some reason the `getAccountLimits` mock in 02-send-lightning-limits didn't work which broke tests.

Card here: https://github.com/GaloyMoney/galoy/projects/9#card-87343623


UncleSamtoshi:
One other comment with this implementation. By having each of the limits be a named field instead of all the limits being a list, we are preventing the app from dynamically changing the limits displayed. In the future it may introduce a migration period if/when the limit types are changed. I still think this is currently the right approach because the limits have different use-cases and we will want to select them individually, however I think before we implement the limits page on the app we may want to be set on our limits schema.

vindard:
> * We may eventually have the same limit type enforced on different timelines (1 month Withdrawl limit, 1 day withdrawal limit). For now, I think it would be good to include the limit duration as a property on the AccountLimit object.

Added this to the top-level `limits` selection. We can shift this around to wherever we think it's best to live: https://github.com/GaloyMoney/galoy/pull/1953/commits/0f31c95e4f905c86a5781d899e3c93199f13210d

vindard:
Some additional considerations I've been thinking about that I'd like to apply here.

### 1. Rename limit types
Rename  from `DailyAccountLimit`/`WeeklyAccountLimit` etc. to `OneDayAccountLimit`/`OneWeekAccountLimit` etc.

I've been thinking that this could be an easier standard to extend from if for some reason in the future we wanted things like `FiveDay`/`TwoWeek`/`ThreeMonth` limits.

### 2. Add a level of nesting for `fixedStart` vs `rolling` limits

Our limits right now are applied as a rolling window but we've discussed that at some point we might want to have limits that apply for a fixed time period.

Having a level of nesting under `limits` now for this would also allow us to create two separate objects to express this

- `rolling` object, has selections:
  - `totalLimit`
  - `remainingLimit`
  - `interval`: which is the amount of time the limit applies for in `Seconds`

- `fixedStart` object, has selections:
  - `totalLimit`
  - `remainingLimit`
  - `startTime`: unix timestamp
  - `duration`: Seconds
  - `elapsedTime`: Seconds
  - (optional) `remainingTime`: Seconds

### 3. (Optional) Adding a `sort` param

@Extheoisah has a question about if we can apply any ordering to the list of limits returned. I believe in graphql this is usually encoded with some kind of a `sort` param that takes a selection name (or multiple?) to sort by and an order (`ASC` or `DESC`).

Is this something we should include at the API level, or is it something we should let the client handle?

daviroo:
> ### 1. Rename limit types
> Rename from `DailyAccountLimit`/`WeeklyAccountLimit` etc. to `OneDayAccountLimit`/`OneWeekAccountLimit` etc.
> 
> I've been thinking that this could be an easier standard to extend from if for some reason in the future we wanted things like `FiveDay`/`TwoWeek`/`ThreeMonth` limits.

Seems fine to me.

> ### 2. Add a level of nesting for `fixedStart` vs `rolling` limits
> Our limits right now are applied as a rolling window but we've discussed that at some point we might want to have limits that apply for a fixed time period.

I think it's fine to add an extra nested level for now if we think we might extend this in the future.  However we should hold off implementing the fixed start object until we come to use them.

> ### 3. (Optional) Adding a `sort` param
> There was [a question](https://github.com/GaloyMoney/galoy-mobile/pull/727#discussion_r1051679430) about if we can apply any ordering to the list of limits returned. I believe in graphql this is usually encoded with some kind of a `sort` param that takes a selection name (or multiple?) to sort by and an order (`ASC` or `DESC`).
> 
> Is this something we should include at the API level, or is it something we should let the client handle?

I think @Extheoisah has implemented a sort on the front end but this would be a good pattern to use going forward I think.

src/app/accounts/account-limit.ts
```diff
@@ -0,0 +1,81 @@
...
+import { AccountLimitsVolumes } from "@domain/accounts/limits-volume"
+import { InvalidAccountLimitTypeError } from "@domain/errors"
+import { LedgerService } from "@services/ledger"
+
+export const accountLimit = async ({
+  account,
+  limitType,
+}: {
+  account: Account
+  limitType: "Intraledger" | "Withdrawal" | "TradeIntraAccount"
```
bodymindarts, line 13: looks like `LimitType` should be a galoy style enum with hard coded values.
```
let const LimitType = {
  IntraLedger: "IntraLedger",
  ...
} as const
```

don't know if ^^ that is the exact syntax we've been using - but I'm sure you know what I'm referring to. Then we derive a type and reference `LimitType.IntraLedger` instead of hard coded strings everywhere.

src/app/payments/helpers.ts
```diff
@@ -164,6 +163,26 @@ const checkLimitsBase = async ({
     (vol): vol is TxBaseVolumeAmount<WalletCurrency> => true,
   )
 
+  return walletVolumes
+}
+
+const checkLimitsBase = async ({
```
bodymindarts, line 169: Can we avoid the word `Base`? This is not OO - hence it's not a precise term. Can easily be misinterpreted.

src/domain/accounts/limits-volume.ts
```diff
@@ -0,0 +1,145 @@
...
+const calc = AmountCalculator()
+
+export const calculateLimitsInUsd = async ({
+  limitName,
+  limitAmount,
+  priceRatio,
+
+  walletVolumes,
+}: {
+  limitName:
```
bodymindarts, line 18: Anther galoy - enum. Don't like the hard coded Strings.

src/domain/accounts/primitives.ts
```diff
@@ -9,3 +9,11 @@ export const AccountStatus = {
   Active: "active",
   Locked: "locked",
 } as const
+
+export const AccountLimitsRange = {
```
bodymindarts, line 13: Yeah like this... everywhere you are hard coding strings that convey type information should be wrapped in something like this...

Oh and can we delete the comments? I think we can figure out how to add another value here without them.

bodymindarts (commented):
Looks good overall... just need to `const`-ify the magic strings and think of a better name than `base` IMO.

src/app/accounts/account-limit.ts
vindard: https://github.com/GaloyMoney/galoy/pull/1953/commits/e82fb51f3f67e7cef82fe468968876c57d16c8d8

src/app/payments/helpers.ts
vindard: Fixed: https://github.com/GaloyMoney/galoy/pull/1953/commits/45dd3d0399cc8aa37a5d7036993c864903c5e7ed

src/domain/accounts/limits-volume.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1953/commits/934dc92450970208dd31a69aa9fed0797b822c61

I tried to do a bigger cleanup around this here (https://github.com/GaloyMoney/galoy/compare/a3583194056d0ac9151b0e33e74c8e397e31e472...0deae6bbe2d86c7ce10d1d3d21fa3617e83f984f) but the mocks for the limits test inexplicably broke.

src/domain/accounts/primitives.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/1953/commits/a3583194056d0ac9151b0e33e74c8e397e31e472

src/app/accounts/account-limit.ts
```diff
@@ -0,0 +1,82 @@
+import { volumesForAccountId } from "@app/payments/helpers"
+import { getMidPriceRatio } from "@app/shared"
+import { getAccountLimits, getDealerConfig, ONE_DAY } from "@config"
+import { AccountLimitsType } from "@domain/accounts"
+import { AccountLimitsVolumes } from "@domain/accounts/limits-volume"
+import { InvalidAccountLimitTypeError } from "@domain/errors"
+import { LedgerService } from "@services/ledger"
+
+export const accountLimit = async ({
```
nicolasburtey, line 9:
```suggestion
export const remainingLimit = async ({
```

maybe this would be more appropriate?

src/graphql/types/object/one-day-account-limit.ts
```diff
@@ -0,0 +1,58 @@
...
+      description: `The amount of cents remaining below the limit for the current 24 hour period.`,
+      resolve: async (source) => {
+        const { account, limitType } = source
+        const volumes = await Accounts.accountLimit({
+          account,
+          limitType,
+        })
+        if (volumes instanceof Error) throw mapError(volumes)
+
+        return normalizePaymentAmount(volumes.volumeRemaining).amount
```
nicolasburtey, line 47: the use case either return all the limits (`accountLimit`) or just the absolute limit (`accountLimitFromConfig`). In the first case we're using only one property. Maybe not a big deal but this seems off.

nicolasburtey (dismissed):
some comments but can be responded or addressed in a later PR.

one thing that will be needed in a future PR is the limit on `deposit`, specifically for level0 accounts.

nicolasburtey:
this should close #1611

nicolasburtey (approved):

PR #1963, opened 2022-11-30
kratos: cleanup 4

nicolasburtey:


vindard (commented):
Should we delete these from the mongoose schema as well [here](https://github.com/GaloyMoney/galoy/blob/eecea6cb8c4f61ba19b6a1b0bffee21d165d3fe9/src/services/mongoose/schema.ts#L291)?

nicolasburtey:
> Should we delete these from the mongoose schema as well [here](https://github.com/GaloyMoney/galoy/blob/eecea6cb8c4f61ba19b6a1b0bffee21d165d3fe9/src/services/mongoose/schema.ts#L291), or is that for a follow-up PR?

good point. better to do it in the same PR

vindard (approved):

PR #1973, opened 2022-12-02
feat: adjust account volumes

vindard:
## Description

This PR does a broader check on account volumes before applying limits.

### DIFF
--- a/src/app/payments/helpers.ts
+++ b/src/app/payments/helpers.ts
@@ -130,6 +130,44 @@ const recipientDetailsFromInvoice = async <R extends WalletCurrency>(
   }
 }
 
+const checkLimitsBase = async <S extends WalletCurrency>({
+  wallet,
+  volumeAmountSinceFn,
+}: {
+  wallet: WalletDescriptor<S>
+  volumeAmountSinceFn: GetVolumeAmountSinceFn
+}) => {
+  const timestamp1Day = timestampDaysAgo(ONE_DAY)
+  if (timestamp1Day instanceof Error) return timestamp1Day
+
+  const account = await AccountsRepository().findById(wallet.accountId)
+  if (account instanceof Error) return account
+
+  const wallets = await WalletsRepository().listByAccountId(wallet.accountId)
+  if (wallets instanceof Error) return wallets
+
+  const walletVolumesWithErrors = await Promise.all(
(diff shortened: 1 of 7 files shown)


src/app/payments/helpers.ts
```diff
@@ -130,6 +130,44 @@ const recipientDetailsFromInvoice = async <R extends WalletCurrency>(
   }
 }
 
+const checkLimitsBase = async <S extends WalletCurrency>({
```
dolcalmi, line 133: if this is to check account volumes then we should pass account id or account. Also, this is the problem with "helpers" now we keep adding new functions and increasing the complexity of a file that give no context

dolcalmi (changes requested):

src/app/payments/helpers.ts
vindard: Good catch, I changed the arg to `accountId` here: https://github.com/GaloyMoney/galoy/pull/1973/commits/6062aa743c33fec0d2176da8abac9745bce8ed24

Good point on the helpers thing too. I think these can probably be their own use-case module now. I can change that in the next PR #1953 where I'm already adding new use-cases and doing some refactoring

src/domain/accounts/new-limits-checker.ts
```diff
@@ -4,9 +4,32 @@ import {
...
+    const outgoingUsdAmount =
+      walletVolume.outgoingBaseAmount.currency === WalletCurrency.Btc
+        ? await priceRatio.convertFromBtc(
+            walletVolume.outgoingBaseAmount as BtcPaymentAmount,
+          )
+        : (walletVolume.outgoingBaseAmount as UsdPaymentAmount)
+    outgoingBaseAmounts.push(outgoingUsdAmount)
+  }
+
+  return calc.addList(outgoingBaseAmounts)
```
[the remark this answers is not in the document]
vindard: Oops you're right, not sure how I ended up making this that complicated 😅  ... simplified here: https://github.com/GaloyMoney/galoy/pull/1973/commits/6a0a9287e45693dadaa01a1e4c2c68ae60d1781f & https://github.com/GaloyMoney/galoy/pull/1973/commits/b0e3e37652410a8fab57c85a1d8e2a52a15cedf0

dolcalmi (approved):

PR #1975, opened 2022-12-03
feature: integrate kratos webhook

nicolasburtey:
before this PR, the flow for creating an Account was the following:

- client send an API call to graphql login endpoint
- api backend would create the kratos Identity and then the Account

after this PR:
- client send an API call to graphql login endpoint
- api backend create the kratos Identity
- kratos callback the backend (`kratos/registration` endpoint) and create the Account

the callback make it possible to eventually create an identity directly to kratos for other user case, like email+password or username+password (that would not be rate limited thought)


ntheile (dismissed):

dev/ory/kratos.yml
```diff
@@ -77,11 +77,23 @@ selfservice:
       after:
         password:
           hooks:
+            - hook: web_hook
+              config:
+                url: http://e2e-tests:4012/kratos/registration
```
nicolasburtey, line 82: this needs to point out to the api server instead of `e2e-tests`

dev/ory/kratos.yml
```diff
@@ -77,11 +77,23 @@ selfservice:
...
+                method: POST
+                can_interrupt: true
+                body: file:///home/ory/body.jsonnet # TODO: use a base64 encoding instead
```
nicolasburtey, line 85: this should include what is in https://github.com/GaloyMoney/galoy/blob/main/dev/ory/body.jsonnet

dev/ory/kratos.yml
```diff
@@ -77,11 +77,23 @@ selfservice:
...
+                auth:
+                  type: api_key
+                  config:
+                    name: Authorization
+                    value: The-Value-of-My-Key
```
nicolasburtey, line 90: we should create an unique key to authenticate the request. this key should also be passed to the api backend.

dev/ory/kratos.yml
krtk6160: Does this need to be the public endpoint or the k8s internal service endpoint of the api server?

nicolasburtey: should be the internal endpoint only. 

this endpoint should not be publicly accessible

nicolasburtey: although probably a big deal if this is (for now) because it's protected by a key

krtk6160: Will the pod that connects to this endpoint be in the same namespace as the api server? We can hardcode it as `api:4012/kratos/registration` if so, no templating needed. Otherwise will have to include the fqdn - `api.namespace.svc.cluster.local:4012/kratos/registration` and that will require templating.

nicolasburtey: kratos will connect to it. afaik it's in the same namespace?

PR #1978, opened 2022-12-05
chore: update migration

nicolasburtey:


src/migrations/2022120510370802-fix-migration.ts
```diff
@@ -0,0 +1,41 @@
+module.exports = {
+  async up(db) {
+    console.log("Begin account to users property migration")
+
+    const users = db.collection("users").find({ id: { $exists: true } })
+
+    let progress = 0
+    for await (const user of users) {
```
dolcalmi, line 8: should we include a delete routine for users created without phone?

src/migrations/2022120510370802-fix-migration.ts
```diff
@@ -0,0 +1,41 @@
...
+      const userId = user.id
+      const phoneMetadata = user.phoneMetadata
+      const language = user.language ?? ""
+      const deviceTokens = user.deviceTokens ?? []
+      const phone = user.phone
+      const createdAt = user.createdAt
+
+      await db.collection("users").updateOne(
+        {
+          userId,
```
dolcalmi, line 20: this will duplicate users because in this case there is no userId, we need to update by id and sent userId too

dolcalmi (changes requested):

PR #1984, opened 2022-12-06
ci(shared): bump vendored ci files

galoybot:
# Bump Shared Tasks

This PR syncs in this repository, shared CI tasks from [concourse-shared](https://github.com/GaloyMoney/concourse-shared).


nicolasburtey:
@sandipndev @krtk6160 can you eventually look into this and fix it?

.github/workflows/nodejs-check-code.yml
```diff
@@ -1,18 +0,0 @@
...
-
-jobs:
-  check-code:
-    name: Nodejs Check Code
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
-        with:
-          node-version: 18
```
sandipndev, line 15: The PR adds `check-code` but not with this node versions.
Since concourse shared syncs across all nodejs repos together, therefore only if all of our nodejs repos support node version 18, then we can make a change there and it should fix the failing tests.

.github/workflows/spelling.yml
```diff
@@ -1,16 +1,16 @@
...
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout Actions Repository
-        uses: actions/checkout@v3
-      - name: Spell Check Repo
-        uses: crate-ci/typos@master
-        with:
-          config: typos.toml
+    - name: Checkout Actions Repository
+      uses: actions/checkout@v2
```
sandipndev, line 12: concourse shared needs to be upgraded to use actions/checkout@v3

sandipndev (changes requested):

PR #1985, opened 2022-12-06
feat: implement database pagination for transactions connections

samerbuna:


src/services/ledger/paginated-ledger.ts
```diff
@@ -0,0 +1,73 @@
...
+import { IAnyObject } from "medici"
+
+import { transactionModel } from "medici/build/models/transaction"
+import { parseFilterQuery } from "medici/build/helper/parse/parseFilterQuery"
+
+import { MainBook } from "./books"
+
+export const DEFAULT_MAX_CONNECTION_LIMIT = 100
+
+interface ITransaction {
```
dolcalmi, line 17: please remove this and use ILedgerTransaction https://github.com/GaloyMoney/galoy/blob/main/src/services/ledger/schema.types.d.ts#L1

src/services/ledger/paginated-ledger.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  if (after) {
+    filterQuery["_id"] = { $lt: new Types.ObjectId(after) }
+  }
+
+  if (before) {
+    filterQuery["_id"] = { $gt: new Types.ObjectId(before) }
+  }
+
+  return transactionModel.collection
```
dolcalmi, line 63: please dont use transactionModel, replace by our Transaction Model https://github.com/GaloyMoney/galoy/blob/main/src/services/ledger/schema.ts#L142

src/services/ledger/paginated-ledger.ts
```diff
@@ -0,0 +1,56 @@
...
+    filterQuery["_id"] = { $lt: new Types.ObjectId(after) }
+  }
+
+  if (before) {
+    filterQuery["_id"] = { $gt: new Types.ObjectId(before) }
+  }
+
+  return Transaction.collection
+    .find(filterQuery, {
+      limit: DEFAULT_MAX_CONNECTION_LIMIT,
```
dolcalmi, line 48: not sure how this param interacts with `first` and `last` params but for eg stablestats use 200 https://github.com/GaloyMoney/stablesats-rs/blob/main/galoy-client/src/client/mod.rs#L182 should we change this or is it overrided by that params?

samerbuna: This makes the default max limit a 100. If you specify more than that in `first` or `last`, the request will error out.

Should we make the default max limit 200? @bodymindarts?

bodymindarts: I think 100 is fine as a default. Seems s

dolcalmi (approved):

PR #1998, opened 2022-12-09
feat: mobile e2e test on concourse ci

ntheile:


ci/pipeline.yml
```diff
@@ -444,6 +469,12 @@ resources:
   -  #@ deps_version_resource()
   -  #@ bundled_deps_resource()
   -  #@ slack_resource()
+  - name: galoy-mobile-repo
+    type: git
+    icon: github
+    source:
+      uri: https://github.com/GaloyMoney/galoy-mobile
+      branch: main
```
ntheile, line 477: @bodymindarts This is probably supposed to be injected in. Do we have a common shared repo for github resources?

ci/pipeline.yml
```diff
@@ -436,6 +436,31 @@ jobs:
           run:
             path: pipeline-tasks/ci/tasks/open-bump-proto-pr.sh
 
+  - name: test-android
+    plan:
+      - in_parallel:
+          - { get: galoy-mobile-repo, passed: "release", trigger: true }
```
ntheile, line 442: @bodymindarts Is this the correct spot to trigger the mobile e2e tests from when "release" passes?

PR #2014, opened 2022-12-14
chore: cookie auth testing

nicolasburtey:


test/e2e/integration-external-services/kratos.spec.ts
```diff
@@ -266,6 +266,56 @@ describe("session revokation", () => {
...
+    const isSession1Revoked = activeSessions.data.find((s) => s.id === session1Id)
+    // session2Id should NOT be in the list
+    const isSession2Revoked = activeSessions.data.find((s) => s.id === session2Id)
+    expect(isSession1Revoked).toBeDefined()
+    expect(isSession2Revoked).toBeUndefined()
+
+    // * validateKratosToken has a weird bug with multiple sessions
+    //  it throws an error on session1 and thinks session2 is valid
+    //  this is the opposite of what should happen
+    const isSession1Valid = await validateKratosToken(session1Token) // * this should be valid (but its the opposite)
```
ntheile, line 313: @nicolasburtey Have you noticed weird results from the `validateKratosToken` specifically the call to kratosPublic.toSession({ xSessionToken: sessionToken }) ? It seems to not handle multiple sessions properly. If you run this test the intention is that session2 should be revoked and therefore the token invalid. But the opposite happens. It shows session1 is valid and session2 is invalid. Also when `validateKratosToken(session2Token)` is called it returned the wrong sessionId (is returns session1).

Any ideas?

ntheile:
related session issue opened here https://github.com/ory/kratos/issues/2921#issuecomment-1357031795

ntheile:
@nicolasburtey ,

Here are some initial ideas on how to get cookies to work for the web wallet and admin panel. I am initially just testing on the some origin domain localhost:4002.

If you run the `make start-main-fast` it will boot up the main graphql server with two new endpoints:

- http://localhost:4002/cookieLoginForm
- http://localhost:4002/cookieLogin

The login flow is as follows:
1. If you open http://localhost:4002/cookieLoginForm it will ask for `Phone` and `code`
2. Hit login and it will post back to /cookieLogin and kick up the `loginWithPhoneReturnCookie` flow
https://github.com/GaloyMoney/galoy/blob/317df824a2b6157f5b792b46b3dbdd25acd72c21/src/app/auth/login.ts#L97-L134
3. Then we pass back those cookies to the client https://github.com/GaloyMoney/galoy/blob/317df824a2b6157f5b792b46b3dbdd25acd72c21/src/servers/graphql-server.ts#L330-L351
4. In the client on `http://localhost:4002/cookieLogin` you will see the two new cookies `ory_session_cookie` and `csrf_token`

dev/ory/email_and_phone_with_password_v0.identity.schema.json
```diff
@@ -32,8 +32,7 @@
           }
         }
       },
-      "required": ["email"],
-      "additionalProperties": false
+      "additionalProperties": true
```
nicolasburtey, line 35: what's the goal of this change?

src/servers/graphql-server.ts
```diff
@@ -184,6 +199,9 @@ export const startApolloServer = async ({
 }): Promise<Record<string, unknown>> => {
   const app = express()
   const httpServer = createServer(app)
+  app.use(cookieParser())
```
nicolasburtey, line 202: is that needed? the server itself doesn't need to parse the cookie? (this is done by oathkeeper?)

src/servers/graphql-server.ts
```diff
@@ -86,16 +94,23 @@ const setGqlContext = async (
...
-
   const ipString = isDev
     ? req.ip
     : req.headers["x-real-ip"] || req.headers["x-forwarded-for"]
-
   const ip = parseIps(ipString)
-
+  const body = req.body ?? null
+  let tokenPayload
+  const kratosCookie = req.cookies.ory_kratos_session
```
nicolasburtey, line 103: after the cookie is set, and the client make request to `/graphql`, oathkeeper should parse the cookie and return the JWT. so there should be no change in the way request are been handled.

src/services/kratos/index.ts
```diff
@@ -32,3 +32,25 @@ export const validateKratosToken = async (
...
+    }
+    return new UnknownKratosError(err)
+  }
+
+  // TODO: should return aal level also
+  return {
+    kratosUserId: session.identity.id,
+    session,
+  }
+}
```
nicolasburtey, line 56: I don't think we need this function

the reason we are parsing the Token when they arrive in the header, is because we have to manage the legacy JWT + new kratos token. so there is some logic in the backend.

for cookie, we can have oathkeeper managing the authentication directly, and returning the user directly through the header (with a signed asymmetric JWT)

nicolasburtey (commented):
some good progress. 

overall the biggest flaw seems to be this one: https://github.com/GaloyMoney/galoy/pull/2014/files#r1054124473

nicolasburtey:
@ntheile can you add the graphql endpoint?

eventually that would be great to also do the mobile part if you have the context.

ntheile:
@nicolasburtey 

Yes I will add the graphql endpoint to revoke sessions. Does a mutation called `userLogout` sound like a good name?

nicolasburtey:
> @nicolasburtey
> 
> Yes I will add the graphql endpoint to revoke sessions. Does a mutation called `userLogout` sound like a good name?

yes but @samerbuna is better at naming things

ntheile:
> @ntheile can you add the graphql endpoint?

Here is a PR for the graphql endpoint `userLogout`. I will probably just continue using this PR title `chore: logout skeleton` for experimenting with cookies

nicolasburtey:
> I will probably just continue using this PR titled `chore: logout skeleton` for experimenting with cookies

can you rename the PR then?

nicolasburtey:
needs to be rebased

PR #2031, opened 2022-12-16
fix: pagination pageInfo

samerbuna:


src/services/ledger/index.types.d.ts
```diff
@@ -203,7 +203,7 @@ type IntraledgerLedgerMetadata =
 
 type SendLedgerMetadata = AddOnchainSendLedgerMetadata | AddLnSendLedgerMetadata
 
-type PaginationArgs = { after?: LedgerTransactionId; before?: LedgerTransactionId }
+type PaginatedResp<T> = { slice: T[]; total: number }
```
samerbuna, line 206: Maybe let's come up with a better name for this. `slice` is kinda confusing because it's also a method on array. @vindard @dolcalmi ?

src/services/ledger/index.types.d.ts
```diff
@@ -203,7 +203,7 @@ type IntraledgerLedgerMetadata =
 
 type SendLedgerMetadata = AddOnchainSendLedgerMetadata | AddLnSendLedgerMetadata
 
-type PaginationArgs = { after?: LedgerTransactionId; before?: LedgerTransactionId }
+type PaginatedArray<T> = { slice: T[]; total: number }
```
samerbuna, line 206: Maybe let's come up with a better name for this. `slice` is kinda confusing because it's also a method on array.

src/services/ledger/paginated-ledger.ts
```diff
@@ -53,4 +52,13 @@ export const paginatedLedger = async (
...
     .toArray()
+
+  const countPromise = Transaction.collection.countDocuments(filterQuery, {
+    session: options.session,
+  })
+
+  return {
+    slice: await findPromise,
+    total: await countPromise,
+  } as PaginatedArray<ILedgerTransaction>
```
nicolasburtey, line 63: can we avoid the casting? having `findPromise` and `countPromise` return the right type?

src/services/ledger/paginated-ledger.ts
```diff
@@ -19,21 +19,20 @@ type IFilterQuery = {
...
-} & Partial<ILedgerTransaction> & {
-    [key: string]: string[] | number | string | Date | boolean | Types.ObjectId
-  }
+} & Partial<ILedgerTransaction>
 
 type IOptions = {
   session?: ClientSession
 }
 
 export const paginatedLedger = async (
```
nicolasburtey, line 28: we generally use 1 argument as an object for multi argument functions in the codebase (ie: `{query, paginationArgs, options}: {query: IFilterQuery, paginationArgs: PaginationArgs, options: IOptions}`)

src/services/ledger/paginated-ledger.ts
```diff
@@ -19,21 +19,20 @@ type IFilterQuery = {
...
-  query: IFilterQuery & PaginationArgs,
+  query: IFilterQuery,
+  paginationArgs = {} as PaginationArgs,
```
nicolasburtey, line 30: can we avoid the casting?

src/services/ledger/paginated-ledger.ts
```diff
@@ -19,21 +19,20 @@ type IFilterQuery = {
...
   options = {} as IOptions,
```
nicolasburtey, line 31: can we avoid the casting?

nicolasburtey: `IOptions` is probably too generic of a type name

src/services/ledger/index.types.d.ts
nicolasburtey: name looks ok to me. it's not exposed as an API or written in DB so would be straightforward to update if we wanted to

src/graphql/types/object/usd-wallet.ts
```diff
@@ -82,19 +87,29 @@ const UsdWallet = GT.Object<Wallet>({
...
         if (error instanceof Error) throw mapError(error)
 
         // Non-null signal to type checker; consider fixing in PartialResult type
-        if (transactions === null) throw error
-        return connectionFromArray<WalletTransaction>(transactions, args)
+        if (!result?.slice) throw error
+
+        return connectionFromArraySlice<WalletTransaction>(
+          result.slice,
+          result?.total,
```
nicolasburtey, line 110: why should total be optional?

src/graphql/types/object/usd-wallet.ts
```diff
@@ -61,15 +61,20 @@ const UsdWallet = GT.Object<Wallet>({
...
           paginationArgs,
         })
         if (error instanceof Error) throw mapError(error)
 
         // Non-null signal to type checker; consider fixing in PartialResult type
-        if (transactions === null) throw error
-        return connectionFromArray<WalletTransaction>(transactions, args)
+        if (!result?.slice) throw error
+
+        return connectionFromArraySlice<WalletTransaction>(
```
nicolasburtey, line 73:
```suggestion
        return connectionFromPaginatedArray<WalletTransaction>(
```

?

nicolasburtey (commented):
few comments. I think we should strive to avoid casting. this has regularly introduced bugs in our type system in the past during refactoring.

nicolasburtey:
need rebase, but LGTM otherwise

nicolasburtey (approved):

PR #2042, opened 2022-12-19
feat: add display currency schema and mutation

dolcalmi:
- Add display currency to account schema
- Add update display currency mutation


src/graphql/types/scalar/display-currency.ts
```diff
@@ -0,0 +1,28 @@
...
+  parseLiteral(ast) {
+    if (ast.kind === GT.Kind.STRING) {
+      return validDisplayCurrencyValue(ast.value)
+    }
+    return new InputValidationError({ message: "Invalid type for Currency" })
+  },
+})
+
+function validDisplayCurrencyValue(value: string): string | InputValidationError {
+  if (!value || value.length < 3) {
```
nicolasburtey, line 22: can it be more than 3 ? or 4? maybe having an upper bound could be nice as well

src/domain/accounts/index.types.d.ts
```diff
@@ -77,7 +77,7 @@ type Account = {
   readonly isEditor: boolean
   readonly quizQuestions: UserQuizQuestion[]
   kratosUserId: UserId
-
+  displayCurrency: string
```
nicolasburtey, line 80: we could make a branded type for it? 

```suggestion
  displayCurrency: DisplayCurrency
```

nicolasburtey, line 80: we could make a branded type for it? 

```suggestion
  displayCurrency: DisplayCurrency
```

src/services/mongoose/accounts.ts
```diff
@@ -226,4 +229,5 @@ const translateToAccount = (result: AccountRecord): Account => ({
       }),
     ) || [],
   kratosUserId: result.kratosUserId as UserId,
+  displayCurrency: result.displayCurrency || DisplayCurrency.Usd,
```
nicolasburtey, line 232: not sure about the syntax

```suggestion
  displayCurrency: result.displayCurrency as DisplayCurrency || DisplayCurrency.Usd,
```

nicolasburtey, line 232: not sure about the syntax

```suggestion
  displayCurrency: result.displayCurrency as DisplayCurrency || DisplayCurrency.Usd,
```

nicolasburtey (approved):
LGTM overall but I think we could have a branded type for it. Could be done in an upcoming PR as we are going to do more integration with the ledger I guess.

src/services/mongoose/accounts.ts
dolcalmi: currently we have a DisplayCurrency type (enum), we will replace it in the next PR because we cant use the same because currencies are dynamic

dolcalmi: currently we have a DisplayCurrency type (enum), we will replace it in the next PR because we cant use the same because currencies are dynamic

src/domain/accounts/index.types.d.ts
dolcalmi: currently we have a DisplayCurrency type (enum), we will replace it in the next PR because we cant use the same because currencies are dynamic

dolcalmi: currently we have a DisplayCurrency type (enum), we will replace it in the next PR because we cant use the same because currencies are dynamic

src/graphql/types/scalar/display-currency.ts
dolcalmi: fixed https://github.com/GaloyMoney/galoy/pull/2042/commits/7e85659b7c5dcc040dfcf3f026d56472308e192b

nicolasburtey (approved):

PR #2054, opened 2022-12-20
fix: ip check fallback following axios regression

nicolasburtey:
axios had a regression that made proxycheck call failed: https://github.com/axios/axios/pull/5306
 
this prior PR fixes the proxycheck issue: https://github.com/GaloyMoney/galoy/commit/eca59ab0833c09d6d53d3aa3600ce162d683944a

now most `IPType` are not populated with the correct info, so as a result the rewards are not working.

this PR will make a call back to proxychecker if the relevant info is not present.

also, it adds proper open telemetry info

### DIFF
--- a/src/app/accounts/update-account-ip.ts
+++ b/src/app/accounts/update-account-ip.ts
@@ -1,9 +1,15 @@
 import { getIpConfig } from "@config"
 import { RepositoryError } from "@domain/errors"
 import { IpFetcherServiceError } from "@domain/ipfetcher"
+import { ErrorLevel } from "@domain/shared"
 import { IpFetcher } from "@services/ipfetcher"
 import { AccountsIpRepository } from "@services/mongoose/accounts-ips"
-import { asyncRunInSpan, SemanticAttributes } from "@services/tracing"
+import {
+  addAttributesToCurrentSpan,
+  asyncRunInSpan,
+  recordExceptionInCurrentSpan,
+  SemanticAttributes,
+} from "@services/tracing"
 
 const accountsIp = AccountsIpRepository()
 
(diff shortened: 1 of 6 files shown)


src/app/accounts/update-account-ip.ts
```diff
@@ -17,28 +23,28 @@ export const updateAccountIPsInfo = async ({
   logger: Logger
 }): Promise<void | RepositoryError> =>
   asyncRunInSpan(
-    "app.users.updateAccountIPsInfo",
+    "app.accounts.updateAccountIPsInfo",
```
dolcalmi, line 26: this wrap is not necessary because in sessionContext we are using `Accounts.updateAccountIPsInfo` which is already a wrapper

src/app/accounts/update-account-ip.ts
```diff
@@ -47,31 +53,80 @@ export const updateAccountIPsInfo = async ({
...
+          return ipFetcherInfo
+        }
+
+        // deep copy
+        const ipFetcherInfoForOtel = JSON.parse(JSON.stringify(ipFetcherInfo))
+
+        for (const key in ipFetcherInfoForOtel) {
+          ipFetcherInfoForOtel["proxycheck." + key] = ipFetcherInfoForOtel[key]
+          delete ipFetcherInfoForOtel[key]
+        }
```
dolcalmi, line 100: this can be replaced by a `.map`

src/app/accounts/update-account-ip.ts
```diff
@@ -47,31 +53,80 @@ export const updateAccountIPsInfo = async ({
...
+
+        addAttributesToCurrentSpan(ipFetcherInfoForOtel)
 
-        if (ipConfig.proxyCheckingEnabled) {
-          const ipFetcher = IpFetcher()
-          const ipFetcherInfo = await ipFetcher.fetchIPInfo(ip)
+        if (
+          ipFetcherInfo.isoCode === undefined ||
+          ipFetcherInfo.proxy === undefined ||
+          ipFetcherInfo.asn === undefined
```
dolcalmi, line 107: why to use `undefined` and not just `!`

dolcalmi (dismissed):

dolcalmi (approved):

PR #2057, opened 2022-12-20
feat: add pending HTLCs and inbound liquidity gauges to the exporter

openoms:


src/app/lightning/get-balances.ts
```diff
@@ -23,6 +23,21 @@ export const getTotalBalance = async (): Promise<Satoshis | ApplicationError> =>
   return sumBalances(balances)
 }
 
+export const getInboundBalance = async (): Promise<Satoshis | ApplicationError> =>
```
nicolasburtey, line 26: I don't see a reason to cache the value here?

src/servers/exporter.ts
```diff
@@ -318,3 +320,23 @@ export const getBookingVersusRealWorldAssets = async () => {
     (lightning + bitcoin)
   ) // value in accounting
 }
+
+createGauge({
+  name: "pendingPaymentsCount",
+  description: "How many pending HTLCs there are in the channels of the active nodes",
+  collect: async () => {
+    const pendingPaymentsCount = await getPendingPaymentCount()
+    if (getPendingPaymentCount instanceof Error) return 0
```
nicolasburtey, line 329: does grafana understand `NaN`? it's better to return `NaN` versus 0 is that is the case. 

ie: it's better to not show anything that an erroneous value. 

showing a balance of 0 when when maybe it's just lnd that is down is misleading.

src/services/lnd/utils-ln-service.ts
```diff
@@ -0,0 +1,35 @@
...
+  pending_payments: number[]
+}
+
+async function getPendingPaymentCountLnd1(): Promise<number> {
+  const channels = await lnService.getChannels({ lnd1 })
+  const pendingPayments = channels.filter(
+    (channel: Channel) => channel.pending_payments.length > 0,
+  )
+  return pendingPayments.length
+}
```
nicolasburtey, line 20: this function is wrong. it just counts how many channels have active htlc. it doens't count how many pending htlcs there can be on each channel.

nicolasburtey (changes requested):
need to fix the function

openoms:
Thank you for the comments, all addressed, but didn't yet get to  test.

src/app/lightning/get-balances.ts
openoms: addressed in [1103899](https://github.com/GaloyMoney/galoy/pull/2057/commits/11038996b32a3643fb143ee24a9dbec792cb7a1a)

src/servers/exporter.ts
openoms: resolved in [ecda354](https://github.com/GaloyMoney/galoy/pull/2057/commits/ecda354e8a3ad8ac3aefae39163af37a5ff55328)

src/services/lnd/utils-ln-service.ts
openoms: addressed in https://github.com/GaloyMoney/galoy/pull/2057/commits/5559d221d8a752cdeef26ad6920f02547826352b

src/services/lnd/utils-ln-service.ts
```diff
@@ -0,0 +1,44 @@
+import lnService from "ln-service"
```
dolcalmi, line 1: please add these methods to the main lnd service interface

dolcalmi (changes requested):

src/services/lnd/utils-ln-service.ts
openoms: thank you for the review, but can you please give more detail about what changes are expected?
Should I add the methods from ln-service or my new methods from this file to which file instead?

openoms:
rebased to the latest main

src/services/lnd/utils-ln-service.ts
dolcalmi: I mean to add the methods `ILightningService` interface and remove this file

src/app/lightning/get-balances.ts
```diff
@@ -23,6 +23,50 @@ export const getTotalBalance = async (): Promise<Satoshis | ApplicationError> =>
...
+      .listActivePubkeys()
+      .map((pubkey) => offChainService.getInboundOutboundBalance(pubkey)),
+  )
+
+  const outbound: Satoshis[] = []
+  inboundOutboundBalance.forEach((balance) => {
+    if ("outbound" in balance) {
+      outbound.push(balance.outbound)
+    }
+  })
```
dolcalmi, line 65: you can replace this by a map 
```suggestion
  const outbound = inboundOutboundBalance.map(b => b.outbound)
```

src/app/lightning/get-balances.ts
```diff
@@ -23,6 +23,50 @@ export const getTotalBalance = async (): Promise<Satoshis | ApplicationError> =>
...
+      .listActivePubkeys()
+      .map((pubkey) => offChainService.getInboundOutboundBalance(pubkey)),
+  )
+
+  const inbound: Satoshis[] = []
+  inboundOutboundBalance.forEach((balance) => {
+    if ("inbound" in balance) {
+      inbound.push(balance.inbound)
+    }
+  })
```
dolcalmi, line 43: you can replace this by a map 
```suggestion
  const inbound = inboundOutboundBalance.map(b => b.inbound)
```

src/services/lnd/utils-ln-service.ts
```diff
@@ -0,0 +1,44 @@
...
+  channels.forEach((channel: { id: string; pending_payments: any[] }) => {
+    pendingHtlcCounts[channel.id] = channel.pending_payments.length
+  })
+  return pendingHtlcCounts
+}
+
+async function getPendingHtlcCountLnd2(): Promise<{ [channelId: string]: number }> {
+  const channels = await lnService.getChannels({ lnd2 })
+  const pendingHtlcCounts: { [channelId: string]: number } = {}
+  channels.forEach((channel: { id: string; pending_payments: any[] }) => {
```
dolcalmi, line 22: here probably you can use a reducer instead the use of 2 methods (unless you need the pending htlcs by channel id), something like:

```
return channels.reduce((total, channel) => {
    return total + channel.pending_payments.length;
}, 0)
```

dolcalmi (changes requested):

src/services/lnd/utils-ln-service.ts
openoms: done in [242366f](https://github.com/GaloyMoney/galoy/pull/2057/commits/242366f414516e45406203ac6ddfc6552867f80e)

src/services/lnd/utils-ln-service.ts
openoms: `getTotalPendingHtlcCount` is one function now

src/app/lightning/get-balances.ts
dolcalmi: this was not updated, is there a reason to use that notation?

src/app/lightning/get-balances.ts
dolcalmi: same here

src/servers/exporter.ts
```diff
@@ -311,3 +312,103 @@ export const getBookingVersusRealWorldAssets = async () => {
...
   ) // value in accounting
 }
+
+createGauge({
+  name: "totalPendingHtlcCount",
+  description: "How many pending HTLCs there are in the channels of the active nodes",
+  collect: async () => {
+    try {
+      const lndService = LndService()
+      if (lndService instanceof Error) throw lndService
```
dolcalmi, line 322: in theory you dont need to wrap this code in try/catch, service should return an error so you can return NaN instead throwing the error and return NaN in the catch (this applies to the new `createGauge` handlers)

```js
if (lndService instanceof Error) return NaN
```

dolcalmi (changes requested):

src/app/lightning/get-balances.ts
openoms: the suggestion was causing a type error: https://github.com/GaloyMoney/galoy/pull/2057/commits/2eb0640a00755b3f41c8ff24e81d40e73e3ae121 so reverted to the original solution

src/app/lightning/get-balances.ts
openoms: the suggestion was causing a type error: https://github.com/GaloyMoney/galoy/pull/2057/commits/b5cb4d345b8549a1242ec53ba9651fb063548b26

src/servers/exporter.ts
openoms: remove the try/catch wrapping here: https://github.com/GaloyMoney/galoy/pull/2057/commits/6402dd14ef3a8438b0297deca8b28c4c08b424df

openoms: the map solution doesn't check if the `inbound` property is present in the `inboundOutboundBalance` array and throws an error if there is no inbound present.

src/app/lightning/get-balances.ts
dolcalmi: is because balance could be an error too, so you need to handle it.

```
const inbound = inboundOutboundBalance.map((balance) => {
    if (balance instanceof Error) return toSats(0) 
    return balance.inbound
  })
```

src/app/lightning/get-balances.ts
dolcalmi:
```
const outbound = inboundOutboundBalance.map((balance) => {
    if (balance instanceof Error) return toSats(0) 
    return balance.outbound
  })
 ```

**Review by dolcalmi — CHANGES_REQUESTED** [review]

**Comment by openoms** [comment]
Running:
```
TEST="01|02" make reset-integration 
make exporter
```
locally now shows:
```
# HELP galoy_totalPendingHtlcCount How many pending HTLCs there are in the channels of the active nodes
# TYPE galoy_totalPendingHtlcCount gauge
galoy_totalPendingHtlcCount 0

# HELP galoy_activeChannels How many active channels there are on the active nodes
# TYPE galoy_activeChannels gauge
galoy_activeChannels 3

# HELP galoy_offlineChannels How many offline channels there are on the active nodes
# TYPE galoy_offlineChannels gauge
galoy_offlineChannels 0

# HELP galoy_publicChannels How many public channels there are on the active nodes
# TYPE galoy_publicChannels gauge
galoy_publicChannels 3

# HELP galoy_privateChannels How many private channels there are on the active nodes
# TYPE galoy_privateChannels gauge
galoy_privateChannels 0

# HELP galoy_inboundBalance How much inbound balance there is on the active nodes
# TYPE galoy_inboundBalance gauge
galoy_inboundBalance 1915460

# HELP galoy_outboundBalance How much outbound balance there is on the active nodes
# TYPE galoy_outboundBalance gauge
galoy_outboundBalance 155200
```

dolcalmi (approved):
looks good, next time please dont forget to use https://www.conventionalcommits.org/ for commits.

PR #2065, opened 2022-12-21
chore: make it possible to filter internal IP address

nicolasburtey:


dolcalmi (commented):
looks good, just a question...dont we want to use something more generic? like https://stackoverflow.com/a/13969691 or use a library like `ipaddr.js` https://github.com/whitequark/ipaddr.js/blob/master/test/ipaddr.test.js#L150

nicolasburtey:
> looks good, just a question...dont we want to use something more generic? like https://stackoverflow.com/a/13969691 or use a library like `ipaddr.js` https://github.com/whitequark/ipaddr.js/blob/master/test/ipaddr.test.js#L150

yeah that would remove the additional configuration parameter I have added. that is a good point. I'll do that.

nicolasburtey:
ok I think that is better. @dolcalmi let me know

test/unit/services/ipfetcher/fetch-ip-info.spec.ts
```diff
@@ -59,6 +60,28 @@ describe("IpFetcher - fetchIPInfo", () => {
   })
 })
 
+describe("ipaddr.js", () => {
```
dolcalmi, line 63: this needs to be moved to test domain folder and description should be isPrivateIp (we are testing domain function here)

dolcalmi (changes requested):
looks good, just a minor change

PR #2068, opened 2022-12-22
feat: add currencyList public query

dolcalmi:
- Add currency list query
- Update price service

https://github.com/GaloyMoney/price/pull/34 is required before merge

How to test:
- Run `make start-deps`
- Run `make start-main-fast`
- Open http://localhost:4002/graphql or use postman to run:
```graphql
query CurrencyList {
  currencyList {
    code
    symbol
    name
    flag
  }
}
```

### DIFF
--- a/src/app/prices/list-currencies.ts
+++ b/src/app/prices/list-currencies.ts
@@ -1,0 +1,14 @@
+import { CacheKeys } from "@domain/cache"
+import { PriceService } from "@services/price"
+import { LocalCacheService } from "@services/cache"
+import { SECS_PER_10_MINS } from "@config"
+
+export const listCurrencies = async (): Promise<PriceCurrency[] | ApplicationError> => {
+  return LocalCacheService().getOrSet({
+    key: CacheKeys.PriceCurrencies,
+    ttlSecs: SECS_PER_10_MINS,
+    getForCaching: () => {
+      return PriceService().listCurrencies()
+    },
+  })
+}
--- a/src/domain/price/index.types.d.ts
+++ b/src/domain/price/index.types.d.ts
@@ -11,10 +11,10 @@ type Tick = {
 }
 
 type PriceCurrency = {
-  readonly code: string
-  readonly symbol: string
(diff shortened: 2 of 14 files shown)


src/domain/price/index.types.d.ts
```diff
@@ -10,12 +10,20 @@ type Tick = {
   readonly price: DisplayCurrencyPerSat
 }
 
+type PriceCurrency = {
```
nicolasburtey, line 13: would be handy to have some description of what this means, maybe with an example.
not clear to me what is the difference between `code` and `symbol`. also what are we going to store in flag? emoji?

dolcalmi: we will store the emoji, aslso the code is the ISO code (3 letters) and symbol the currency symbol 

https://github.com/GaloyMoney/price/blob/074e7ca3890a00d94e7acb28089d6c98a132d1b4/realtime/default.yaml#L5

dolcalmi: Comments added https://github.com/GaloyMoney/galoy/pull/2068/commits/dc5aa7c48351e03d62127cdf7cbe1adf5293f32b

daviroo:
@dolcalmi this looks good to me.

src/app/prices/list-currencies.ts
```diff
@@ -0,0 +1,14 @@
...
+import { LocalCacheService } from "@services/cache"
+import { SECS_PER_10_MINS } from "@config"
+
+export const listCurrencies = async (): Promise<PriceCurrency[] | ApplicationError> => {
+  return LocalCacheService().getOrSet({
+    key: CacheKeys.PriceCurrencies,
+    ttlSecs: SECS_PER_10_MINS,
+    getForCaching: () => {
+      return PriceService().listCurrencies()
+    },
```
nicolasburtey, line 12:
```suggestion
    getForCaching: PriceService().listCurrencies
```

nicolasburtey (approved):

PR #2071, opened 2022-12-22
feat: device account

nicolasburtey:
this is an example on how an user could get started with a bearer token that would be generated from the mobile app, and then the user can upgrade by adding his phone number.

it's a question whether the bearer token could be exported or not to be use by another phone. I think that might be something we want to do, it introduces a lot of complexity and is not really part of the use case of receiving $1 or $10 before having to put the phone number

EDIT: writing this, I think we don't even need to pass a token from the mobile app to the backend. maybe there should still be a deviceId been sent to the backend to be able to do some rateLimit on this criteria.


src/config/yaml.ts
```diff
@@ -318,6 +326,7 @@ export const getRewardsConfig = () => {
 export const getDefaultAccountsConfig = (config = yamlConfig): AccountsConfig => ({
   initialStatus: config.accounts.initialStatus as AccountStatus,
   initialWallets: config.accounts.initialWallets,
+  initialLevel: AccountLevel.One,
```
vindard, line 329: Can be a config too?

dev/ory/oathkeeper.yml
```diff
@@ -9,6 +9,15 @@ tracing:
       local_agent_address: otel-agent:6831
 
 authenticators:
+  jwt:
+    enabled: true
+    config:
+      trusted_issuers:
+        - https://firebaseappcheck.googleapis.com/1076318732193
```
ntheile, line 16: @nicolasburtey need to swap out the number `1076318732193` for the value from galoy's `firebase > project settings > Project Number`

dev/ory/oathkeeper.yml
```diff
@@ -9,6 +9,15 @@ tracing:
...
+      target_audience:
+        - projects/1076318732193
```
ntheile, line 18: swap Project number here too

dev/ory/oathkeeper_rules.yaml
```diff
@@ -37,6 +37,14 @@
     url: "<(http|https)>://<[a-zA-Z0-9-.:]+>/graphql<.*>"
     methods: ["POST"]
   authenticators:
+    - handler: jwt
+      config:
+        trusted_issuers:
+          - https://firebaseappcheck.googleapis.com/1076318732193
+        target_audience:
+          - projects/1076318732193
```
ntheile, line 45: and here

dev/appcheck/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+fs.writeFileSync('.well-known/jwks.json', JSON.stringify(jwks, null, 2));
+console.log('JWKS file generated: .well-known/jwks.json');
+Run the script:
+bash
+Copy code
+node jwks-generator.js
+After running the script, you should have a .well-known/jwks.json file that contains your JWKS. You can then serve this file using an HTTP server (such as Express) or upload it to a web server where your identity provider is hosted.
+
+Keep in mind that the JWKS file should be accessible from the URL specified in your Oathkeeper configuration. Update the jwks_urls and trusted_issuers fields in the Oathkeeper configuration to match the actual URL of your identity provider and JWKS file.
```
ntheile, line 42: nice! i was trying to solve this yesterday. so with this method we won't need to update the config in the firebase console with the sha256 hash of our local debug.keystore?

ntheile:
@nicolasburtey ,

After discussing with @UncleSamtoshi we thought it makes more sense to just use kratos session tokens instead of jwt's. This will simplify also of the edge case logic with the device account history we were brainstorming. We simply think of the app-check jwt as a password and call the `userLoginDevice` mutation to login (or create new account) and get a session token back. The same flow as userLogin for phone. Here is the mutations:

```graphql
mutation {
  userLoginDevice(input:{
    jwt: "{{MY_APP_CHECK_JWT}}"
  }) {
    errors {
      message
    }
    authToken
  }
}
```

dev/ory/device_account_v0.identity.schema.json
```diff
@@ -0,0 +1,40 @@
+{
```
nicolasburtey, line 1: I guess the title of the file should be device_account_or_phone_v0?

package.json
```diff
@@ -87,6 +86,7 @@
     "ipaddr.js": "^2.0.1",
     "js-yaml": "^4.1.0",
     "jsonwebtoken": "^9.0.0",
+    "jwk-to-pem": "^2.0.5",
```
nicolasburtey, line 89: it should just be in `devDependencies`?

package.json
```diff
@@ -99,6 +99,7 @@
     "medici": "^6.1.0",
     "mongoose": "^6.9.2",
     "node-cache": "^5.1.2",
+    "node-jose": "^2.2.0",
```
nicolasburtey, line 102: it should just be in devDependencies?

src/app/accounts/index.ts
```diff
@@ -35,6 +35,12 @@ export const getAccountFromUserId = async (
   return accounts.findByUserId(kratosUserId)
 }
 
+export const getAccountFromDeviceId = async (
+  deviceId: DeviceId,
+): Promise<Account | RepositoryError> => {
+  return accounts.findByUserId(deviceId as unknown as UserId)
```
nicolasburtey, line 41: this casting doesn't sound right to me.

are we not using a kratos identity also for device account? is it a legacy function?

src/config/schema.ts
```diff
@@ -342,6 +357,16 @@ export const configSchema = {
           duration: 3600,
           blockDuration: 86400,
         },
+        createDeviceAccountPerIp: {
+          points: 4,
+          duration: 86400,
+          blockDuration: 86400 * 10,
+        },
+        createDeviceAccountGlobal: {
+          points: 180,
```
nicolasburtey, line 366:
```suggestion
          points: 500,
```

I think this value can be higher now that Appcheck has been integrated, it was set pre-Appcheck capability in mind

src/app/accounts/index.ts
ntheile: yeah this a legacy function. I will delete it

src/app/auth/login.ts
```diff
@@ -113,6 +132,331 @@ export const loginWithPhoneCookie = async ({
...
+      /// Scenario 1.5 - has txns on device account but phone account exists
+      return new PhoneAccountAlreadyExistsNeedToSweepFunds(
+        `Error phone account already exists. You need to manually sweep funds
+        to your phone account. Here are the steps to do this:
+        1. login with phone account and create an invoice (save this invoice)
+        2. logout of the phone account
+        3. login with device account (click "Explore Wallet Instead" button)
+        4  Pay the invoice with max wallet amount.
+        5. Logout of the device account
+        6. Log into the phone account and check for the funds`,
```
nicolasburtey, line 382: this should be in the frontend eventually

src/app/auth/login.ts
```diff
@@ -113,6 +132,331 @@ export const loginWithPhoneCookie = async ({
...
+  if (accountLevelOne instanceof Error) return accountLevelOne
+
+  // returning a new session token
+  const authPhone = AuthWithPhonePasswordlessService()
+  const kratosResult = await authPhone.loginToken({ phone })
+  if (kratosResult instanceof Error) return kratosResult
+  return kratosResult.sessionToken
+}
+
+export const loginWithDevice = async ({
```
nicolasburtey, line 423: it is meant to only work when the deviceAlready exist? 

how one know when to go here, versus https://github.com/GaloyMoney/galoy/pull/2071/files#diff-4b614afac7a950ff856e0a8f9ed1c3b92c23281e1e1ebe3b297d983b77b5b280R8 ?

src/domain/accounts/index.ts
```diff
@@ -22,11 +22,32 @@ export * from "./primitives"
...
+  // kratosUserId is a firebase jwt subject
+  if (isDeviceId(userId)) return userId as UserId
+
+  return new InvalidUserId(userId)
+}
+
+export const isDeviceId = (userId: string) => {
+  // user id format from AppCheck: 1:72279297366:android:35666807ae916c5aa75af7
+  if (userId.match(/^[0-9]+:[0-9]+:[a-z]+:[0-9a-z]+$/i)) {
+    return true
```
nicolasburtey, line 48:
```suggestion
  return userId.match(/^[0-9]+:[0-9]+:[a-z]+:[0-9a-z]+$/i))
```

src/domain/accounts/index.ts
```diff
@@ -22,11 +22,32 @@ export * from "./primitives"
...
+    return userId as UserId
+  }
+
+  // kratosUserId is a firebase jwt subject
+  if (isDeviceId(userId)) return userId as UserId
+
+  return new InvalidUserId(userId)
+}
+
+export const isDeviceId = (userId: string) => {
```
nicolasburtey, line 45: this way to check for deviceId is not how we typically do check. 

of this is a deviceId, then we should return a typed value. (ie: `return id as DeviceId`) so that we don't have to cast outside like we're doing [here](https://github.com/GaloyMoney/galoy/pull/2071/files#diff-651d74d3e907cd7859cee097d1d418d4d434273dc9f3392e4faabc6446eb3974R159)

src/graphql/root/mutation/user-login-device.ts
```diff
@@ -0,0 +1,58 @@
...
+}>({
+  extensions: {
+    complexity: 120,
+  },
+  type: GT.NonNull(JwtPayload),
+  args: {
+    input: { type: GT.NonNull(UserLoginDeviceInput) },
+  },
+  resolve: async (_, args, { ip }) => {
+    const { jwt } = args.input
```
nicolasburtey, line 30: it's not great to have to pass the jwt around (as opposed to passing a checked value from the header from Oathkeeper mutator, because Oathkeeper would have checked the value already) but I guess this is how this new design architecture works

we should make sure the jwt is excluded from open telemetry

src/migrations/20230502222224-add-user-device-id.ts
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  async up(db) {
+    try {
+      const result = db
+        .collection("users")
+        .createIndex({ device: 1 }, { unique: true, sparse: true })
+      console.log({ result }, "index created for users/device")
+    } catch (error) {
+      console.log({ result: error }, "error creating index for users/device")
```
nicolasburtey, line 9: we should probably throw here

nicolasburtey (commented):
I've gone through half of the PR. will continue tomorrow.

dev/ory/device_account_v0.identity.schema.json
ntheile: the phone part was legacy. i removed that. the device and phone schemas are separate.

src/migrations/20230502222224-add-user-device-id.ts
ntheile: i just copied how to other migrations were written

src/domain/accounts/index.ts
ntheile: updated to use `checkedToDeviceId` here https://github.com/GaloyMoney/galoy/pull/2071/commits/1bc785dfa9c9a1ff71bef82d6bb236d7a7c0fe9a

src/app/auth/login.ts
ntheile: it's meant to follow the flow we already have for the phone login. If the device account already exists it logs you in, otherwise it creates the account.

i just removed the legacy device flow in this commit https://github.com/GaloyMoney/galoy/pull/2071/commits/57df85ec4364d20a3d8141ec3e6f42c74c86ded5

src/graphql/root/mutation/user-login-device.ts
ntheile: The idea behind this was to make it similar to the phone login flow. The jwt is equivalent to the twilio code. The userLogin and userLoginDevice mutations are UnAuth'd called ( i.e no oathkeeper bearer).  So the userLoginDevice call is verfied in the `jwtVerfy` function similar to how the twilio code is validated with `isCodeValid`.

All `Auth'd` mutations and queries still get the jwt validated in oathkeeper and the subject is  passed to the backend for authenicated calls.

src/app/auth/login.ts
```diff
@@ -113,6 +128,171 @@ export const loginWithPhoneCookie = async ({
...
+  ip,
+}: {
+  jwt: string
+  ip: IpAddress
+}): Promise<SessionToken | ApplicationError> => {
+  {
+    const limitOk = await checkFailedLoginAttemptPerIpLimits(ip)
+    if (limitOk instanceof Error) return limitOk
+  }
+
```
nicolasburtey, line 270: we are no longer using `createDeviceAccountGlobal` in the code right now.

src/domain/accounts/index.ts
```diff
@@ -22,11 +23,34 @@ export * from "./primitives"
 const UserIdRegex =
   /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i
 
+// device id format from AppCheck: 1:72279297366:android:35666807ae916c5aa75af7
+const DeviceIdRegex = /^[0-9]+:[0-9]+:[a-z]+:[0-9a-z]+$/i
+
+export const checkedToDeviceAccountUserId = (userId: string) => {
+  if (userId.match(/^[0-9]+:[0-9]+:[a-z]+:[0-9a-z]+$/i)) {
```
nicolasburtey, line 30:
```suggestion
  if (userId.match(DeviceIdRegex) {
```

is this what you meant here?

src/domain/accounts/index.ts
```diff
@@ -22,11 +23,34 @@ export * from "./primitives"
...
+  }
+
+  return new InvalidUserId(userId)
+}
+
 export const checkedToUserId = (userId: string): UserId | ValidationError => {
-  if (!userId.match(UserIdRegex)) {
-    return new InvalidUserId(userId)
+  // kratosUserId is a uuid v4
+  if (userId.match(UserIdRegex)) {
```
nicolasburtey, line 39: even the `UserIdRegex` regex is giving hindsight that there is some inconsistency here with https://github.com/GaloyMoney/galoy/pull/2071/files#diff-f365a6e339ebf8a923ca76401012e4718a4ee3e4b8a2bd460ededbdf41467dc4R31

src/graphql/main/schema.graphql
```diff
@@ -387,6 +388,11 @@ enum InvoicePaymentStatus {
   PENDING
 }
 
+type JwtPayload {
```
nicolasburtey, line 391: do we need a new type here? shouldn't this be `AuthTokenPayload` ?

src/graphql/main/schema.graphql
```diff
@@ -1346,6 +1354,11 @@ type User {
   contacts: [UserContact!]! @deprecated(reason: "will be moved to account")
   createdAt: Timestamp!
   defaultAccount: Account!
+
+  """
+  Device ID of the user's device.
+  """
+  device: String
```
nicolasburtey, line 1361: how about `deviceId` as the variable name?

src/servers/graphql-server.ts
```diff
@@ -142,31 +142,39 @@ export const sessionContext = ({
   return addAttributesToCurrentSpanAndPropagate(
     {
       "token.sub": tokenPayload?.sub,
+      "token.iss": tokenPayload?.iss,
```
nicolasburtey, line 145: I think we may not need this right now?

src/services/mongoose/users.ts
```diff
@@ -106,10 +121,28 @@ export const UsersRepository = (): IUsersRepository => {
     }
   }
 
+  const unsetDeviceIdForUser = async (id: UserId): Promise<User | RepositoryError> => {
```
nicolasburtey, line 124: this is very (too) specific for a database layer, wdyt @bodymindarts or @dolcalmi ?

src/services/mongoose/schema.types.d.ts
```diff
@@ -117,6 +117,7 @@ interface UserRecord {
   deviceTokens: string[]
   phoneMetadata?: TwilioObjectForUser
   phone?: PhoneNumber
+  device?: DeviceId
```
nicolasburtey, line 120:
```suggestion
  deviceId?: DeviceId
```

src/services/mongoose/schema.ts
```diff
@@ -340,6 +343,12 @@ const UserSchema = new Schema(
       unique: true,
       sparse: true,
     },
+    device: {
```
nicolasburtey, line 346:
```suggestion
    deviceId: {
```

src/servers/index.files.d.ts
```diff
@@ -13,6 +13,7 @@ type GraphQLContext = {
   domainAccount: Account | undefined
   geetest: GeetestType
   ip: IpAddress | undefined
+  sub: string
```
nicolasburtey, line 16: we're not using this right now?

src/servers/index.files.d.ts
```diff
@@ -22,6 +23,7 @@ type GraphQLContextAuth = {
   domainAccount: Account
   geetest: GeetestType
   ip: IpAddress
+  sub: string
```
nicolasburtey, line 26: we're not using this right now?

src/services/kratos/auth-device-account.ts
```diff
@@ -0,0 +1,249 @@
...
+      return new UnknownKratosError(err.message || err)
+    }
+  }
+
+  const createDeviceIdentity = async ({
+    deviceId,
+  }: {
+    deviceId: DeviceId
+  }): Promise<CreateKratosUserForDeviceSchemaResponse | KratosError> => {
+    const traits = { device: deviceId }
```
nicolasburtey, line 80:
```suggestion
    const traits = { deviceId: deviceId }
```

I would also be consistent at this layer, ie: using `deviceId` instead of `device`.

src/services/kratos/auth-device-account.ts
```diff
@@ -0,0 +1,249 @@
...
+      traits,
+    }
+    try {
+      const identityResults = await kratosAdmin.createIdentity({
+        createIdentityBody: identityBody,
+      })
+      console.log(identityResults.status)
+      const result = await loginDeviceAccount({ deviceId })
+      if (result instanceof KratosError) return result
+      const sessionToken = result.sessionToken
```
nicolasburtey, line 94: the `sessionToken` here will never expired in the current PR? isn't that a big drawback of this new flow?

src/services/kratos/auth-device-account.ts
```diff
@@ -0,0 +1,249 @@
...
+            ...identity.metadata_admin,
+            { device: identity.traits.device, updated_at },
+          ]
+        } else {
+          // is json object
+          identity.metadata_admin = [
+            identity.metadata_admin,
+            { device: identity.traits.device, updated_at },
+          ]
+        }
```
nicolasburtey, line 155: do we need to maintain 2 scenario where it's an array OR an object?

currently we don't use `metadata_admin`, so if we device it should only be an object, we don't need this branching.

src/services/kratos/auth-device-account.ts
```diff
@@ -0,0 +1,249 @@
...
+    phone: PhoneNumber
+  }) => {
+    let identity: Identity
+
+    try {
+      ;({ data: identity } = await kratosAdmin.getIdentity({ id: kratosUserId }))
+    } catch (err) {
+      if (err.message === "Request failed with status code 400") {
+        return new LikelyUserAlreadyExistError(err)
+      }
```
nicolasburtey, line 129: not sure this error message make most sense here, we expect the user to already exist?

src/services/kratos/auth-device-account.ts
```diff
@@ -0,0 +1,249 @@
...
+            identity.metadata_admin,
+            { device: identity.traits.device, updated_at },
+          ]
+        }
+      } else {
+        // new data
+        identity.metadata_admin = [{ device: identity.traits.device, updated_at }]
+      }
+    } catch (err) {
+      console.log(err)
```
nicolasburtey, line 161: use baseLogger

src/services/kratos/auth-device-account.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    const { data: newIdentity } = await kratosAdmin.updateIdentity({
+      id: kratosUserId,
+      updateIdentityBody: adminIdentity,
+    })
+
+    return toDomainIdentityPhone(newIdentity)
+  }
+
+  const verifyJwt = async (token: string) => {
```
nicolasburtey, line 184: is that not an (better?) option to call Oathkeeper api (the internal API) to verify the JWT?

src/services/mongoose/users.ts
```diff
@@ -55,12 +57,25 @@ export const UsersRepository = (): IUsersRepository => {
     }
   }
 
+  const findByDeviceId = async (device: DeviceId): Promise<User | RepositoryError> => {
+    try {
+      const result = await User.findOne({ device })
+      if (!result) return new CouldNotFindUserFromPhoneError()
```
nicolasburtey, line 63:
```suggestion
      if (!result) return new CouldNotFindUserFromDeviceIdError()
```

src/services/mongoose/users.ts
```diff
@@ -55,12 +57,25 @@ export const UsersRepository = (): IUsersRepository => {
...
+
+      return translateToUser(result)
+    } catch (err) {
+      //  else
```
nicolasburtey, line 67:
```suggestion
```

src/services/mongoose/users.ts
bodymindarts: Agree - this looks like a layer violation. Putting use case code into a repository implementation leads to very strong coupling.
There may be exceptions when an operation must be atomic (and we don't have transaction handling)- so  it might be acceptable. We do this in at least 1 case I can think of.
But I don't see any indication of that being the case here.

nicolasburtey: probably https://github.com/GaloyMoney/galoy/pull/2071/files#diff-501a242d86bf5d6bdcab743568228d988d5a13898af716c7fae13064b6f16033R106 is similarly a layer boundary violation

cc @krtk6160

src/servers/graphql-server.ts
```diff
@@ -142,31 +142,39 @@ export const sessionContext = ({
...
       [SemanticAttributes.HTTP_CLIENT_IP]: ip,
     },
     async () => {
       // note: value should match (ie: "anon") if not an accountId
       // settings from dev/ory/oathkeeper.yml/authenticator/anonymous/config/subjet
-      const maybeUserId = checkedToUserId(tokenPayload?.sub || "")
+      const sub = tokenPayload?.sub || ""
+
+      const maybeUserId = checkedToUserId(sub)
+      const maybeDeviceId = checkedToDeviceId(sub)
```
nicolasburtey, line 154: ok I now understand why we have all those unexpected, and what feel weirdly designed, casting between DeviceId and UserId. 

this line/file is the main reason: we can receive from Oathkeeper either a UserId or a DeviceId from the same field.

I don't believe we are using Oathkeeper properly here. We should not receive both a DeviceId and a UserId from the sub. we should always receive a UserId from kratos. 

the backend should, at this stage, not have to bother what type of identity it's receiving. it should always be the same.

how can we manage this properly?

We can use an [hydrator](https://www.ory.sh/docs/oathkeeper/pipeline/mutator#hydrator) as an additional mutator in the oathkeeper pipeline. we should conditional call a new API endpoint in the backend when the request comes from a JWT, and make the translation deviceId to UserId, so that the request at the backend always comes from a kratos userId. this way there would be no need for custom logic. 

-- 
that said, there might be one reason to not do this. we previously had a lot of 503 when we had a custom authenticator that was pointing to the backend. we never really debug the issue, and instead got rid of the backend request we remove the legacy JWT.

src/migrations/20230502222224-add-user-device-id.ts
nicolasburtey: ahah. it's not because I made dump migration that you have to copy them :)

src/servers/middlewares/kratos-router.ts
```diff
@@ -163,4 +164,40 @@ kratosRouter.post(
...
+        } else if (!(maybeDeviceId instanceof ValidationError)) {
+          const deviceId = maybeDeviceId
+          const deviceUser = await UsersRepository().findByDeviceId(deviceId)
+          if (!(deviceUser instanceof Error)) userId = deviceUser.id
+        }
+        res.status(200).json({
+          subject, // subject: userId, // <=== this breaks if you try to mutate the subject (i.e changing deviceId subject to userId)
+          extra: {
+            userId, // <==== need to use extra
+          },
```
ntheile, line 193: @nicolasburtey ,

you cannot mutate the subject (a JSON validation error is thrown by oathkeeper) so I added an extra field called userId. 

Ory added an AI bot in their docs! The bot helped answer the question why changing the subject caused errors. (some keys, like subject cannot be changes)
Screen Shot 2023-05-16 at 8 43 30 PM

nicolasburtey: then we can do a `userId = token.userId ?? token.sub` I guess. not ideal but probably better flow that what we had previously.

src/graphql/error-map.ts
```diff
@@ -549,6 +563,8 @@ export const mapError = (error: ApplicationError): CustomApolloError => {
     case "KratosError":
     case "AuthenticationKratosError":
     case "ExtendSessionKratosError":
+    case "PhoneAccountAlreadyExists":
+    case "PhoneAccountAlreadyExistsNeedToSweepFunds":
```
UncleSamtoshi, line 567: Can you map these to a differentiated error, so that the frontend is capable of parsing it? `PhoneAccountAlreadyExistsError` could be the name.

src/graphql/main/schema.graphql
ntheile: yep, updated to use deviceId

src/services/kratos/auth-device-account.ts
ntheile: i don't think using oathkeeper will work because this is only called when creating a new device account, the kratos account does yet exist, so we can't call `http://localhost:4456/decisions/graphql` with the MeQuery for example.

to make this maybe work, i would have to create a separate oathkeeper rule for a specific graphql or kratos router endpoint to simply pass back the verified jwt data. seems like a lot of extra hops and config when we can simply verify the jwt with the jwks.

src/graphql/error-map.ts
ntheile: https://github.com/GaloyMoney/galoy/pull/2071/commits/93a078bc5207eb4432150d47f80bcd4657c7dd22

dolcalmi (changes requested):

src/services/mongoose/users.ts
```diff
@@ -55,12 +62,28 @@ export const UsersRepository = (): IUsersRepository => {
     }
   }
 
+  const findByDeviceId = async (deviceId: DeviceId): Promise<User | RepositoryError> => {
+    try {
+      const deviceIdChecked = checkedToDeviceId(deviceId)
+      if (deviceIdChecked instanceof ValidationError)
+        return new RepositoryError("Invalid format for deviceId")
```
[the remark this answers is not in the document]
ntheile: oh yeah i forgot to remove. i was just testing if that would get rid of the security code scanning error above, but it did not. any ideas on getting rid of the scanning error (might just be a false positive)?

src/services/mongoose/users.ts
ntheile: fixed here https://github.com/GaloyMoney/galoy/pull/2071/commits/674327722dc4ba8834690dea5d32621d84e93359

package.json
ntheile: i am using it here https://github.com/GaloyMoney/galoy/pull/2071/files#diff-8c1f181fc0b0efef5b124341ef62eed90f9dcf7f640ecaf7420996ce4d281439R6

src/app/auth/login.ts
ntheile: added it as an error type here https://github.com/GaloyMoney/galoy/pull/2071/commits/9a6050ee9bcc834111ff442c34cec9ca932bee25 for @UncleSamtoshi to handle on the frontend

src/config/yaml.ts
```diff
@@ -326,6 +334,7 @@ export const getDefaultAccountsConfig = (config = yamlConfig): AccountsConfig =>
   initialStatus: config.accounts.initialStatus as AccountStatus,
   initialWallets: config.accounts.initialWallets,
   randomizeDefaultWallet: config.accounts.randomizeDefaultWallet,
+  initialLevel: AccountLevel.One,
```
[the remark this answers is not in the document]
ntheile: @nicolasburtey @vindard should we make this configuration and split into two configs, one for device account and for a standard account. 

```
initialLevelDevice: AccountLevel.Zero,
initialLevelStandard: AccountLevel.Zero,
```

src/app/auth/login.ts
ntheile: removed here https://github.com/GaloyMoney/galoy/pull/2071/commits/abc51de5b4c012442eeb7f9bd1800fbdb2c7efbf

src/graphql/main/schema.graphql
ntheile: the jwt is a different length, so the AuthToken validator fails here https://github.com/GaloyMoney/galoy/blob/main/src/graphql/types/scalar/auth-token.ts#L17

src/servers/graphql-server.ts
ntheile: not sure, appears to always be "galoy.io"

src/servers/index.files.d.ts
ntheile: removed

src/services/kratos/auth-device-account.ts
ntheile: device flow documents in mermaid sequence diagram here https://github.com/GaloyMoney/galoy/blob/abc51de5b4c012442eeb7f9bd1800fbdb2c7efbf/dev/appcheck/README.md#device-login-flow

src/services/kratos/auth-device-account.ts
ntheile: i wanted to avoid accidentally over-writing any data in the metadata_admin, so i check for both types. i want the metadata_admin to be an array of object as a history log. just being extra cautious checking for existing data (even tho there should not be any)

src/services/kratos/auth-device-account.ts
ntheile: yep. removed that error

nicolasburtey:
@ntheile I have been thinking about the non mutability of the `sub`, and also the fact we have to use an hydrator... and this make me think the idea of having a kratos identity for device may not be the best path forward imo. 

if our backend relies on kratos identity for normal (non auth related) request, then maybe we should work on removing this dependency instead of adding some more complexity with the device account PR.

Let's talk some more about that point maybe tomorrow.

ntheile:
> if our backend relies on kratos identity for normal (non auth related) request, then maybe we should work on removing this dependency instead of adding some more complexity with the device account PR.

Do you suggest updating all references of `kratosUserId` in the codebase to `userId` and change the `UsersRepository().findById` to search for either kratos users id or device id? FYI there is about 200 results when you search `kratosUserId` in the code base, so we would have to be careful to not cause regressions. I will be out tomorrow for Memorial Day Holiday.

src/app/auth/login.ts
```diff
@@ -113,6 +131,138 @@ export const loginWithPhoneCookie = async ({
...
+  if (hasPhoneAccount) {
+    // is there still txns left over on the device account?
+    const deviceWallets = await WalletsRepository().listByAccountId(account.id)
+    if (deviceWallets instanceof Error) return deviceWallets
+    const ledger = LedgerService()
+    let deviceAccountHasBalance = false
+    for (const wallet of deviceWallets) {
+      const balance = await ledger.getWalletBalance(wallet.id)
+      if (balance instanceof Error) return balance
+      if (balance > 0) {
```
bodymindarts, line 207: Potential race condition.
Balance report 0 here but is updated in parallel.

src/domain/accounts/index.ts
```diff
@@ -22,11 +22,22 @@ export * from "./primitives"
...
   /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i
 
+// device id format from AppCheck: 1:72279297366:android:35666807ae916c5aa75af7
+const DeviceIdRegex = /^[0-9]+:[0-9]+:[a-z]+:[0-9a-z]+$/i
+
 export const checkedToUserId = (userId: string): UserId | ValidationError => {
-  if (!userId.match(UserIdRegex)) {
-    return new InvalidUserId(userId)
+  // kratosUserId is a uuid v4
+  if (userId.match(UserIdRegex)) {
```
bodymindarts, line 30: Here it becomes very obvious our domain model is insufficient.

Imo it should be something like this:
```
type AuthSubjectRef = AppCheckDeviceId | KratosId
```
and `UserId` is a mongoldb database reference that is immutable.

src/domain/authentication/index.types.d.ts
```diff
@@ -4,6 +4,7 @@ type AuthenticationError = import("./errors").AuthenticationError
 
 type IdentityPassword = string & { readonly brand: unique symbol }
 
+// can be either kratosUserId or deviceId subjects
```
bodymindarts, line 7: This comment indicates missing domain concepts.

src/servers/graphql-server.ts
bodymindarts: Adding information to spans can't hurt.

src/services/kratos/index.ts
```diff
@@ -54,3 +54,19 @@ export const validateKratosCookie = async (
...
+export const revokeKratosToken = async (token: string): Promise<void | KratosError> => {
+  try {
+    const res = await kratosPublic.performNativeLogout({
+      performNativeLogoutBody: {
+        session_token: token,
+      },
+    })
+    if (res.status !== 204) {
+      // TODO
+      return new UnknownKratosError("unsuccessful token revocation")
```
bodymindarts, line 67: Can we pass more information here. No idea what res.status is for example.

test/e2e/integration-services/device-account.spec.bak
```diff
@@ -0,0 +1,320 @@
+import { kratosAdmin } from "@services/kratos/private"
```
bodymindarts, line 1: Why `.bak`?

src/services/kratos/auth-device-account.ts
```diff
@@ -0,0 +1,79 @@
...
+    // Decode the token
+    const decodedToken = jsonwebtoken.decode(token, { complete: true })
+    if (!decodedToken) return Error("problem with jwt")
+
+    // Find the kid (Key Id) used to sign the token
+    const kid = decodedToken.header.kid
+    let kidJwk = jwkJson.find((j) => j.kid === kid)
+
+    // if in dev environment, use local jwks.json if you dont have access to firebase appcheck debug
+    if (!kidJwk && (isDev || isCI)) {
```
bodymindarts, line 51: This check shouldn't be made here at runtime every call. Just once at startup time and injected.

bodymindarts (commented):
Its looking good overall - structure is pretty clean.


To me it's obvious we need an iteration on the domain model. `UserId` is overloaded and insufficient to cleanly represent this use case. If we are otherwise close to shipping I guess we can have the domain model iteration as a 'post-feature refactoring' type of task. But certainly needs to be addressed before adding additional complexity.

test/e2e/integration-services/device-account.spec.bak
ntheile: removed it. was keeping it temporarily as a reference to older code to look back at.

src/services/kratos/auth-device-account.ts
ntheile: refactored this code to use existing function sendOathkeeperRequest

src/services/kratos/auth-device-account.ts
ntheile: i figured it out using sendOathkeeperRequest

src/app/auth/login.ts
ntheile: yeah, this just returns a message to the user suggesting what they should do, so i don't think it's a big deal.

src/domain/accounts/index.ts
ntheile: I defer to @nicolasburtey . maybe we can do a data modeling session in a follow up PR. Below nicolas created a diagram of the existing model.  This shows the Account collection as the source of truth (top of the heirarchy). Our current data model has some duplication and can probably be normalized in a follow up PR. i.e user id is duplicated in the user collection as the foreign key link, but in a normalized database should probably be AccountId 
** Entity represents the kratos database
Screen Shot 2023-06-01 at 4 56 58 PM

dev/appcheck/README.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+Run the script:
+```bash
+node jwks-generator.js
+```
+
+After running the script, you should have a .well-known/jwks.json file that contains your JWKS. You can then serve this file using an HTTP server (such as Express) or upload it to a web server where your identity provider is hosted.
+
+Keep in mind that the JWKS file should be accessible from the URL specified in your Oathkeeper configuration. Update the jwks_urls and trusted_issuers fields in the Oathkeeper configuration to match the actual URL of your identity provider and JWKS file.
```
nicolasburtey, line 48:
```suggestion
The JWKS file should be accessible from the URL specified in your Oathkeeper configuration. Update the `jwks_urls` and `trusted_issuers` fields in the Oathkeeper configuration to match the actual URL of your identity provider and JWKS file.
```

dev/appcheck/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+```mermaid
+sequenceDiagram
+    Mobile->>Firebase: (1) Firebase, here is my unique device id
+    Firebase->>Attestation Provider: (2) Thanks device, let me ask if this is authentic from Play Integrity (android) or App Attest (ios)
+    Attestation Provider-->>Firebase: (3) Yep this device is authentic and unique
+    Firebase-->>Mobile: (4) Here's your jwt, valid for 1 hour
+    Mobile-->>Oathkeeper: (5) Let's have Oathkeeper verify the JWT against the JWKS
+    Oathkeeper-->>Backend: (6) Yep, he's real, here is the subject (deviceId)
+    Backend-->Mongo: (7) Hey mongo does this user exist in the Accounts collection? (check userId field for the deviceId subject)
```
nicolasburtey, line 61:
```suggestion
sequenceDiagram
    Mobile->>Firebase: (1) Unique device id
    Firebase->>Attestation Provider: (2) Check device authenticity
    Attestation Provider-->>Firebase: (3) Device authenticated
    Firebase-->>Mobile: (4) Here's jwt (1 hour validity)
    Mobile-->>Oathkeeper: (5) Verify jwt against JWKS
    Oathkeeper-->>Backend: (6) User authenticated, here's deviceId
    Backend-->Mongo: (7) Check user in Accounts collection
```

a GPT4 optimized version that is a bit more compact.

nicolasburtey: (and therefore display more nicely on the README.md page)

package.json
nicolasburtey: I only see it used in `dev/ory/gen-test-jwt.ts`

galoy.yaml
```diff
@@ -102,6 +108,10 @@ rateLimits:
     points: 8
     duration: 3600
     blockDuration: 86400
+  createDeviceAccountPerIp:
```
nicolasburtey, line 111: this is currently not used right now. probably dropped in some refactoring. this was implemented pre appcheck. maybe this could be dropped for now?

src/app/auth/login.ts
```diff
@@ -113,6 +135,136 @@ export const loginWithPhoneCookie = async ({
...
+
+  const validCode = await isCodeValid({ phone, code })
+  if (validCode instanceof Error) return validCode
+
+  await rewardFailedLoginAttemptPerIpLimits(ip)
+  await rewardFailedLoginAttemptPerPhoneLimits(phone)
+
+  // Scenario 1) Phone account already exists (no txn with device acct) - User logs in with device jwt then he
+  //             wants to upgrade to a phone account that already exists. there are no txns on the device account
+  //             Tell the user to logout and log back in with the phone account
```
nicolasburtey, line 167: shouldn't we auto upgrade to the phone account here? seems awkward to ask the user log out if there is no tx.

I guess it requires whether the existing device has a tx. but that would be a common scenario and the ux would be a lot better.

galoy.yaml
nicolasburtey: I guess we're using `checkFailedLoginAttemptPerIpLimits` instead which is good.

src/app/auth/login.ts
```diff
@@ -228,3 +380,16 @@ export const isCodeValid = async ({
...
+
+export const verifyJwt = async (token: string) => {
+  const newToken = await sendOathkeeperRequest(token as SessionToken)
+  if (newToken instanceof Error) return newToken
+  const keyJwks = await jwksRsa(getJwksArgs()).getSigningKey()
+  const verifiedToken = jsonwebtoken.verify(newToken, keyJwks.getPublicKey(), {
+    algorithms: ["RS256"],
+  })
+  if (typeof verifiedToken === "string") {
+    throw new Error("tokenPayload should be an object")
```
nicolasburtey, line 392: A more precise error should be defined that should be a subclass of the `Error` type, so we could filter it out at the graphql level and provide proper alerting

src/app/auth/login.ts
```diff
@@ -113,6 +135,136 @@ export const loginWithPhoneCookie = async ({
...
+    const limitOk = await checkFailedLoginAttemptPerIpLimits(ip)
+    if (limitOk instanceof Error) return limitOk
+  }
+
+  const verifiedJwt = await verifyJwt(jwt)
+  if (verifiedJwt instanceof Error) {
+    await rewardFailedLoginAttemptPerIpLimits(ip)
+    return verifiedJwt
+  }
+  const deviceId = verifiedJwt.sub as UserId
```
nicolasburtey, line 251: `verifiedJwt.sub` can be `undefined` based on the type information. we should do a check on it.

src/app/auth/login.ts
```diff
@@ -113,6 +135,136 @@ export const loginWithPhoneCookie = async ({
...
+  if (accountExist instanceof CouldNotFindAccountFromKratosIdError) {
+    const levelZeroAccountsConfig = getDefaultAccountsConfig()
+    levelZeroAccountsConfig.initialLevel = AccountLevel.Zero
+    const account = await createAccountForDeviceAccount({
+      userId: deviceId,
+      config: levelZeroAccountsConfig,
+    })
+    if (account instanceof Error) return account
+  }
+  return jwt as SessionToken
```
nicolasburtey, line 265: do we need to return jwt? could we just return `Success: true` instead. it will make it easier to understand for someone looking into the API what is expected. otherwise someone could assume we get a different JWT here?

At least that my thoughts. wdyt @UncleSamtoshi ?

src/domain/errors.ts
```diff
@@ -34,6 +34,7 @@ export class CouldNotFindWalletInvoiceError extends CouldNotFindError {}
 export class CouldNotFindUserError extends CouldNotFindError {}
 export class CouldNotFindUserFromIdError extends CouldNotFindError {}
 export class CouldNotFindUserFromPhoneError extends CouldNotFindError {}
+export class CouldNotFindUserFromDeviceError extends CouldNotFindError {}
```
nicolasburtey, line 37:
```suggestion
export class CouldNotFindUserFromDeviceIdError extends CouldNotFindError {}
```
?

src/app/auth/login.ts
nicolasburtey: I guess this would break the rollout

src/graphql/root/mutation/user-login-device.ts
nicolasburtey: note: this is an indication we should probably have a normal rest endpoint for this, and other typical authentication endpoint. ok, you can resolve back the conversation :)

src/servers/graphql-server.ts
```diff
@@ -148,32 +148,35 @@ export const sessionContext = ({
...
-          logger,
-        })
-
-        const userRes = await UsersRepository().findById(account.kratosUserId)
-        if (userRes instanceof Error) throw mapError(userRes)
-        user = userRes
-
-        addAttributesToCurrentSpan({ [ACCOUNT_USERNAME]: domainAccount?.username })
+        if (account instanceof CouldNotFindAccountFromKratosIdError) {
+          // do nothing, the jwt is valid but the account does not exist
```
nicolasburtey, line 163: if we were to move this endpoint https://github.com/GaloyMoney/galoy/pull/2071/files#diff-2a3744191b13e4b2375d4bc44df6706cd453f9b3ccbf3970f801ad4ce97219aaR16 outside of graphql, then we would not need to have this exception. 

what do you think @bodymindarts @ntheile ? could be part of a future refactoring.

src/services/kratos/auth-phone-no-password.ts
```diff
@@ -173,6 +173,40 @@ export const AuthWithPhonePasswordlessService = (): IAuthWithPhonePasswordlessSe
...
+    const traits = { phone, deviceId }
+    const method = "password"
+    try {
+      const flow = await kratosPublic.createNativeRegistrationFlow()
+      const result = await kratosPublic.updateRegistrationFlow({
+        flow: flow.data.id,
+        updateRegistrationFlowBody: {
+          traits,
+          method,
+          password,
```
nicolasburtey, line 194: it seems you could also use `transient_payload` instead of `deviceId`. one of the benefits is that the kratos entity would not be bloated with deviceId because it won't be use anymore, and this would simplify any further migration of the kratos entity in the future.

deviceId should still be stored somewhere, but this could be in the User mongodb collection. I think this would be a better place than here, because it's just metadata. and adding a `traits` doesn't feels the ideal place for metadata as this is directly related to how our authentication works.


documentation of `transient_payload`:
    /**
     * Transient data to pass along to any webhooks
     * @type {object}
     * @memberof UpdateRegistrationFlowWithOidcMethod
     */
    'transient_payload'?: object;

src/services/kratos/index.ts
```diff
@@ -54,3 +58,20 @@ export const validateKratosCookie = async (
     session,
   }
 }
+
+export const revokeKratosToken = async (token: string): Promise<void | KratosError> => {
```
nicolasburtey, line 62: this is not used in this PR for info. we could have another PR and provide explicit test for it (it may be buggy on the kratos side so some tests would be actually useful)

src/services/mongoose/schema.ts
```diff
@@ -302,6 +304,7 @@ const UserSchema = new Schema(
       default: Date.now,
     },
     userId: {
+      // this would need to be migrated
```
nicolasburtey, line 307:
```suggestion
```

nicolasburtey:
2 remaining main comments on my end:
- https://github.com/GaloyMoney/galoy/pull/2071/files#r1214573711
- https://github.com/GaloyMoney/galoy/pull/2071/files#r1214606851

src/graphql/root/mutation/user-login-device.ts
ntheile: redacted jwt here in telemetry https://github.com/GaloyMoney/charts/pull/3158/commits/53a98af477acd9d9566ba42c9f856d729c0945ab

src/services/mongoose/users.ts
```diff
@@ -88,6 +86,36 @@ export const UsersRepository = (): IUsersRepository => {
...
+    currentUserIdSubject: UserId
+    newUserIdSubject: UserId
+    phone: PhoneNumber
+  }): Promise<User | RepositoryError> => {
+    const filter = { userId: { $eq: currentUserIdSubject } }
+    const update = {
+      $set: {
+        userId: newUserIdSubject,
+        phone,
+        deviceId: currentUserIdSubject,
```
nicolasburtey, line 103: could be even easier to just set it at the time of id creation?

I guess it'll lead to some data duplicate, but the pros is that one query would be able to give us all deviceId, otherwise we need to do some join to get all the deviceId with current scheme.

src/app/auth/login.ts
```diff
@@ -113,6 +136,138 @@ export const loginWithPhoneCookie = async ({
...
+    for (const wallet of deviceWallets) {
+      const balance = await ledger.getWalletBalance(wallet.id)
+      if (balance instanceof Error) return balance
+      if (balance > 0) {
+        deviceAccountHasBalance = true
+      }
+    }
+    if (deviceAccountHasBalance) {
+      // Scenario 1.5 - has txns on device account but phone account exists
+      return new PhoneAccountAlreadyExistsNeedToSweepFundsError()
```
nicolasburtey, line 217: seems you may not want to do the balance swipe? maybe as an alternative to do it, we set up a dedicated honeycomb board to easily display at which frequency this arise and then we can decide if this is worth implementing or not in a few weeks.

src/app/accounts/upgrade-device-account.ts
```diff
@@ -0,0 +1,33 @@
...
+  accountDevice.kratosUserId = userId
+  accountDevice.level = AccountLevel.One
+  const accountUpdated = await AccountsRepository().update(accountDevice)
+  if (accountUpdated instanceof Error) return accountUpdated
+
+  // 2. update user
+  const userUpdated = await UsersRepository().migrateUserIdSubject({
+    currentUserIdSubject: deviceId,
+    newUserIdSubject: userId,
+    phone,
```
nicolasburtey, line 28: we are fetching phone metadata in the "normal" flow. I'm not sure we're doing it in this upgrade path. something maybe to add in a follow up PR? we could add a TODO here in the meantime.

src/domain/authentication/index.types.d.ts
nicolasburtey: maybe we can add another TODO here. we should tackle this when this PR is merged I think.

src/servers/middlewares/kratos-router.ts
```diff
@@ -122,8 +129,16 @@ kratosRouter.post(
       }
 
       let account: Account | RepositoryError
-      // phone+code flow
-      if (phoneRaw) {
+
+      // upgrade device to phone flow if both traits are present
+      if (transient_payload?.deviceId && phoneRaw) {
+        account = await upgradeAccountFromDeviceToPhone({
```
nicolasburtey, line 135: shouldn't we have to upgrade both Account and User collection here?

nicolasburtey: nevermind, underlying function indeed is doing both

src/services/kratos/tests-but-not-prod.ts
dolcalmi, line 1: remove and move to test utils if it is necessary

src/services/kratos/auth-device-account.ts
dolcalmi, line 1: why is this not an interface implementation?

src/services/mongoose/accounts.ts
```diff
@@ -166,7 +166,10 @@ export const AccountsRepository = (): IAccountsRepository => {
     kratosUserId: UserId,
   ): Promise<Account | RepositoryError> => {
     try {
-      const result = await Account.findOne({ kratosUserId })
+      const result = await Account.findOne(
+        { kratosUserId: { $eq: kratosUserId } },
+        { lastIPs: 0 },
```
dolcalmi, line 171: not sure about this... why findByUserId should depends on lastIPs?

src/services/mongoose/users.ts
```diff
@@ -88,6 +86,36 @@ export const UsersRepository = (): IUsersRepository => {
...
+  const migrateUserIdSubject = async ({
+    currentUserIdSubject,
+    newUserIdSubject,
+    phone,
+  }: {
+    currentUserIdSubject: UserId
+    newUserIdSubject: UserId
+    phone: PhoneNumber
+  }): Promise<User | RepositoryError> => {
+    const filter = { userId: { $eq: currentUserIdSubject } }
```
dolcalmi, line 98: is not the same to use `{ userId: currentUserIdSubject }`

src/services/rate-limit/index.ts
```diff
@@ -49,7 +49,7 @@ export const consumeLimiter = async ({
   keyToConsume,
 }: {
   rateLimitConfig: RateLimitConfig
-  keyToConsume: IpAddress | PhoneNumber | EmailAddress | AccountId
+  keyToConsume: IpAddress | PhoneNumber | EmailAddress | AccountId | ""
```
dolcalmi, line 52: keyToConsume must have a valid param, having `""` is not the intended behaviour

test/e2e/servers/graphql-main-server/mutations/user-login-upgrade.gql
dolcalmi, line 1: why do we need this? is not in generated file?

test/e2e/servers/ory/kratos.spec.ts
```diff
@@ -370,7 +372,7 @@ it("extend session", async () => {
   const initialExpiresAt = new Date(session.expires_at)
 
   await extendSession({ session })
-
+  await sleep(200)
```
dolcalmi, line 375: why do we need to sleep here if you are awaiting extendSession?

test/helpers/index.ts
```diff
@@ -90,7 +90,7 @@ export const amountByPriceAsMajor = <
 export const chunk = (a, n) =>
   [...Array(Math.ceil(a.length / n))].map((_, i) => a.slice(n * i, n + n * i))
 
-export const graphqlAdmin = <
+export const graphqlAdmin = async <
```
dolcalmi, line 93: why do you need to add async if internal code is not using await?

src/app/accounts/create-account.ts
```diff
@@ -64,12 +71,38 @@ const initializeCreatedAccount = async ({
...
 
   return updatedAccount
 }
 
+export const createAccountForDeviceAccount = async ({
+  userId,
+}: {
+  userId: UserId
+}): Promise<Account | RepositoryError> => {
+  const user = await UsersRepository().update({ id: userId })
```
dolcalmi, line 88: not sure to understand this... what are you updating?

src/app/accounts/create-account.ts
```diff
@@ -64,12 +71,38 @@ const initializeCreatedAccount = async ({
...
+  const levelZeroAccountsConfig = getDefaultAccountsConfig()
+  levelZeroAccountsConfig.initialLevel = AccountLevel.Zero
+
+  const account = await initializeCreatedAccount({
+    account: accountNew,
+    config: levelZeroAccountsConfig,
+  })
+  if (account instanceof Error) return account
+
+  return account
```
dolcalmi, line 103: if you are not doing nothing with the error then just return.
```
return initializeCreatedAccount({
    account: accountNew,
    config: levelZeroAccountsConfig,
  })
  ```

File: galoy.yaml [hunk 70]
```diff
@@ -207,3 +217,7 @@ skipFeeProbeConfig:
   chanId: []
 smsAuthUnsupportedCountries: []
 whatsAppAuthUnsupportedCountries: []
+appcheckConfig:
```
dolcalmi, line 220: if this is related to firebase and depending on the service that you use is better to rename it. maybe firebaseAppcheckConfig?

dev/ory/gen-test-jwt.ts
dolcalmi, line 1: and we need because? I didnt find references in doc, package script or in make file... if it just a helper for tests move it to utils or create the script (similar to e2e-codege)

.envrc
```diff
@@ -89,3 +89,7 @@ export KRATOS_PG_PORT="5433"
 if test "$(uname -m)" = "arm64"; then
   export MONGODB_VERSION="4.4"
 fi
+
+export APPCHECK_AUDIENCE="projects/72279297366"
+export APPCHECK_ISSUER="https://firebaseappcheck.googleapis.com/72279297366"
```
dolcalmi, line 94: is this generic? can we expose this?

src/app/auth/login.ts
```diff
@@ -113,6 +136,138 @@ export const loginWithPhoneCookie = async ({
...
+    } else {
```
dolcalmi, line 218: else is not needed, you are returning in the if

src/app/auth/login.ts
```diff
@@ -113,6 +136,138 @@ export const loginWithPhoneCookie = async ({
...
+    if (limitOk instanceof Error) return limitOk
+  }
+
+  const validCode = await isCodeValid({ phone, code })
+  if (validCode instanceof Error) return validCode
+
+  await rewardFailedLoginAttemptPerIpLimits(ip)
+  await rewardFailedLoginAttemptPerPhoneLimits(phone)
+
+  // Scenario 1) Phone account already exists (no txn with device acct) - User logs in with device jwt then he
```
dolcalmi, line 166: if tthis is intentional please move it to method docs

src/app/auth/login.ts
```diff
@@ -113,6 +136,138 @@ export const loginWithPhoneCookie = async ({
...
+  //             Option 2 - create some kind of recovery code process?
+
+  // Scenario 1 - does phone account already exist?
+  const phoneAccount = await UsersRepository().findByPhone(phone)
+
+  if (phoneAccount instanceof CouldNotFindUserFromPhoneError) {
+    // nothing to do, we'll go to scenario 2
+  } else if (phoneAccount instanceof Error) {
+    return phoneAccount
+  } else {
```
dolcalmi, line 202: else is not needed

src/app/auth/login.ts
```diff
@@ -113,6 +136,138 @@ export const loginWithPhoneCookie = async ({
...
+  // Scenario 3) Unhappy path - User sells phone and forgets to sweep funds to phone account
+  //             Option 1 - too bad, we can't help them
+  //             Option 2 - create some kind of recovery code process?
+
+  // Scenario 1 - does phone account already exist?
+  const phoneAccount = await UsersRepository().findByPhone(phone)
+
+  if (phoneAccount instanceof CouldNotFindUserFromPhoneError) {
+    // nothing to do, we'll go to scenario 2
+  } else if (phoneAccount instanceof Error) {
```
dolcalmi, line 200: if nothing to do then we dont need it, please remove

src/app/auth/login.ts
```diff
@@ -113,6 +136,138 @@ export const loginWithPhoneCookie = async ({
...
+
+  const verifiedJwt = await verifyJwt(jwt)
+  if (verifiedJwt instanceof Error) {
+    await rewardFailedLoginAttemptPerIpLimits(ip)
+    return verifiedJwt
+  }
+  const deviceId = verifiedJwt.sub as UserId
+  if (!deviceId) return new JwtSubjectUndefinedError("deviceId sub is undefined")
+
+  // 1. Does account exist in mongo?
```
dolcalmi, line 259: why do we need a comment on something that is evident?

src/app/auth/login.ts
```diff
@@ -228,3 +383,16 @@ export const isCodeValid = async ({
...
+
+export const verifyJwt = async (token: string) => {
+  const newToken = await sendOathkeeperRequest(token as SessionToken)
+  if (newToken instanceof Error) return newToken
+  const keyJwks = await jwksRsa(getJwksArgs()).getSigningKey()
+  const verifiedToken = jsonwebtoken.verify(newToken, keyJwks.getPublicKey(), {
+    algorithms: ["RS256"],
+  })
+  if (typeof verifiedToken === "string") {
+    throw new JwtVerifyTokenError("tokenPayload should be an object")
```
dolcalmi, line 395: why are we throwing in app layer? please use a proper return and include return type

src/domain/accounts/index.ts
```diff
@@ -19,14 +19,24 @@ export * from "./limits-checker"
 export * from "./account-validator"
 export * from "./primitives"
 
-const UserIdRegex =
+const KratosUserIdRegex =
```
dolcalmi, line 22: we have a big problem with this (in general) we integrated kratos in our domain and that is not good

src/domain/authentication/errors.ts
```diff
@@ -5,3 +5,5 @@ export class LikelyNoUserWithThisPhoneExistError extends AuthenticationError {}
 export class LikelyUserAlreadyExistError extends AuthenticationError {}
 export class PhoneIdentityDoesNotExistError extends AuthenticationError {}
 export class UserWithPhoneAlreadyExistsError extends AuthenticationError {}
+export class JwtVerifyTokenError extends AuthenticationError {}
+export class JwtSubjectUndefinedError extends AuthenticationError {}
```
dolcalmi, line 9: none of these are critical?

dolcalmi (changes requested):
is really big, @bodymindarts coud you please check?, if we approve as it is we will have a lot of technical debt

src/services/mongoose/accounts.ts
nicolasburtey: it does not. ip is fetched from another service (`accounts-ip`). the fact it's not excluded is a just a miss. it's been excluded from other queries in this file.

bodymindarts:
> is really big, @bodymindarts coud you please check?, if we approve as it is we will have a lot of technical debt

I agree - we certainly need an internal facing iteration to improve our domain model to handle this (and future) complexity.
I think it makes sense to push this out to iterate quickly. I am sure there will be issues in prod (like with bria integration) that we need to find out about asap.

Boundary issues and other comments you make should be addressed either pre or shortly post merge. Domain refactoring should be its own task to prepare whatever the next thing is we are going to do that touches auth.

src/app/auth/login.ts
ntheile: yeah I think it will be safer that way as this use case will probably rarely get hit

src/services/mongoose/users.ts
ntheile: it fails the codeql code scan without the `$eq`

src/services/kratos/tests-but-not-prod.ts
ntheile: cc @nicolasburtey

src/services/mongoose/users.ts
```diff
@@ -88,6 +86,36 @@ export const UsersRepository = (): IUsersRepository => {
...
+      $set: {
+        userId: newUserIdSubject,
+        phone,
+        deviceId: currentUserIdSubject,
+      },
+    }
+    const options = { upsert: true, returnOriginal: false }
+
+    try {
+      const result = await User.findOneAndUpdate(filter, update, options)
```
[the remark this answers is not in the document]
ntheile: yeah i tried that originally but it did not work because of this special use-case...updating the userId. calling something like this in the domain does not work:

```
UsersRepository().update({
    id,
    language,
    deviceTokens,
    phoneMetadata,
    phone,
    createdAt,
    deviceId,
  })
```

because downstream the service calls mongo like this:

```
const result = await User.findOneAndUpdate(
        { userId: id },
        {
          userId: id, // THIS DOES NOT UPDATE
          deviceTokens,
          phoneMetadata,
          language,
          phone,
          deviceId,
          createdAt, 
        },
        {
          new: true,
          upsert: true,
        },
      )
```

mongoose does not allow you to update the `userId` because we queried by it.

@nicolasburtey suggested we do it this way for now until we have the domain modeling session next time we touch the Auth code. just one of the trade-off with the architectural decision of making a  `userId` field either a `KratosId` or `DeviceId`

src/services/rate-limit/index.ts
ntheile: cc @nicolasburtey

test/e2e/integration-services/device-account.spec.ts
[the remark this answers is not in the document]
ntheile: @nicolasburtey does this belong up one directory in `test/e2e` ? I just copied where you originally had it

src/app/accounts/create-account.ts
ntheile: code was updated to include deviceId. also it looks like the update uses the config below

```
{
  new: true,
  upsert: true,
}
```

galoy.yaml
ntheile: the product is called App Check but you can write custom providers (i.e for hauwei )

dev/ory/gen-test-jwt.ts
ntheile: yeah its used in local device account testing. so you can run `gen-test-jwt` and get a jwt token to test locally in the graphql playground. i will move to dev scripts and create a helper in the make file.

.envrc
ntheile: its just a reference to the audience.

src/app/auth/login.ts
ntheile: i'll move it to a readme

src/app/auth/login.ts
ntheile: forget to delete it. it helps prompt github co-pilot to write the code for me lol

src/graphql/admin/root/query/all-levels.ts
```diff
@@ -1,13 +1,13 @@
...
-import { levels } from "@config"
+import { Levels } from "@config"
 
 import AccountLevel from "@graphql/types/scalar/account-level"
 
 const AllLevelsQuery = GT.Field({
   type: GT.NonNullList(AccountLevel),
   resolve: () => {
-    return levels
+    return Levels
```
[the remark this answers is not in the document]
ntheile: cc @nicolasburtey

src/domain/authentication/errors.ts
ntheile: not to alert on, jwts expires

src/app/accounts/create-account.ts
```diff
@@ -64,12 +71,38 @@ const initializeCreatedAccount = async ({
...
+  userId: UserId
+}): Promise<Account | RepositoryError> => {
+  const user = await UsersRepository().update({ id: userId })
+  if (user instanceof Error) return user
+
+  const accountNew = await AccountsRepository().persistNew(userId)
+  if (accountNew instanceof Error) return accountNew
+
+  const levelZeroAccountsConfig = getDefaultAccountsConfig()
+  levelZeroAccountsConfig.initialLevel = AccountLevel.Zero
```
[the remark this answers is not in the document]
ntheile: @nicolasburtey any ideas? not sure what to do here

test/e2e/servers/graphql-main-server/mutations/user-login-upgrade.gql
ntheile: removed

src/app/accounts/create-account.ts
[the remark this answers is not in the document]
nicolasburtey: because default config currently is for level 1. the denomination/naming could be improved.

src/app/auth/login.ts
UncleSamtoshi: Right now in the mobile application we use the sessionToken being present as a proxy for success, so we have to keep it to be back backwards compatible, but could migrate over time to `success: true`.

nicolasburtey:
approved on my end

dolcalmi (approved):

PR #2072, opened 2022-12-22
feat: add onchain stablesats send 🎉 

vindard:
## Description

This PR builds on the work already done in #1408.

It removes the USD code-path blockers, adds new queries/mutations and integration tests for USD use-cases.

### Open `TODO`s
- [x] add tests for fee probe from USD wallet (done: https://github.com/GaloyMoney/galoy/pull/2072/commits/dc44f7b3aa0d94bebffcae2c84bd7e5907827e90)
- [x] add new USD gql mutations and enable existing ones; test

### Follow-up PR
- cherry-pick https://github.com/GaloyMoney/galoy/pull/2094/commits/ff5cca9a708e4bb5c9a8c142b57ccc6e279ef52a for independent PR
- move validate functions to app layer (https://github.com/GaloyMoney/galoy/pull/2072#discussion_r1058588580)
- add hedge cases for onchain where applicable

### DIFF
--- a/src/graphql/error.ts
+++ b/src/graphql/error.ts
@@ -153,7 +153,7 @@ export class DustAmountError extends CustomApolloError {
   constructor(errData: CustomApolloErrorData) {
     super({
       code: "ENTERED_DUST_AMOUNT",
-      message: `Use lightning to send amounts less than ${onChainWalletConfig.dustThreshold}`,
+      message: `Use lightning to send amounts less than ${onChainWalletConfig.dustThreshold} sats`,
       forwardToClient: true,
       ...errData,
     })
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -48,7 +48,19 @@ export const validateIsBtcWalletForQuery = async (
   const wallet = await WalletsRepository().findById(walletId)
   if (wallet instanceof Error) return mapError(wallet)
 
-  if (wallet.currency === WalletCurrency.Usd) {
+  if (wallet.currency !== WalletCurrency.Btc) {
+    return QueryDoesNotMatchWalletCurrencyError
+  }
+  return true
+}
+
+export const validateIsUsdWalletForQuery = async (
+  walletId: WalletId,
+): Promise<true | CustomApolloError> => {
+  const wallet = await WalletsRepository().findById(walletId)
+  if (wallet instanceof Error) return mapError(wallet)
+
+  if (wallet.currency !== WalletCurrency.Usd) {
     return QueryDoesNotMatchWalletCurrencyError
   }
   return true
--- a/src/graphql/main/mutations.ts
+++ b/src/graphql/main/mutations.ts
@@ -26,6 +26,7 @@ import AccountUpdateDefaultWalletIdMutation from "@graphql/root/mutation/account
 import UserContactUpdateAliasMutation from "@graphql/root/mutation/user-contact-update-alias"
 import UserQuizQuestionUpdateCompletedMutation from "@graphql/root/mutation/user-quiz-question-update-completed"
 import OnChainPaymentSendMutation from "@graphql/root/mutation/onchain-payment-send"
+import OnChainUsdPaymentSendMutation from "@graphql/root/mutation/onchain-usd-payment-send"
 import OnChainPaymentSendAllMutation from "@graphql/root/mutation/onchain-payment-send-all"
 import CaptchaRequestAuthCodeMutation from "@graphql/root/mutation/captcha-request-auth-code"
 import CaptchaCreateChallengeMutation from "@graphql/root/mutation/captcha-create-challenge"
@@ -72,6 +73,7 @@ const fields = {
   onChainAddressCreate: OnChainAddressCreateMutation,
   onChainAddressCurrent: OnChainAddressCurrentMutation,
   onChainPaymentSend: OnChainPaymentSendMutation,
+  onChainUsdPaymentSend: OnChainUsdPaymentSendMutation,
   onChainPaymentSendAll: OnChainPaymentSendAllMutation,
 }
 
--- a/src/graphql/main/queries.ts
+++ b/src/graphql/main/queries.ts
@@ -8,6 +8,7 @@ import BtcPriceListQuery from "@graphql/root/query/btc-price-list"
 import QuizQuestionsQuery from "@graphql/root/query/quiz-questions"
 import MobileVersionsQuery from "@graphql/root/query/mobile-versions"
 import OnChainTxFeeQuery from "@graphql/root/query/on-chain-tx-fee-query"
+import OnChainUsdTxFeeQuery from "@graphql/root/query/on-chain-usd-tx-fee-query"
 import UsernameAvailableQuery from "@graphql/root/query/username-available"
 import BusinessMapMarkersQuery from "@graphql/root/query/business-map-markers"
 import AccountDefaultWalletQuery from "@graphql/root/query/account-default-wallet"
@@ -27,6 +28,7 @@ const fields = {
   btcPrice: BtcPriceQuery,
   btcPriceList: BtcPriceListQuery,
   onChainTxFee: OnChainTxFeeQuery,
+  onChainUsdTxFee: OnChainUsdTxFeeQuery,
   lnInvoicePaymentStatus: LnInvoicePaymentStatusQuery,
 } as const
 
--- a/src/graphql/main/schema.graphql
+++ b/src/graphql/main/schema.graphql
@@ -708,6 +708,7 @@ type Mutation {
   onChainAddressCurrent(input: OnChainAddressCurrentInput!): OnChainAddressPayload!
   onChainPaymentSend(input: OnChainPaymentSendInput!): PaymentSendPayload!
   onChainPaymentSendAll(input: OnChainPaymentSendAllInput!): PaymentSendPayload!
+  onChainUsdPaymentSend(input: OnChainUsdPaymentSendInput!): PaymentSendPayload!
   userContactUpdateAlias(
     input: UserContactUpdateAliasInput!
   ): UserContactUpdateAliasPayload! @deprecated(reason: "will be moved to AccountContact")
@@ -782,6 +783,14 @@ type OnChainUpdate {
   txHash: OnChainTxHash!
   txNotificationType: TxNotificationType!
   usdPerSat: Float! @deprecated(reason: "updated over displayCurrencyPerSat")
+  walletId: WalletId!
+}
+
+input OnChainUsdPaymentSendInput {
+  address: OnChainAddress!
+  amount: CentAmount!
+  memo: Memo
+  targetConfirmations: TargetConfirmations = 1
   walletId: WalletId!
 }
 
--- a/src/graphql/main/schema.graphql
+++ b/src/graphql/main/schema.graphql
@@ -923,6 +923,12 @@ type Query {
     targetConfirmations: TargetConfirmations = 1
     walletId: WalletId!
   ): OnChainTxFee!
+  onChainUsdTxFee(
+    address: OnChainAddress!
+    amount: CentAmount!
+    targetConfirmations: TargetConfirmations = 1
+    walletId: WalletId!
+  ): OnChainTxFee!
   quizQuestions: [QuizQuestion]
   userDefaultWalletId(username: Username!): WalletId!
     @deprecated(reason: "will be migrated to AccountDefaultWalletId")
--- a/src/graphql/root/mutation/onchain-usd-payment-send.ts
+++ b/src/graphql/root/mutation/onchain-usd-payment-send.ts
@@ -1,0 +1,86 @@
+import { GT } from "@graphql/index"
+import Memo from "@graphql/types/scalar/memo"
+import { mapAndParseErrorForGqlResponse } from "@graphql/error-map"
+import WalletId from "@graphql/types/scalar/wallet-id"
+import CentAmount from "@graphql/types/scalar/cent-amount"
+import OnChainAddress from "@graphql/types/scalar/on-chain-address"
+import PaymentSendPayload from "@graphql/types/payload/payment-send"
+import TargetConfirmations from "@graphql/types/scalar/target-confirmations"
+import { validateIsUsdWalletForMutation } from "@graphql/helpers"
+import { Wallets } from "@app"
+
+const OnChainUsdPaymentSendInput = GT.Input({
+  name: "OnChainUsdPaymentSendInput",
+  fields: () => ({
+    walletId: { type: GT.NonNull(WalletId) },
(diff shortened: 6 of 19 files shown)


src/graphql/error.ts
```diff
@@ -153,7 +153,7 @@ export class DustAmountError extends CustomApolloError {
   constructor(errData: CustomApolloErrorData) {
     super({
       code: "ENTERED_DUST_AMOUNT",
-      message: `Use lightning to send amounts less than ${onChainWalletConfig.dustThreshold}`,
+      message: `Use lightning to send amounts less than ${onChainWalletConfig.dustThreshold} sats`,
```
nicolasburtey, line 156: probably unrelated to this PR but surprising to see this message twice one time with `state` argument and another one with `onChainWalletConfig`

src/graphql/main/schema.graphql
```diff
@@ -778,6 +779,14 @@ type OnChainUpdate {
   walletId: WalletId!
 }
 
+input OnChainUsdPaymentSendInput {
+  address: OnChainAddress!
+  amount: CentAmount!
+  memo: Memo
+  targetConfirmations: TargetConfirmations = 1
```
nicolasburtey, line 786: I don't think I would expose this. eventually batching will expose more an enum, like `now` or `within 24 hours` or `within a week`.

src/graphql/root/mutation/onchain-usd-payment-send.ts
```diff
@@ -0,0 +1,86 @@
...
+  }),
+})
+
+const OnChainUsdPaymentSendMutation = GT.Field<
+  {
+    input: {
+      walletId: WalletId | InputValidationError
+      address: OnChainAddress | InputValidationError
+      amount: number
+      memo: Memo | InputValidationError | null
```
nicolasburtey, line 29: if nothing is passed, it default to `null`, not `undefined`?

src/graphql/error.ts
vindard: Looks like one is the default message for that CustomApolloError from global config and the other is because we inject the global config (and `dustThreshold`) into the OnChainPaymentFlowBuilder

src/graphql/root/mutation/onchain-usd-payment-send.ts
vindard: Not sure which is appropriate here. This was copied from the existing other 2 onchain-send mutations and it looks like we use `null` in a few other places in the graphql layer.

src/graphql/main/schema.graphql
vindard: This one is in all our other send and probe mutations as well. We can probably remove it in a follow-up PR? (might be a schema change)

src/graphql/root/mutation/onchain-usd-payment-send.ts
```diff
@@ -0,0 +1,86 @@
...
+    if (memo instanceof Error) {
+      return { errors: [{ message: memo.message }] }
+    }
+
+    if (targetConfirmations instanceof Error) {
+      return { errors: [{ message: targetConfirmations.message }] }
+    }
+
+    const usdWalletValidated = await validateIsUsdWalletForMutation(walletId)
+    if (usdWalletValidated != true) return usdWalletValidated
```
nicolasburtey, line 63:
```suggestion
    if (usdWalletValidated !== true) return usdWalletValidated
```

nicolasburtey: maybe for consistency this should be mapped same as fees?, ie:

```
if (usdWalletValidated instanceof Error) throw usdWalletValidated
```

src/graphql/root/query/on-chain-usd-tx-fee-query.ts
```diff
@@ -0,0 +1,50 @@
...
+  },
+  resolve: async (_, args, { domainAccount }) => {
+    const { walletId, address, amount, targetConfirmations } = args
+
+    for (const input of [walletId, address, amount, targetConfirmations]) {
+      if (input instanceof Error) throw input
+    }
+
+    const usdWalletValidated = await validateIsUsdWalletForQuery(walletId)
+    if (usdWalletValidated instanceof Error) throw usdWalletValidated
```
nicolasburtey, line 32: throw/return inconsistency across query/mutation? maybe it's on purpose, throw on query and return on mutation?

nicolasburtey (commented):
skimmed through the PR and looks very good

test/integration/02-user-wallet/02-send-onchain.spec.ts
```diff
@@ -195,147 +241,237 @@ describe("UserWallet - onChainPay", () => {
...
+  expect(senderSettledTx.settlementAmount).toBe(-senderAmount)
+  expect(senderSettledTx.displayCurrencyPerSettlementCurrencyUnit).toBeGreaterThan(0)
+
+  // Check txn details for received wallet
+  // ===
+  const { result: txsRecipient, error: errorUserA } =
+    await Wallets.getTransactionsForWalletId({
+      walletId: recipientWalletId,
+    })
+  if (errorUserA instanceof Error || txsRecipient === null) {
```
nicolasburtey, line 407: if `txsRecipient === null` it won't be catch as an error

test/integration/02-user-wallet/02-send-onchain.spec.ts
```diff
@@ -195,147 +241,237 @@ describe("UserWallet - onChainPay", () => {
...
-      }
-      const pendingTxs = txResult.result.filter(
-        ({ status }) => status === TxStatus.Pending,
-      )
-      expect(pendingTxs.length).toBe(0)
+  {
+    const txResult = await Wallets.getTransactionsForWalletId({
+      walletId: senderWalletId,
+    })
+    if (txResult.error instanceof Error || txResult.result === null) {
```
nicolasburtey, line 252: if `txResult.result === null` the error won't be catched

test/integration/02-user-wallet/02-send-onchain.spec.ts
```diff
@@ -116,75 +124,113 @@ afterAll(async () => {
...
-    if (satsPrice instanceof Error) throw satsPrice
+    if (satsPrice instanceof Error) return satsPrice
 
-    const paymentAmount = { amount: BigInt(amount), currency: WalletCurrency.Btc }
+    const paymentAmount = { amount: BigInt(amount), currency: senderWallet.currency }
     const displayPaymentAmount = {
-      amount: amount * satsPrice,
+      amount: senderWallet.currency === WalletCurrency.Btc ? amount * satsPrice : amount,
       currency: DefaultDisplayCurrency,
     }
```
nicolasburtey, line 236: `createPushNotificationContent` is only in the non send all branch. there is no notification for sendAll?

src/graphql/root/mutation/onchain-usd-payment-send.ts
vindard: We'd decided that mutations return an `errors` field as part of  a`Payload` type while queries throw at the top level.

I think there's a bigger question about why the onchain fee probe is a query and not a mutation (like the lightning one is), but that would be a schema-breaking change.

src/graphql/root/query/on-chain-usd-tx-fee-query.ts
vindard: Yup exactly https://github.com/GaloyMoney/galoy/pull/2072#discussion_r1058361966

test/integration/02-user-wallet/02-send-onchain.spec.ts
vindard: This is more for the type catcher here. We've done this in a number of places in the tests. With the way the `Partial` return from `getTransactionsForWalletId` works there can never be a `null` txsRecipient value without there being an `errorUserA`.

We could probably clean up how we call that method in follow-up PR if we want to make that clearer.

test/integration/02-user-wallet/02-send-onchain.spec.ts
vindard: https://github.com/GaloyMoney/galoy/pull/2072#discussion_r1058366152

test/integration/02-user-wallet/02-send-onchain.spec.ts
vindard: Good catch, I was supposed to circle back to that. Fixed https://github.com/GaloyMoney/galoy/pull/2072/commits/e3848a1d8c84f8483ce46c823c8b279674092196

nicolasburtey (approved):

nicolasburtey (dismissed):

test/integration/02-user-wallet/02-send-onchain.spec.ts
```diff
@@ -926,3 +901,89 @@ describe("UserWallet - onChainPay", () => {
...
+    })
+
+    it("sends from btc wallet to usd wallet", async () => {
+      const res = await testInternalSend({
+        senderAccount: accountA,
+        senderWalletId: walletIdA,
+        recipientWalletId: walletIdUsdB,
+        senderAmount: amount,
+      })
+      if (res instanceof Error) throw res
```
dolcalmi, line 934: what are you testing here? test should have expects if you are hiding it the helper function then is not a good practice.

test/integration/02-user-wallet/02-send-onchain.spec.ts
```diff
@@ -198,144 +300,192 @@ describe("UserWallet - onChainPay", () => {
...
-      const finalBalance = await getBalanceHelper(walletIdE)
-      expect(finalBalance).toBe(0)
+  it("sends all in a successful payment", async () => {
+    const res = await testExternalSend({
+      senderAccount: accountE,
+      senderWalletId: walletIdE,
+      amount,
+      sendAll: true,
+    })
+    if (res instanceof Error) throw res
```
dolcalmi, line 476: same problem here, this is not a good test

src/graphql/root/mutation/onchain-usd-payment-send.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    if (memo instanceof Error) {
+      return { errors: [{ message: memo.message }] }
+    }
+
+    if (targetConfirmations instanceof Error) {
+      return { errors: [{ message: targetConfirmations.message }] }
+    }
+
+    const usdWalletValidated = await validateIsUsdWalletForMutation(walletId)
```
dolcalmi, line 62: this validation should be part of app layer, also `validateIsUsdWalletForMutation` is using wallet repository directly so that method should be removed and be included in use cases

src/graphql/root/mutation/onchain-usd-payment-send.ts
```diff
@@ -0,0 +1,86 @@
...
+import TargetConfirmations from "@graphql/types/scalar/target-confirmations"
+import { validateIsUsdWalletForMutation } from "@graphql/helpers"
+import { Wallets } from "@app"
+
+const OnChainUsdPaymentSendInput = GT.Input({
+  name: "OnChainUsdPaymentSendInput",
+  fields: () => ({
+    walletId: { type: GT.NonNull(WalletId) },
+    address: { type: GT.NonNull(OnChainAddress) },
+    amount: { type: GT.NonNull(CentAmount) },
```
dolcalmi, line 17: for lighting invoices we dont have this problem but if an user is reading a qr with amount we will need to add sats amount. Probably a problem for mobile wallet and to fix in an additional PR but just to keep in mind @UncleSamtoshi @daviroo

src/graphql/main/mutations.ts
```diff
@@ -72,6 +73,7 @@ const fields = {
   onChainAddressCreate: OnChainAddressCreateMutation,
   onChainAddressCurrent: OnChainAddressCurrentMutation,
   onChainPaymentSend: OnChainPaymentSendMutation,
+  onChainUsdPaymentSend: OnChainUsdPaymentSendMutation,
```
dolcalmi, line 76: you are missing shield update, this mutation should use `isAuthenticated` in graphql-main-server

src/graphql/main/queries.ts
```diff
@@ -27,6 +28,7 @@ const fields = {
   btcPrice: BtcPriceQuery,
   btcPriceList: BtcPriceListQuery,
   onChainTxFee: OnChainTxFeeQuery,
+  onChainUsdTxFee: OnChainUsdTxFeeQuery,
```
dolcalmi, line 31: you are missing shield update, this query should use isAuthenticated in graphql-main-server

src/graphql/helpers.ts
```diff
@@ -48,7 +48,19 @@ export const validateIsBtcWalletForQuery = async (
...
+  if (wallet.currency !== WalletCurrency.Btc) {
+    return QueryDoesNotMatchWalletCurrencyError
+  }
+  return true
+}
+
+export const validateIsUsdWalletForQuery = async (
+  walletId: WalletId,
+): Promise<true | CustomApolloError> => {
+  const wallet = await WalletsRepository().findById(walletId)
```
dolcalmi, line 60: this validation is responsibility of app layer use cases

dolcalmi (changes requested):
looks good, just a couple of blockers

test/integration/02-user-wallet/02-send-onchain.spec.ts
vindard: This is to de-duplicate the tests that run in that helper function. The function is re-used 8 times in this file. I'm not sure of better ways to de-duplicate these here, but open to suggestions.

test/integration/02-user-wallet/02-send-onchain.spec.ts
vindard: https://github.com/GaloyMoney/galoy/pull/2072#discussion_r1058587750

src/graphql/root/mutation/onchain-usd-payment-send.ts
vindard: Oh yea good point. These are used in lots of mutation/queries here. Maybe we can change all of these in a follow-up PR to keep the diff in this one clear?

src/graphql/helpers.ts
vindard: https://github.com/GaloyMoney/galoy/pull/2072#discussion_r1058588580

src/graphql/main/queries.ts
vindard: Good catch! Done: https://github.com/GaloyMoney/galoy/pull/2072/commits/400ba13a8c53504207c1c21484874e6538c07e00

src/graphql/main/mutations.ts
vindard: https://github.com/GaloyMoney/galoy/pull/2072#discussion_r1058591644

dolcalmi (approved):

PR #2076, opened 2022-12-23
chore: remove no longer needed condition

nicolasburtey:
this condition was here during the upgrade of not using kratos to using kratos, where the user would be upgrade if the migration script had not completed yet

this is no longer needed. all accounts has an associated kratos identity


ntheile:
Looks like the e2e test is failing with `CouldNotFindUserFromPhoneError()`.

PR #2077, opened 2022-12-24
test: remove okex redis mock & use stablesats in-memory internal mock

vindard:
## Description

This PR is to configure stablesats to use its new in-memory price feed (mocked) and to remove unused redis price publish mock functions.


nicolasburtey:
we may also be able to remove `stablesats` from the dependancy on `integrations` test: https://github.com/GaloyMoney/galoy/blob/main/docker-compose.yml#L28?

bodymindarts (approved):

PR #2078, opened 2022-12-24
feat: kratos userLogout mutation

ntheile:
New graphql mutation for user logout. It only logs the user out of the current session. If a user is logged in on two devices the other device will remain logged in.

```gql
mutation {
  userLogout( input: {
    authToken: "AUTH_TOKEN"
  }) { 
  errors {
    message
  }
    authToken
  }
}
```

### DIFF
--- a/src/graphql/root/mutation/user-logout.ts
+++ b/src/graphql/root/mutation/user-logout.ts
@@ -8,7 +8,7 @@ import { mapAndParseErrorForGqlResponse } from "@graphql/error-map"
 const UserLogoutInput = GT.Input({
   name: "UserLogoutInput",
   fields: () => ({
-    token: {
+    authToken: {
       type: GT.NonNull(AuthToken),
     },
   }),
@@ -16,7 +16,7 @@ const UserLogoutInput = GT.Input({
 
 const UserLogoutMutation = GT.Field<{
   input: {
-    token: string | InputValidationError
+    authToken: string | InputValidationError
   }
 }>({
   extensions: {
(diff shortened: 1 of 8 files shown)


src/graphql/root/mutation/user-logout.ts
```diff
@@ -0,0 +1,40 @@
...
+  type: GT.NonNull(AuthTokenPayload),
+  args: {
+    input: { type: GT.NonNull(UserLogoutInput) },
+  },
+  resolve: async (_, args) => {
+    const { authToken } = args.input
+    if (authToken instanceof Error) return
+    const authService = AuthWithPhonePasswordlessService()
+    const logoutResp = await authService.logout(authToken)
+    if (logoutResp instanceof Error)
```
nicolasburtey, line 34: You need to create a use case for it in apps. Graphql code should not call service code directly

nicolasburtey (dismissed):
Hexagonal pattern boundary issues to fix

src/graphql/root/mutation/user-logout.ts
```diff
@@ -0,0 +1,39 @@
...
+  extensions: {
+    complexity: 120,
+  },
+  type: GT.NonNull(AuthTokenPayload),
+  args: {
+    input: { type: GT.NonNull(UserLogoutInput) },
+  },
+  resolve: async (_, args) => {
+    const { authToken } = args.input
+    if (authToken instanceof Error) return
```
nicolasburtey, line 31: you should return an error like line 34 or 35

nicolasburtey (approved):

PR #2083, opened 2022-12-25
wip phone prefix

nicolasburtey:
this should remove some of the yield farming earn rewards, and sms flooding attacks

let me know what you think about the approach. I'll do some tests if this sounds good.

### DIFF
--- a/galoy.yaml
+++ b/galoy.yaml
@@ -90,6 +90,10 @@ rateLimits:
     points: 4
     duration: 3600
     blockDuration: 10800
+  requestPhoneCodePerPhonePrefix:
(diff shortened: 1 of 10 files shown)


dolcalmi:
I dont think this is a good solution.. we will end with a really long list in the config and also this does not protect against voip or temp phone number providers

nicolasburtey:
> I dont think this is a good solution.. we will end with a really long list in the config and also this does not protect against voip or temp phone number providers

voip can be dealt with another way: those phone type can be removed from the rewards program. temp phone number providers also typically give back voip number.

PR #2103, opened 2022-12-29
feat: enable cookie auth and phone+code or email+password kratos flows

ntheile:


src/servers/middlewares/kratos-router.ts
```diff
@@ -37,34 +41,47 @@ kratosRouter.post(
...
+        })
+      } else if (email) {
+        // email+password flow
+        // kratos user likely exists from self registration flow
+        account = await createAccountForEmailIdentifier({
+          kratosUserId: userIdChecked,
+          config: getDefaultAccountsConfig(),
+        })
+      } else {
+        // insert new flow, such as email with code
```
nicolasburtey, line 82: should probably throw/return error in the meantime.

src/servers/middlewares/kratos-router.ts
```diff
@@ -37,34 +41,47 @@ kratosRouter.post(
...
+          res.status(400).send("invalid phone")
+          return
+        }
+        account = await createAccountWithPhoneIdentifier({
+          newAccountInfo: { phone, kratosUserId: userIdChecked },
+          config: getDefaultAccountsConfig(),
+        })
+      } else if (email) {
+        // email+password flow
+        // kratos user likely exists from self registration flow
```
nicolasburtey, line 76:
```suggestion
        // kratos user exists from self registration flow
```

registration is a private API that should also be called by kratos which mean kratos user is always in the process of been created when this is been called through kratos webhook

nicolasburtey (commented):
LGTM

nicolasburtey (approved):

PR #2105, opened 2022-12-30
refactor: move validate wallet methods from gql to app layer

vindard:
Based on this feedback: https://github.com/GaloyMoney/galoy/pull/2072#discussion_r1058548679

### DIFF
--- a/src/app/accounts/add-earn.ts
+++ b/src/app/accounts/add-earn.ts
@@ -76,12 +76,7 @@ export const addEarn = async ({
   const shouldGiveReward = await RewardsRepository(accountId).add(quizQuestionId)
   if (shouldGiveReward instanceof Error) return shouldGiveReward
 
-  const intraledgerPaymentSendFn =
-    funderWallet.currency === WalletCurrency.Btc
-      ? Payments.intraledgerPaymentSendWalletIdForBtcWallet
-      : Payments.intraledgerPaymentSendWalletIdForUsdWallet
-
-  const payment = await intraledgerPaymentSendFn({
+  const payment = await Payments.intraledgerPaymentSendWalletIdForBtcWallet({
     senderWalletId: funderWalletId,
     recipientWalletId,
     amount,
--- a/src/app/accounts/add-earn.ts
+++ b/src/app/accounts/add-earn.ts
@@ -76,7 +76,12 @@ export const addEarn = async ({
   const shouldGiveReward = await RewardsRepository(accountId).add(quizQuestionId)
   if (shouldGiveReward instanceof Error) return shouldGiveReward
 
-  const payment = await Payments.intraledgerPaymentSendWalletId({
+  const intraledgerPaymentSendFn =
+    funderWallet.currency === WalletCurrency.Btc
+      ? Payments.intraledgerPaymentSendWalletIdForBtcWallet
+      : Payments.intraledgerPaymentSendWalletIdForUsdWallet
+
+  const payment = await intraledgerPaymentSendFn({
     senderWalletId: funderWalletId,
     recipientWalletId,
     amount,
--- a/src/graphql/root/mutation/intraledger-payment-send.ts
+++ b/src/graphql/root/mutation/intraledger-payment-send.ts
@@ -1,7 +1,6 @@
 import { Accounts, Payments } from "@app"
 import { checkedToWalletId } from "@domain/wallets"
 import { mapAndParseErrorForGqlResponse } from "@graphql/error-map"
-import { validateIsBtcWallet } from "@app/wallets"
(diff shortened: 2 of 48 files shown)


src/graphql/root/mutation/intraledger-payment-send.ts
```diff
@@ -44,7 +44,9 @@ const IntraLedgerPaymentSendMutation = GT.Field({
     }
 
     const btcWalletValidated = await validateIsBtcWalletForMutation(walletId)
-    if (btcWalletValidated != true) return btcWalletValidated
+    if (btcWalletValidated instanceof Error) {
```
nicolasburtey, line 47: shoudn't this check be in `app/`?

vindard: The validation is tied to the specific query/mutation it's being called in. Not sure if app layer would have this context to apply the check.

image

vindard: Actually, I guess we could differentiate this by making new app layer use-case methods. I can do that here.

vindard: Done: https://github.com/GaloyMoney/galoy/pull/2105/commits/02fe1376f8d159b9ade51184449d025dd1e7c8ba

src/app/accounts/add-earn.ts
```diff
@@ -76,7 +76,12 @@ export const addEarn = async ({
   const shouldGiveReward = await RewardsRepository(accountId).add(quizQuestionId)
   if (shouldGiveReward instanceof Error) return shouldGiveReward
 
-  const payment = await Payments.intraledgerPaymentSendWalletId({
+  const intraledgerPaymentSendFn =
```
nicolasburtey, line 79: my assumption was that funder wallet would always be a Btc wallet but I guess you want to remove this assumption?

test/helpers/user.ts
```diff
@@ -240,7 +240,11 @@ export const createAndFundNewWallet = async <S extends WalletCurrency>({
   if (wallet instanceof Error) throw wallet
   // Fund new wallet if a non-zero balance is passed
   if (balanceAmount.amount === 0n) return wallet
-  const lnInvoice = await Wallets.addInvoiceForSelf({
+  const addInvoiceFn =
```
nicolasburtey, line 243: why did we had not a segregation per wallet before?

I found the function below pretty "hard" to read compare to the previous flow. not an issue really just an impression

nicolasburtey (commented):
some questions

src/app/accounts/add-earn.ts
vindard: No strong opinions on my end. I just wasn't sure what the assumption was. Fixed here: https://github.com/GaloyMoney/galoy/pull/2105/commits/7dadb9ad6e3e9cf0426612db7fe1de8d1c35ce78

test/helpers/user.ts
vindard: Yea the changes here (https://github.com/GaloyMoney/galoy/compare/4e85159da684722efece05fff5baba28d9c5a9cf...move-validate-wallet-methods) ended up being quite a lot.

I did it this way because moving the wallet validation into the use-cases meant that the undifferentiated functions should be private to those modules and we should probably only expose the wallet-specific methods. This meant that the undifferentiated methods were no longer available for import into the tests and we'd now have to do explicit differentiation like with here.

nicolasburtey: maybe I'm not a fan of a ternary operator also. the main reason is we're assuming only two base currency, but this is just a temporary thing. eventually we will manage more currencies (not just display currency but base currency) with fiat integration.

nicolasburtey: so a `switch` of explicitly one liner `if` could be not only easier to read but also easier to expand.

vindard: I'm not sure I'd include a switch here to handle 3+ base currencies. I think at that point we'd have a lot more refactoring to do to pass in amounts + currencies instead of having separate use-cases for each base currency. Including a switch here feels like partial/premature optimization for that.

An ok path to keep readability too could be to revert the last 3 commits and leave the validation call in the queries/mutations. Since this was supposed to be a simple PR to move those methods into the app layer, that could potentially be an ok middle-ground without having to get into refactoring too many other things prematurely.

nicolasburtey: > I'm not sure I'd include a switch here to handle 3+ base currencies. I think at that point we'd have a lot more refactoring to do to pass in "amounts + currencies" instead of having separate use-cases for each base currency. Including a switch here feels like partial/premature optimization for that

this is fair. approved.

PR #2107, opened 2022-12-31
chore: make logs more meaningful

nicolasburtey:
currently I'm using the logs when doing local development.

I don't believe we have any meaningful use of logs in prod (given we rely on honeycomb), so local dev is probably the most relevat

with `wrapSerializers: false`, the logs are very long.

really the things that matter for me is to get the body query+variables requests

this PR is reducing drastically the number of output and just show the relevant outputs


vindard (approved):
Oh this is much cleaner for local dev!

PR #2108, opened 2022-12-31
feat: combine both admin and public endpoint in the same server

nicolasburtey:
there is no reason to launch a dedicated admin server. it makes local dev more memory/cpu heavy.

this is what kratos is doing and I think this is a preferrable approach - the only downside is that if there is a crash on the admin, currently it won't affect public users. but it will with this refactoring

right now this is just a test, we have to do some work on the CI before doing a broader refactoring of the /server files.

### DIFF
--- a/src/servers/graphql-main-server.ts
+++ b/src/servers/graphql-main-server.ts
@@ -13,6 +13,7 @@ import { gqlMainSchema, mutationFields, queryFields } from "@graphql/main"
 
 import { isAuthenticated, startApolloServer } from "./graphql-server"
 import { walletIdMiddleware } from "./middlewares/wallet-id"
+import { startApolloServerForAdminSchema } from "./graphql-admin-server"
 
 const graphqlLogger = baseLogger.child({ module: "graphql" })
 
@@ -56,7 +57,10 @@ if (require.main === module) {
   setupMongoConnection(true)
     .then(async () => {
       activateLndHealthCheck()
-      await startApolloServerForCoreSchema()
+      await Promise.race([
+        startApolloServerForCoreSchema(),
+        startApolloServerForAdminSchema(),
+      ])
     })
     .catch((err) => graphqlLogger.error(err, "server error"))
 }


src/servers/graphql-main-server.ts
```diff
@@ -56,7 +57,10 @@ if (require.main === module) {
   setupMongoConnection(true)
     .then(async () => {
       activateLndHealthCheck()
-      await startApolloServerForCoreSchema()
+      await Promise.race([
```
vindard, line 60: Why not `Promise.all` here?

nicolasburtey: because I think we want to quit as soon as one critical/non recoverable arise?

nicolasburtey: maybe this should be the behavior for an error, but not for a clean shutdown, in that case we should wait for both servers to stop?

vindard: Ah right, I hadn't thought about if either of those promises gets rejected. I'm not actually sure what graceful shutdown looks like for these servers currently. Maybe @dolcalmi has some ideas too

dolcalmi (commented):
does this requires changes in charts? @krtk6160

nicolasburtey:
> does this requires changes in charts? @krtk6160

not necessarily at first with this PR only. because the admin server will still be built in addition to the public+server which will be expose solely on the public API.

but eventually yes the idea is to remove one Pod from the cluster.

nicolasburtey:
> > does this requires changes in charts? @krtk6160
> 
> not necessarily at first with this PR only. because the admin server will still be built in addition to the public+server which will be expose solely on the public API.
> 
> but eventually yes the idea is to remove one Pod from the cluster.

so to make it a smooth transition we can:
1. approve and deploy this PR
2. do the chart upgrade (without any need to sync with 1.)
3. refactor the code to only have server on the dev side

krtk6160:
> does this requires changes in charts? @krtk6160

I've built the docker image and tested it locally, looks like we won't need any changes in charts for this

krtk6160 (approved):

nicolasburtey:
> > does this requires changes in charts? @krtk6160
> 
> I've built the docker image and tested it locally, looks like we won't need any changes in charts for this

there will be an improvement eventually on this: we don't need to launch 3 instances (2 public + 1 admin) anymore. we can just launch 2 and admin would also have redundancy.

we can talk about it on daily dev.

krtk6160:
> > > does this requires changes in charts? @krtk6160
> > 
> > 
> > I've built the docker image and tested it locally, looks like we won't need any changes in charts for this
> 
> there will be an improvement eventually on this: we don't need to launch 3 instances (2 public + 1 admin) anymore. we can just launch 2 and admin would also have redundancy.
> 
> we can talk about it on daily dev.

Yep I agree. My response was to @dolcalmi in that this particular PR doesn't required a change on charts and can be merged independently.

PR #2109, opened 2022-12-31
chore: combine auth and unauth runtime object

nicolasburtey:
small refactoring removing some smelly code. still not perfect but at least more readable and remove some duplicate.


vindard (approved):
Looks pretty straightforward to me

PR #2110, opened 2022-12-31
feat: ws upgrade (attempt 4)

nicolasburtey:
previous attempt (https://github.com/GaloyMoney/galoy/pull/1780) was trying to both upgrade the ws socket library to `graphql-ws` and also segregates the ws socket server from the main server (a recommended best practice to get apollo federation to work). 

the second part is the harder part (more code/graphql schema moving around, more to do on the infra side to keep things to map the backend code), and is the reason this PS has not been completed yet.

I instead go with this PR with [this code snippet](https://github.com/enisdenjo/graphql-ws#ws-backwards-compat) that let run both the old and new websocket protocol simultaneously. the goal of this PR is now to only upgrade the websocket library. 

the seggregation of the HTTP server and ws server can be done in another PR later.

### DIFF
--- a/src/servers/graphql-server.ts
+++ b/src/servers/graphql-server.ts
@@ -1,15 +1,14 @@
 import { createServer } from "http"
 
-import cors from "cors"
 import { Accounts } from "@app"
 import { getApolloConfig, getGeetestConfig, getJwksArgs, isDev } from "@config"
 import Geetest from "@services/geetest"
 import { baseLogger } from "@services/logger"
 import {
   ACCOUNT_USERNAME,
+  SemanticAttributes,
   addAttributesToCurrentSpan,
   addAttributesToCurrentSpanAndPropagate,
-  SemanticAttributes,
 } from "@services/tracing"
 import {
   ApolloServerPluginDrainHttpServer,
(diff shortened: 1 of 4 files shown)


src/servers/graphql-server.ts
```diff
@@ -292,10 +297,31 @@ export const startApolloServer = async ({
...
         ignore: (req) => req.url === "/healthz",
       },
+      serializers: {
+        res: (res) => ({ statusCode: res.statusCode }),
+        req: (req) => ({
+          id: req.id,
+          method: req.method,
+          url: req.url,
+          remoteAddress: req.remoteAddress,
+          // headers: req.headers,
```
daviroo, line 322: Why don't we need the headers here?

src/servers/graphql-server.ts
```diff
@@ -320,77 +346,167 @@ export const startApolloServer = async ({
...
+      ? connectionParams?.ip
+      : connectionParams?.["x-real-ip"] || connectionParams?.["x-forwarded-for"]
+
+    const ip = parseIps(ipString)
+
+    const authz = (connectionParams.authorization || connectionParams.Authorization) as
+      | string
+      | undefined
+
+    // TODO: also manage the case where there is a cookie in the request
```
daviroo, line 423: Is this a copy of the above code at line 362?

daviroo (changes requested):
Nothing major just some questions on a few things.

src/servers/graphql-server.ts
nicolasburtey: wrong commit, this commit should be part of this PR: https://github.com/GaloyMoney/galoy/pull/2107

but the answer to your question is that header are leaking out credentials

src/servers/graphql-server.ts
nicolasburtey: yes, I want to sync with @ntheile on this

daviroo (approved):

PR #2117, opened 2023-01-03
chore: remove legacy entry builder

vindard:
## Description

This PR swaps that last remaining ledger `addOnChainTxReceipt` for the corresponding LedgerFacade method, and then removes the now unused `LegacyEntryBuilder`.


src/app/wallets/update-on-chain-receipt.ts
```diff
@@ -140,20 +147,60 @@ const processTxForWallet = async (
...
-            sats,
-            fee,
-            amountDisplayCurrency,
+          const feeBtc = paymentAmountFromNumber({
+            amount: fee,
+            currency: WalletCurrency.Btc,
+          })
+          if (feeBtc instanceof Error) return feeBtc
+
+          const walletAddressReceiver = await WalletAddressReceiver({
```
bodymindarts, line 162: I don't know if this is from this PR or already in the code base but an `async` constructor function is very suspicious to me.

src/domain/wallet-invoices/index.types.d.ts
```diff
@@ -84,8 +89,15 @@ type WalletInvoiceReceiver = WalletInvoice & {
   receivedAmount: () => BtcPaymentAmount | UsdPaymentAmount
 }
 
-type WalletInvoiceReceiverArgs = {
-  walletInvoice: WalletInvoice
+type WalletAddressReceiver<S extends WalletCurrency> = WalletAddress<S> & {
```
bodymindarts, line 92: Something seems wrong with this type. Don't quite understand what yet.
Not sure of the name or the bundling of the arguments.

src/domain/wallet-on-chain-addresses/wallet-address-receiver.ts
```diff
@@ -0,0 +1,62 @@
...
+      btcBankFee: feeBtc,
+    }
+
+    return {
+      ...walletAddress,
+      btcToCreditReceiver: calc.sub(receivedBtc, feeBtc),
+      usdToCreditReceiver: calc.sub(receivedUsd, bankFee.usdBankFee),
+      recipientWalletDescriptor: walletAddress.recipientWalletDescriptor,
+      ...bankFee,
+      receivedAmount: () =>
```
bodymindarts, line 33: Why is this lazy. the `xxxDescriptor.currency` can never change during the lifecycle of this state.

bodymindarts (commented):
Needs follow up conversation - no unit tests.

src/app/wallets/update-on-chain-receipt.ts
vindard: I pretty much modeled this from the `WalletInvoiceReceiver`. Async in both cases is because we call conversion methods from the dealer async. Could potentially be a code smell 🤔

src/domain/wallet-on-chain-addresses/wallet-address-receiver.ts
vindard: It's a pattern we seem to use in the `WalletInvoiceReceiver` and places like `PaymentFlow` to access internal values.

vindard:
> no unit tests.

Ah sorry missed this! I had this in mind to implement and got distracted I think. Done now: https://github.com/GaloyMoney/galoy/pull/2117/commits/83f67c103df941b1d102e980872d398cd2016261

src/domain/wallet-invoices/index.types.d.ts
vindard: I looked at the original `WalletInvoiceReceiver` type again and made some changes. Do these help? https://github.com/GaloyMoney/galoy/compare/2c87ed30f2d0be6b86b3057683d467fde0ed77f9...60fb5fa5d03bf55d513c5bf69d5e72d6405e47b6

PR #2138, opened 2023-01-06
chore: add migration for 'usd' prop in existing onchain sends

vindard:
## Description

**_Supports #2129_**

This migration only targets onchain sends (BTC & USD) that have the new centsAmount & centsFee properties. These would have started with the new onchain-payment-builder flow.

**This will only be merged and deployed after #2129 has been deployed.**

### DIFF
--- a/src/migrations/20230106142545-fix-onchain-send-usd.ts
+++ b/src/migrations/20230106142545-fix-onchain-send-usd.ts
@@ -1,0 +1,28 @@
+module.exports = {
+  async up(db) {
+    const collection = db.collection("medici_transactions")
+    const newUsd = {
+      $round: [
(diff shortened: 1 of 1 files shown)


src/migrations/20230106142545-fix-onchain-send-usd.ts
```diff
@@ -0,0 +1,28 @@
+module.exports = {
```
dolcalmi, line 1: did you test it locally?

vindard: Yup, both on old-code-version and new-code-version txns

dolcalmi (approved):

PR #2149, opened 2023-01-10
feat: kratos cookie flow

samerbuna:
- feat: kratos cookie flow
- feat: cookieLogout endpoint
- fix: update login endpoint url
- refactor: use phoneNumber/authCode and return identity


ntheile:
Todo

1. test cookieLogouts, especially bad cookie state
2. cookie login tests
3. rate limiting code

ntheile:
@samerbuna ,

The code still need a bunch of cleanup and refactoring but here is the logout flow. I also added an endpoint that should help clear out any lingering cookies:

1. `/auth/clearCookies` Clear out any lingering cookies
2. `/auth/login` Login and get cookies
3.  `auth/logout`  Logout and revoke session 

1) Before you log in clear out any lingering cookies `/auth/clearCookies`
```
var requestOptions = {
  method: 'GET',
  redirect: 'follow',
  credentials: 'include'
};
fetch("http://localhost:4002/auth/clearCookies", requestOptions)
  .then(response => response.text())
  .then(result => console.log(result))
  .catch(error => console.log('error', error));
```

2) Login and get cookies `/auth/login`
```
var myHeaders = new Headers();
myHeaders.append("Content-Type", "application/json");
var raw = JSON.stringify({
  "phoneNumber": "+1XXXXXXX",
  "authCode": "XXXXXX"
});
var requestOptions = {
  method: 'POST',
  headers: myHeaders,
  body: raw,
  redirect: 'follow',
  credentials: 'include'
};
fetch("http://localhost:4002/auth/login", requestOptions)
  .then(response => response.text())
  .then(result => console.log(result))
  .catch(error => console.log('error', error));
```

3) you can test if you are logged by hitting this endpoint in your browser
it will return data if you have a kratos cookie
```
http://localhost:4433/sessions/whoami
```

4) Logout and revoke session `auth/logout` * I made this a GET request just in case the cookie state get messy on the client, the user will at least be able to hit this endpoint in a browser tab to clean cookie state
```
var requestOptions = {
  method: 'GET',
  redirect: 'follow',
  credentials: 'include'
};
fetch("http://localhost:4002/auth/logout", requestOptions)
  .then(response => response.text())
  .then(result => console.log(result))
  .catch(error => console.log('error', error));
```

ntheile:
@samerbuna ,

I added an oathkeeper rule to allow the `/auth/clearCookies` endpoint to be hit even with expired or revoked cookies, to allow the server to delete them, here: https://github.com/GaloyMoney/galoy/pull/2149/commits/67d6a7e0eb4e73207d90f3b1013f5a9d036636f4
 
One caveat is you need to call the `4012` endpoint to hit oathkeeper like this (not sure why the 4002 endpoint does not work, any ideas @nb1) :

```
var requestOptions = {
  method: 'GET',
  redirect: 'follow',
  credentials: 'include'
};
fetch("http://localhost:4012/auth/clearCookies", requestOptions)
  .then(response => response.text())
  .then(result => console.log(result))
  .catch(error => console.log('error', error));
```

So now you should be able to handle 401 errors in the web wallet with a global http handler to hit the `/auth/clearCookies` in the event that the server revokes the cookies

src/servers/middlewares/auth-router.ts
```diff
@@ -114,4 +119,82 @@
...
+  }
+
+  res.send({
+    identity: {
+      id: loginRes.kratosUserId,
+      uid: loginRes.kratosUserId,
+      phoneNumber,
+    },
+  })
+})
```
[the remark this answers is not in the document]
ntheile: fixed it

dev/ory/oathkeeper_rules.yaml
```diff
@@ -10,6 +10,22 @@
     handler: allow
   mutators:
     - handler: noop
+
+- id: allow-clear-cookies-endpoint
+  upstream:
+    url: "http://e2e-tests:4012"
```
nicolasburtey, line 16: request going to `<(http|https)>://<[a-zA-Z0-9-.:]+>/auth/clearCookies` won't go to the upstream as I understand it?

src/domain/authentication/index.types.d.ts
```diff
@@ -23,17 +26,30 @@ type WithSessionResponse = {
   kratosUserId: UserId
 }
 
+type WithCookieResponse = {
+  cookiesToSendBackToClient: Array<KratosCookie>
```
nicolasburtey, line 30: do we send back different type of Cookie, and those changes over time? if it's all the same, we could type of the cookie we need more specifically instead of having an `Array`, ie: 
```
{ 
  cookieSession: CookieSession
  cookieXSS:  CookieXSS
  kratosUserId: UserId
}
```

src/servers/middlewares/auth-router.ts
```diff
@@ -114,4 +121,135 @@ authRouter.post(
...
+    namespace: "auth-router",
+    fn: async (req: express.Request, res: express.Response) => {
+      try {
+        const ipString = isDev ? req?.ip : req?.headers["x-real-ip"]
+        const ip = parseIps(ipString)
+        if (ip === undefined) {
+          throw new Error("IP is not defined")
+        }
+
+        const phone = req.body.phoneNumber as PhoneNumber
```
nicolasburtey, line 136: should verify type with `checkedToPhoneNumber`. this is typically done in the use case not in the network layer.

src/servers/middlewares/auth-router.ts
```diff
@@ -114,4 +121,135 @@ authRouter.post(
...
+        const code = req.body.authCode
+
+        const loginResp = (await Auth.loginWithPhone({
```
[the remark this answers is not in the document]
nicolasburtey: agree. would make sense to get a `loginwithPhoneToken` and `loginWithPhoneCookie` use cases

dev/ory/oathkeeper_rules.yaml
```diff
@@ -10,6 +10,22 @@
...
+  match:
+    url: "<(http|https)>://<[a-zA-Z0-9-.:]+>/auth/clearCookies"
+    methods: ["GET", "POST", "OPTIONS"]
+  authenticators:
+    - handler: anonymous
+  authorizer:
+    handler: allow
+  mutators:
+    - handler: delete_cookies
```
nicolasburtey, line 25: where is there a documentation for this one?

can't find anything here: https://www.ory.sh/docs/oathkeeper/pipeline/mutator#cookie-configuration

nicolasburtey: this settings doesn't sound right to me. I think if the goal is to have `cookie`, it is meant to inject a cookie from the authenticator. ie: for rewriting a header to a cookie.

I don't see any `delete_cookies` mutator

src/services/kratos/auth-phone-no-password.ts
```diff
@@ -189,8 +281,10 @@ export const AuthWithPhonePasswordlessService = (): IAuthWithPhonePasswordlessSe
 
   return {
     login,
```
nicolasburtey, line 283: now we're introducing cookie, this function should be also probably updated as `loginToken`

test/e2e/servers/ory/kratos.spec.ts
```diff
@@ -517,3 +517,40 @@ describe("decoding link header", () => {
...
+  it("login with cookie", async () => {
+    const res = await authService.createIdentityWithCookie(phone)
+    if (res instanceof Error) throw res
+    expect(res).toHaveProperty("kratosUserId")
+    expect(res).toHaveProperty("cookiesToSendBackToClient")
+    cookies = res.cookiesToSendBackToClient
+  })
+
+  it("logout and revoke cookies", async () => {
+    for (const c of cookies) {
```
nicolasburtey, line 536: I know it's just used for 2 lines, but AFAIK, we try to not use 1 letter variable in the codebase. 

`for (const cookie of cookies) {` would probably still fit in on a single line.

nicolasburtey (dismissed):
few suggestions for improvement

dev/ory/oathkeeper_rules.yaml
```diff
@@ -10,6 +10,22 @@
     handler: allow
   mutators:
     - handler: noop
+
+- id: allow-clear-cookies-endpoint
```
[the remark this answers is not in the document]
ntheile: here https://github.com/GaloyMoney/charts/pull/2393

src/domain/authentication/index.types.d.ts
ntheile: should be just the csrf token and kratos session cookies, unless kratos decides to change the api on us :)

dev/ory/oathkeeper_rules.yaml
ntheile: got it from chatgpt 
Screen Shot 2023-01-17 at 9 50 32 PM

src/app/auth/login.ts
```diff
@@ -54,23 +54,65 @@ export const loginWithPhone = async ({
...
+    const limitOk = await checkFailedLoginAttemptPerIpLimits(ip)
+    if (limitOk instanceof Error) return limitOk
+  }
+
+  {
+    const limitOk = await checkFailedLoginAttemptPerPhoneLimits(phone)
+    if (limitOk instanceof Error) return limitOk
+  }
+
+  // TODO:
```
bodymindarts, line 90: Is this the right place for this?

src/app/auth/login.ts
```diff
@@ -54,23 +54,65 @@ export const loginWithPhone = async ({
...
+  const validCode = await isCodeValid({ phone, code })
+  if (validCode instanceof Error) return validCode
+
+  await rewardFailedLoginAttemptPerIpLimits(ip)
+  await rewardFailedLoginAttemptPerPhoneLimits(phone)
+
+  const authService = AuthWithPhonePasswordlessService()
+
+  let kratosResult = await authService.loginCookie(phone)
+  // FIXME: this is a fuzzy error.
```
bodymindarts, line 103: What is there to fix?

bodymindarts (approved):
Too many comments for my taste - but it's not a big deal. Might want to bring it up at ACE to see if other ppl feel they are justified. Approving to move on as that is more of a style question. Production code LGTM

src/app/auth/login.ts
ntheile: i copied that code from @nicolasburtey loginToken. will defer to him

src/app/auth/login.ts
ntheile: cc @nicolasburtey

ntheile:
> Too many comments for my taste - but it's not a big deal. Might want to bring it up at ACE to see if other ppl feel they are justified. Approving to move on as that is more of a style question. Production code LGTM

Yeah, i probably over-commented because there is  a lot of nuances with Kratos and the custom way in which we are using it. Will make a `docs/kratos.md` in a follow up PR

PR #2151, opened 2023-01-10
fix: handle pending payments when failed payments with same hash exist

vindard:
## Description

This PR fixes an issue where our `updatePendingPayments` isn't able to progress past the `isLnTxRecorded` step because it wasn't able to register if pending ledger transactions **_also exist_** alongside settled failed ones.

- The first step was to adjust the `isLnTxRecorded` check

- The next step was to then fail a pending payment if its amount did not match the amount we got back from lnd (in the case of different amounts used for no-amount invoices)

### Note
This PR assumes that there can't be 2 simultaneous **pending** payments (see https://github.com/GaloyMoney/galoy/pull/2151/commits/a4f9e06578ad73a7046e5a4d71309c37b407118b) for the same hash with the same correct amount. It assumes there can only be 1 pending payment for the correct amount and multiple other failed payments for the same hash. 

If this assumption does not hold then there is the risk that multiple payments out get settled for the same single lnd payment which would mean that users would notice an extra payment out. If we think this assumption isn't fair, we can consider adding some provision to account for potentially multiple pending payments (or we can wait until a user observes this in actual usage).


src/app/payments/update-pending-payments.ts
```diff
@@ -179,31 +184,11 @@ const updatePendingPayment = wrapAsyncToRunInSpan({
...
-            paymentFlow,
-            journalId: pendingPayment.journalId,
-            actualFee: roundedUpFee,
-            revealedPreImage,
-          })
-        } else if (status === PaymentStatus.Failed) {
+        if (
+          status === PaymentStatus.Failed ||
+          // pendingPayment is a different version to latest payment from lnd
+          satsAmount !== toSats(paymentFlow.btcPaymentAmount.amount)
```
dolcalmi, line 190: this comparison should be made with invoice amount because fees can change depending on the route

src/app/payments/update-pending-payments.ts
```diff
@@ -229,6 +214,30 @@ const updatePendingPayment = wrapAsyncToRunInSpan({
               return setErrorCritical(reimbursed)
             }
           }
+        } else if (status === PaymentStatus.Settled) {
```
dolcalmi, line 217: if we return `else` is not necessary

src/app/payments/update-pending-payments.ts
```diff
@@ -240,31 +249,44 @@ const updatePendingPayment = wrapAsyncToRunInSpan({
...
   const senderAccount = await AccountsRepository().findById(senderWallet.accountId)
   if (senderAccount instanceof Error) return senderAccount
 
+  const filteredPayments = payments.filter((tx) =>
+    senderWallet.currency === WalletCurrency.Btc
+      ? tx.satsAmount === inputAmount
+      : tx.centsAmount === inputAmount,
+  )
+  if (!(filteredPayments && filteredPayments.length))
+    return new CouldNotFindTransactionError()
```
dolcalmi, line 287: not sure about this.... so I cant re-try a failed invoice? an example can be done with OBW, create an invoice, close the app, try a payment (it will fail because is not connected) then open OBW wait to connect and try payment again (should succeed)

dolcalmi (changes requested):

src/app/payments/update-pending-payments.ts
vindard: I think both sides of this _is_ the invoice amount... is it?

src/app/payments/update-pending-payments.ts
vindard: Clean up here https://github.com/GaloyMoney/galoy/pull/2151/commits/242ce03ad013963d01e656030fb27b6f7739d1a1

src/app/payments/update-pending-payments.ts
vindard: I don't understand how this affects retry. As-is now now can always retry as tested here? https://github.com/GaloyMoney/galoy/pull/2151/files#diff-15e9812156b4ff1c33faaaf99ffcc368a8af2756ccb7a5e81acdcfc46782db9fR2363

src/services/ledger/index.ts
```diff
@@ -319,11 +319,16 @@ export const LedgerService = (): ILedgerService => {
...
         pending: false,
         hash: paymentHash,
       })
-      return total > 0
+      const { total: totalPending } = await MainBook.ledger({
+        pending: true,
+        hash: paymentHash,
+      })
+
+      return totalNotPending > 0 && totalPending === 0
```
vindard, line 331: This is the material change that allows us to retry if there are both pending and settled txns (from multiple attempts)

dolcalmi (approved):
Approved with small comment, also, please let me know to test this in staging.

dolcalmi (approved):

PR #2157, opened 2023-01-11
chore: add missing amount metadata props

vindard:
## Description

This PR adds the `satsAmount` set of properties to the transaction types that currently don't have it, and it adds tests for metadata on all non-admin transaction types.

The intention is to eventually rely on these and to remove the `fee`, `feeUsd` and `usd` properties from transactions.

### DIFF
--- a/src/app/wallets/send-on-chain.ts
+++ b/src/app/wallets/send-on-chain.ts
@@ -278,8 +278,8 @@ const executePaymentViaIntraledger = async <
     const converter = NewDisplayCurrencyConverter(displayCentsPerSat)
 
     let metadata:
-      | NewAddOnChainIntraledgerSendLedgerMetadata
-      | NewAddOnChainTradeIntraAccountLedgerMetadata
+      | AddOnChainIntraledgerSendLedgerMetadata
+      | AddOnChainTradeIntraAccountLedgerMetadata
     let additionalDebitMetadata: { [key: string]: string | undefined } = {}
     if (senderWallet.accountId === recipientWallet.accountId) {
       ;({ metadata, debitAccountAdditionalMetadata: additionalDebitMetadata } =
--- a/src/app/wallets/send-on-chain.ts
+++ b/src/app/wallets/send-on-chain.ts
@@ -286,7 +286,7 @@ const executePaymentViaIntraledger = async <
         LedgerFacade.OnChainTradeIntraAccountLedgerMetadata({
           payeeAddresses,
           sendAll,
-          paymentFlow,
+          paymentAmounts: paymentFlow,
 
           amountDisplayCurrency: converter.fromUsdAmount(paymentFlow.usdPaymentAmount),
           feeDisplayCurrency: 0 as DisplayCurrencyBaseAmount,
@@ -299,7 +299,7 @@ const executePaymentViaIntraledger = async <
         LedgerFacade.OnChainIntraledgerLedgerMetadata({
           payeeAddresses,
           sendAll,
-          paymentFlow,
+          paymentAmounts: paymentFlow,
 
           amountDisplayCurrency: converter.fromUsdAmount(paymentFlow.usdPaymentAmount),
           feeDisplayCurrency: 0 as DisplayCurrencyBaseAmount,
@@ -453,7 +453,7 @@ const executePaymentViaOnChain = async <
     const metadata = LedgerFacade.OnChainSendLedgerMetadata({
       // we need a temporary hash to be able to search in admin panel
       onChainTxHash: crypto.randomBytes(32).toString("hex") as OnChainTxHash,
-      paymentFlow,
+      paymentAmounts: paymentFlow,
 
       amountDisplayCurrency: converter.fromUsdAmount(paymentFlow.usdPaymentAmount),
       feeDisplayCurrency: converter.fromUsdAmount(paymentFlow.usdProtocolFee),
--- a/src/app/wallets/update-on-chain-receipt.ts
+++ b/src/app/wallets/update-on-chain-receipt.ts
@@ -112,7 +112,7 @@ export const updateOnChainReceipt = async ({
   return onChainTxs.length
 }
 
-const processTxForWallet = async (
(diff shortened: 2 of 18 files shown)


src/app/wallets/update-on-chain-receipt.ts
```diff
@@ -179,11 +179,20 @@ const processTxForWallet = async (
...
+
           const metadata = LedgerFacade.OnChainReceiveLedgerMetadata({
             onChainTxHash: tx.rawTx.txHash,
-            fee: walletAddressReceiver.btcBankFee,
+            paymentFlow: {
+              btcPaymentAmount: walletAddressReceiver.btcToCreditReceiver,
+              usdPaymentAmount: walletAddressReceiver.usdToCreditReceiver,
+              btcProtocolFee: walletAddressReceiver.btcBankFee,
+              usdProtocolFee: walletAddressReceiver.usdBankFee,
+            } as OnChainPaymentFlowState<S, R>,
```
bodymindarts, line 191: Don't like the cast in a use-case function. If it must be a cast then at least put it behind a simple function. If it works for the desired dependency graph perhaps it could even be `walletAddressReceiver.intoPaymentFlowState` or the other way around `paymentFlowStateFromReceiver(..)`. Basically an explicit conversion function (which hides the cast).

vindard: I was able to change that type to remove the cast completely
https://github.com/GaloyMoney/galoy/pull/2157/commits/d5dcccde0bfe364c408c055d28bd550277a4ba24

test/integration/02-user-wallet/02-receive-lightning.spec.ts
```diff
@@ -181,6 +183,34 @@ describe("UserWallet - Lightning", () => {
...
+    // Check ledger transaction metadata for BTC 'LedgerTransactionType.Invoice'
+    // ===
+    const usdPaymentAmount = await usdFromBtcMidPriceFn({
+      amount: BigInt(sats),
+      currency: WalletCurrency.Btc,
+    })
+    if (usdPaymentAmount instanceof Error) throw usdPaymentAmount
+    const centsAmount = Number(usdPaymentAmount.amount)
+
+    const expectedFields = {
```
dolcalmi, line 196: should not check display currency too? if so, please check all ledger `expectedFields` validations

src/domain/wallet-invoices/wallet-invoice-receiver.ts
```diff
@@ -1,53 +1,77 @@
...
+      btcBankFee: feeBtc,
+    }
 
     return {
-      btcToCreditReceiver,
-      usdToCreditReceiver,
-      ...ZERO_BANK_FEE,
+      btcToCreditReceiver: calc.sub(receivedBtc, feeBtc),
+      usdToCreditReceiver: calc.sub(receivedUsd, bankFee.usdBankFee),
+      ...bankFee,
```
dolcalmi, line 29: does not worth desctructure 2 properties

src/domain/payments/index.types.d.ts
```diff
@@ -11,7 +11,18 @@ type XorPaymentHashProperty = XOR<
...
 >
 
-type PaymentFlowCommonState<S extends WalletCurrency, R extends WalletCurrency> = {
+type AmountsAndFees = {
+  btcPaymentAmount: BtcPaymentAmount
+  btcProtocolFee: BtcPaymentAmount
+
+  usdPaymentAmount: UsdPaymentAmount
+  usdProtocolFee: UsdPaymentAmount
+}
```
dolcalmi, line 20: please add a comment to clarify that these are protocol + bank fees (while we decide this at ACE meeting)

src/app/wallets/send-on-chain.ts
```diff
@@ -278,15 +278,15 @@ const executePaymentViaIntraledger = async <
...
+      | AddOnChainIntraledgerSendLedgerMetadata
+      | AddOnChainTradeIntraAccountLedgerMetadata
     let additionalDebitMetadata: { [key: string]: string | undefined } = {}
     if (senderWallet.accountId === recipientWallet.accountId) {
       ;({ metadata, debitAccountAdditionalMetadata: additionalDebitMetadata } =
         LedgerFacade.OnChainTradeIntraAccountLedgerMetadata({
           payeeAddresses,
           sendAll,
-          paymentFlow,
+          paymentAmounts: paymentFlow,
```
dolcalmi, line 289: what is there a difference between paymentFlow and paymentAmounts? or renaming is a big change?

dolcalmi (changes requested):

test/integration/02-user-wallet/02-receive-lightning.spec.ts
vindard: Done: https://github.com/GaloyMoney/galoy/pull/2157/commits/d3afbb66f198a2e3cb8143fa4d8810153d63401b

src/domain/wallet-invoices/wallet-invoice-receiver.ts
vindard: Like this? https://github.com/GaloyMoney/galoy/pull/2157/commits/67ab149edd4e558dc302b801d5257da70ebc9d04

src/app/wallets/send-on-chain.ts
vindard: `paymentAmounts` is a subset of `paymentFlow`. The hack before this was to cast the param type to get the 4 properties needed. Now we've explicitly declared a new subset type to avoid the cast. Since it's a subset of `paymentFlow`, I believe we can pass the whole object here instead of selecting the 4 properties from it that we actually need

... or?

src/domain/payments/index.types.d.ts
vindard: Thinking about this again, we can probably just rename these in the code. I've made a card for this in the project to do next
https://github.com/GaloyMoney/galoy/projects/9#card-87595055

dolcalmi (approved):

PR #2175, opened 2023-01-15
chore: make quiz best practice graphql-wise

nicolasburtey:
Adding a new graphql type for Quiz that is cache-enabled.

first previous Quiz object were scattered to both:
- Root Query 
- under me 

and both had to be combined to get an accurate picture of the quiz completed and how much is the reward for each quizzes.

also, because the object return for me was 
```
    quizQuestions {
      completed
      question {
        id
        earnAmount
      }
    }
```

basically it was not cache friend because `quizQuestions` were not returning an id. only question was. but to be cacheable `quizQuestions` also needs an id

similarly, the mutation now return the whole object to be able to be cached automatically.

Additional, this object is now only only under `ConsumerAccount`.


src/config/index.types.d.ts
```diff
@@ -0,0 +1,25 @@
+type QuizQuestionId =
```
bodymindarts, line 1: Shouldn't this be an `as const` object like we generally do for enums.

bodymindarts (approved):

src/config/index.types.d.ts
nicolasburtey: yeah I could do that. will do as part of another PR, more work is needed for the earn use case as we're expending it.

PR #2181, opened 2023-01-16
feat: add update phone usecase

krtk6160:


src/app/admin/update-user-phone.ts
```diff
@@ -0,0 +1,29 @@
...
+  const account = await accountsRepo.findById(id as AccountId)
+  if (account instanceof Error) return account
+  const kratosUserId = account.kratosUserId
+
+  const usersRepo = UsersRepository()
+  const user = await usersRepo.findById(kratosUserId)
+  if (user instanceof Error) return user
+
+  user.phone = phone
+  await usersRepo.update(user)
```
dolcalmi, line 22: you need to handle errors properly

dolcalmi (changes requested):

src/app/admin/update-user-phone.ts
krtk6160: Fixed

src/services/mongoose/users.ts
```diff
@@ -85,15 +89,19 @@ export const UsersRepository = (): IUsersRepository => {
...
-    if (!result) {
-      return new RepositoryError("Couldn't delete phone from user")
+    try {
+      const result = await User.findOneAndUpdate(
+        { userId: id },
+        { $unset: { phone: true } },
+        { new: true },
+      )
+      if (!result) {
+        return new CouldNotDeletePhoneFromUser("Couldn't delete phone from user")
```
bodymindarts, line 101: This should be `CouldNotUnsetPhoneFromUser` and don't pass a msg. If you want to pass a message pass it in the overridden constructor where you inherit. But actually the type should be expressive enough to not need a message.

bodymindarts (approved):

bodymindarts (approved):

src/graphql/error-map.ts
```diff
@@ -481,6 +482,8 @@ export const mapError = (error: ApplicationError): CustomApolloError => {
     case "LikelyNoUserWithThisPhoneExistError":
     case "LikelyUserAlreadyExistError":
     case "PhoneIdentityDoesNotExistError":
+    case "UserWithPhoneAlreadyExistsError":
```
vindard, line 485: We could probably add a friendly message for this error since it is very specific and relevant to the calling user

src/services/kratos/auth-phone-no-password.ts
```diff
@@ -141,6 +141,42 @@ export const AuthWithPhonePasswordlessService = (): IAuthWithPhonePasswordlessSe
...
+    let identity: Identity
+
+    try {
+      ;({ data: identity } = await kratosAdmin.getIdentity({ id: kratosUserId }))
+    } catch (err) {
+      return new UnknownKratosError(err)
+    }
+
+    if (identity.state === undefined)
+      throw new KratosError("state undefined, probably impossible state") // type issue
```
vindard, line 160: We always return errors in application layer and below. We so far only throw in gql query resolvers.

We also always use curly braces if an if statement is more than 1 line https://blog.codecentric.de/curly-braces

```suggestion
    if (identity.state === undefined) {
      return new KratosError("state undefined, probably impossible state") // type issue
    }
```

src/services/kratos/auth-phone-no-password.ts
```diff
@@ -141,6 +141,42 @@ export const AuthWithPhonePasswordlessService = (): IAuthWithPhonePasswordlessSe
...
+
+    identity.traits = { phone }
+
+    const adminIdentity: UpdateIdentityBody = {
+      ...identity,
+      credentials: { password: { config: { password } } },
+      state: identity.state,
+    }
+
+    const { data: newIdentity } = await kratosAdmin.updateIdentity({
```
vindard, line 170: This can probably throw? If yes, we should wrap in try-catch (may also be an issue in existing `upgradeToPhoneAndEmailSchema` method)

src/domain/errors.ts
```diff
@@ -13,13 +13,16 @@ export class DuplicateError extends RepositoryError {}
...
 export class CouldNotFindError extends RepositoryError {}
+export class CouldNotUpdateError extends RepositoryError {}
 export class CannotConnectToDbError extends RepositoryError {
   level = ErrorLevel.Critical
 }
 export class DbConnectionClosedError extends RepositoryError {
   level = ErrorLevel.Critical
 }
 
+export class CouldNotUnsetPhoneFromUser extends CouldNotUpdateError {}
```
vindard, line 24: Missing `Error` suffix

src/graphql/error-map.ts
krtk6160: Fixed here - 6df07700713571070501cc450952a51b96e5b031

vindard (commented):
Few small nits

src/graphql/error-map.ts
```diff
@@ -335,6 +335,9 @@ export const mapError = (error: ApplicationError): CustomApolloError => {
       message = "Invalid pagination arguments"
       return new ValidationInternalError({ message, logger: baseLogger })
 
+    case "UserWithPhoneAlreadyExistsError":
+      message =
+        "The phone is associated with an existing wallet that has a non zero balance. Sweep the funds and try again."
```
vindard, line 340: Missing return for this `case`

```suggestion
        "The phone is associated with an existing wallet that has a non zero balance. Sweep the funds and try again."
      return new ValidationInternalError({ message, logger: baseLogger })

```

krtk6160: Fixed here - e96b93fc9fa1ae4a0609410d4b64b77c7f073e70

src/services/kratos/auth-phone-no-password.ts
vindard: I just realized, this probably needs a more specific error type (new error type like `MissingStateForIdentityError`) and we can then leave out the message argument here

src/graphql/error-map.ts
vindard: We generally keep the CustomApolloErrors pretty generic, but get very specific with the domain errors. I'd lean more to `ValidationInternalError` instead of making a new specific apollo error

test/e2e/servers/ory/kratos.spec.ts
```diff
@@ -517,3 +528,106 @@ describe("decoding link header", () => {
...
+      query: ACCOUNT_DETAILS_BY_USER_PHONE,
+      variables: { phone: newPhone },
+    })
+
+    await disposeAdminClient()
+
+    expect(result.data.accountDetailsByUserPhone.id).toBe(uid)
+  })
+
+  it("updates phone even if new phone is associated with a zero balance account", async () => {
```
vindard, line 582: We may need a failure mode test for this too. Can probably add a test for where the user has an existing balance in the account.

krtk6160: That is already being done at the end of that test, perhaps I should improve the title of the test

src/graphql/error-map.ts
krtk6160: Resolved

src/domain/errors.ts
krtk6160: Resolved

src/services/kratos/auth-phone-no-password.ts
krtk6160: That throw was copied over from the `upgradeToPhoneAndEmailSchema` method. I've updated it to a return now. I'm not adding another error class in this PR as this function will only be consumed via the admin panel. Will introduce a new error type in a follow up pr to fix it in `upgradeToPhoneAndEmailSchema` as well

src/services/kratos/auth-phone-no-password.ts
krtk6160: Fixed

PR #2182, opened 2023-01-16
chore: migrate 'onchain_payment' sats amounts properties

vindard:
## Description

This PR is to add the new `satsAmount`/`centsAMount` etc. properties to legacy `onchain_payment` transactions that were created before the onchain payment flow builder was deployed.

The migration works by:

- isolating the `"BTC"` entries for `onchain_payment` type txns that don't have the `satsAmount` property
  (works for both USD & BTC onchain payments, even though all USD payments will have `satsAmount` prop and likely won't be matched in our environments)

- finding the max BTC debit for that txn group (for BTC will be from sender wallet, for USD will be from dealer wallet)

- calculating the new properties using the `debit`, `fee`, `feeUsd` & `usd` properties

- writing back the new properties **by `hash`** to all matching txns

_Note: an add'l `satsAmountMigration` property is added to track changes and facilitate any potential rollbacks needed_


src/migrations/20230116132109-add-sats-amounts-onchain-payments.ts
```diff
@@ -0,0 +1,91 @@
...
+            {
+              $set: {
+                satsAmount,
+                satsFee: fee,
+                centsAmount,
+                centsFee,
+                displayAmount: centsAmount,
+                displayFee: centsFee,
+                displayCurrency: "USD",
+                satsAmountMigration: true,
```
dolcalmi, line 47: maybe instead of `satsAmountMigration` we can have a field with the current display amount (in that way we can validate or rollback because this is set to empty if we rollback so we will loose the value if that happens)

dolcalmi (changes requested):

src/migrations/20230116132109-add-sats-amounts-onchain-payments.ts
vindard: There isn't a property called `displayAmount` for these transactions I think. We use the `usd` property to derive that, which isn't being touched in these migrations

image

dolcalmi (approved):

PR #2185, opened 2023-01-17
fix: kratos extend session cron

ntheile:
This PR fixes a bug in kratos `listIdentities` pagination function. The  `response.headers.link` does not return the last page properly.

### DIFF
--- a/src/services/kratos/identity.ts
+++ b/src/services/kratos/identity.ts
@@ -63,6 +63,18 @@ export const IdentityRepository = (): IIdentityRepository => {
         hasNext = page !== undefined && totalCount !== 0
       }
 
+      // * fixes bug in kratos res.headers.link not returning last page properly
+      const pageCount = totalCount / perPage
+      const isPageCountWholeNumber = pageCount % 1 // 0 means its a whole number, therefore, we can skip because no partial last page of data exists
+      if (isPageCountWholeNumber !== 0) {
+        const lastPageNum = Math.ceil(totalCount / perPage)
+        const lastPageRes = await kratosAdmin.listIdentities({
+          perPage,
+          page: lastPageNum,
+        })
+        identities.push(...lastPageRes.data)
+      }
+
       // FIXME(nb) function above return duplicated query for the first 2 calls, so removing them here
       const uniqueIdentities = identities.filter(
         (value, index, self) => index === self.findIndex((t) => t.id === value.id),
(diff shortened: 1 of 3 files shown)


src/services/kratos/identity.ts
```diff
@@ -63,6 +63,18 @@ export const IdentityRepository = (): IIdentityRepository => {
         hasNext = page !== undefined && totalCount !== 0
       }
 
+      // * fixes bug in kratos res.headers.link not returning last page properly
+      const pageCount = totalCount / perPage
+      const isPageCountWholeNumber = pageCount % 1 // 0 means its a whole number, therefore, we can skip because no partial last page of data exists
+      if (isPageCountWholeNumber !== 0) {
+        const lastPageNum = Math.ceil(totalCount / perPage)
```
dolcalmi, line 70: check page description (is not sequential) https://www.ory.sh/docs/kratos/reference/api#tag/identity/operation/listIdentities so not sure if this will fix the issue

nicolasburtey: yup. although it's on the description not sure if this is the case in practice... I don't have a lot more context @ntheile than what @dolcalmi said, so feel free to try it if you think this is worth it.

src/services/kratos/identity.ts
```diff
@@ -63,6 +63,18 @@ export const IdentityRepository = (): IIdentityRepository => {
         hasNext = page !== undefined && totalCount !== 0
       }
 
+      // * fixes bug in kratos res.headers.link not returning last page properly
```
bodymindarts, line 66: Has this been raised upstream? Have you contacted them on slack? Given:
> This value is currently an integer, but it is not sequential. The value is not the page number, but a reference. The next page can be any number and some numbers might return an empty list.

I don't see how this will fix the issue.

src/services/kratos/identity.ts
ntheile: @dolcalmi yeah the documentation does not match how to works in practice (maybe they have not updated the documentation yet). The `request.headers.link` has many bugs (it's actually not supported in some browsers). Here is a thread on kratos slack about the topic https://ory-community.slack.com/archives/C012RJ2MQ1H/p1673618866406469

src/services/kratos/identity.ts
ntheile: yep just posted this topic in slack https://ory-community.slack.com/archives/C012RJ2MQ1H/p1673618866406469

ntheile:
I added tracing, so now we can see the count of how many sessions were extended, total identities, number of inactive sessions, and total error count at the root span  `name = cron.extendSessions` 

image

bodymindarts (approved):

ntheile:
related issue here https://github.com/ory/kratos/issues/2834

PR #2192, opened 2023-01-17
chore: remove redundant makefile commands

krtk6160:


Makefile
```diff
@@ -16,10 +16,6 @@ start-main:
 start-main-fast:
 	yarn run watch-main | yarn pino-pretty -c -l
 
-start-admin:
-	. ./.envrc && yarn tsnd --respawn --files -r tsconfig-paths/register -r src/services/tracing.ts \
-		src/servers/graphql-admin-server.ts | yarn pino-pretty -c -l
```
dolcalmi, line 21: `src/servers/graphql-admin-server.ts` still is not deleted

nicolasburtey: `start-main` can then be renamed to `start`

nicolasburtey (approved):

PR #2212, opened 2023-01-24
fix: update fee calc rounding in tests

dolcalmi:
attempt to fix 1 sat (rounding) issue in tests using same strategy used in lightning library for safe tokens https://github.com/alexbosworth/lightning/blob/master/bolt00/safe_tokens.js

purpose: reduce dev, github actions and CI errors produced by millisat fee rounding issue


dolcalmi:
converting to draft in favor of https://github.com/GaloyMoney/galoy/pull/2210

vindard:
Seems like the issue is resolved for now. Should we close this @dolcalmi?

dolcalmi:
the real issue is a misconfig in channel's fees setup.

PR #2224, opened 2023-01-25
fix: actually saving the displayCurrency to accounts

nicolasburtey:


src/services/mongoose/schema.ts
```diff
@@ -285,6 +285,8 @@ const AccountSchema = new Schema<AccountRecord>(
       type: String,
       index: true,
     },
+
+    displayCurrency: String, // FIXME: should be an enum
```
vindard, line 289: Needs to be added to AccountRecord type as well?

nicolasburtey: already there

vindard (approved):

PR #2226, opened 2023-01-26
fix: update gql settlementPrice type

dolcalmi:
Update settlementPrice type to avoid issues with apollo client cache. This should not be a breaking change but must be tested first in staging

### DIFF
--- a/src/graphql/admin/schema.graphql
+++ b/src/graphql/admin/schema.graphql
@@ -313,16 +313,6 @@ enum PhoneCodeChannelType {
   WHATSAPP
 }
 
-"""
-Price amount expressed in base/offset. To calculate, use: `base / 10^offset`
-"""
-type Price {
-  base: SafeInt!
-  currencyUnit: String!
-  formattedAmount: String!
-  offset: Int!
-}
-
 type PsbtDetail {
   fee: SatAmount!
   transaction: String!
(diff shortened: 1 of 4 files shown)


src/graphql/main/schema.graphql
```diff
@@ -1116,7 +1127,7 @@ type Transaction {
   """
   Price in USDCENT/SETTLEMENTUNIT at time of settlement.
   """
-  settlementPrice: Price!
+  settlementPrice: SettlementPrice!
```
nicolasburtey, line 1130: I don't think it's backward compatible. Are we using it already?

I'm not sure if there is an easy way on the backend to see what field are been used already. I think it was one of the benefits of Apollo Studio. would be relevant to know if we can create this capability back with honeycomb

src/graphql/admin/schema.graphql
```diff
@@ -349,6 +339,17 @@ scalar SafeInt
 """
 scalar SatAmount
 
+"""
+Price amount expressed in base/offset. To calculate, use: `base / 10^offset`
+"""
+type SettlementPrice {
+  base: SafeInt!
+  currency: DisplayCurrency!
```
nicolasburtey, line 347: should probably also have the underlying currency, USD or BTC dpendanding of the wallet used.

nicolasburtey (commented):
some questions/comments

src/graphql/main/schema.graphql
dolcalmi: I didnt see validations in settlement price with `__typename`. also, this is a read only field.... we can merge it and test it in staging, I think it is not a problem

src/graphql/admin/schema.graphql
dolcalmi: why? we already have `settlementCurrency`

src/graphql/main/schema.graphql
nicolasburtey: true. we should check with the released version of the mobile app also.

nicolasburtey (approved):

dolcalmi:
closed in favor of https://github.com/GaloyMoney/galoy/pull/2439/

PR #2241, opened 2023-01-27
ci(shared): bump vendored ci files

galoybot:
# Bump Shared Tasks

This PR syncs in this repository, shared CI tasks from [concourse-shared](https://github.com/GaloyMoney/concourse-shared).


krtk6160 (changes requested):
This PR doesn't look quite right. Two checks are failing, and it looks like a fix is needed in concourse-shared

sandipndev:
paused shared execution for galoy for now.

PR #2252, opened 2023-01-30
feat: resolve txn metadata with dataloader

vindard:
## Description

This PR introduces `dataloader` as a way to expose data from our `TransactionsMetadata` collection when returning transactions in the API.

Benefits:
- gets around n+1 problem when querying for lists of transactions
- TransactionsMetadata collection isn't queried at all if no `settlementVia` selection is included in the gql query

Downsides:
- TransactionsMetadata reads are now separate from the LedgerService and only get combined in to wallet transactions lazily in the graphql layer
- A new dataloader will need to be added for each new selection we add in future that pulls from the metadata collection

_**Note:** Dataloaders are usually included at the selection level, and for this one we've decided to include it at the `settlementVia` layer in the `Transaction` object and not one level lower on the `revealedPreImage` selection._

**This PR supersedes #1427**


src/servers/index.files.d.ts
```diff
@@ -1,5 +1,12 @@
+interface Loaders {
+  /* eslint @typescript-eslint/ban-ts-comment: "off" */
+  // @ts-ignore-next-line no-implicit-any
+  txnMetadata
```
dolcalmi, line 4: is there no type for DataLoader in the imported package? also... cant we use same strategy used in trace service?

src/app/transactions/get-transactions-metadata-by-ids.ts
```diff
@@ -0,0 +1,6 @@
+import { TransactionsMetadataRepository } from "@services/ledger/services"
```
dolcalmi, line 1: is this not part of wallet folder?

vindard: It can be in either `wallet` or `account` because we have transactions use-case in both. I wasn't sure where to put these given that 🤔

src/servers/index.files.d.ts
vindard: There's the class, but I had a really hard time trying to import it into a `.d.ts` file because of a generics issue (chatGPT was no help either 😅).

We can probably pair to take a look at this actually.

vindard: We figured this out here https://github.com/GaloyMoney/galoy/pull/2252/commits/1d28952427c72f3beb53a227b06c389d85c209ad

PR #2260, opened 2023-01-31
refactor: remove use of slowFindByPhone

krtk6160:
`slowFindByPhone` does not return all the results. I couldn't find exactly how many, but the cli tool with the same api returns 250 results. Hence, I'm now using the user collection for obtaining the kratosUserId instead.

### DIFF
--- a/src/app/admin/update-user-phone.ts
+++ b/src/app/admin/update-user-phone.ts
@@ -1,9 +1,5 @@
-import {
-  PhoneIdentityDoesNotExistError,
-  UserWithPhoneAlreadyExistsError,
-} from "@domain/authentication/errors"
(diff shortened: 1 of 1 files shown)


nicolasburtey (approved):

nicolasburtey:
notes: slowFindByPhone should return more than 250 entities. if that is not the case, then the function is broken again @ntheile

ntheile:
> 

Yeah it is most likely not returning the last page of identities. If we change the code to use the service wrapper in `IdentityRepository` it should work

```
import {IdentityRepository } from "@services/kratos"

 const identites = await IdentityRepository().listIdentities()
```

PR #2267, opened 2023-02-01
chore: add new twilio restricted region error

vindard:
Trace: [link](https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/cVqfHtLWBFD/trace/8mFGP6wPPWJ?fields[]=c_name&fields[]=c_service.name&span=3173ead06736557b)

### DIFF
--- a/src/services/twilio.ts
+++ b/src/services/twilio.ts
@@ -119,6 +119,7 @@ const handleCommonErrors = (err: Error | string) => {
       return new InvalidPhoneNumberPhoneProviderError(errMsg)
 
     case match(KnownTwilioErrorMessages.RestrictedRegion):
+    case match(KnownTwilioErrorMessages.CountryNeedsVetting):
     case match(KnownTwilioErrorMessages.BlockedRegion):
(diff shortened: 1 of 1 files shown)


src/services/twilio.ts
```diff
@@ -144,6 +145,7 @@ export const KnownTwilioErrorMessages: { [key: string]: RegExp } = {
   InvalidMobileNumber: /not a mobile number/,
   InvalidPhoneNumberParameter: /Invalid parameter `To`/,
   RestrictedRegion: /has not been enabled for the region/,
+  RestrictedCountryChina: /China require use case vetting/,
```
dolcalmi, line 148: can we just validate against `require use case vetting` ? having an error type for each country is not an option (is not only china)

dolcalmi (changes requested):

src/services/twilio.ts
nicolasburtey: agree I would not make it country specific. we don't know which other country twilio requires "vetting"

dolcalmi (approved):

PR #2271, opened 2023-02-01
fix: oathkeeper matcher rules more specific

ntheile:
@krtk6160 , @nicolasburtey 

After troubleshooting with Kartik this should fix the e2e test errors. Oathkeeper was throwing an error if a rule was double matched.

2 main fixes here:

1. The `graphql-backend` rule was too broad with the wildcard.
2. I converted the `oathkeeper_rule` file from yaml to json (to stay consistent with charts)

The main change is to make the `galoy-backend` rule more specific. I changed it to this matcher below (i tried to get all the possible routes, let me know if i missed any).
```
<(http|https)>://<[a-zA-Z0-9-.:]+>/<(graphql|kratos|browser|healthz|metrics)(.*)>
``` 
code here https://github.com/GaloyMoney/galoy/pull/2271/files#diff-49a6c246937a7f41bcbb61c72b158afe30328853165655aa64a63bcc431e6da4R103 

By making the rules more specific it will be better for security too.


krtk6160 (commented):
I think the rules schema is much more readable in the yaml form. Otherwise LGTM.

krtk6160 (approved):

PR #2282, opened 2023-02-03
chore: schema requests

UncleSamtoshi:


src/graphql/main/schema.graphql
```diff
@@ -877,6 +877,7 @@ scalar PaymentHash
 
 type PaymentSendPayload {
   errors: [Error!]!
+  transaction: Transaction
```
UncleSamtoshi, line 880: It would be nice if PaymentSendPayload returned a full transaction object, that way the transaction enters the cache and we can add it to our list of transactions without having to requery transactions from the home screen. It also allows us to link directly to the transaction details in the application. Currently we dont have a transaction id returned from the mutation so we cant link within the app to the correct page.

src/graphql/main/schema.graphql
```diff
@@ -1354,7 +1355,12 @@ input UserRequestAuthCodeInput {
   phone: Phone!
 }
 
-union UserUpdate = IntraLedgerUpdate | LnUpdate | OnChainUpdate | Price
+type TransactionUpdate {
+  transaction: Transaction
```
UncleSamtoshi, line 1359: In general it would be helpful to return full objects that are changing from the subscriptions. If we do this, our cache is automatically updated and anything within the application that is currently looking at that object will get automatically refreshed without having to manually handled

src/graphql/main/schema.graphql
```diff
@@ -877,6 +877,8 @@ scalar PaymentHash
 
 type PaymentSendPayload {
   errors: [Error!]!
+  transaction: Transaction
+  wallet: Wallet
```
UncleSamtoshi, line 881: Since the wallet object is changing (the balance is different and the history) as a result of this mutation, it would be helpful to return the Wallet object and have the cache be automatically updated.

src/graphql/main/schema.graphql
```diff
@@ -1354,7 +1362,13 @@ input UserRequestAuthCodeInput {
   phone: Phone!
 }
 
-union UserUpdate = IntraLedgerUpdate | LnUpdate | OnChainUpdate | Price
+type TransactionUpdate {
+  transaction: Transaction
+  wallet: Wallet
```
UncleSamtoshi, line 1367: Same as above.

src/graphql/main/schema.graphql
```diff
@@ -730,9 +730,9 @@ type Mutation {
   lnUsdInvoiceFeeProbe(input: LnUsdInvoiceFeeProbeInput!): SatAmountPayload!
   onChainAddressCreate(input: OnChainAddressCreateInput!): OnChainAddressPayload!
   onChainAddressCurrent(input: OnChainAddressCurrentInput!): OnChainAddressPayload!
-  onChainPaymentSend(input: OnChainPaymentSendInput!): PaymentSendPayload!
+  onChainPaymentSend(input: OnChainPaymentSendInput!): PaymentSendPayload! // This should allow for usd wallets
   onChainPaymentSendAll(input: OnChainPaymentSendAllInput!): PaymentSendPayload!
-  onChainUsdPaymentSend(input: OnChainUsdPaymentSendInput!): PaymentSendPayload!
+  onChainNoAmountUsdPaymentSend(input: OnChainUsdPaymentSendInput!): PaymentSendPayload!
```
UncleSamtoshi, line 735: Not in love with calling this `NoAmount` however it is `NoAmount` in the sense that any btc address that has an amount specified will denominate that amount in sats and since this is denominated in cents it can only be used to send an amount that is chosen by the sender (a `NoAmount`).

src/graphql/main/schema.graphql
```diff
@@ -730,9 +730,9 @@ type Mutation {
   lnUsdInvoiceFeeProbe(input: LnUsdInvoiceFeeProbeInput!): SatAmountPayload!
   onChainAddressCreate(input: OnChainAddressCreateInput!): OnChainAddressPayload!
   onChainAddressCurrent(input: OnChainAddressCurrentInput!): OnChainAddressPayload!
-  onChainPaymentSend(input: OnChainPaymentSendInput!): PaymentSendPayload!
+  onChainPaymentSend(input: OnChainPaymentSendInput!): PaymentSendPayload! // This should allow for usd wallets
```
UncleSamtoshi, line 733: `onChainPaymentSend` needs to be able to accept a UsdWallet because to pay an onchain invoice with an amount, that amount will be denominated in sats, so we need a mutation that accepts a usd wallet while denominating the amount in sats.

src/graphql/main/schema.graphql
```diff
@@ -964,7 +966,13 @@ type Query {
     targetConfirmations: TargetConfirmations = 1
     walletId: WalletId!
   ): OnChainTxFee!
-  onChainUsdTxFee(
+  onChainUsdFee(
```
UncleSamtoshi, line 969: This is to get the fee for an invoice that is denominated in sats but paid from a usd wallet.

src/graphql/main/schema.graphql
nicolasburtey: here, transaction probably optional

src/graphql/main/schema.graphql
nicolasburtey: wallet. mandatory?

PR #2289, opened 2023-02-06
chore: remove legacy display currency converter

vindard:
## Description

This PR removes remaining instances of `DisplayCurrencyConverter` and then removes `DisplayCurrencyConverter` implementation.

### DIFF
--- a/src/domain/fiat/display-currency.ts
+++ b/src/domain/fiat/display-currency.ts
@@ -1,5 +1,8 @@
 export const CENTS_PER_USD = 100
 
+export const usdMinorToMajorUnit = (amount: number | bigint) =>
+  Number((Number(amount) / CENTS_PER_USD).toFixed(2))
+
 export const toDisplayCurrencyBaseAmount = (amount: number) =>
   amount as DisplayCurrencyBaseAmount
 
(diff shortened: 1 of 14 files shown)


test/mocks/get-current-price.ts
```diff
@@ -21,3 +24,12 @@ export const getCurrentUsdCentPrice = ({
...
 }
+
+export const getCurrentPriceAsPriceRatio = async ({
+  currency,
+}: GetCurrentSatPriceArgs): Promise<PriceRatio | PriceServiceError> => {
+  const price = await getCurrentSatPrice({ currency })
+  if (price instanceof Error) return price
+
+  return toPriceRatio(price.price * CENTS_PER_USD)
+}
```
dolcalmi, line 35: were you getting an error without this? in theory if code did not change this is not necessary

src/domain/fiat/display-currency.ts
```diff
@@ -1,29 +1,11 @@
 export const CENTS_PER_USD = 100
 
+export const usdMinorToMajorUnit = (amount: number | bigint) =>
+  Number((Number(amount) / CENTS_PER_USD).toFixed(2))
+
 export const toDisplayCurrencyBaseAmount = (amount: number) =>
   amount as DisplayCurrencyBaseAmount
 
-export const DisplayCurrencyConverter = (
```
dolcalmi, line 6: will we remove/rename `NewDisplayCurrencyConverter` in a different PR?

dolcalmi (changes requested):
looks good, just minor questions

test/mocks/get-current-price.ts
vindard: Yup tests fail with "undefined" function for that method

dolcalmi (approved):

PR #2227, opened 2023-01-26
feat: add realtime price query

dolcalmi:
- Add realtimePrice query
  - Avoid issues with apollo client cache. 
  - Add support for wallet currencies (needed for the multi-currency support)
  - Add timestamp and id
- Add realtimePrice subscription

## Query
  ```graphql
  query {
    realtimePriceSAT: realtimePrice(walletCurrency: BTC, displayCurrency: "EUR"){
      id
      timestamp
      base
      offset
      walletCurrency
      displayCurrency
      currencyUnit
      formattedAmount
    }
    realtimePriceUSD: realtimePrice(walletCurrency: USD, displayCurrency: "EUR"){
      id
      timestamp
      base
      offset
      walletCurrency
      displayCurrency
      currencyUnit
      formattedAmount
    }
  }
  ```
## Subscription
  ```graphql
  subscription prices {
    realtimePrice(input: {walletCurrency: BTC, displayCurrency: "EUR"}) {
      errors {
        message
      }
      price {
        id
        timestamp
        base
        offset
        walletCurrency
        displayCurrency
        currencyUnit
        formattedAmount
      }
    }
  }
  ```

PR #2291, opened 2023-02-07
feat: update price service interface

dolcalmi:
Update price service interface according to https://github.com/GaloyMoney/galoy/pull/2227#issuecomment-1412204380


src/app/prices/get-current-price.ts
```diff
@@ -13,24 +13,24 @@ export const getCurrentPrice = async ({
...
+    displayCurrency: checkedDisplayCurrency,
   })
   if (realtimePrice instanceof Error)
     return getCachedPrice({ currency: checkedDisplayCurrency })
 
   let cachedPrices = await getCachedPrices()
   cachedPrices = cachedPrices instanceof Error ? {} : cachedPrices
 
-  cachedPrices[currency] = realtimePrice
+  cachedPrices[currency] = realtimePrice.price as DisplayCurrencyPerSat
```
dolcalmi, line 25: this cast will be replaced here https://github.com/GaloyMoney/galoy/pull/2292/

test/unit/app/prices/get-current-price.spec.ts
```diff
@@ -26,12 +26,19 @@ describe("Prices", () => {
         .spyOn(PriceServiceImpl, "PriceService")
         .mockImplementationOnce(() => ({
           listHistory: jest.fn(),
-          getRealTimePrice: () => Promise.resolve(0.05 as DisplayCurrencyPerSat),
+          getUsdCentRealTimePrice: jest.fn(),
```
vindard, line 29: Are we adding new tests to check `getUsdCentRealTimePrice` too?

vindard (approved):
Just one comment, but not a blocker

PR #2292, opened 2023-02-07
feat: update current price cache implementation by wallet currency

dolcalmi:
Update current price cache implementation to support wallet currencies and new price service interface

Review/Merge after https://github.com/GaloyMoney/galoy/pull/2291


vindard (approved):
Looks pretty straightforward

PR #2298, opened 2023-02-07
feat: add realtimePrice query and subscription

dolcalmi:
Add realtimePrice query/subscription according to https://github.com/GaloyMoney/galoy/pull/2227#issuecomment-1412204380

Review/merge after https://github.com/GaloyMoney/galoy/pull/2292/

## How to test

```
make start-deps
make start-main-fast
# and if you want updates after running subscription
. ./.envrc && yarn tsnd --respawn --files -r tsconfig-paths/register -r src/services/tracing.ts \
                src/servers/trigger.ts | yarn pino-pretty -c -l
```

### Query
```graphql
query {
  realtimePrice (currency: "EUR") {
    id
    timestamp
    denominatorCurrency
    btcSatPrice {
      base
      offset
      currencyUnit
    }
    usdCentPrice {
      base
      offset
      currencyUnit
    }
  }
}
```

### Subscription
```graphql
subscription  {
  realtimePrice(input: { currency: "EUR" }) {
    errors {
      message
    }
    realtimePrice {
      id
      timestamp
      denominatorCurrency
      btcSatPrice {
        base
        offset
        currencyUnit
      }
      usdCentPrice {
        base
        offset
        currencyUnit
      }
    }
  }
}
```


src/graphql/root/subscription/realtime-price.ts
```diff
@@ -0,0 +1,165 @@
...
+        payload: { errors: [{ message: displayCurrency.message }] },
+      })
+      return pubsub.createAsyncIterator({ trigger: immediateTrigger })
+    }
+
+    const currencies = await Prices.listCurrencies()
+    if (currencies instanceof Error) {
+      pubsub.publishImmediate({
+        trigger: immediateTrigger,
+        payload: { errors: [mapError(currencies)] },
```
vindard, line 119: Can use this alternative method to also get the `code` property

```suggestion
        payload: { errors: [mapAndParseErrorForGqlResponse(currencies)] },
```

src/graphql/root/subscription/realtime-price.ts
```diff
@@ -0,0 +1,165 @@
...
+      return pubsub.createAsyncIterator({ trigger: immediateTrigger })
+    }
+
+    const pricePerSat = await Prices.getCurrentSatPrice({
+      currency: checkedDisplayCurrency,
+    })
+    const pricePerUsdCent = await Prices.getCurrentUsdCentPrice({
+      currency: checkedDisplayCurrency,
+    })
+    if (!(pricePerSat instanceof Error) && !(pricePerUsdCent instanceof Error)) {
```
vindard, line 141: What happens in case where `pricePerSat` or `pricePerUsdCent` is an Error. Is subscription silent?

vindard (commented):
Few small comments

src/graphql/main/schema.graphql
```diff
@@ -81,6 +81,12 @@ type AccountUpdateDisplayCurrencyPayload {
   errors: [Error!]!
 }
 
+interface AnyPrice {
```
nicolasburtey, line 84: could `PriceInterface` be a better name than `AnyPrice`? `AnyPrice` make me think about something where the buyer can decide what he want to spend... maybe that's just how I feel about it but wondering if interface doesn't convey better the intent?

src/graphql/main/schema.graphql
```diff
@@ -974,6 +998,11 @@ type Query {
     @deprecated(
       reason: "TODO: remove. we don't need a non authenticated version of this query. the users can only do the query while authenticated"
     )
+
+  """
+  Returns 1 Sat and 1 Usd Cent price for the given currency
+  """
+  realtimePrice(currency: DisplayCurrency! = "USD"): RealtimePrice
```
nicolasburtey, line 1005: this would be use by `galoy-pay`?

nicolasburtey: the `!` means mandatory? I'm confused the combination of this + a default argument

nicolasburtey (commented):
for the mobile app, I think we would need:

```
Query {
  me {
     realtimePrice {
       usd: RealtimePricePayload
       btc: RealtimePricePayload
    }
  }
}

subscription { # behind auth
   realtimePrice {
       usd: RealtimePricePayload
       btc: RealtimePricePayload
    }
}
```

because both of those endpoint would be authenticated, we can directly used the preferred display currency served for the user.

we need both price simultaneously because it will make it straightforward to do the conversion in any combination of

btc <> usd <> display

src/graphql/main/schema.graphql
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/2298/commits/a0abbc6b716ae40839384a059442fbf5909f4abf

src/graphql/root/subscription/realtime-price.ts
dolcalmi: this is just the first time so the worst case is to wait 30 seconds for trigger update

src/graphql/root/query/realtime-price.ts
```diff
@@ -0,0 +1,55 @@
...
+    })
+    if (btcPrice instanceof Error) throw mapError(btcPrice)
+
+    const usdPrice = await Prices.getCurrentUsdCentPrice({
+      currency,
+    })
+    if (usdPrice instanceof Error) throw mapError(usdPrice)
+
+    const centsPerSat = 100 * btcPrice.price
+    const centsPerUsdCent = 100 * usdPrice.price
```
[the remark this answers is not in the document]
dolcalmi: :) yep, we have Mayor/Minor for base currency (i.e wallet currency) and we have Major/Minor for quote currency (i.e display currency). I did an update to use a domain function https://github.com/GaloyMoney/galoy/pull/2298/commits/8b2fe7e6169dcf8f3eecfc25ad65f54d1e53a66c

src/graphql/root/subscription/realtime-price.ts
dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/2298/commits/8b2fe7e6169dcf8f3eecfc25ad65f54d1e53a66c

src/graphql/main/schema.graphql
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/2298/commits/46de91a0de0d3b19db1751129dfd056c7167a506

dolcalmi:
@nicolasburtey `my-updates` subscription update here https://github.com/GaloyMoney/galoy/pull/2298/commits/5d907b32d6d469671d3e82586a39815cf7e0e6a2 and account update with realtimePrice here https://github.com/GaloyMoney/galoy/pull/2298/commits/116348a7865aec73c7a1fe4ee803302e00b006ec

src/graphql/root/subscription/my-updates.ts
```diff
@@ -102,13 +120,15 @@ type MeResolveIntraLedger = {
 
 type MeResolveSource = {
   errors: IError[]
+  realtimePrice?: MePayloadPrice
   price?: MePayloadPrice
```
vindard, line 124: Should we add a deprecated for this or are we keeping both?

vindard (commented):
Looks ok to me. Will leave for Nicolas to check his comments against as well.

src/graphql/main/schema.graphql
```diff
@@ -4,6 +4,7 @@ interface Account {
   displayCurrency: DisplayCurrency!
   id: ID!
   limits: AccountLimits!
+  realtimePrice: RealtimePrice!
```
nicolasburtey, line 7: https://github.com/GaloyMoney/galoy/blob/116348a7865aec73c7a1fe4ee803302e00b006ec/src/graphql/main/schema.graphql#L977 should be marked as deprecated

nicolasburtey: (or even removed if not used?)

src/graphql/main/schema.graphql
```diff
@@ -1005,6 +1036,27 @@ type QuizQuestion {
...
+  timestamp: Timestamp!
+  usdCentPrice: PriceOfOneUsdCent!
+}
+
+input RealtimePriceInput {
+  currency: DisplayCurrency = "USD"
+}
+
+type RealtimePricePayload {
+  errors: [Error!]!
```
nicolasburtey, line 1056: why does this have an error array? isn't that the pattern for mutations only?

src/graphql/main/schema.graphql
```diff
@@ -974,6 +1000,11 @@ type Query {
     @deprecated(
       reason: "TODO: remove. we don't need a non authenticated version of this query. the users can only do the query while authenticated"
     )
+
+  """
+  Returns 1 Sat and 1 Usd Cent price for the given currency
+  """
+  realtimePrice(currency: DisplayCurrency = "USD"): RealtimePrice
```
nicolasburtey, line 1007: the return type should be mandatory (instead of optional)

src/graphql/main/schema.graphql
```diff
@@ -211,6 +212,7 @@ type ConsumerAccount implements Account {
   List the quiz questions of the consumer account
   """
   quiz: [Quiz!]!
+  realtimePrice: RealtimePrice!
```
nicolasburtey, line 215: the one at the root query should have an input to select the type?

src/graphql/root/subscription/my-updates.ts
dolcalmi: It is part of a union really couldn't find a way to do it.

src/graphql/main/schema.graphql
dolcalmi: this is part of subscriptions too because you dont want to throw and close the connection but you want to inform the user that we had an issue (specially with subscriptions like myUpdates)

src/graphql/main/schema.graphql
dolcalmi: Cant be removed https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/rKUW377pxbg/a/yz7oZw5Fw5g?hideMarkers

Updated https://github.com/GaloyMoney/galoy/pull/2298/commits/3304a7639402a306e8126a52f2c8bda5b6ea5690

src/graphql/main/schema.graphql
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/2298/commits/220c4ae26f4ec62588809307892cf8559ce486ef

src/graphql/main/schema.graphql
dolcalmi: this is part of `ConsumerAccount` so there is no need for input. We use the display currency from user account

src/graphql/main/schema.graphql
nicolasburtey: ok I see the error in `MyUpdatesPayload`.

we're not using this at all in the mobile repo (another example of something  exposed to the mobile that is not been used).

What action would you suggest on `errors`? 

it's a string/unstructured object so not very convenient to parse

nicolasburtey (approved):

PR #2290, opened 2023-02-06
chore: improving tests with eslint jest plugin

nicolasburtey:

PR #2306, opened 2023-02-09
fix: tests lint issues

dolcalmi:
fixes based on PR https://github.com/GaloyMoney/galoy/pull/2290

workarounds:
- turn off jest/no-disabled-tests
- disable `jest/no-conditional-expect` in swap tests. @ntheile if you have an alternative let me know
- @nicolasburtey `elevatingSessionWithTotp` is not working (returns an exception) but is not a blocker


nicolasburtey (commented):
LGTM but probably Arvin or Nick can dive into it

ntheile:
this looks good to me

nicolasburtey (approved):

PR #2327, opened 2023-02-14
chore: add new 'DestinationMissingDependentFeatureError'

vindard:
This happens when the destination misses a feature that a feature our node requires depends on.

More: https://github.com/lightningnetwork/lnd/blob/f245591e4bfc18d50326938d1f50bf8dc5fe4771/routing/payment_session.go#L63


krtk6160 (approved):
LGTM assuming the tests pass

PR #2329, opened 2023-02-15
fix: update currency list return to be mandatory

dolcalmi:
Update currencyList query to return a mandatory list. This helps to work with cache in mobile app


nicolasburtey:
> Update currencyList query to return a mandatory list. This helps to work with cache in mobile app

not cache, but just parsing the data.

it's also a general best practice from what I read

nicolasburtey (approved):

PR #2334, opened 2023-02-16
feat: add DisplayPriceRatio for handling 'DisplayAmount' type

vindard:
## Description

This PR adds a new PriceRatio type to specifically handle display amounts and return display-related data.


src/domain/payments/index.types.d.ts
```diff
@@ -1,11 +1,27 @@
...
+
+type WalletPriceRatio = {
   convertFromUsd(convert: UsdPaymentAmount): BtcPaymentAmount
   convertFromBtc(convert: BtcPaymentAmount): UsdPaymentAmount
   convertFromBtcToFloor(convert: BtcPaymentAmount): UsdPaymentAmount
   convertFromBtcToCeil(convert: BtcPaymentAmount): UsdPaymentAmount
   usdPerSat(): DisplayCurrencyBasePerSat
 }
 
+type DisplayPriceRatio<S extends WalletCurrency, T extends DisplayCurrency> = {
```
vindard, line 17: New object type coming in here that outputs `DisplayCurrencyObect<T>` types

src/domain/shared/index.types.d.ts
```diff
@@ -19,6 +23,11 @@ type DisplayAmount<T extends DisplayCurrency> = {
   amount: number
 }
 
+type DisplayCurrencyObject<T extends DisplayCurrency> = {
```
vindard, line 26: New object type coming in here that's used in `DisplayPriceRatio`

src/app/cold-storage/rebalance-to-cold-wallet.ts
```diff
@@ -29,7 +29,7 @@ export const rebalanceToColdWallet = async (): Promise<boolean | ApplicationErro
   const offChainService = LndService()
   if (offChainService instanceof Error) return offChainService
 
-  const displayPriceRatio = await getCurrentPriceAsPriceRatio({
+  const displayPriceRatio = await getCurrentPriceAsWalletPriceRatio({
```
dolcalmi, line 32: this should be `getCurrentPriceAsDisplayPriceRatio`, `getCurrentPriceAsWalletPriceRatio` must only be used in tx related to usd wallets

src/app/accounts/send-default-wallet-balance-to-users.ts
```diff
@@ -26,7 +26,7 @@ export const sendDefaultWalletBalanceToAccounts = async () => {
       const balanceAmount = await LedgerService().getWalletBalanceAmount(wallet)
       if (balanceAmount instanceof Error) return balanceAmount
 
-      const displayPriceRatio = await getCurrentPriceAsPriceRatio({
+      const displayPriceRatio = await getCurrentPriceAsWalletPriceRatio({
```
dolcalmi, line 29: same here `getCurrentPriceAsDisplayPriceRatio` then we will change this to account display currency (which is display currency)

src/app/accounts/active-accounts.ts
```diff
@@ -16,7 +16,7 @@ export const getRecentlyActiveAccounts = async function* ():
   const unlockedAccounts = AccountsRepository().listUnlockedAccounts()
   if (unlockedAccounts instanceof Error) return unlockedAccounts
 
-  const displayPriceRatio = await getCurrentPriceAsPriceRatio({
+  const displayPriceRatio = await getCurrentPriceAsWalletPriceRatio({
```
dolcalmi, line 19: if this is not using `getCurrentPriceAsDisplayPriceRatio` then this should be renamed

src/app/wallets/update-on-chain-receipt.ts
```diff
@@ -257,7 +257,7 @@ const processTxForHotWallet = async ({
 
   const ledger = LedgerService()
 
-  const displayPriceRatio = await getCurrentPriceAsPriceRatio({
+  const displayPriceRatio = await getCurrentPriceAsWalletPriceRatio({
```
dolcalmi, line 260: same here should be `getCurrentPriceAsDisplayPriceRatio`

src/domain/shared/index.types.d.ts
dolcalmi: and if we only modify DisplayAmount?

src/domain/payments/index.types.d.ts
dolcalmi: maybe is better to modify DisplayAmount

src/domain/payments/index.types.d.ts
```diff
@@ -1,11 +1,27 @@
-type PriceRatio = {
+type PriceRatio<S extends WalletCurrency> = {
+  convertFromOther(other: bigint): PaymentAmount<S>
+  convertFromWallet(btcWalletAmount: PaymentAmount<S>): bigint
+  convertFromWalletToFloor(btcWalletAmount: PaymentAmount<S>): bigint
+  convertFromWalletToCeil(btcWalletAmount: PaymentAmount<S>): bigint
+  otherUnitPerWalletUnit(): DisplayCurrencyBasePerSat
```
dolcalmi, line 6: - not sure if should be responsability of this Floor/Ceil 
- `FromWallet` but param is `PaymentAmount` should we use `convertFromPaymentAmount`?
- `Other` is a bit confusing... `convertFromBigInt` or `convertToPaymentAmount`? :thinking:

src/app/cold-storage/rebalance-to-cold-wallet.ts
vindard: Yup planning to do this in a follow-up PR. This PR is just to establish the new `DisplayPriceRatio` object and separate its implementation for the existing `PriceRatio` one that now becomes `WalletPriceRatio`. So places like this remain functionally unchanged, but will be cleaned up to return the new types in a follow-up PR.

src/app/accounts/send-default-wallet-balance-to-users.ts
vindard: https://github.com/GaloyMoney/galoy/pull/2334#discussion_r1108847733

src/app/accounts/active-accounts.ts
vindard: https://github.com/GaloyMoney/galoy/pull/2334#discussion_r1108847733

src/app/wallets/update-on-chain-receipt.ts
vindard: https://github.com/GaloyMoney/galoy/pull/2334#discussion_r1108847733

src/domain/payments/index.types.d.ts
vindard: Yup, but this would be a transition step while we have the two separate types in our codebase still. I haven't changed any of the application code as yet to use this new type. Just establishing the pattern first and looking for feedback on that as our eventual target.

src/domain/shared/index.types.d.ts
vindard: Yup, but this would be a transition step while we have the two separate types in our codebase still. I haven't changed any of the application code as yet to use this new type. Just establishing the pattern first and looking for feedback on that as our eventual target.

src/domain/payments/index.types.d.ts
vindard: https://github.com/GaloyMoney/galoy/pull/2334/files#r1108850976

src/domain/payments/index.types.d.ts
vindard: > * `FromWallet` but param is `PaymentAmount` should we use `convertFromPaymentAmount`?

Planning to rename `PaymentAmount` to `WalletAmount` in a follow-up PR

dolcalmi (approved):
looks good, only a concern with `const displayPriceRatio = await Prices.getCurrentPriceAsWalletPriceRatio` will be difficult to track later but is not a blocker.

Also, all payments/receive flows must be tested in staging

PR #2347, opened 2023-02-21
tests: make build/test work on m1/m2

nicolasburtey:
compilation and test of the backend on m1/m2 breaks some months ago.
the issue with the docker image from bitnami is that they only contains amd64 binary.
so I'm moving to use the "official" images from mongo and redis as those contains arm64 that works correctly on new Mac.

the downside of using the official images is that there is less configurability. for mongodb, it seems (based on very quick search) that adding user needs some dedicated script and I didn't want to do that/maintain it. 

so I created a branch such that we don't need to have username/password in the docker-compose environement for mongodb.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -73,7 +73,7 @@ services:
       - DB_DEBUG=false
     depends_on: [postgres-price-history]
   redis:
-    image: bitnami/redis:7.0.7
+    image: redis:7.0.8
     environment:
       - ALLOW_EMPTY_PASSWORD=yes
       - REDIS_DISABLE_COMMANDS=FLUSHDB,FLUSHALL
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -85,14 +85,9 @@ services:
     environment:
       - MONGODB_ADDRESS=mongodb
   mongodb:
-    image: bitnami/mongodb:5.0.10-debian-11-r3
+    image: mongo:4.4
```
krtk6160, line 88: Should we downgrade this? We're running mongo 5 in staging and prod

nicolasburtey: which version? I just pick this because this was the last version published on docker. 

I thought bitnami version wasn't following the version of the underlying app

krtk6160: We're using `MongoDB server version: 5.0.10` atm

PR #2349, opened 2023-02-21
chore: test using cache

nicolasburtey:
@dolcalmi it seems we could speed up our test quite a bit with the cache. assigning it to you in case you want to have a deeper look.


dolcalmi (approved):
@nicolasburtey I did some tests and there is a small improvement but when test fails the re-run improves like 30 seconds so I think is good enough to merge

PR #2351, opened 2023-02-22
fix: correct receive-side settlementAmount calculation for API

vindard:
## Description

This fixes the issue where in mongo `satsAmount` and `centsAmount` reflect receive amounts less fees which no longer matches with the `credit` value which includes fees when doing display price calculations.


vindard:
Discovered that was only a bug that affected Dealer's transaction view

PR #2363, opened 2023-02-24
fix: add id to currencyList

dolcalmi:
Update currencyList query to use id instead of code

### DIFF
--- a/src/graphql/main/schema.graphql
+++ b/src/graphql/main/schema.graphql
@@ -253,8 +253,8 @@ type Coordinates {
 }
 
 type Currency {
-  code: String!
   flag: String!
+  id: String!
   name: String!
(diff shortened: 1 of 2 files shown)


src/graphql/main/schema.graphql
```diff
@@ -253,8 +253,8 @@ type Coordinates {
 }
 
 type Currency {
-  code: String!
   flag: String!
+  id: String!
```
nicolasburtey, line 257: `id` should be type `ID`. it's an alias for string expect I think it means it should be unique.

nicolasburtey: which is how it's automatically cached by the client afaiu

nicolasburtey (approved):

PR #2366, opened 2023-02-24
chore: need iteration displayCurrency

nicolasburtey:


src/services/notifications/create-push-notification-content.ts
```diff
@@ -26,7 +26,7 @@ export const createPushNotificationContent = ({
     { walletCurrency: baseCurrency },
   )
   const baseCurrencyName = baseCurrency === WalletCurrency.Btc ? "sats" : ""
-  const baseCurrencySymbol = baseCurrency === WalletCurrency.Usd ? "$" : ""
+  const baseCurrencySymbol = baseCurrency === WalletCurrency.Usd ? fiatSymbol : ""
```
vindard, line 29: This might be ok here since `$` is hardcoded to USD WalletCurrency (not display currency). The display currency symbols are set below on line 54.

nicolasburtey: probably need a space between `+₡500` and `CRC`

PR #2372, opened 2023-02-27
chore: removing unused files

nicolasburtey:


jest.config.js
```diff
@@ -1,26 +0,0 @@
-module.exports = {
```
dolcalmi, line 1: thought it was loaded automatically by jest https://jestjs.io/docs/configuration

nicolasburtey: are we not passing a custom path to the configuration file for every test we do? 

ie: https://github.com/GaloyMoney/galoy/blob/main/package.json#L20

dolcalmi (approved):

PR #2375, opened 2023-02-27
fix: cookie expiration date

ntheile:
This PR fixes the cookie session expiration date. Before it would expire after a few hours. Now it will not expire until the kratos session expires (kratos session lifespan setting, typically 30 days/720h)

### DIFF
--- a/src/servers/middlewares/auth-router.ts
+++ b/src/servers/middlewares/auth-router.ts
@@ -23,6 +23,7 @@ import setCookie from "set-cookie-parser"
 import cookieParser from "cookie-parser"
 import { logoutCookie } from "@app/auth"
 import { checkedToPhoneNumber } from "@domain/users"
+import libCookie from "cookie"
 
 const authRouter = express.Router({ caseSensitive: true })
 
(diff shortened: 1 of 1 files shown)


src/servers/middlewares/auth-router.ts
```diff
@@ -168,21 +169,36 @@ authRouter.post(
...
+            sameSite: "none",
+            secure: kratosSessionCookie.secure,
+            httpOnly: kratosSessionCookie.httpOnly,
+            path: kratosSessionCookie.path,
+          },
+        )
+        const session = await kratosPublic.toSession({ cookie: kratosCookieStr })
+        const expiresAt = session.data.expires_at
+          ? new Date(session.data.expires_at)
+          : new Date(new Date().setDate(new Date().getDate() + 30))
```
nicolasburtey, line 187: should avoid the magic number, or document it with a dedicated const

nicolasburtey:
> This PR fixes the cookie session expiration date. Before it would expire after a few hours. Now it will not expire until the kratos session expires (kratos session lifespan setting, typically 30 days/720h)

the idea of shorter session for cookie (versus token) actually make sense for me? but I understand this should be a setting from kratos not from our code ideally.

but I don't think it's possible from kratos to say that cookie should have short duration than token

src/servers/middlewares/auth-router.ts
ntheile: yeah chatgpt wrote that for me lol. it stands for 30 days from now. where should i put the dedicated const?

ntheile:
> 

yeah it was a bug, the maxAge is used in browsers and the expires field in cookies are deprecated now https://mrcoles.com/blog/cookies-max-age-vs-expires/#:~:text=Quick%20Answer%3A,mostly)%20all%20browsers%20support%20expires

bodymindarts (approved):

PR #2378, opened 2023-02-28
test: fix flaky channel-fee setting by skipping test conditionally

vindard:
## Description

There is a recently introduced routed-fee test that fails sometimes if channel-fee setting doesn't work properly. The first alternative would be to skip this test since it was recently introduced to test an edge case. I opted instead to explicitly check for flakiness and skip the test only if flakiness is detected.


test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -1235,6 +1236,70 @@ describe("UserWallet - Lightning Pay", () => {
...
+
+          const { public_key: partnerPubKey } = policies[1]
+          expect(partnerPubKey).toBe(process.env.LND_OUTSIDE_2_PUBKEY)
+
+          return true
+        }
+
+        // Skip rest of test if fee-setting fails (flaky)
+        const updateResult = await flakyUpdateLndOutside1ChannelFee()
+        if (updateResult instanceof Error) return
```
dolcalmi, line 1298: tests needs to be deterministic.. not sure if just skipping the test is a solution

dolcalmi (changes requested):

test/integration/02-user-wallet/02-send-lightning.spec.ts
vindard: Not sure what else to do from here. Only other alternative I can see is to `it.skip` the entire test if we want to get rid of flakiness

vindard: I just tried an alternative where we set the fees on channel open. This didn't work either.

Not sure how else to set those channel fees in a non-flaky way 🤷🏾‍♂️ ... if we can't find a way our options would be to:
- keep the flakiness
- skip the routed-fee test entirely

PR #2398, opened 2023-03-02
chore: use latest stablesats in docker-compose

bodymindarts:


dolcalmi (approved):

dev/stablesats.yml
```diff
@@ -1,6 +1,10 @@
 pubsub:
   host: redis
 
+exchanges:
+  okex:
+    weight: 1.0
```
dolcalmi, line 6: is this enough.. integration test is returning `UnknownDealerPriceServiceError`

dolcalmi (changes requested):

dolcalmi (approved):

PR #2414, opened 2023-03-09
feat: add minorUnitToMajorUnitOffset to realtime price query/subscription

dolcalmi:
- ~currencyUnit now is an Enum (default is `MINOR` but adds the option to include/return `MAJOR` based on a param)~
- Add minorUnitToMajorUnitOffset to realtimePrice query and subscriptions

### DIFF
--- a/src/domain/fiat/display-currency.ts
+++ b/src/domain/fiat/display-currency.ts
@@ -2,6 +2,7 @@ export const CENTS_PER_USD = 100
 
 export const MajorExponent = {
   STANDARD: 2,
+  ZERO: 0,
   ONE: 1,
   THREE: 3,
 } as const
@@ -25,5 +26,34 @@ export const majorToMinorUnit = ({
   displayMajorExponent: CurrencyMajorExponent
 }) => Number(Number(amount) * 10 ** displayMajorExponent)
 
-export const usdMajorToMinorUnit = (amount: number | bigint) =>
-  majorToMinorUnit({ amount, displayMajorExponent: MajorExponent.STANDARD })
+export const currencyMajorToMinorUnit = ({
+  amount,
+  displayCurrency,
+}: {
+  amount: number | bigint
+  displayCurrency: DisplayCurrency
+}) => {
+  const displayMajorExponent = getCurrencyMajorExponent(displayCurrency)
+  return majorToMinorUnit({ amount, displayMajorExponent })
+}
+
+export const getCurrencyMajorExponent = (
+  currency: DisplayCurrency,
+): CurrencyMajorExponent => {
+  try {
+    const formatter = new Intl.NumberFormat("en-US", { style: "currency", currency })
+    const { minimumFractionDigits } = formatter.resolvedOptions()
+    switch (minimumFractionDigits) {
+      case 0:
+        return MajorExponent.ZERO
+      case 1:
+        return MajorExponent.ONE
+      case 3:
+        return MajorExponent.THREE
+      default:
+        return MajorExponent.STANDARD
+    }
+  } catch {
+    return MajorExponent.STANDARD
+  }
+}
--- a/src/graphql/main/schema.graphql
+++ b/src/graphql/main/schema.graphql
@@ -260,6 +260,11 @@ type Currency {
   symbol: String!
 }
 
+enum CurrencyUnit {
+  MAJOR
+  MINOR
+}
+
 input DeviceNotificationTokenCreateInput {
   deviceToken: String!
 }
@@ -929,8 +934,7 @@ input PriceInput {
 
 interface PriceInterface {
   base: SafeInt!
-  currencyUnit: String!
-    @deprecated(reason: "Deprecated in favor of minorUnitToMajorUnitOffset")
+  currencyUnit: CurrencyUnit!
   minorUnitToMajorUnitOffset: Int!
   offset: Int!
 }
@@ -940,8 +944,7 @@ Price of 1 sat in base/offset. To calculate, use: `base / 10^offset`
 """
 type PriceOfOneSat implements PriceInterface {
   base: SafeInt!
-  currencyUnit: String!
-    @deprecated(reason: "Deprecated in favor of minorUnitToMajorUnitOffset")
+  currencyUnit: CurrencyUnit!
   minorUnitToMajorUnitOffset: Int!
   offset: Int!
 }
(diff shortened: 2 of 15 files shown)


src/graphql/main/schema.graphql
```diff
@@ -929,7 +934,8 @@ input PriceInput {
 
 interface PriceInterface {
   base: SafeInt!
-  currencyUnit: String!
+  currencyUnit: CurrencyUnit!
```
nicolasburtey, line 937: is this not used in prod already?

nicolasburtey: I don't quite get why we want to return Major or Minor?  Why no already return Minor?

src/graphql/main/schema.graphql
```diff
@@ -260,6 +260,11 @@ type Currency {
   symbol: String!
 }
 
+enum CurrencyUnit {
+  MAYOR
```
nicolasburtey, line 264:
```suggestion
  MAJOR
```

src/graphql/main/schema.graphql
dolcalmi: this only a change for realtimePrice, I asked @UncleSamtoshi about this field and is not used also I am changing it because does not make sense to use `${currency}CENTS` for all currencies (specially for currencies that dont use CENTS as minor unit).

src/domain/fiat/display-currency.ts
```diff
@@ -25,5 +26,34 @@ export const majorToMinorUnit = ({
...
+    const { minimumFractionDigits } = formatter.resolvedOptions()
+    switch (minimumFractionDigits) {
+      case 0:
+        return MajorExponent.ZERO
+      case 1:
+        return MajorExponent.ONE
+      case 3:
+        return MajorExponent.THREE
+      default:
+        return MajorExponent.STANDARD
```
vindard, line 54: Do we think it's worth it to do something like this in case something weird comes back from the library?:

```
case 2:
        return MajorExponent.STANDARD
default:
        recordExceptionInCurrentSpan({
            error: new UnexpectedMinimumFractionDigitsError(`${minimumFractionDigits}`),
            level: ErrorLevel.Critical,
        })
        return MajorExponent.STANDARD
```

src/graphql/main/schema.graphql
```diff
@@ -929,7 +934,8 @@ input PriceInput {
 
 interface PriceInterface {
   base: SafeInt!
-  currencyUnit: String!
+  currencyUnit: CurrencyUnit!
+  minorUnitToMajorUnitOffset: Int!
```
vindard, line 938: I like this name. It's longer than something like `currencyMinorUnitOffset` but it removes all ambiguity.

src/graphql/main/schema.graphql
vindard: I think the more important information here might be more which currency this is (`USD`/`EUR`/`CRC`) and less whether it's `MAJOR`/`MINOR`. I can't think of a case where we'd need `MAJOR` sometime soon... and if no-one else can either, then maybe we can deprecate this and re-introduce it when we need it again

src/graphql/root/subscription/my-updates.ts
```diff
@@ -250,8 +271,8 @@ const MeSubscription = {
       const priceData = {
         timestamp: pricePerSat.timestamp,
         displayCurrency,
-        centsPerSat: usdMajorToMinorUnit(pricePerSat.price),
```
vindard, line 253: Did this used to be major unit coming out of `pricePerSat` before?

vindard (commented):
Few questions/comments

src/domain/fiat/display-currency.ts
dolcalmi: :thinking: domain layer should not depend on service layer...

src/graphql/root/subscription/my-updates.ts
dolcalmi: price service returns price per sat in major unit.

src/graphql/main/schema.graphql
dolcalmi: Add deprecation reason to currency unit https://github.com/GaloyMoney/galoy/pull/2414/commits/f48c847ae9cdff7666dd0d079aee0717b5b5633b

src/domain/fiat/display-currency.ts
bodymindarts: Domain layer shouldn't depend on external dependencies at all. If it does (eg as a wrapper like with the `domain/bitcoin` module) we need to deal with its possible return values explicitly to make it safe for other parts of the code to use.

Even if we don't want to return an exception we should at least add some attribute to make it visible that the library we are using isn't working as expected.

src/domain/fiat/display-currency.ts
```diff
@@ -25,5 +26,34 @@ export const majorToMinorUnit = ({
...
+    }
+  } catch {
+    return MajorExponent.STANDARD
```
bodymindarts, line 57: Silent catching is not good! Either this is a -we want it to crash- situation (ie remove try/catch) or we should be explicit about what we are catching and at least add it to the trace.

src/graphql/main/schema.graphql
```diff
@@ -938,7 +945,9 @@ Price of 1 sat in base/offset. To calculate, use: `base / 10^offset`
 """
 type PriceOfOneSat implements PriceInterface {
   base: SafeInt!
-  currencyUnit: String!
+  currencyUnit: CurrencyUnit!
+    @deprecated(reason: "Deprecated in favor of minorUnitToMajorUnitOffset")
```
bodymindarts, line 949: Which line is the `@deprecated` applying to? The `currencyUnit`?
Why are we changing the type if we are deprecating it?

src/graphql/types/scalar/currency-unit.ts
```diff
@@ -0,0 +1,11 @@
+import { GT } from "@graphql/index"
+
+const CurrencyUnit = GT.Enum({
```
bodymindarts, line 3: why is this added if its also deprecated instantly?

src/graphql/main/schema.graphql
dolcalmi: now `${currency}CENTS` does not make sense... the deprecation was requested in the last display currency meeting with mobile devs

src/graphql/types/scalar/currency-unit.ts
dolcalmi: https://github.com/GaloyMoney/galoy/pull/2414#discussion_r1135669342

src/domain/fiat/display-currency.ts
dolcalmi: so tracing service is an exception? I can add a comment but this is necessary for alias currencies (we dont want to crash to get the major exponent)

dolcalmi:
closed in favor of https://github.com/GaloyMoney/galoy/pull/2439/

PR #2419, opened 2023-03-09
chore: update default mobile min build versions

ntheile:
checked with sam on the app/play store that version 362 is a recent stable version of the apk and ipa.

### DIFF
--- a/galoy.yaml
+++ b/galoy.yaml
@@ -14,11 +14,11 @@ dealer:
 ratioPrecision: 1000000
 buildVersion:
   android:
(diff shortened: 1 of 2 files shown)


nicolasburtey (approved):

nicolasburtey:
Approved

Have you checked if the wallet show a message that an update is available from the old app?

ntheile:
Tested on a really old android version 1.7.7 build 310 (April 2022)
Tested on ios version 2.0.61 build 361 (Dec 2022)

PR #2425, opened 2023-03-10
fix: backend starting up on mac m1/m2

nicolasburtey:


docker-compose.yml
```diff
@@ -85,7 +85,7 @@ services:
     environment:
       - MONGODB_ADDRESS=mongodb
   mongodb:
-    image: mongo:5.0.10
+    image: mongo:${MONGODB_VERSION}
```
bodymindarts, line 88: Can we set a default here?

.env.ci
```diff
@@ -59,3 +58,4 @@ KRATOS_MASTER_USER_PASSWORD=${KRATOS_MASTER_USER_PASSWORD}
 KRATOS_PUBLIC_API=http://kratos:4433
 KRATOS_ADMIN_API=http://kratos:4434
 KRATOS_CALLBACK_API_KEY=${KRATOS_CALLBACK_API_KEY}
+MONGODB_VERSION=${MONGODB_VERSION}
```
bodymindarts, line 61: This should be hard coded or omitted if we can set a default directly in the docker-compose file.

docker-compose.yml
bodymindarts:
```suggestion
    image: mongo:${MONGODB_VERSION:-5.0.10}
```
Should work

.env.ci
bodymindarts: Need to check if the default works

PR #2426, opened 2023-03-10
fix: backend starting up on mac m1/m2

nicolasburtey:


bodymindarts (approved):
Default works - merge when tests are green.

PR #2427, opened 2023-03-12
feat: creating dedicated websocket server

nicolasburtey:
Pulling code from graphql-server to make the websocket server independent

This has multiple benefits:

- Because websockets are traditional http connection (at least until the `upgrade` protocol exchange has happened), it is not really possible to know when a request arrive from nginx whenever they are meant to be a traditional query/mutation, or a websocket. 
By having a different server, we can point the websocket server to `ws.domain.com` (while the current API would stay at `api.domain.com`) and this would make it much easier to filter query and understand when some issues arise whether they are related to websocket queries or not.
- This is a pre-requirement to get Apollo Federation. Because Apollo Federation doesn't support subscription, it is require to get a separate subscription
- minor point: it generally will make the http server code less convoluted (to me at least)

Note: the current code for graphql-server also support legacy websocket protocol (that is incompatible with the new protocol). the code for this new server only support the new protocol. 

This will also make it easier to eventually properly instrument the server for open telemetry.

To make the transition, we will need to:
- [ ] update the charts to have this new server. https://github.com/GaloyMoney/charts/pull/2719
- update the mobile app, so that the websocket point to the new url
- after all mobile clients have updated to a version that point to the new url, we can deprecated the websocket code from the current server


src/servers/ws-server.ts
```diff
@@ -0,0 +1,175 @@
...
+
+import { setupMongoConnection } from "@services/mongodb"
+
+import { baseLogger } from "@services/logger"
+
+import { sessionContext } from "./graphql-server"
+
+const schema = gqlMainSchema
+
+const port = 4000
```
ntheile, line 26: should we make this configurable?

nicolasburtey: Yup. Feel free to take over this Pr if you'd like @ntheile 

I can do it but if you want a short term task this is probably a  good one.

ntheile: yeah i can work on it!

nicolasburtey:
approved @ntheile .  I can't approved myself but you can approve on your end and merge it

ntheile (approved):

PR #2433, opened 2023-03-13
fix: add higher precision walletPriceRatio to PaymentFlow objects

vindard:
## Description

The WalletPriceRatio encoded via the payment amounts inside a `PaymentFlow` sometimes lose some or all precision if the payment amounts are too small. This PR is to add an explicit high precision WalletPriceRatio to PaymentFlow.

### DIFF
--- a/src/app/payments/get-protocol-fee.ts
+++ b/src/app/payments/get-protocol-fee.ts
@@ -5,7 +5,6 @@ import {
   LnPaymentRequestNonZeroAmountRequiredError,
   LnPaymentRequestZeroAmountRequiredError,
   SkipProbeForPubkeyError,
-  WalletPriceRatio,
 } from "@domain/payments"
 import { LndService } from "@services/lnd"
 
@@ -167,7 +166,7 @@ const estimateLightningFee = async ({
       : "undefined",
   })
 
-  const priceRatio = WalletPriceRatio({ usd: usdPaymentAmount, btc: btcPaymentAmount })
+  const priceRatio = await builder.walletPriceRatio()
   if (priceRatio instanceof Error) return PartialResult.err(priceRatio)
 
   let paymentFlow: PaymentFlow<WalletCurrency, WalletCurrency> | ApplicationError
--- a/src/domain/payments/price-ratio.ts
+++ b/src/domain/payments/price-ratio.ts
@@ -212,3 +212,31 @@ export const toDisplayPriceRatio = <S extends WalletCurrency, T extends DisplayC
     displayMajorExponent,
   })
 }
+
+export const priceRatioFromUsdFromBtc = async (usdFromBtc: UsdFromBtcMidPriceFn) => {
+  const btcForPriceRatio = {
+    amount: BigInt(RATIO_PRECISION),
+    currency: WalletCurrency.Btc,
+  }
+  const usdForPriceRatio = await usdFromBtc(btcForPriceRatio)
+  if (usdForPriceRatio instanceof Error) return usdForPriceRatio
+
+  return WalletPriceRatio({
+    usd: usdForPriceRatio,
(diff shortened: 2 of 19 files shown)


src/domain/payments/price-ratio.ts
```diff
@@ -212,3 +212,31 @@ export const toDisplayPriceRatio = <S extends WalletCurrency, T extends DisplayC
     displayMajorExponent,
   })
 }
+
+export const priceRatioFromUsdFromBtc = async (usdFromBtc: UsdFromBtcMidPriceFn) => {
```
dolcalmi, line 216: please include return type

src/domain/payments/price-ratio.ts
```diff
@@ -212,3 +212,31 @@ export const toDisplayPriceRatio = <S extends WalletCurrency, T extends DisplayC
...
+  const usdForPriceRatio = await usdFromBtc(btcForPriceRatio)
+  if (usdForPriceRatio instanceof Error) return usdForPriceRatio
+
+  return WalletPriceRatio({
+    usd: usdForPriceRatio,
+    btc: btcForPriceRatio,
+  })
+}
+
+export const priceRatioFromBtcFromUsd = async (btcFromUsd: BtcFromUsdMidPriceFn) => {
```
dolcalmi, line 230: please include return type

src/services/mongoose/schema.ts
```diff
@@ -363,6 +363,8 @@ const paymentFlowStateSchema = new Schema<PaymentFlowStateRecord>(
     usdPaymentAmount: { type: Number, required: true },
     inputAmount: { type: Number, required: true },
 
+    walletPriceRatio: { type: String, required: true },
```
dolcalmi, line 366: do we need a migration?

src/services/mongoose/payment-flow.ts
```diff
@@ -221,6 +225,9 @@ const paymentFlowFromRaw = <S extends WalletCurrency, R extends WalletCurrency>(
   })
   if (usdProtocolAndBankFee instanceof Error) return usdProtocolAndBankFee
 
+  const walletPriceRatio = toWalletPriceRatio(Number(paymentFlowState.walletPriceRatio))
```
dolcalmi, line 228: what happen to previous payment flows?

src/app/payments/get-protocol-fee.ts
```diff
@@ -167,7 +166,7 @@ const estimateLightningFee = async ({
       : "undefined",
   })
 
-  const priceRatio = WalletPriceRatio({ usd: usdPaymentAmount, btc: btcPaymentAmount })
+  const priceRatio = await builder.walletPriceRatio()
```
dolcalmi, line 169: is this builder responsability? we use builder calling specific methods to get a final/structured object at the end.. I understand this but not sure if should be part of builder

dolcalmi (changes requested):

src/app/payments/get-protocol-fee.ts
vindard: It's reading values out at an intermediate state of the builder which we already do for payment amounts as well.

src/services/mongoose/schema.ts
vindard: Maybe not because they only live for 5 mins in the collection and get deleted after. There will be an intermediate state though where if someone has done a fee query and then we deploy and then they try to pay that they'll see an error in the 5 mins after their fee query. 

This might be an edge case that may not be worth explicitly accommodating for?

bodymindarts:
Don't quite understand what issue you are solving here. We should respect the volume sensitivity of stablesats price response.

vindard:
Closed in favour of #2481 which used the mid-price ratio precedent for checks

PR #2439, opened 2023-03-14
refactor: realtime and settlement price types

dolcalmi:


src/graphql/types/object/transaction.ts
```diff
@@ -81,21 +84,29 @@ const Transaction = GT.Object<WalletTransaction>({
...
-          base: Math.round(
-            displayCurrencyPerSettlementCurrencyUnitInCents *
-              10 ** SAT_PRICE_PRECISION_OFFSET,
-          ),
-          offset: SAT_PRICE_PRECISION_OFFSET,
-          currencyUnit: "USDCENT",
+          base: Math.round(displayCurrencyPriceInMinorUnit * 10 ** offset),
+          offset,
+          currencyUnit: "MINOR",
+          formattedAmount: displayCurrencyPriceInMinorUnit.toString(),
```
vindard, line 106: Would `toString()` always work, or should we do something like `${displayCurrencyPriceInMinorUnit}`

src/graphql/main/schema.graphql
```diff
@@ -1172,9 +1182,9 @@ type Transaction {
   settlementFee: SignedAmount!
 
   """
-  Price in USDCENT/SETTLEMENTUNIT at time of settlement.
+  Price in WALLETCURRENCY/SETTLEMENTUNIT at time of settlement.
   """
-  settlementPrice: Price!
+  settlementPrice: SettlementPriceInMinorUnit!
```
vindard, line 1187: Tongue-in-cheek suggestion, would `PriceOfOneSettlementMinorUnitInDisplayMinorUnit` be too long for here?

vindard (commented):
This looks ok to me based on what we discussed. I'd leave the approve for one more person to review (maybe someone from frontend).

src/graphql/main/schema.graphql
bodymindarts: Yeah I think its a good suggestion - cannot think of anything better really.

dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/2439/commits/1dfb099095a98c412a0e220e2342ab0ccf04b6a3

src/graphql/types/object/transaction.ts
dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/2439/commits/1dfb099095a98c412a0e220e2342ab0ccf04b6a3

PR #2446, opened 2023-03-15
chore: more mobile compatibility

nicolasburtey:
couple of updates:
- ensuring the schema.graphql are always up to date (needed to make sure the tooling doing the comparison are using an up to date files)
- adding the generated.gql from the current deployed application (2.1.66). need more CI/CD iteration on that
- adding associated github actions


src/graphql/main/index.ts
```diff
@@ -1,28 +1,37 @@
...
-
 export const gqlMainSchema = new GraphQLSchema({
   query: QueryType,
   mutation: MutationType,
   subscription: SubscriptionType,
   types: ALL_INTERFACE_TYPES,
 })
+
+const updateSchemaGraphql = async () => {
+  console.log("updating schema.graphql")
```
dolcalmi, line 22: shouldnt we use logger?

src/graphql/admin/index.ts
```diff
@@ -9,17 +9,26 @@ import { ALL_INTERFACE_TYPES } from "./types"
...
-}
-
 export const gqlAdminSchema = new GraphQLSchema({
   query: QueryType,
   mutation: MutationType,
   types: ALL_INTERFACE_TYPES,
 })
+
+const updateSchemaGraphql = async () => {
+  console.log("updating schema.graphql")
```
dolcalmi, line 19: same here

bodymindarts:
I would take a different approach to this. Lets sync

bodymindarts:
How about this: https://github.com/GaloyMoney/galoy/pull/2448? It depends on https://github.com/GaloyMoney/galoy-mobile-deployments/tree/main/deployed-schemas which gets updated every 'beta' deploy. If we want to drop compatibility we just delete a file in that folder.

src/graphql/admin/index.ts
nicolasburtey: no need. it's just used as output info when running github action/locally (it's a script) so no need to wrap it in object

dolcalmi (approved):

PR #2448, opened 2023-03-16
ci: add mobile-schema-compatibility action

bodymindarts:


ci/tasks/check-mobile-schema-compatibility.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+curl -fsS https://raw.githubusercontent.com/GaloyMoney/galoy-mobile/main/app/graphql/generated.gql -o main.gql
+
+echo Checking compatibility with current main branch
+npx @graphql-inspector/cli validate ./main.gql repo/src/graphql/main/schema.graphql --apollo --noStrictFragments
```
nicolasburtey, line 6: it should not be tied to main branch on the backend. it should be running in the current branch.

bodymindarts: it is the current branch. Its main branch of Galoy-mobile

nicolasburtey (commented):
this looks good expect:
- we're missing the check schema.graphql is up to date
- we should do the check with local branch, not main

nicolasburtey (approved):

PR #2450, opened 2023-03-16
fix: major exponent

dolcalmi:
Update minorToMajorUnit and majorToMinorUnit to use currency major exponent

Merge after https://github.com/GaloyMoney/galoy/pull/2439


src/app/cold-storage/rebalance-to-cold-wallet.ts
```diff
@@ -81,19 +81,21 @@ export const rebalanceToColdWallet = async (): Promise<boolean | ApplicationErro
   })
   if (rebalanceBtcAmount instanceof Error) return rebalanceBtcAmount
   const amountDisplayCurrencyAmount = displayPriceRatio.convertFromBtc(rebalanceBtcAmount)
-  const amountDisplayCurrency = usdMinorToMajorUnit(
-    amountDisplayCurrencyAmount.amount,
-  ) as DisplayCurrencyBaseAmount
+  const amountDisplayCurrency = minorToMajorUnit({
```
vindard, line 84: I'm noticing here now that there's some cleanup I need to do with calling WalletPriceRatio/DisplayPriceRatio still. I'll probably start building it on these PRs

src/app/wallets/update-on-chain-receipt.ts
```diff
@@ -249,7 +244,7 @@ const processTxForWallet = async (
             amount: Number(
               minorToMajorUnit({
                 amount: creditAccountAdditionalMetadata.displayAmount,
-                displayMajorExponent: MajorExponent.STANDARD,
+                displayCurrency: creditAccountAdditionalMetadata.displayCurrency,
```
vindard, line 247: Like here. I think after my refactor (after this branch) the `displayCurrency` could come directly from `amount` field if it's typed as `DisplayAmount` and then we could remove the `displayCurrency` param here

src/domain/fiat/display-currency.ts
```diff
@@ -5,40 +5,44 @@ export const MajorExponent = {
...
   amount,
   displayCurrency,
 }: {
   amount: number | bigint
   displayCurrency: DisplayCurrency
-}) => {
+}): number => {
   const displayMajorExponent = getCurrencyMajorExponent(displayCurrency)
-  return majorToMinorUnit({ amount, displayMajorExponent })
+  return Number(Number(amount) * 10 ** displayMajorExponent)
```
vindard, line 45: Could `Number(amount) * 10 ** displayMajorExponent`

src/domain/fiat/display-currency.ts
```diff
@@ -5,40 +5,44 @@ export const MajorExponent = {
...
-
-// TODO: remove after remove hardcoded DisplayCurrency.Usd
-export const usdMajorToMinorUnit = (amount: number | bigint) =>
-  majorToMinorUnit({ amount, displayMajorExponent: MajorExponent.STANDARD })
+  displayCurrency: DisplayCurrency
+  fixed?: boolean
+}): number => {
+  const displayMajorExponent = getCurrencyMajorExponent(displayCurrency)
+  const majorAmount = Number(amount) / 10 ** displayMajorExponent
+  return fixed ? Number(majorAmount.toFixed(displayMajorExponent)) : majorAmount
```
vindard, line 34: I think the two values in this conditional work out to the same, redundant conditional maybe. Looks like the string stuff is covered with `minorToMajorUnitFormatted` anyway too.

vindard (commented):
Left a few small comments

src/domain/fiat/display-currency.ts
dolcalmi: `fixed: false` it is needed for displayCurrencyPerSettlementCurrencyMajorUnit

src/domain/fiat/display-currency.ts
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/2450/commits/d0583cab39d5ca7c0c4ebe6daf5d4ab0c2f529e6

src/domain/fiat/display-currency.ts
vindard: Oh right, the `fixed` name isn't intuitive though. ChatGPT recommended the following:
- [ ] useRounded
- [ ] applyPrecision
- [ ] roundResult
- [x] exactPrecision
- [ ] isRounded

I like the `exactPrecision` one

vindard (changes requested):
Just one requested change, but otherwise this look good

vindard (approved):

PR #2454, opened 2023-03-17
feat:save contacts after making intra ledger payment

tonyguesswho:
Save contacts After making lightning payment between two bbw users

### DIFF
--- a/src/app/payments/helpers.ts
+++ b/src/app/payments/helpers.ts
@@ -1,4 +1,5 @@
 import { btcFromUsdMidPriceFn, usdFromBtcMidPriceFn } from "@app/prices"
+import { Accounts } from "@app"
 import {
   getAccountLimits,
   getPubkeysToSkipProbe,
@@ -289,3 +290,33 @@ export const getPriceRatioForLimits = wrapAsyncToRunInSpan({
     return WalletPriceRatio(paymentAmounts)
   },
 })
+
+export const addContactsAfterSend = async ({
+  senderAccount,
+  recipientAccount,
+}: {
+  senderAccount: Account
+  recipientAccount: Account
+}): Promise<true | ApplicationError> => {
+  if (!(senderAccount.contactEnabled && recipientAccount.contactEnabled)) {
+    return true
+  }
+
+  if (recipientAccount.username) {
+    const addContactToPayerResult = await Accounts.addNewContact({
+      accountId: senderAccount.id,
+      contactUsername: recipientAccount.username,
+    })
+    if (addContactToPayerResult instanceof Error) return addContactToPayerResult
+  }
+
+  if (senderAccount.username) {
+    const addContactToPayeeResult = await Accounts.addNewContact({
+      accountId: recipientAccount.id,
+      contactUsername: senderAccount.username,
+    })
+    if (addContactToPayeeResult instanceof Error) return addContactToPayeeResult
+  }
+
+  return true
+}
--- a/src/app/payments/helpers.ts
+++ b/src/app/payments/helpers.ts
@@ -1,5 +1,5 @@
 import { btcFromUsdMidPriceFn, usdFromBtcMidPriceFn } from "@app/prices"
-import { Accounts } from "@app"
+import { addNewContact } from "@app/accounts/add-new-contact"
 import {
   getAccountLimits,
   getPubkeysToSkipProbe,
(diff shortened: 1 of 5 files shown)


vindard (commented):
Thanks for this @tonyguesswho! The implementation is already looking pretty good.

To round it off though, could you maybe take a look at `test/integration/02-user-wallet/02-send-lightning.spec.ts` and add tests where appropriate for a test that initiates via lightning but settles intraledger? We already have contacts tests in some of the tests that are intraledger-initiated/intraledger-settled in that same file for reference.

---

**_Edit 1_**

Ah actually some of the checks are failing
- prettier check failed, you should be able to check this and fix locally by following from [here](https://github.com/GaloyMoney/galoy/blob/bd487c2acf15e7fd23ee967782706ab9c58a6790/DEV.md#running-checks)
- one of the integration tests failed. You can see with `TEST="01|02" make reset-integration` for initial full test run and then to isolate you can change the specific broken test to `it.only(...)` and re-run with `TEST="02-s.*l.*g\." make integration` to iterate when you start working on the broken test

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -1234,7 +1236,13 @@ describe("UserWallet - Lightning Pay", () => {
         // }))
         // await paymentOtherGaloyUser({walletPayee: userWalletB, walletPayer: userwalletC})
         expect(accountRecordA.contacts).toEqual(
-          expect.not.arrayContaining([expect.objectContaining({ id: usernameC })]),
+          expect.arrayContaining([expect.objectContaining({ id: usernameB })]),
+        )
+        expect(accountRecordB.contacts).toEqual(
```
vindard, line 1241: For this test, we may not be testing the new functionality completely since we aren't checking if the contact already existed on the user before we did this payment.

Ideally we should load each account before and test for the absence of the contact we're interested in, execute the payment, and then load the account again and confirm that the contact is now added.

Also if we take this approach, we may not be able to use this specific test since it runs twice (for each of the test cases) and so on the 2nd run the absence check would fail.

I'd recommend:
- adding the absence check as well before payment & presence check
- moving these contact checks to a new `it` test outside of the forEach loop

vindard (changes requested):
Added one comment about the effectiveness of the test we added.

src/app/payments/helpers.ts
```diff
@@ -1,5 +1,5 @@
 import { btcFromUsdMidPriceFn, usdFromBtcMidPriceFn } from "@app/prices"
-import { Accounts } from "@app"
+import { addNewContact } from "@app/accounts/add-new-contact"
```
vindard, line 2: Unless this was somehow necessary I wouldn't change this. We have some wrapping on functions for opentelemetry when they're accessed like `import { Accounts } from "@app"`

src/domain/accounts/index.types.d.ts
```diff
@@ -73,7 +73,7 @@ type Account = {
   title: BusinessMapTitle | null
   coordinates: Coordinates | null
   contactEnabled: boolean
-  readonly contacts: AccountContact[]
```
vindard, line 76: We shouldn't be changing types in the domain to facilitate tests

test/integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -1208,15 +1208,31 @@ describe("UserWallet - Lightning Pay", () => {
...
           ).not.toBeInstanceOf(Error)
         }
-
-        const accountRecordA = await getAccountRecordByTestUserRef("A")
+        accountA.contacts = []
+        accountB.contacts = []
+        accountC.contacts = []
+        await AccountsRepository().update(accountA)
+        await AccountsRepository().update(accountB)
+        await AccountsRepository().update(accountC)
```
vindard, line 1216: Resetting the contacts array was a good way of not having to move this test outside the for loop 👍🏾 

This is another way to do this by importing `Account` from `"@services/mongoose/utils"` directly. It gets around having to change the domain type.

```suggestion
        await Account.updateMany(
          { _id: { $in: [accountA.id, accountB.id, accountC.id].map(toObjectId) } },
          { $set: { contacts: [] } },
        )
```

vindard (changes requested):
One requested change set, but otherwise looks good

vindard (approved):

PR #2466, opened 2023-03-22
chore: cleanup wallet & display price ratios

vindard:
## Description

This is mostly a housekeeping PR to:
- ensure we use wallet vs. display price ratio in appropriate places
- clean up the interface for `minorToMajorUnit` interface to not expose `NewDisplayAmount` internals

### DIFF
--- a/src/domain/fiat/display-currency.ts
+++ b/src/domain/fiat/display-currency.ts
@@ -1,3 +1,5 @@
+import { safeBigInt } from "@domain/shared"
+
 export const CENTS_PER_USD = 100
 
 export const MajorExponent = {
@@ -66,5 +68,26 @@ export const getCurrencyMajorExponent = (
   } catch {
     // this is necessary for non-standard currencies
     return MajorExponent.STANDARD
+  }
+}
+
+export const newDisplayAmountFromNumber = <T extends DisplayCurrency>({
+  amount,
+  currency,
+}: {
+  amount: number
+  currency: T
+}): NewDisplayAmount<T> | ValidationError => {
+  const amountInMinor = safeBigInt(amount)
+  if (amountInMinor instanceof Error) return amountInMinor
+
+  const displayMajorExponent = getCurrencyMajorExponent(currency)
+
+  return {
+    amountInMinor,
+    currency,
+    displayInMajor: (Number(amountInMinor) / 10 ** displayMajorExponent).toFixed(
+      displayMajorExponent,
+    ),
   }
 }
--- a/src/domain/fiat/display-currency.ts
+++ b/src/domain/fiat/display-currency.ts
@@ -20,17 +20,17 @@ export const minorToMajorUnitFormatted = ({
   return majorAmount.toFixed(displayMajorExponent)
 }
 
-export const minorToMajorUnit = ({
-  amount,
-  displayCurrency,
+export const minorToMajorUnit = <T extends DisplayCurrency>({
+  displayAmount,
   fixed = true,
 }: {
-  amount: number | bigint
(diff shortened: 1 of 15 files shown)


src/domain/fiat/display-currency.ts
```diff
@@ -68,3 +68,40 @@ export const getCurrencyMajorExponent = (
     return MajorExponent.STANDARD
   }
 }
+
+export const newDisplayAmountFromNumber = <T extends DisplayCurrency>({
+  amount,
+  currency,
+}: {
+  amount: number
```
bodymindarts, line 76: Why would it every be a number?

src/domain/shared/index.types.d.ts
```diff
@@ -29,6 +29,12 @@ type NewDisplayAmount<T extends DisplayCurrency> = {
   displayInMajor: DisplayCurrencyMajorAmount
 }
 
+type PriceAmount<T extends DisplayCurrency> = {
+  priceOfOneSatInMinorUnit: number
```
bodymindarts, line 33: Why are we using `number` for money?

dolcalmi (changes requested):

src/domain/shared/index.types.d.ts
vindard: It's the one place where `number` is still left I believe. This PR does away with the `DisplayAmount` type (which contained an internal `number`) in favour of the  `NewDisplayAmount` type (which contains internal `bigint`).

We have one property still on `WalletTransaction` type, `displayCurrencyPerSettlementCurrencyUnit`, which is a decimal number that eventually gets converted to an integer `base` and `offset` in the API. I think we can remove this `number` now and convert to `BigInt` in the domain for the `WalletTransaction` type instead of in the API layer. I'd prefer to do that in a follow-up PR though since it's a slightly different concern from what this PR is addressing which is an in-place refactor to swap all instances of `DisplayAmount` for `NewDisplayAmount` which implicitly already removes a lot of `number` values and replaces them with `bigint` ones.

src/domain/fiat/display-currency.ts
```diff
@@ -68,3 +68,40 @@ export const getCurrencyMajorExponent = (
...
+  return {
+    amountInMinor,
+    currency,
+    displayInMajor: (Number(amountInMinor) / 10 ** displayMajorExponent).toFixed(
+      displayMajorExponent,
+    ),
+  }
+}
+
+export const priceAmountFromNumber = <T extends DisplayCurrency>({
```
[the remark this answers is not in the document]
vindard: Implementing because we've swapped in `NewDisplayAmount` now and we use bigints everywhere for display amounts, but this price value still needs to be a `number` for now. That's why I made the new `PriceAmount` type and this complementary method.

src/domain/fiat/display-currency.ts
vindard: It's from a part of the code that we haven't refactored away from `number` as yet. All `...Metadata` methods still take display amounts as numbers (branded `DisplayCurrencyBaseAmount` type) so this gets passed around quite a bit in different use-cases.

This could be addressed if we decide to include removing the `DisplayCurrencyBaseAmount` type from the code as part of our post-cleanup, which currently has 150 instances of being used in the codebase.

dolcalmi (approved):

src/domain/shared/index.types.d.ts
vindard: > I'd prefer to do that in a follow-up PR though

Done here: https://github.com/GaloyMoney/galoy/pull/2473

bodymindarts (approved):

PR #2472, opened 2023-03-23
refactor: swap legacy display amount type for new one, fix notifications

vindard:
## Description

_Merge after #2466_

This PR Removes all usages of `DisplayAmount` and swaps in `NewDisplayAmount` instead. In the process of doing this it also fixes notifications to accommodate display currencies.

`NewDisplayAmount` will be renamed in a follow-up PR so as to not confuse the diff here


bodymindarts (approved):
Good cleanup

PR #2488, opened 2023-03-30
feat!: include route channel ids in skipFeeProbe checks

vindard:
## Description

This is to address the issue where fee probes fail to Breez when an on-the-fly channel needs to be created for the end-user. Something we noticed is that these invoices have route hints where the non-existent channel's id is `1x0x0`. This PR is to add this heuristic to our code when deciding which fee probes to skip.

### Note

Breez's LSP has a solution for handling probes [here](https://github.com/breez/lspd#probing-support) where they ask that the payment hash included in the probe is formatted in a certain way. This would also help in the case where for e.g. the mobile app is offline, but this could potentially add significant complexity to our fee-probing logic and we still wouldn't be able to make the actual payment if the app is offline.

### DIFF
--- a/galoy.yaml
+++ b/galoy.yaml
@@ -198,11 +198,8 @@ oathkeeperConfig:
   decisionsApi: http://127.0.0.1:4456/decisions/
 captcha:
   mandatory: false
-# deprecated: use skipFeeProbeConfig instead
 skipFeeProbe:
   - 038f8f113c580048d847d6949371726653e02b928196bad310e3eda39ff61723f6
 skipFeeProbeConfig:
-  pubkey:
-    - 038f8f113c580048d847d6949371726653e02b928196bad310e3eda39ff61723f6
-  chanId:
-    - 1x0x0
+  pubkey: []
+  chanId: []
--- a/galoy.yaml
+++ b/galoy.yaml
@@ -198,8 +198,6 @@ oathkeeperConfig:
   decisionsApi: http://127.0.0.1:4456/decisions/
 captcha:
   mandatory: false
-skipFeeProbe:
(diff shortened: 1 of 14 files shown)


bodymindarts (dismissed):

src/app/payments/helpers.ts
```diff
@@ -44,9 +45,14 @@ export const constructPaymentFlowBuilder = async <
 }): Promise<LPFBWithConversion<S, R> | ApplicationError> => {
   const lndService = LndService()
   if (lndService instanceof Error) return lndService
+  const skipProbePubkeysLegacy = getPubkeysToSkipProbe()
```
nicolasburtey, line 48: why `Legacy`?

galoy.yaml
```diff
@@ -199,3 +199,6 @@ captcha:
   mandatory: false
 skipFeeProbe:
   - 038f8f113c580048d847d6949371726653e02b928196bad310e3eda39ff61723f6
```
nicolasburtey, line 201: why keep this and maintain some `Legacy` code? we don't need backward compatibility, we can just update the associated yaml? just make sure to have the ! in the commit message to signal a breaking change.

nicolasburtey (commented):
couple of comments on the PR

nicolasburtey:
LGTM. I guess this needs a chart update? cc @openoms @krtk6160

vindard:
> LGTM. I guess this needs a chart update? cc @openoms @krtk6160

Already opened here: https://github.com/GaloyMoney/galoy-deployments/pull/3343

openoms:
There are a lot of config values not added to charts. As the default is empty it shouldn't be a problem, just need to be added in deployments as in https://github.com/GaloyMoney/galoy-deployments/pull/3343

nicolasburtey (approved):

PR #2491, opened 2023-04-04
feat: add mutation for onchain usd send denominated in sats

vindard:
## Description

This PR modifies the `OnchainPaymentFlowBuilder` and adds a new mutation to support sending onchain from USD wallet using an amount denominated in sats .


test/unit/payments/onchain-payment-flow-builder.spec.ts
```diff
@@ -287,21 +304,24 @@ describe("OnChainPaymentFlowBuilder", () => {
 
                 const paymentLowest = await withBtcWalletBuilder
                   .withoutRecipientWallet()
-                  .withAmount(51) // Close to 1 cent
+                  .withAmount({ amount: BigInt(51), currency: amountCurrency }) // Close to 1 cent
```
bodymindarts, line 307: I think you can just use `51n` or some other postfix like that to signify `BigInt`

bodymindarts (approved):

src/domain/payments/onchain-payment-flow-builder.ts
```diff
@@ -187,8 +181,14 @@ const OPFBWithAmount = <S extends WalletCurrency, R extends WalletCurrency>(
         const updatedStateFromBtcProposedAmount = async (
           btcProposedAmount: BtcPaymentAmount,
         ): Promise<OPFBWithConversionState<S, R> | DealerPriceServiceError> => {
-          const convertedAmount = await mid.usdFromBtc(btcProposedAmount)
-          if (convertedAmount instanceof Error) return convertedAmount
+          const convertedAmountRaw = await mid.usdFromBtc(btcProposedAmount)
+          if (convertedAmountRaw instanceof Error) return convertedAmountRaw
+
+          // Round up USD if sats is non-zero in intraledger (mid-price) case and cents is zero
```
vindard, line 187: Note on this, it's only to not fail in the edge-case that one user sends 1 sat from their own **usd** wallet to another user's **usd** wallet

image

PR #2521, opened 2023-04-12
chore: exposing levels to the api

nicolasburtey:
exposing Level would be useful for multiple reasons for front end related logic, notably for not letting user go to earn if they are level 0, ie: a device account user.


src/graphql/types/scalar/level.ts
```diff
@@ -0,0 +1,25 @@
+import { GT } from "@graphql/index"
+
+const AccountLevel = GT.Enum({
+  name: "AccountLevel",
+  values: {
+    ZERO: {},
+    ONE: {},
+    TWO: {},
```
dolcalmi, line 8: shouldnt we use AccountLevel enum here ? (check src/domain/accounts/primitives.ts)

src/graphql/types/scalar/level.ts
dolcalmi, line 1: we already have account-level

dolcalmi (dismissed):

src/graphql/types/scalar/level.ts
nicolasburtey: oh. thanks. it's on admin I didn't see it.

dolcalmi (approved):

PR #2534, opened 2023-04-16
feat: add expired to invoice status

dolcalmi:
Add EXPIRED to InvoicePaymentStatus, this is required in galoy pay to give feedback to the user specially for USD invoices (when default wallet is stablesats)

### DIFF
--- a/test/e2e/servers/graphql-main-server/galoy-pay-testing.spec.ts
+++ b/test/e2e/servers/graphql-main-server/galoy-pay-testing.spec.ts
@@ -3,6 +3,8 @@ import crypto from "crypto"
 import { ApolloClient, NormalizedCacheObject } from "@apollo/client/core"
 import { toSats } from "@domain/bitcoin"
 
+import { sleep } from "@utils"
+
 import LN_INVOICE_CREATE_ON_BEHALF_OF from "./mutations/ln-invoice-create-on-behalf-of-recipient.gql"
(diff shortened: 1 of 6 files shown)


test/e2e/servers/graphql-main-server/galoy-pay-testing.spec.ts
```diff
@@ -231,6 +224,12 @@ describe("galoy-pay", () => {
       })
       expect(makePayment.data.lnInvoicePaymentSend.status).toEqual("SUCCESS")
 
+      // Subscribe to the invoice
```
vindard, line 227: Why did this have to be moved to after the invoice is paid?

vindard (commented):
Looks ok. Just one question from the tests change and whether this signals a change in behaviour of the subscription?

test/e2e/servers/graphql-main-server/galoy-pay-testing.spec.ts
dolcalmi: because of a race condition.. I had added sleep but in this way is cleaner. I had random errors with github actions

dolcalmi: adding status query before subscription paid validation gives enough time to fix the race condition with `notificationsService.lightningTxReceived` (no await)
https://github.com/GaloyMoney/galoy/pull/2534/commits/c76a2fbc0ffc579a303efcd86bc9d0213323c5d0

vindard (approved):

PR #2540, opened 2023-04-17
tests: use typed query for e2e test (same as with mobile)

nicolasburtey:
- make it easier to code on e2e on the backend and mobile (same file structures/pattern)
- type will help catch bug earlier
- note: because the there are some conflicts between the type in the public api and the admin api if I try to combine them into the same generated.ts, the admin api is still using the "old" untyped way currently. there are only 2 queries using the admin api so it can be refactored later.


Makefile
```diff
@@ -171,3 +171,8 @@ lncli-outside-2:
 	container_id=$$(docker ps -q -f status=running -f name="galoy-lnd-outside-2"); \
 	docker exec -it "$$container_id" /bin/sh -c 'lncli -n regtest ${command}'
 
+kill-graphql:
+	kill $(lsof -t -i:4001) & kill $(lsof -t -i:4012) & kill $(lsof -t -i:4000)
+
+redis-cli:
+	docker-compose exec redis redis-cli
```
bodymindarts, line 178: missing new line

bodymindarts (dismissed):

PR #2558, opened 2023-04-21
fix: update tracing service and trigger record errors properly

dolcalmi:
Fix crash in trigger because of invalid error reported by subscription + remove `// @ts-nocheck` from tracing service

```
/app/lib/services/tracing.js:207
        span.setAttribute("error.name", exception["name"]);
                                                 ^

TypeError: Cannot read properties of undefined (reading 'name')
    at recordException (/app/lib/services/tracing.js:207:50)
    at recordExceptionInCurrentSpan (/app/lib/services/tracing.js:193:5)
    at EventEmitter.<anonymous> (/app/lib/servers/trigger.js:354:52)
    at EventEmitter.emit (node:events:513:28)
    at finished (/app/node_modules/lightning/lnd_methods/invoices/subscribe_to_invoices.js:106:22)
    at ClientReadableStreamImpl.<anonymous> (/app/node_modules/lightning/lnd_methods/invoices/subscribe_to_invoices.js:158:34)
    at /app/node_modules/@opentelemetry/context-async-hooks/build/src/AbstractAsyncHooksContextManager.js:50:55
    at AsyncLocalStorage.run (node:async_hooks:330:14)
    at AsyncLocalStorageContextManager.with (/app/node_modules/@opentelemetry/context-async-hooks/build/src/AsyncLocalStorageContextManager.js:33:40)
    at ClientReadableStreamImpl.contextWrapper (/app/node_modules/@opentelemetry/context-async-hooks/build/src/AbstractAsyncHooksContextManager.js:50:32)
```

### DIFF
--- a/src/app/accounts/active-accounts.ts
+++ b/src/app/accounts/active-accounts.ts
@@ -34,7 +34,7 @@ export const getRecentlyActiveAccounts = async function* ():
     const wallets = await WalletsRepository().listByAccountId(account.id)
     if (wallets instanceof Error) {
       recordExceptionInCurrentSpan({
-        error: "impossible to listByAccountId",
+        error: wallets,
         level: ErrorLevel.Critical,
         attributes: { account: account.id },
       })
@@ -44,7 +44,7 @@ export const getRecentlyActiveAccounts = async function* ():
     const isActive = await activityChecker.aboveThreshold(wallets)
     if (isActive instanceof Error) {
       recordExceptionInCurrentSpan({
-        error: "impossible to get aboveThreshold value",
+        error: isActive,
         level: ErrorLevel.Critical,
         attributes: { account: account.id, wallets: JSON.stringify(wallets) },
       })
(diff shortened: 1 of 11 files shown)


src/app/accounts/active-accounts.ts
```diff
@@ -34,7 +34,7 @@ export const getRecentlyActiveAccounts = async function* ():
     const wallets = await WalletsRepository().listByAccountId(account.id)
     if (wallets instanceof Error) {
       recordExceptionInCurrentSpan({
-        error: "impossible to listByAccountId",
+        error: wallets,
```
UncleSamtoshi, line 37: What was the motivation for switching out some of the strings with the underlying errors?

dolcalmi: I think was a mistake.. otel include string as a possible value for Exception but that is not our intention. We need a Domain error to do tracing properly

UncleSamtoshi (approved):

PR #2563, opened 2023-04-22
chore: upgrade to kratos v13

nicolasburtey:


src/servers/middlewares/kratos-router.ts
```diff
@@ -19,6 +20,61 @@ const { callbackApiKey } = getKratosPasswords()
...
+          console.log("invalid phone")
+          res.status(400).send("invalid phone")
+          return
+        }
+
+        const user = await UsersRepository().findByPhone(phone)
+
+        // we expect the phone number to not exist
+        if (user instanceof Error) {
+          console.log("phone doesn't already exist")
```
bodymindarts, line 60: Can we send an OTEL event instead of log?

src/servers/middlewares/kratos-router.ts
```diff
@@ -19,6 +20,61 @@ const { callbackApiKey } = getKratosPasswords()
...
+          res.sendStatus(200)
+          return
+        }
+
+        console.error("phone already exist")
```
bodymindarts, line 65: Otel event vs log?

src/servers/middlewares/kratos-router.ts
```diff
@@ -19,6 +20,61 @@ const { callbackApiKey } = getKratosPasswords()
...
+        res.status(500).send(`phone already exist`)
+        return
+      }
+
+      // TODO
```
bodymindarts, line 70: Is this really needed? I don't think we will loose site of the fact that we may want to implement email+pw

src/servers/middlewares/kratos-router.ts
```diff
@@ -51,16 +107,18 @@ kratosRouter.post(
 
       const userIdChecked = checkedToUserId(userId)
       if (userIdChecked instanceof Error) {
+        // TODO: log this error as critical to honeycomb
```
bodymindarts, line 110: Then we should probably do it pre-merge or?

src/servers/middlewares/kratos-router.ts
```diff
@@ -83,6 +141,7 @@ kratosRouter.post(
       }
 
       if (account instanceof Error) {
+        // TODO: log this error as critical to honeycomb
```
bodymindarts, line 144: Then we should do it pre-merge

bodymindarts (approved):
I'm approving but think we should remove / handle comments.

PR #2575, opened 2023-04-26
chore: deprecate legacy header

nicolasburtey:
can one of you take this ticket and make sure this header in our chart/deployment is removed? we no longer use/need it

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -297,7 +297,7 @@ services:
     volumes:
       - ./:/repo
   oathkeeper:
-    image: oryd/oathkeeper:v0.40.1
+    image: oryd/oathkeeper:v0.40.2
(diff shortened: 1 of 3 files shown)


openoms (approved):

docker-compose.yml
```diff
@@ -297,7 +297,7 @@ services:
     volumes:
       - ./:/repo
   oathkeeper:
-    image: oryd/oathkeeper:v0.40.1
+    image: oryd/oathkeeper:v0.40.2
```
openoms, line 300: confirmed that this brings the oathkeeper version to the same as used in charts: https://github.com/ory/k8s/tree/master/helm/charts/oathkeeper

openoms:
the changes here are in prod now

PR #2580, opened 2023-05-01
refactor: remove update pending payments from get balance

dolcalmi:
Remove `updatePendingPaymentsByWalletId` from getBalance

----
Pending:
- [x] Lighting library fix https://github.com/alexbosworth/lightning/pull/137
- [x] Tests - remove manual updates for payments with settle and cancel hold invoices

### DIFF
--- a/src/servers/trigger.ts
+++ b/src/servers/trigger.ts
@@ -9,6 +9,8 @@ import {
   SubscribeToInvoiceInvoiceUpdatedEvent,
   subscribeToInvoices,
   SubscribeToInvoicesInvoiceUpdatedEvent,
+  subscribeToPayments,
+  SubscribeToPaymentsPaymentEvent,
   subscribeToTransactions,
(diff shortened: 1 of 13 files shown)


src/servers/trigger.ts
```diff
@@ -451,11 +454,46 @@ export const setupInvoiceSubscribe = ({
...
+    recordExceptionInCurrentSpan({
+      error: new SubscriptionInterruptedError((err && err.message) || "subPayments"),
+      level: ErrorLevel.Warn,
+      attributes: { ["error.subscription"]: "subPayments" },
+    })
+    subPayments.removeAllListeners()
+  })
+
+  // Update existing pending payments
+  Payments.updatePendingPayments(logger)
```
vindard, line 487: Is this just for in case some previous updates failed?

vindard (approved):
Looks good to me. Iiuc we're adding a new pending payments subscription so that we don't have to do the update when a user calls getBalance.

PR #2581, opened 2023-05-01
chore: randomly assign default wallet currencies

UncleSamtoshi:
- if userId is an even number assign BTC default wallet, otherwise assign USD
- this will be used as part of an A/B test to analyze retention and spending for different default currencies

### DIFF
--- a/src/app/accounts/create-account.ts
+++ b/src/app/accounts/create-account.ts
@@ -9,6 +9,13 @@ import {
 } from "@services/mongoose"
 import { TwilioClient } from "@services/twilio"
 
+function isUserIdEven(userId: UserId): boolean {
+  const lastChar = userId[userId.length - 1].toLowerCase()
+  const evenChars = new Set(["0", "2", "4", "6", "8", "a", "c", "e"])
+
+  return evenChars.has(lastChar)
+}
+
 const initializeCreatedAccount = async ({
   account,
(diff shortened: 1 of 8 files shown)


src/app/accounts/create-account.ts
```diff
@@ -35,10 +42,17 @@ const initializeCreatedAccount = async ({
     enabledWallets[currency] = wallet
   }
 
-  // Set default wallet explicitly as BTC, or implicitly as 1st element in
+  // If enabled, randomly assign a default wallet based on the user id in order to A/B test the default wallet
+  const desiredDefaultWalletCurrency =
+    isUserIdEven(account.kratosUserId) || !config.randomizeDefaultWallet
```
nicolasburtey, line 47: Why reliance on userId? A Math.random() > .5 function will work just as well and be more concise

UncleSamtoshi: It allows for holistics to easily know the default wallet an account was created with. @jireva thought this would be the simplest way for constructing the dashboard, but potentially there are other ways.

jireva (approved):

src/app/accounts/create-account.ts
UncleSamtoshi: I believe that is the same as the current logic.

PR #2596, opened 2023-05-08
feat: introduce bria to backend

vindard:
## Description

This PR adds bria as a dev dependency, connects it via grpc and implements a method in the IOnChainService interface against the bria instance.


src/domain/bitcoin/bria/index.types.d.ts
```diff
@@ -0,0 +1,39 @@
+type ClientReadableStream<T> = import("@grpc/grpc-js").ClientReadableStream<T>
+
+type UtxoDetectedEvent = {
```
vindard, line 3: I would've leaned more to having these as service types instead of being here in the domain, but I needed `UtxoEvent` to pass in as a type argument to `ClientReadableStream` in the service type definition lower down

test/integration/02-user-wallet/02-receive-onchain.spec.ts
```diff
@@ -507,6 +508,43 @@ describe("UserWallet - On chain", () => {
   })
 })
 
+describe("BriaService", () => {
```
dolcalmi, line 511: should we create a new file for this?

src/services/bria-api/grpc.ts
dolcalmi, line 1: will we replace this with the automatic generated code? similar to stablesats?

src/domain/bitcoin/onchain/index.types.d.ts
```diff
@@ -111,3 +111,17 @@ interface IOnChainService {
     targetConfirmations,
   }: PayToAddressArgs): Promise<OnChainTxHash | OnChainServiceError>
 }
+
+interface OnChainEvent {
+  payload: string
+  sequence: string
+  recorded_at: number
+}
```
dolcalmi, line 119: should this be part of domain?

src/domain/bitcoin/onchain/index.types.d.ts
```diff
@@ -111,3 +111,17 @@ interface IOnChainService {
...
+
+type OnChainEventHandler = (event: OnChainEvent) => true | ApplicationError
```
dolcalmi, line 121: ApplicationError is not part of domain layer

src/config/process.ts
```diff
@@ -11,6 +11,16 @@ type TwilioConfig = {
 export const GALOY_API_PORT = process.env.GALOY_API_PORT || 4012
 export const GALOY_ADMIN_PORT = process.env.GALOY_ADMIN_PORT || 4001
 
+const BRIA_PROFILE_API_KEY =
+  process.env.BRIA_PROFILE_API_KEY || "bria_dev_000000000000000000000"
+export const BRIA_HEADERS = {
+  "x-bria-api-key": BRIA_PROFILE_API_KEY,
+}
```
dolcalmi, line 18: this must be handled internally in the service

src/services/bria-api/index.ts
dolcalmi, line 1: shouldnt be only `bria`? (folder name)

dev/lnd/lnd1.tpub
dolcalmi, line 1: was not possible to get this from lnd?

dolcalmi (changes requested):

test/integration/02-user-wallet/02-receive-onchain.spec.ts
bodymindarts: yes please

src/services/bria-api/grpc.ts
bodymindarts: yes @krtk6160 is working on the automation for that - every release will open a PR to add the auto-generated stuff.

src/domain/bitcoin/onchain/index.types.d.ts
bodymindarts: We talked about this and the events themselves are part of the translation / adaptor layer. Its the handler that calls the use cases which is where we enter the app layer (and thus domain).

src/domain/bitcoin/onchain/index.types.d.ts
bodymindarts: Think of the event handler as a similar to the graphql layer. It calls into the app.

src/services/bria-api/index.ts
bodymindarts: agree

dev/lnd/lnd1.tpub
bodymindarts: Makes bootstrapping more complicated. This does come from lnd but as its deterministic its easier to bring up bria with this on disk - at least I believe that's what going on cc @vindard

vindard: Yes, we could reach into lnd and get this but it adds a lot more steps

test/integration/02-user-wallet/02-receive-onchain.spec.ts
vindard: Done https://github.com/GaloyMoney/galoy/pull/2596/commits/6d94f8babba07fd50dad32dd3b5086e9e8f45ea5

src/domain/bitcoin/onchain/index.types.d.ts
vindard: Yup, lots of refactoring around this in subsequent branches

docker-compose.yml
```diff
@@ -303,13 +303,16 @@ services:
     image: us.gcr.io/galoy-org/bria:edge
     environment:
       - PG_CON=postgres://user:password@postgres-bria:5432/pg
-    command: bash -c "
-      bria daemon \
-      --config bria.yml \
-      dev \
-      -x tpubDD4vFnWuTMEcZiaaZPgvzeGyMzWe6qHW8gALk5Md9kutDvtdDjYFwzauEFFRHgov8pAwup5jX88j5YFyiACsPf3pqn5hBjvuTLRAseaJ6b4
-      -d m/84h/0h/0h"
-    working_dir: /repo/dev
```
bodymindarts: @vindard this was the issue - the release image is hardened to a specific user that only has access to certain paths in the image.

bodymindarts (approved):

PR #2598, opened 2023-05-09
ci: only include tagged commits for proto sync

krtk6160:


bodymindarts:
I didn't see we already have the commit filter - that seems like it should work too and we don't need the tagged filter.

krtk6160:
> I didn't see we already have the commit filter - that seems like it should work too and we don't need the tagged filter. 

It was added in to be tested and didn't work, but tag regex works as expected.

bodymindarts:
What about bria and price-history?

krtk6160:
> What about bria and price-history?

Added it for bria but not price history because price repo does not have any [tagging system](https://github.com/GaloyMoney/price/tags).

bodymindarts (dismissed):

PR #2599, opened 2023-05-10
feat: add pending incoming utxo repo

dolcalmi:


src/domain/bitcoin/onchain/index.types.d.ts
```diff
@@ -111,3 +112,27 @@ interface IOnChainService {
     targetConfirmations,
   }: PayToAddressArgs): Promise<OnChainTxHash | OnChainServiceError>
 }
+
+type PendingIncomingOnChainTransaction = {
+  id: PendingIncomingOnChainTransactionId
```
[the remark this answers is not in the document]
dolcalmi: why? current behavior is search by address this is similar to lnPayments should not be related to domain schemas

src/domain/bitcoin/onchain/index.types.d.ts
```diff
@@ -111,3 +112,27 @@ interface IOnChainService {
...
+  createdAt: Date
+}
+
+type PersistNewPendingIncomingOnChainTransactionArgs = Omit<
+  PendingIncomingOnChainTransaction,
+  "id" | "createdAt"
+>
+
+interface IPendingIncomingUTXORepository {
+  listByAddresses(
```
bodymindarts, line 131: I think `listByAccountId` would instead of `listByAddresses` would be better here.

dolcalmi: This should not be related to account. Is just a "cache" of the pending txs

bodymindarts: Why can't the cache be account aware? It makes for more efficient recovery and is safer. What is the domain centric reason why it should be like this? Just the fact that its being used as a cache isn't reason enough (imo).

dolcalmi: We talked about this in the first sync meeting... it should be similar to lnPayments and that makes sense because is only related to bitcoin domain, also if we add account id then this should not be in bitcoin domain and finally if that is the case should be wallet id and not account id (also the performance is the same specially in a collection with few records)

bodymindarts: Okay that makes sense.
You are right if we did that it should not be account id but wallet id and it should not be in bitcoin domain.
If you think performance is the same then I don't care either way. I just thought if its by wallet id we don't need to lookup all the addresses when we go to list transactions so it would save us 1 round trip. But perhaps I'm wrong.

PR #2605, opened 2023-05-10
test: remove lnd2 from environment

vindard:
## Description

This is to explore removing lnd2 from our test environment. We will lose 2 tests:
- internal lnd1 to lnd2 connection test
- internal rebalancing test


dolcalmi:
why do we want to do this?

nicolasburtey:
> why do we want to do this?

shouldn't we try to remove one of the external node instead? there are 3. I think a particular test required the third but maybe it could be refactored also?

bodymindarts (dismissed):

PR #2610, opened 2023-05-10
chore: bria handler boilerplate

bodymindarts:

PR #2622, opened 2023-05-12
feat: add WalletOnChainPendingReceiveRepository

dolcalmi:
split if necessary for https://github.com/GaloyMoney/galoy/pull/2610


src/services/mongoose/schema.types.d.ts
```diff
@@ -163,3 +163,18 @@ type PaymentFlowStateRecordIndex = XOR<
   senderWalletId: string
   inputAmount: number
 }
+
+type WalletOnChainPendingReceiveRecord = {
+  walletId: string
+  address: string
+  transactionHash: string
+  amount: number
```
vindard, line 171: If these are wallet amount & fee I might rename to `walletAmount`, `walletFee` and `walletCurrency` just to be clear that they're not sats denominated.

Also, should we persist satsAmount and satsFee explicitly too?

src/services/mongoose/schema.types.d.ts
```diff
@@ -163,3 +163,18 @@ type PaymentFlowStateRecordIndex = XOR<
   senderWalletId: string
   inputAmount: number
 }
+
+type WalletOnChainPendingReceiveRecord = {
+  walletId: string
+  address: string
+  transactionHash: string
```
vindard, line 170: We should probably add `vout` as well

src/services/mongoose/wallet-onchain-pending-receive.ts
```diff
@@ -0,0 +1,111 @@
...
+import { TxStatus } from "@domain/wallets"
+import { WalletCurrency } from "@domain/shared"
+import { CouldNotFindWalletOnChainPendingReceiveError } from "@domain/errors"
+
+import { parseRepositoryError } from "./utils"
+import { WalletOnChainPendingReceive } from "./schema"
+
+export const WalletOnChainPendingReceiveRepository =
+  (): IWalletOnChainPendingReceiveRepository => {
+    const listByWalletId = async ({
```
vindard, line 12: Should we change this to `listByWalletIds` to take a `walletId` array instead? It'll match out ledger service `getTransactionsByWalletIds` method and make it easier to fetch records for multiple wallets like we do in the `getTransactionsForWallets` use-case

vindard (changes requested):
Left a few comments about potential changes to the schema and a repo method

src/services/mongoose/schema.types.d.ts
dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/2622/commits/903ca44ded470fca59c481cbf8fcebccb121948a

src/services/mongoose/schema.types.d.ts
dolcalmi: Probably we will need to fix the vout issue first in current implementation but... I added it https://github.com/GaloyMoney/galoy/pull/2622/commits/903ca44ded470fca59c481cbf8fcebccb121948a

src/services/mongoose/wallet-onchain-pending-receive.ts
dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/2622/commits/903ca44ded470fca59c481cbf8fcebccb121948a

src/services/mongoose/wallet-onchain-pending-receive.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+}
+
+const translateToDbRecord = (
+  tx: PersistWalletOnChainPendingReceiveArgs,
+): WalletOnChainPendingReceiveRecord => ({
+  walletId: tx.walletId || "",
+  address: tx.initiationVia.address,
+  transactionHash: tx.settlementVia.transactionHash,
+  vout: tx.settlementVia.vout || 0,
```
vindard, line 108: I'm not sure that we should default to `0` here. 

If a transaction happens to get to here without a vout (should never happen, since we only take Bria events now), we should persist as undefined instead since the default vout isn't actually`0` in reality.

I can probably change this after we merge though.

vindard (approved):
Looks good to me now

bodymindarts (approved):

vindard (approved):

PR #2624, opened 2023-05-15
feat: add vout to onchain receipt

dolcalmi:
required for bria integration. check https://github.com/GaloyMoney/galoy/pull/2622#discussion_r1192947785 and https://github.com/GaloyMoney/galoy/pull/2610


vindard (approved):
Looks really good

PR #2626, opened 2023-05-15
feat: add bria trigger event handling code

vindard:


vindard:
There's some little things we can clean up here still, but I'd like to merge this branch and continue iterating on follow-up branches we've already started refactoring things in (see https://github.com/GaloyMoney/galoy/compare/main...add-bria-onchain-service)

PR #2651, opened 2023-05-19
chore: implement plugin to add partial account transactions to response

vindard:
## Description

This is to help prevent instances where we get an error and then don't show any transactions at all back to the user.

image


src/servers/graphql-server.ts
```diff
@@ -215,6 +216,56 @@ export const startApolloServer = async ({
...
+        ...edge,
+        node: {
+          ...edge.node,
+          settlementDisplayPrice: {
+            ...edge.node.settlementDisplayPrice,
+            base: Number(edge.node.settlementDisplayPrice.base),
+            offset: Number(edge.node.settlementDisplayPrice.offset),
+          },
+        },
+      })
```
dolcalmi, line 231: this hide query behavior, if we can move this to a "common" place for txs would be better

dolcalmi (changes requested):

src/servers/graphql-server.ts
dolcalmi: also.. is really an user facing error if we cant get the pending onchain txs?

vindard: > this hide query behavior, if we can move this to a "common" place for txs would be better

Like this you mean? https://github.com/GaloyMoney/galoy/pull/2651/commits/16a739127cd8bcec3e45c7ab774f2078515e8967

vindard: > also.. is really an user facing error if we cant get the pending onchain txs?

If it's a partial error then ideally it doesn't reach the user, but maybe it's useful for the frontend to know via the API still?

PR #2653, opened 2023-05-20
fix: change rpc bitcoind library

dolcalmi:
fix an issue with request package in nodejs 20 (is used in almost all bitcoin core libraries) https://www.npmjs.com/package/request it is deprecated so there is no way to do a PR to fix it.

`rpc-bitcoin` is not actively maintained but only has 1 dependency (rpc-request)


package.json
```diff
@@ -154,6 +154,7 @@
     "@typescript-eslint/parser": "^5.59.5",
     "@xascode/tscpaths": "0.1.4",
     "axios-mock-adapter": "^1.21.2",
+    "bitcoin-core-ts": "^3.0.3",
```
nicolasburtey, line 157: is it still used?

dolcalmi: yes, in tests... this PR has been replaced by https://github.com/GaloyMoney/galoy/pull/2655 but I will create another to clean tests

PR #2655, opened 2023-05-20
test: update coldstorage test to use only app layer

dolcalmi:
Update coldstorage tests to use only app layer methods (reduces dependency on bitcoin-core lib)


vindard (approved):
Looks good to me. I tested curl the metrics endpoint as well and am not getting that error anymore locally

PR #2662, opened 2023-05-23
feat: switch new-address to bria and add new receive listener

vindard:
## Description

This PR adds the bria listener to the existing trigger and it switches our "create new address" implementation to using bria and our bitcoind signer wallet.

Next steps:
- enable stablesats for onchain receive
- pass walletDescriptor instead of walletId to bria createAdress metadata
- clean up additional resetLimits calls added near end of PR to tests


src/servers/trigger.ts
dolcalmi, line 1: I thought we didnt want to do this. if that is the case probably we want to move "bria-server" to app layer

dolcalmi (changes requested):

src/servers/trigger.ts
vindard: We've already implemented the relevant methods (`addPendingTransaction`, `addSettledTransaction`) in app layer. This is just orchestration code for trigger similar to how we have things like `onChainBlockEventHandler` at this server layer.

I kept it separate in case we wanted to make bria a separate trigger later on. Do you think it's better if everything was in the same `trigger.ts` file instead?

src/servers/bria-trigger.ts
dolcalmi, line 1: if we will not have a different pod we need to create a structure for these handlers (similar to middlewares)

test/integration/02-user-wallet/02-receive-onchain.spec.ts
```diff
@@ -95,745 +119,1540 @@ afterAll(async () => {
...
-describe("FunderWallet - On chain", () => {
-  it("receives on-chain transaction", async () => {
-    const funderWalletId = await getFunderWalletId()
-    await sendToWalletTestWrapper({
-      walletId: funderWalletId,
-      amountSats: getRandomAmountOfSats(),
-    })
-  })
-})
+describe("With Bria", () => {
```
dolcalmi, line 122: didnt we say that this should have a new test file?

test/integration/02-user-wallet/02-receive-onchain.spec.ts
```diff
@@ -95,745 +119,1540 @@ afterAll(async () => {
...
+    if (error instanceof Error || initTransactions === null) {
+      throw error
+    }
 
-describe("UserWallet - On chain", () => {
-  it("get last on chain address", async () => {
-    const address = await Wallets.createOnChainAddressForBtcWallet(walletIdA)
-    const lastAddress = await Wallets.getLastOnChainAddress(walletIdA)
+    // Get address and send txn
+    // ===
```
dolcalmi, line 141: does this mean something? please remove

test/integration/02-user-wallet/02-receive-onchain.spec.ts
```diff
@@ -95,745 +119,1540 @@ afterAll(async () => {
...
+    // Receive transaction with bria trigger methods
+    // ===
+    const detectedEvent = await onceBriaSubscribe({
+      type: BriaPayloadType.UtxoDetected,
+      txId,
+    })
+    if (detectedEvent?.payload.type !== BriaPayloadType.UtxoDetected) {
+      throw new Error(`Expected ${BriaPayloadType.UtxoDetected} event`)
+    }
+    const resultPending = await Wallets.addPendingTransaction(detectedEvent.payload)
```
dolcalmi, line 163: please use the handlers, this should not be part of the test

test/integration/02-user-wallet/02-bria.spec.ts
```diff
@@ -62,7 +64,8 @@ describe("BriaSubscriber", () => {
             receivedEvents.push(event)
             if (receivedEvents.length === nExpectedEvents) {
               setTimeout(() => {
-                resolver(receivedEvents)
+                resolve(receivedEvents)
+                clearTimeout(timeoutId)
```
dolcalmi, line 68: uhmm please use timeout promise, similar to lnd https://github.com/GaloyMoney/galoy/pull/2664/files#diff-f82eb815ee086a501049a55fd44d719dd309efdae7d7c812731e7bc5048ab766R1135 or waitFor https://chat.galoy.io/galoy/pl/ys1eq6uzwi8afeoe76j67e9dih

test/e2e/servers/trigger/trigger.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+        resolve(result)
+        break
+      }
+
+      await sleep(1000)
+    }
+  })
+}
+
+const waitForNoAddress = async ({ fn, args }: waitForAddressArgs): Promise<true> => {
```
dolcalmi, line 117: this seems to be a generic method, why waitForNoAddress?

test/e2e/servers/trigger/trigger.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+})
+
+type waitForAddressArgs = {
+  fn: (
+    ListWalletOnChainPendingReceiveByAddressesArgs,
+  ) => Promise<WalletOnChainSettledTransaction[] | Error>
+  args: ListWalletOnChainPendingReceiveByAddressesArgs
+}
+
+const waitForAddress = async ({
```
dolcalmi, line 98: same here

test/helpers/bria.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  const timeout = 30_000
+  let wrapper
+  const promise = new Promise(async (resolve, reject) => {
+    const timeoutId = setTimeout(() => {
+      reject(new Error(`Promise timed out after ${timeout} ms`))
+    }, timeout)
+    wrapper = await bria.subscribeToAll(eventHandler({ resolve, timeoutId }))
+  })
```
dolcalmi, line 37: same coment here, use timeout promise or waitFor

src/config/process.ts
```diff
@@ -11,7 +11,10 @@ type TwilioConfig = {
 export const GALOY_API_PORT = process.env.GALOY_API_PORT || 4012
 export const GALOY_ADMIN_PORT = process.env.GALOY_ADMIN_PORT || 4001
 
-export const BRIA_WALLET_NAME = process.env.BRIA_WALLET_NAME || "dev"
+const briaUrl = process.env.BRIA_HOST ?? "localhost"
+const briaPort = process.env.BRIA_PORT ?? "2742"
+export const BRIA_ENDPOINT = `${briaUrl}:${briaPort}`
```
dolcalmi, line 16: please create a function to return all bria configs

src/config/yaml.ts
```diff
@@ -82,6 +82,9 @@ export const USER_ACTIVENESS_MONTHLY_VOLUME_THRESHOLD = toCents(
   yamlConfig.userActivenessMonthlyVolumeThreshold,
 )
 
+export const BRIA_WALLET_NAME = yamlConfig.bria.hotWalletName
+export const BRIA_QUEUES = yamlConfig.bria.queues
```
dolcalmi, line 86: same here, maybe in the main file to consolidate process and yaml configs

src/config/schema.types.d.ts
```diff
@@ -53,6 +53,14 @@ type YamlSchema = {
     onChainWallet: string
     targetConfirmations: number
   }
+  bria: {
+    hotWalletName: string
+    queues: {
+      fast: string
+      medium: string
+      slow: string
+    }
```
dolcalmi, line 62: @bodymindarts this is what you mean with inject queues? I thought was inject a list of queues

src/domain/bitcoin/onchain/incoming-tx-handler.ts
```diff
@@ -80,3 +82,78 @@ export const IncomingOnChainTxHandler = (
...
   }
 }
+
+export const NewIncomingOnChainTxHandler = <S extends WalletCurrency>(
+  txns: WalletOnChainSettledTransaction[],
+): NewIncomingOnChainTxHandler<S> | ValidationError => {
+  const walletCurrency = txns[0].settlementCurrency
+  const settlementCurrencies = new Set(txns.map((tx) => tx.settlementCurrency))
+  if (settlementCurrencies.size !== 1) {
+    return new MultipleCurrenciesForSingleCurrencyOperationError()
```
dolcalmi, line 92: this does not match logic in `briaGetPendingOnChainBalanceForWallets`

src/services/bria/index.ts
```diff
@@ -125,7 +132,38 @@ export const NewOnChainService = (): INewOnChainService => {
...
+        request.setExternalId(requestId)
+      }
+
+      const newAddressWithMetadataOverload = (
+        { request, metadata }: { request: NewAddressRequest; metadata: Metadata },
+        callback: (error: ServiceError | null, response: NewAddressResponse) => void,
+      ) => {
+        const newAddress = bitcoinBridgeClient.newAddress.bind(bitcoinBridgeClient)
+        return newAddress(request, metadata, callback)
+      }
```
dolcalmi, line 151: why to use this structure if you will promisify later?

src/app/wallets/add-settled-transaction.ts
dolcalmi, line 1: is this for all transactions or only for onchain? if it is only for onchain please update file name

src/app/wallets/get-pending-onchain-balance-for-wallet.ts
```diff
@@ -24,3 +30,40 @@ export const getPendingOnChainBalanceForWallets = async (
...
+): Promise<{ [key: WalletId]: PaymentAmount<S> } | ApplicationError> => {
+  const balancesFromLnd = await lndGetPendingOnChainBalanceForWallets(wallets)
+  if (balancesFromLnd instanceof Error) return balancesFromLnd
+
+  const balancesFromBria = await briaGetPendingOnChainBalanceForWallets<S>(wallets)
+  if (balancesFromBria instanceof Error) return balancesFromBria
+
+  const walletIds = Array.from(
+    new Set([...Object.keys(balancesFromLnd), ...Object.keys(balancesFromLnd)]),
+  ) as WalletId[]
```
dolcalmi, line 60: this confirms the issue of having two methods... we just need to replace/update getOnChainTxs (and persistence) with `WalletOnChainPendingReceiveRepository` and only have 1 method/ source of truth

dolcalmi (changes requested):

src/services/ledger/facade.ts
```diff
@@ -64,7 +64,7 @@ export const recordSend = async ({
   return persistAndReturnEntry({ entry, hash: metadata.hash })
 }
 
-export const recordReceive = async ({
+export const recordReceiveContraLnd = async ({
```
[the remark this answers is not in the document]
vindard: I did this in payout PR (https://github.com/GaloyMoney/galoy/pull/2684/commits/4917f6e874d9de8dabaff37e813130a9cf8e1c0e) because all offchain went against lnd account, but in this PR there's still some onchain places where we're writing things against the lnd account.

But now I'm thinking, we'll eventually import lnd xpubs to bria as well anyway, so maybe we should change all receives (lnd and bria) to be agains that new `OnChain` accounting path now?

src/services/ledger/index.types.d.ts
```diff
@@ -32,6 +32,7 @@ type RecordReceiveArgs = {
     usd: UsdPaymentAmount
     btc: BtcPaymentAmount
   }
+  isBria?: boolean
```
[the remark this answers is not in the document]
vindard: Missed this one in an earlier change, fixed now

src/services/mongoose/schema.ts
```diff
@@ -116,10 +116,7 @@ const WalletSchema = new Schema<WalletRecord>({
   onchain: {
     type: [
       {
-        pubkey: {
-          type: String,
-          required: true,
-        },
+        pubkey: String,
```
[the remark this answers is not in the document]
vindard: DO you mean a dummy string or an actual pubkey? Actual pubkey won't make sense since bria doesn't really have this concept.

src/utils/index.ts
```diff
@@ -13,6 +13,32 @@ export function timeout(delay: MilliSeconds | number, msg: string) {
   })
 }
 
+export function timeoutWithCancel(
```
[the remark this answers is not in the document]
vindard: I changed all instances of `timeout` to `timeoutWithCancel` instead, to help with us not having open async handles in application internal code as well -> https://github.com/GaloyMoney/galoy/pull/2662/commits/c5d1e53249883f2be8fa1a3d4aa40ac32cfcef45

src/app/wallets/create-on-chain-address.ts
```diff
@@ -1,15 +1,18 @@
...
+
+import { NewOnChainService } from "@services/bria"
 import { WalletOnChainAddressesRepository, WalletsRepository } from "@services/mongoose"
+import { OnChainService } from "@services/lnd/onchain-service"
 import { consumeLimiter } from "@services/rate-limit"
 
 import { validateIsBtcWallet, validateIsUsdWallet } from "./validate"
 
-const createOnChainAddress = async (
+export const lndCreateOnChainAddress = async (
```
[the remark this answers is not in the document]
vindard: Those tests will all be removed as well once we move everything to bria, so it's all meant to be a temporary duplicate for now. I can move it, but it was easier to track the parallels if I left it in the main file with the other bria use-cases for now.

(I can move to test helper if we still think this is absolutely necessary)

src/app/wallets/create-on-chain-address.ts
```diff
@@ -34,28 +37,70 @@ const createOnChainAddress = async (
...
-  wallet: Wallet,
-): Promise<OnChainAddress | ApplicationError> => {
+const createOnChainAddress = async ({
+  walletId,
+  requestId,
+}: {
+  walletId: WalletId
+  requestId?: OnChainAddressRequestId
+}) => {
+  const wallet = await WalletsRepository().findById(walletId)
```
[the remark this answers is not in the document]
vindard: How do you mean? I thought we checked for authentication [here](https://github.com/GaloyMoney/galoy/blob/2b9d949aabd7973a99dde2dde9b6c96d849befe2/src/graphql/main/mutations.ts#L84)

src/app/wallets/create-on-chain-address.ts
```diff
@@ -34,28 +37,70 @@ const createOnChainAddress = async (
...
-  return validated instanceof Error ? validated : createOnChainAddress(walletId)
+  return validated instanceof Error
+    ? validated
+    : createOnChainAddress({ walletId, requestId })
 }
 
-export const createOnChainAddressForUsdWallet = async (
-  walletId: WalletId,
-): Promise<OnChainAddress | ApplicationError> => {
+export const createOnChainAddressForUsdWallet = async ({
```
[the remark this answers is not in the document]
vindard: I think it's used [here](https://github.com/GaloyMoney/galoy/blob/c5d1e53249883f2be8fa1a3d4aa40ac32cfcef45/src/app/wallets/create-on-chain-address.ts#L77) which in turn gets used in `add-settled-transaction` and `update-on-chain-receipt`. 

Hm, for payouts it's because of the `amount` argument but I can't remember why for this one. I can double-check if we validated wallet currency for a reason here, and if not maybe we can remove the separation (but maybe in a separate PR since that was in here before).

src/app/wallets/get-last-on-chain-address.ts
```diff
@@ -10,7 +10,7 @@ export const getLastOnChainAddress = async (
   const lastOnChainAddress = await onChainAddressesRepo.findLastByWalletId(walletId)
 
   if (lastOnChainAddress instanceof CouldNotFindError)
-    return createOnChainAddressForBtcWallet(walletId)
+    return createOnChainAddressForBtcWallet({ walletId })
```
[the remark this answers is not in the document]
vindard: See https://github.com/GaloyMoney/galoy/pull/2662#discussion_r1209334388

test/integration/app/trigger/trigger.fn.spec.ts
```diff
@@ -96,7 +101,141 @@ const getWalletState = async (walletId: WalletId): Promise<WalletState> => {
 }
 
 describe("onchainBlockEventHandler", () => {
-  it("should process block for incoming transactions", async () => {
+  const waitForSettledTxs = async ({ wallets, addresses }): Promise<true> => {
+    return new Promise<true>(async (resolve) => {
+      // eslint-disable-next-line no-constant-condition
+      while (true) {
```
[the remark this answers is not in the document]
vindard: Good point, we do already have an e2e that does exactly this at `test/e2e/servers/trigger/trigger.spec.ts`.

Do you think I can remove the `"should process pending invoices on invoice update event"` test as well since this is probably already covered in receive tests?

test/integration/02-user-wallet/02-receive-onchain.spec.ts
vindard: This file is pretty much already all bria. We will remove the `With Lnd` describe block once we switch to handling lnd xpubs from bria

src/config/process.ts
vindard: Done https://github.com/GaloyMoney/galoy/pull/2662/commits/fa660d125254e88c9e2f78a68d47fa1d34a763c7

src/config/yaml.ts
vindard: https://github.com/GaloyMoney/galoy/pull/2662#discussion_r1209421804

src/services/bria/index.ts
vindard: It's because promisify expects 2 arguments where the 2nd one is a callback, so I had to collect all the arguments from `newAddress` before the callback argument into a single one to pass on to promisify

src/app/wallets/add-settled-transaction.ts
vindard: Done https://github.com/GaloyMoney/galoy/pull/2662/commits/7d9ee28a25ec119c30abf784676234d1d3f7d7da

src/app/wallets/create-on-chain-address.ts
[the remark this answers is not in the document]
dolcalmi: if so, then get transactions has a problem :)

src/app/wallets/create-on-chain-address.ts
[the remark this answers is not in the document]
dolcalmi: but what is the difference if we enable it for stablesats?... is just the creation of an onchain address

src/app/wallets/get-last-on-chain-address.ts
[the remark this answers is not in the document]
dolcalmi: same thing, there is no diff between usd and btc onchain address creation (even if it is not enabled today)

test/integration/app/trigger/trigger.fn.spec.ts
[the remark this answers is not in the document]
dolcalmi: you are right, already covered in receive, you can remove it if subscription is covered

src/services/bria/index.ts
dolcalmi: yes, what I mean is that you dont need to prmisify, just create an async function

src/app/wallets/create-on-chain-address.ts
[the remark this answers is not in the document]
vindard: It's just that we haven't done it as yet and I don't think we should do that changeset here in addition to introducing bria

src/app/wallets/get-last-on-chain-address.ts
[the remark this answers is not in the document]
vindard: Yes there's no difference, but this is how we're currently gating that we don't create any usd wallet onchain addresses and allow onchain receives

test/integration/app/trigger/trigger.fn.spec.ts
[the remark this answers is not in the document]
vindard: Done https://github.com/GaloyMoney/galoy/pull/2662/commits/880fdddf90f97a75e5a6ef5b4b73b2da21a92e71

test/e2e/servers/trigger/trigger.spec.ts
vindard: One method starts with errors being returned and wait for a valid result, and the other does the opposite (starts with valid results and waits for errors).

I made it specific to `Address` functions just to make the type signatures easier. I figure we can make it more generic when we need to wait for something else.

test/e2e/servers/trigger/trigger.spec.ts
vindard: https://github.com/GaloyMoney/galoy/pull/2662#discussion_r1210134250

src/domain/bitcoin/onchain/incoming-tx-handler.ts
vindard: True, I added some changes to codify that the lnd handler should not need to handle non-BTC-wallet addresses/transactions because we haven't turned on stablesats receives as yet https://github.com/GaloyMoney/galoy/pull/2662/commits/516c2dbfd6a04a694f56b1bf6b435132d9d4ded4

We can decide if we need to do some more work to listen to lnd and add their pendings to pending-repo and remove when we settled those transactions (maybe ACE topic).

src/services/bria/index.ts
vindard: Don't we need promisify still to make the callback functionality into a promise?

dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/2662/commits/8f0a19a4d2ac11aae61fdb5a6f1eb66be95473af

src/servers/trigger.ts
```diff
@@ -555,6 +556,8 @@ const listenerSwapMonitor = async () => {
 }
 
 const main = () => {
+  listenerBria()
```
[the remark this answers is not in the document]
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/2662/commits/4f7e4dfd91d7cc2aa9bfa1853f3492979ca7dfde

src/services/ledger/facade.ts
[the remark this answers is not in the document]
vindard: Resolved with new handling https://github.com/GaloyMoney/galoy/pull/2662/commits/9a914bc45b9ef0a46d62f75f4636a801a1c8037d

src/app/wallets/create-on-chain-address.ts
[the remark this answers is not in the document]
vindard: Agreed we'll enable stablesats onchain receives after this PR is rolled out

src/domain/bitcoin/onchain/incoming-tx-handler.ts
vindard: `IncomingTxHandler` was removed https://github.com/GaloyMoney/galoy/pull/2662/commits/4a7451d19a8c01acf01f36e4410d31cd46bb057a

src/app/wallets/get-pending-onchain-balance-for-wallet.ts
vindard: Resolved by passing everything thru WalleterPendingRepo

src/servers/middlewares/healthz.ts
```diff
@@ -47,3 +53,14 @@ const isRedisAvailable = async (): Promise<boolean> => {
     return false
   }
 }
+
+const isBriaAvailable = async (): Promise<boolean> => {
```
bodymindarts, line 57: I don't understand why bria not being healthy should cause health check to fail.
In this case bria being down would cause k8s to count the backend as being down and that would make the whole service unusable.

Instead we should be able to handle partial system degradation (eg bria down) gracefully in the api.

cc @dolcalmi

bodymindarts (approved):

src/servers/middlewares/healthz.ts
dolcalmi: is only mandatory where is required. for ex in main api is not https://github.com/GaloyMoney/galoy/pull/2662/files#diff-651d74d3e907cd7859cee097d1d418d4d434273dc9f3392e4faabc6446eb3974R292

PR #2663, opened 2023-05-23
chore: cleanup initial bria integration

vindard:


src/servers/bria-trigger.ts
```diff
@@ -30,11 +23,25 @@ const utxoDetectedEventHandler = async ({
 }: {
   event: UtxoDetected
 }): Promise<true | ApplicationError> => {
-  return Wallets.addPendingTransaction(event)
+  const res = await Wallets.addPendingTransaction(event)
+  if (res instanceof CouldNotFindWalletFromOnChainAddressError) {
```
bodymindarts, line 27: What happens to tracing here? Don't want this error to trigger an alert.

src/services/bria/errors.ts
```diff
@@ -0,0 +1,19 @@
+import { DomainError, ErrorLevel } from "@domain/shared"
+
+export class BriaEventError extends DomainError {}
+export class EventAugmentationMissingError extends BriaEventError {}
+
+export class BriaPayloadError extends BriaEventError {}
```
bodymindarts, line 6: Aren't all these events critical?

bodymindarts (dismissed):

src/servers/bria-trigger.ts
vindard: It shouldn't trigger an alert, its level inherits from `DomainError` as `ErrorLevel.Info`.

bodymindarts (approved):

PR #2675, opened 2023-05-26
feat: add invoice expiration

dolcalmi:
Use case from mattermost

> we want to add lightning to our list of crypto/fiat options available for customers to make payments for services/products via a generated QR code.  For a payment, a fixed amount is generated which must be paid before a set time (e.g. 10mins) elapses. If it is not paid, we invalidate it and regenerate an updated amount if the user is still interested in paying. Now because an expired payment has been invalidated, a customer will not be able to make a successful payment even if they somehow still have the initial QR code.
> 
> However, because we cannot set an expiry for an lnInvoice, a customer that still has a QR code generated from the lightning's payment request will be able to make payment way past our desired expiry time.

- Update invoice builder (including tests)
- Update public and private create invoice mutations
- Update tests + domain validations

### DIFF
--- a/src/app/wallets/add-invoice-for-wallet.ts
+++ b/src/app/wallets/add-invoice-for-wallet.ts
@@ -1,19 +1,26 @@
 import { toSats } from "@domain/bitcoin"
-import { RateLimitConfig } from "@domain/rate-limit"
-import { RateLimiterExceededError } from "@domain/rate-limit/errors"
 import { checkedToWalletId } from "@domain/wallets"
-import { DealerPriceService } from "@services/dealer-price"
-import { LndService } from "@services/lnd"
-import { WalletInvoicesRepository, WalletsRepository } from "@services/mongoose"
-import { consumeLimiter } from "@services/rate-limit"
+import { RateLimitConfig } from "@domain/rate-limit"
+import { checkedToMinutes } from "@domain/primitives"
+import { RateLimiterExceededError } from "@domain/rate-limit/errors"
+import { DEFAULT_EXPIRATIONS } from "@domain/bitcoin/lightning/invoice-expiration"
 import { WalletInvoiceBuilder } from "@domain/wallet-invoices/wallet-invoice-builder"
 
+import { LndService } from "@services/lnd"
+import { consumeLimiter } from "@services/rate-limit"
+import { DealerPriceService } from "@services/dealer-price"
+import { WalletInvoicesRepository, WalletsRepository } from "@services/mongoose"
+
 import { validateIsBtcWallet, validateIsUsdWallet } from "./validate"
+
+const defaultBtcExpiration = DEFAULT_EXPIRATIONS["BTC"].delayMinutes
+const defaultUsdExpiration = DEFAULT_EXPIRATIONS["USD"].delayMinutes
 
 const addInvoiceForSelf = async ({
   walletId,
   amount,
   memo = "",
+  expiresIn,
 }: AddInvoiceForSelfArgs): Promise<LnInvoice | ApplicationError> =>
   addInvoice({
     walletId,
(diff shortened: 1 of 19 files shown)


src/app/wallets/add-invoice-for-wallet.ts
```diff
@@ -84,35 +116,57 @@ const addInvoiceForRecipient = async ({
...
+  const walletId = checkedToWalletId(recipientWalletId)
+  if (walletId instanceof Error) return walletId
+
+  let defaultExpiresIn = defaultBtcExpiration
+  const validated = await validateIsBtcWallet(walletId)
+  if (validated instanceof Error) {
+    defaultExpiresIn = defaultUsdExpiration
+  }
+
+  const expiresInChecked = checkedToMinutes(expiresIn || defaultExpiresIn)
```
nicolasburtey, line 165: we should not allow longer expiration time for USD invoice. this opens an attack vector by enabling longer time frame thus making easy to arbitrage the trade.

nicolasburtey (dismissed):
USD invoice duration should not be extended beyond defined values in configuration

src/app/wallets/add-invoice-for-wallet.ts
dolcalmi: yes, this is covered https://github.com/GaloyMoney/galoy/pull/2675/files#diff-8c5feb755004a1236e79796941161c08424552caf5fb730112c429a5b513d3a4R30

test/unit/domain/bitcoin/lightning/invoice-expiration.spec.ts
```diff
@@ -1,15 +1,63 @@
...
+    delay = toSeconds(0)
+    expiresAt = invoiceExpirationForCurrency(BTC, now, delay)
+    expect(expiresAt).toEqual(expectedExpiration)
+
+    delay = toSeconds(2 * SECS_PER_DAY)
+    expiresAt = invoiceExpirationForCurrency(BTC, now, delay)
+    expect(expiresAt).toEqual(expectedExpiration)
+  })
+
+  it("should return expiration for USD currency with default delay", () => {
```
nicolasburtey, line 30: ok. but why do we allow the override of the default value then here https://github.com/GaloyMoney/galoy/pull/2675/files#r1206365122?

dolcalmi: not sure to understand, the default value that we have right now became the max value.. so for example for USD invoices could be from 1 to 5 minutes (which is even better for us) and for btc invoices from 1 min to 24 hours.

nicolasburtey: ok then that's good

test/unit/domain/bitcoin/lightning/invoice-expiration.spec.ts
```diff
@@ -1,15 +1,63 @@
...
+    const expiration = invoiceExpirationForCurrency(currency, now, delay)
+    const expectedExpiration = new Date("2000-01-01T00:30:00Z")
+    expect(expiration).toEqual(expectedExpiration)
   })
 
-  it("USD invoice expires in 5 minutes", () => {
-    const creationDate = new Date("2000-01-01T00:00:00.000Z")
-    const expiresAt = invoiceExpirationForCurrency("USD", creationDate)
-    expect(expiresAt).toEqual(new Date("2000-01-01T00:05:00.000Z"))
+  it("should return expiration for USD currency with provided delay", () => {
```
vindard, line 56: Should we add tests for if they try to set an expiration above the max delay?

vindard (commented):
Just one comment

test/unit/domain/bitcoin/lightning/invoice-expiration.spec.ts
dolcalmi: it is in `"should return expiration for USD currency with default delay"` https://github.com/GaloyMoney/galoy/pull/2675/files#diff-8c5feb755004a1236e79796941161c08424552caf5fb730112c429a5b513d3a4R43

vindard (approved):

PR #2679, opened 2023-05-26
feat: add supported countries api

UncleSamtoshi:


src/graphql/main/schema.graphql
```diff
@@ -309,6 +314,11 @@ type Globals {
   This can be used to know if an invoice belongs to one of our nodes.
   """
   nodesIds: [String!]!
+
+  """
+  A list of supported CCA2 country codes for this Galoy instance
+  """
+  supportedCountries: [CountryCode!]!
```
dolcalmi, line 321: uhmm now I see why you wanted unsupported countries. if you only return country code you still need to use another source to render the component... alternative is to create a country entity with all the needed data (country code as id, prefix, flag, name).. not sure if the libphonelibrary has the fields but there are other sources too https://github.com/LucianoGanga/country-codes-list/blob/master/countriesData.js

UncleSamtoshi: We already use another source to render the phone code picker, so it is not a big deal.

dolcalmi: uhm I guess this requires more review then @bodymindarts could you please validate the schema?... we will have a breaking change if we decide to add different providers for country.. at least we need to have a `Country` type (even if it only has countryCode attr)

dolcalmi (changes requested):

UncleSamtoshi:
> we will have a breaking change if we decide to add different providers for country
Why will we have a breaking change if we use another country provider? CCA2 is a standard for country codes.

UncleSamtoshi:
This PR precedes the mobile component https://github.com/GaloyMoney/galoy-mobile/actions/runs/5125730731/jobs/9219293994?pr=2246

dolcalmi:
> > we will have a breaking change if we decide to add different providers for country
> 
> Why will we have a breaking change if we use another country provider? CCA2 is a standard for country codes.

this is for validation, I meant  verification code provider, for example the african provider does not offer whatsapp so we need to inform mobile wallet what to offer. Please create a Country Type with a CountryCode attribute this will give us more flexibility in case we need to add more attributes.

src/graphql/root/query/globals.ts
```diff
@@ -22,6 +24,8 @@ const GlobalsQuery = GT.Field({
       lightningAddressDomain: getLightningAddressDomain(),
       lightningAddressDomainAliases: getLightningAddressDomainAliases(),
       buildInformation: getGaloyBuildInformation(),
+      smsAuthSupportedCountries: getSmsAuthSupportedCountries(),
+      whatsAppAuthSupportedCountries: getWhatsAppAuthSupportedCountries(),
```
dolcalmi, line 28: didnt see this before my last comment but.... if you create a Country type (or use the name that you prefer) we will be able to include the supported methods and will only need 1 query

dolcalmi (changes requested):

src/graphql/root/query/globals.ts
UncleSamtoshi: I can switch to that schema. Do you think it needs to be in the domain or is it okay just being a type in gql layer?

dolcalmi (dismissed):
looks good to me, please validate with @nicolasburtey

src/graphql/main/schema.graphql
```diff
@@ -259,6 +259,17 @@ type Coordinates {
   longitude: Float!
 }
 
+type Country {
+  code: CountryCode!
+  id: ID!
```
nicolasburtey, line 264: what is the difference between `code` and `id`?

nicolasburtey: assuming there is no difference we can only keep `Id`, and add a graphql comment explaining this is a countryCode. 

`Id` helps with out-of-the-box-caching so this is why we should favor it.

UncleSamtoshi: No difference at the moment. Code is the correct semantic name but it can work for an ID as well. I believe we do the same thing with the Currency object.

UncleSamtoshi: I suppose we can get rid of ID since it is not used.

UncleSamtoshi: Sorry that comment was not initially showing up for me. I will switch to only use id

UncleSamtoshi: Would you keep the type as NonNullId or CountryCode for the id field

UncleSamtoshi: I will stick with CountryCode for right now as it is more descriptive

src/graphql/main/schema.graphql
```diff
@@ -259,6 +259,16 @@ type Coordinates {
   longitude: Float!
 }
 
+type Country {
+  id: CountryCode!
```
nicolasburtey, line 263: I think `id` need to be of type `ID` for the cache to work.

but you can maybe verify if that's the case.

useful reference: https://www.apollographql.com/docs/react/caching/overview#2-generate-cache-ids

UncleSamtoshi: Per ChatGPT it shouldn't make any difference, I also didn't see any mention of it in that documentation. Either way our current behavior is to cache the entire Globals object.

nicolasburtey (approved):

PR #2688, opened 2023-05-30
ci: exit gracefully on no proto diff

krtk6160:
The combination of tag_regex and paths seems to not be working correctly and the sync job is being triggered on each tagged commit, instead of only on the tagged commits that change the proto files.


bodymindarts:
What about adding a `path` key to the resource?

krtk6160:
> What about adding a `path` key to the resource?

Already tried that here #2637 by applying it to concourse (it's still applied) but it doesn't work. Either the path works or tag_regex, but not both in combination.

PR #2695, opened 2023-06-01
fix: expand address DuplicateKey error regex

vindard:


dolcalmi:
possible bria issue, please dont merge until confirmed with @bodymindarts

vindard:
Very likely a Bria issue, possibly same addresses being generated for unique external_id values passed

PR #2699, opened 2023-06-02
fix: avoid duplicate pending tx ids by deriving from 'txHash:vout'

vindard:
## Description

We derived from just `transactionHash` before which is no longer unique as we parse each vout as a unique transaction now. This PR fixes that.

### DIFF
--- a/src/services/mongoose/wallet-onchain-pending-receive.ts
+++ b/src/services/mongoose/wallet-onchain-pending-receive.ts
@@ -1,3 +1,5 @@
+import getUuidByString from "uuid-by-string"
+
 import { toSats } from "@domain/bitcoin"
 import { toCents } from "@domain/fiat"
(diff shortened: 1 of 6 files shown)


src/services/mongoose/wallet-onchain-pending-receive.ts
```diff
@@ -100,7 +102,7 @@ const translateToWalletOnChainTransaction = (
   const walletCurrency = result.walletCurrency as WalletCurrency
   const toCurrency = walletCurrency === WalletCurrency.Btc ? toSats : toCents
   return {
-    id: result.transactionHash as OnChainTxHash,
+    id: getUuidByString(`${result.transactionHash}:${result.vout}`) as OnChainTxHash,
```
bodymindarts, line 105: why wrap it in `getUuidByString`?

bodymindarts (approved):

PR #2708, opened 2023-06-02
fix: debounce block handler

dolcalmi:


vindard (approved):
Just leaving some notes:
- this is to help with double events from this issue: https://github.com/lightningnetwork/lnd/issues/7745
- there's an edge case where if 4 blocks get mined in under 1000ms (debounce period) then some settled transactions don't get recorded, but these will be picked up by cron anyway

PR #2713, opened 2023-06-04
feat: email login and 2fa with totp

nicolasburtey:


ntheile:
Looks good so far. I am reading the ory documentation here https://www.ory.sh/docs/kratos/self-service/flows/verify-email-account-activation specifically the section on `Attempted verification notifications` and they reference this troy hunt article about `account enumeration attacks` https://www.troyhunt.com/website-enumeration-insanity-how-our-personal-data-is-leaked/ . We might want to consider also enabling a few more settings to prevent this. 

```
flows:
  verification:
    enabled: true
    lifespan: 15m # Defines how much time the user has to complete the verification flow in the UI. Default: 1h.
    use: code # Defines which method is used, one of 'code' or 'link'.
    notify_unknown_recipients: true # Defines if the system sends attempted verification notifications to unregistered addresses.
```

test/e2e/servers/ory/kratos.spec.ts
```diff
@@ -432,40 +390,88 @@ describe("phone+email schema", () => {
...
+      email,
+    })
+    if (!(res instanceof AuthenticationError)) throw new Error("wrong type")
+    expect(res.message).toBe("Request failed with status code 409")
+  })
+
+  it("can't authenticate until email has been verified", async () => {
+    // this check should be at a higher level
+    expect(true).toBe(true)
   })
```
ntheile, line 414: could somebody add an email address that does not belong to them and never verify it. Then when the rightful owner tries to add it, would they be blocked?

ntheile:
should we expose the `addEmailToIdentity` via a graphql mutation or HTTP POST `/auth/add/email`?

test/e2e/servers/ory/kratos.spec.ts
nicolasburtey: I think currently we can only add one email with the admin api we are using. it will need to be deleted before attempting to add another one (for a typo for instance). 

this is because the email is attached to the identity before it has been verified. it's a flaw somewhat but I'm not sure there is an alternative in how we use the API.

so in effect yes it could be an issue, but an "email spammer" would be able to add only one email. 

I've been thinking about a way to clean this up eventually, with a cronjob that would remove unconfirmed email every day or so. maybe not very important at first.

nicolasburtey:
> should we expose the `addEmailToIdentity` via a graphql mutation or HTTP POST `/auth/add/email`?

not sure. I think the one of the heuristic I want to think about when doing auth related query is whether it requires an existing token or not. if it does, maybe using the graphql makes more sense because it simplify client side implementation? (ie: the upgrade path). but the login path should be more rest based.

test/e2e/servers/ory/kratos.spec.ts
ntheile: Good idea, a cron job would be a nice simple solution. I think the email verification code expires after 1 hour by default so a daily cron should be good enough to mitigate

src/services/kratos/totp.ts
```diff
@@ -0,0 +1,89 @@
...
+      updateSettingsFlowBody: {
+        method: "totp",
+        totp_code: totpCode,
+      },
+      xSessionToken: authToken,
+    })
+  } catch (err) {
+    return new UnknownKratosError(err)
+  }
+}
```
ntheile, line 50: should we check the result of the response to be cautious if the kratos api changes? the happy path appears to return `resp.data.state === "success"`.

src/services/kratos/auth-email-no-password.ts
```diff
@@ -24,50 +73,133 @@ export const AuthWithEmailPasswordlessService = (): IAuthWithEmailPasswordlessSe
...
+      //
+      // we do not need to do that because we are passwordless
+
+      if (
+        isAxiosError(err) &&
+        err.response?.data?.error?.id === "browser_location_change_required"
+      ) {
+        const cookies: string[] | undefined = err.response?.headers["set-cookie"]
+        if (cookies === undefined) return new UnknownKratosError("cookies undefined")
+
```
ntheile, line 115: does this catch path always get called? i set a breakpoint here after running the tests and the `browser_location_change_required` path was always called. are cookies just a requirement for this custom flow?

src/services/kratos/totp.ts
nicolasburtey: good point, I'll make the changes

src/services/kratos/auth-email-no-password.ts
nicolasburtey: > does this catch path always get called when validating the email code?

yes I think, but only for the happy path. the idea is that in the normal flow you're been askedm at this point, to change your password. but we don't need on our end to complete this flow, as long as we know the user successfully went to this flow, meaning the code was correct, we can authorize the user. 

> are cookies just a requirement for this custom flow?

I think so, until https://github.com/ory/kratos/issues/3163 is been implemented

src/app/auth/email.ts
```diff
@@ -0,0 +1,79 @@
...
+  userId,
+}: {
+  email: EmailAddress
+  userId: UserId
+}): Promise<AddEmailToIdentityResult | KratosError> => {
+  const authServiceEmail = AuthWithEmailPasswordlessService()
+
+  const hasEmail = await authServiceEmail.hasEmail({ kratosUserId: userId })
+  if (hasEmail instanceof Error) return hasEmail
+  if (hasEmail) return new EmailAlreadyExistsError()
```
bodymindarts, line 17: This has issues with idempotency.
If the `authServiceEmail.sendEmailWithCode ` call fails then the user entity hasn't been persisted yet. The end user might try again and hit the early return even though the `user` wasn't updated.

src/app/auth/index.types.d.ts
```diff
@@ -2,3 +2,23 @@ type NewAccountWithPhoneIdentifier = {
...
   phone: PhoneNumber
 }
+
+type LoginDeviceUpgradeWithPhoneResult = {
+  success: true
+  sessionToken?: SessionToken
+}
+
+type initiateTotpRegistrationResult = {
+  flow: FlowId
```
bodymindarts, line 12: Can we have more granular / better naming around flow id. I don't like that we are just conforming to kratos semantics here.

src/domain/admin-users/index.ts
```diff
@@ -1,9 +1,9 @@
 export const adminUsers = [
   {
-    phone: "+16505554327" as PhoneNumber,
+    phone: "+16505554350" as PhoneNumber,
```
bodymindarts, line 3: Changing this will have impact on live environments.

src/servers/trigger.ts
```diff
@@ -53,7 +53,7 @@ import { NotificationsService } from "@services/notifications"
 import { activateLndHealthCheck, lndStatusEvent } from "@services/lnd/health"
 import { recordExceptionInCurrentSpan, wrapAsyncToRunInSpan } from "@services/tracing"
 
-import healthzHandler from "./middlewares/healthz"
+import healthzHandler from "./rest/healthz"
```
bodymindarts, line 56: I don't really see `rest` being a useful subcategory. It's one of those functional vs semantic groupings. how about:
```
server
- auth
  - router.ts
- health.ts
```
Ie I don't see the need for the rest subfolder - especially since `health` doesn't even conform to rest.

src/services/kratos/index.types.d.ts
```diff
@@ -25,3 +25,8 @@ type KratosPublicMetadata = {
 type KratosAdminMetadata = {
   phoneMetadata: PhoneMetadata | undefined
 }
+
+type FlowId = string & { readonly brand: unique symbol }
```
bodymindarts, line 29: I saw this type being used in domain code so it should not be defined here. Serivces depend domain not the other way around.
Apart from that I also don't like the naming (as commented elsewhere).

src/services/kratos/private.ts
```diff
@@ -22,7 +19,74 @@ export const toDomainSession = (session: KratosSession): Session => {
...
+    case "phone_no_password_v0":
+      return SchemaIdType.PhoneNoPasswordV0
+    case "phone_email_no_password_v0":
+      return SchemaIdType.PhoneEmailNoPasswordV0
+    case "email_no_password_v0":
+      return SchemaIdType.EmailNoPasswordV0
+    case "username_password_deviceid_v0":
+      return SchemaIdType.UsernamePasswordDeviceIdV0
+    default:
+      throw new Error(`Invalid schemaId: ${schemaId}`)
```
bodymindarts, line 38: Are you sure we want to throw here? Shouldn't this return a critical `UnknownSchemaError`?

nicolasburtey: it's really to silent type error from typescript: `Function lacks ending return statement and return type does not include 'undefined'.ts(2366)`

it will only arise in production is kratos.yml is misconfigured, so I think that is fair to throw and not return an error. returning an error would have a cascading effect into checking for everything that depend a schema whether there is an error or not, and so this has a lot of boilerplate for really just a type safety that I want to ignore because I know the `switch/case` is meant to be extensive.

test/unit/domain/authentication/index.spec.ts
```diff
@@ -1,4 +1,4 @@
-import { getSupportedCountries } from "@domain/authentication"
+import { getSupportedCountries } from "@domain/auth"
```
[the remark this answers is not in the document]
nicolasburtey: that make sense. I did that to align with the `auth` we had in `app`. I'll rename it to `authorization`

src/graphql/main/schema.graphql
```diff
@@ -1533,6 +1580,31 @@ type UserContactUpdateAliasPayload {
...
+  me: User
+}
+
+input UserEmailSetInput {
+  email: EmailAddress!
+}
+
+type UserEmailSetPayload {
+  errors: [Error!]!
+  flow: Flow
```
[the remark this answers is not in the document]
nicolasburtey: > SetEmaiOpRef and VerifyEmailOpRef for example. That makes it type safe not to put the wrong flow into the wrong process.

I'm not against using another wording here, but I'm not sure to understand the concern on `That makes it type safe not to put the wrong flow into the wrong process.`

we are just have one meaning for flow currently, it comes with a code to validate an email.

src/app/auth/email.ts
nicolasburtey: I'm not sure to understand the concern? if `hasEmail === true` we don't want to add another email.

src/servers/trigger.ts
nicolasburtey: renamed to `authorization` and move back health into middleware even thought it's not a middleware per `express` definition.

src/app/auth/email.ts
bodymindarts: You are updating 2 places - auth service + user entity. I'm just talking about the potential state / inconsistency that can arise when auth service update succeeds but user entity update fails. IE a likely outcome in a distributed system.

src/servers/authorization/email.ts
```diff
@@ -0,0 +1,135 @@
...
+  EmailValidationSubmittedTooOftenError,
+} from "@domain/authentication/errors"
+
+import { UserLoginIpRateLimiterExceededError } from "@domain/rate-limit/errors"
+
+import { authRouter } from "./router"
+
+// TODO: should code request behind captcha or device token?
+authRouter.post(
+  "/email/code-request",
```
bodymindarts, line 30: In http the verb maps to  the http operation and the path maps to the resource. perhaps use `/email/auth-code`?

src/servers/authorization/email.ts
```diff
@@ -0,0 +1,135 @@
...
+      baseLogger.info("/email/login")
+
+      const ipString = isProd ? req?.headers["x-real-ip"] : req?.ip
+      const ip = parseIps(ipString)
+
+      if (!ip) {
+        return res.status(500).send({ error: "IP is not defined" })
+      }
+
+      const flowRaw = req.body.flow
```
bodymindarts, line 86: `loginId`?

src/graphql/main/schema.graphql
```diff
@@ -330,6 +340,11 @@ type FeesInformation {
   deposit: DepositFeesInformation!
 }
 
+"""
+A flow to be passed between request code and logging api when using email
+"""
+scalar Flow
```
bodymindarts, line 346: `EmailVerificationId`?

src/app/auth/email.ts
nicolasburtey: how would you address this issue here?

src/services/kratos/auth-email-no-password.ts
```diff
@@ -1,18 +1,68 @@
...
+  kratosAdmin,
+  kratosPublic,
+  toDomainIdentityEmail,
+  toDomainIdentityEmailPhone,
+  toDomainIdentityPhone,
+} from "./private"
+import { SchemaIdType } from "./schema"
+
+const getKratosKnex = () =>
+  knex({
```
bodymindarts, line 31: Why not use a simpler client like https://node-postgres.com?

src/services/kratos/auth-email-no-password.ts
```diff
@@ -86,21 +230,243 @@ export const AuthWithEmailPasswordlessService = (): IAuthWithEmailPasswordlessSe
...
+        new UnknownKratosError(err)
+      )
+    }
+
+    if (identity.schema_id !== SchemaIdType.PhoneEmailNoPasswordV0) {
+      return new IncompatibleSchemaUpgradeError()
+    }
+
+    if (identity.state === undefined)
+      throw new KratosError("state undefined, probably impossible state") // type issue
```
bodymindarts, line 319: This should be a specific error `KratodIdentityUndefined` or something like that.

src/services/kratos/auth-email-no-password.ts
```diff
@@ -86,21 +230,243 @@ export const AuthWithEmailPasswordlessService = (): IAuthWithEmailPasswordlessSe
...
+    }
+
+    if (identity.schema_id !== "phone_email_no_password_v0") {
+      return new IncompatibleSchemaUpgradeError(
+        "identity does not have phone_email_no_password_v0 schema",
+      )
+    }
+
+    if (identity.state === undefined)
+      throw new KratosError("state undefined, probably impossible state") // type issue
```
bodymindarts, line 364: Specific error.

src/services/kratos/auth-email-no-password.ts
nicolasburtey: no specific reason, but mainly because we use knex in other repos (ie: https://github.com/GaloyMoney/price/blob/3cec863ef696e331d2e4313f83a32e6ede666770/history/src/config/database.ts#L1)

src/services/kratos/auth-email-no-password.ts
```diff
@@ -1,18 +1,68 @@
...
+} from "./private"
+import { SchemaIdType } from "./schema"
+
+const getKratosKnex = () =>
+  knex({
+    client: "pg", // specify the database client
+    connection: process.env.KRATOS_POSTGRES_DNS,
+  })
+
+export const getIdentityIdFromFlowId = async (flowId: string) => {
```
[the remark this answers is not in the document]
nicolasburtey: nope. good point.

src/services/kratos/auth-email-no-password.ts
```diff
@@ -86,21 +230,243 @@ export const AuthWithEmailPasswordlessService = (): IAuthWithEmailPasswordlessSe
...
+  }
+
+  const removeEmailFromIdentity = async ({ kratosUserId }: { kratosUserId: UserId }) => {
+    let identity: KratosIdentity
+
+    try {
+      ;({ data: identity } = await kratosAdmin.getIdentity({ id: kratosUserId }))
+    } catch (err) {
+      return (
+        (isAxiosError(err) && new UnknownKratosError(err.message)) ||
```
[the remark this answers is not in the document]
nicolasburtey: could you clarify your point?

src/app/authentication/email.ts
```diff
@@ -0,0 +1,82 @@
...
+  })
+  if (res instanceof Error) return res
+
+  const emailRegistrationId = await authServiceEmail.sendEmailWithCode({ email })
+  if (emailRegistrationId instanceof Error) return emailRegistrationId
+
+  const user = await UsersRepository().findById(userId)
+  if (user instanceof Error) return user
+
+  const updatedUser = await UsersRepository().update({ ...user, email })
```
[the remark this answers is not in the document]
nicolasburtey: I don't really have a good answer, and I think this behavior is prevalent in many places in the codebase already. could be a good code engineering topic to get some perspective on how to approach this in a better more systemic way.

bodymindarts (approved):

PR #2723, opened 2023-06-07
ci: use test_on_docker_host

bodymindarts:
already applied

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -7,7 +7,8 @@
 #@   "install_yarn_deps",
 #@   "nodejs_check_code",
 #@   "nodejs_audit",
-#@   "nodejs_integration_test",
+#@   "test_on_docker_host",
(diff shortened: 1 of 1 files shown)


krtk6160 (approved):

ci/pipeline.yml
```diff
@@ -48,8 +49,8 @@ source:
 groups:
   - name: galoy
     jobs:
-      - test-integration
-      - test-e2e
+      - integration-tests
+      - e2e-tests
       - test-unit
```
krtk6160, line 54: This should also be changed to maintain uniformity in naming?

bodymindarts:
```
 - #@ test_on_docker_host("e2e-tests")
 ```
 
 the way the fragment works is it uses the argument (`e2e-tests`) both as the concourse job name as well the name of the container it will run.

PR #2725, opened 2023-06-07
feat: add account delete mutation

dolcalmi:


src/app/accounts/delete-account.ts
```diff
@@ -0,0 +1,27 @@
...
+  accountId: AccountId
+}): Promise<true | ApplicationError> => {
+  const accountsRepo = AccountsRepository()
+  const account = await accountsRepo.findById(accountId)
+  if (account instanceof Error) return account
+
+  const { kratosUserId } = account
+
+  const usersRepo = UsersRepository()
+  const user = await usersRepo.findById(kratosUserId)
```
bodymindarts, line 16: I don't think `kratosUserId` is defined here

src/app/accounts/delete-account.ts
```diff
@@ -0,0 +1,27 @@
+import { IdentityRepository } from "@services/kratos"
+import { AccountsRepository, UsersRepository } from "@services/mongoose"
+
+export const deleteAccount = async ({
```
nicolasburtey, line 4: should we rename to `markAccountForDeletion`? Because the account won't be deleted immediately

nicolasburtey (approved):

PR #2733, opened 2023-06-09
fix: markAccountForDeletion for device accounts

dolcalmi:


src/app/accounts/mark-account-for-deletion.ts
```diff
@@ -16,8 +16,10 @@ export const markAccountForDeletion = async ({
   const user = await usersRepo.findById(kratosUserId)
   if (user instanceof Error) return user
 
-  const result = await usersRepo.adminUnsetPhoneForUserPreservation(kratosUserId)
-  if (result instanceof Error) return result
+  if (user.phone) {
+    const result = await usersRepo.adminUnsetPhoneForUserPreservation(kratosUserId)
+    if (result instanceof Error) return result
```
nicolasburtey, line 21: should we `return true` instead? if the user has a device Id it has been succesfully market for deletion, we should return true unless we ran into thing like database connection issues.
I think we might want to unset some `deviceId` also

dolcalmi: updated with deviceId logic to preserver correctly the user (also because deviceId users are not present in kratos)

ntheile: this does not work in the current design (since userId is overloaded with kratosId | deviceId) . the only way i can think of to soft delete a device account is to prepend "deleted" to the Account and user collections like this: 

```
Account.kratosUserId = "deleted" + kratosUserId
User.userId = "deleted" + userId
```

Might want to hold off on this until we do some more domain modeling...it's seems to be getting a bit hacky imo

dolcalmi: `usersRepo.adminUnsetDeviceIdForUserPreservation(kratosUserId)` this rename deviceId to deletedDeviceId so is not a problem

src/services/mongoose/users.ts
```diff
@@ -122,7 +122,7 @@ export const UsersRepository = (): IUsersRepository => {
   ): Promise<User | RepositoryError> => {
     try {
       const result = await User.findOneAndUpdate(
-        { userId: id },
+        { userId: id, phone: { $exists: true } },
```
nicolasburtey, line 125: this throw an error if we don't have `, phone: { $exists: true }`?

nicolasburtey (approved):

ntheile:
Closing because we implemented this in the device account v3 PR here https://github.com/GaloyMoney/galoy/pull/2747/commits/4ce82be80b523154e93873a526db03dfa09b8cb6

PR #2746, opened 2023-06-13
fix: pending incoming balance for zero txns incoming

vindard:
## Description

This PR handles the cases where there are no pending incoming txns for the queried wallets

### DIFF
--- a/src/app/wallets/get-pending-onchain-balance-for-wallet.ts
+++ b/src/app/wallets/get-pending-onchain-balance-for-wallet.ts
@@ -1,7 +1,11 @@
-import { CouldNotFindError } from "@domain/errors"
+import {
+  CouldNotFindError,
+  MultipleCurrenciesForSingleCurrencyOperationError,
+} from "@domain/errors"
 
 import { WalletOnChainPendingReceiveRepository } from "@services/mongoose"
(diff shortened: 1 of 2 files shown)


dolcalmi (approved):

src/app/wallets/get-pending-onchain-balance-for-wallet.ts
```diff
@@ -1,15 +1,35 @@
...
 export const getPendingOnChainBalanceForWallets = async <S extends WalletCurrency>(
   wallets: Wallet[],
 ): Promise<{ [key: WalletId]: PaymentAmount<S> } | ApplicationError> => {
   const pendingIncoming = await WalletOnChainPendingReceiveRepository().listByWalletIds({
     walletIds: wallets.map((wallet) => wallet.id),
   })
-  if (pendingIncoming instanceof CouldNotFindError) return {}
+
+  if (pendingIncoming instanceof CouldNotFindError) {
+    const settlementCurrencies = new Set(wallets.map((wallet) => wallet.currency))
```
bodymindarts, line 18: Shouldn't this validation happen before we even call `listByWalletIds`?

bodymindarts: Also if `Wallet` was parameterised I guess this would be caught by the compiler right?

PR #2753, opened 2023-06-15
fix: fix+test for account upgrade

nicolasburtey:


src/app/auth/login.ts
```diff
@@ -241,7 +241,7 @@ export const loginWithDevice = async ({
     username,
     password,
   })
-  if (res instanceof Error) throw res
+  if (res instanceof Error) return res
```
nicolasburtey, line 244: @ntheile we should not throw in the app layer (afair at least, @dolcalmi or @bodymindarts  can confirm - maybe this is my mistake did look at blame)

bodymindarts (approved):

PR #2759, opened 2023-06-15
fix: auth-router ip error handling

ntheile:


nicolasburtey (approved):

src/servers/middlewares/auth-router.ts
```diff
@@ -213,8 +213,8 @@ authRouter.post(
       const ipString = isProd ? req?.headers["x-real-ip"] : req?.ip
       const ip = parseIps(ipString)
 
-      if (ip === undefined) {
-        throw new Error("IP is not defined")
+      if (!ip) {
+        return res.status(502).send({ error: "IP is not defined" })
```
bodymindarts, line 217: this is not Bad Gateway (nginx is not acting as a gateway here). IMO `Internal Server Error` is appropriate.

PR #2770, opened 2023-06-19
chore!: upgrade otlp-agent dependencies

nicolasburtey:


src/config/process.ts
```diff
@@ -31,8 +31,6 @@ if (!!btcNetwork && !networks.includes(btcNetwork)) {
 export const BTC_NETWORK = btcNetwork as BtcNetwork
 
 export const tracingConfig = {
-  jaegerHost: process.env.JAEGER_HOST || "localhost",
```
bodymindarts, line 34: I'm sure we still have to inject the host/port of the agent - even if its no longer jaeger.

src/services/tracing.ts
```diff
@@ -215,11 +215,15 @@ class SpanProcessorWrapper extends SimpleSpanProcessor {
     super.onStart(span, parentContext)
   }
 }
+
+const otlpHost = process.env.OTLP_HOST || "localhost"
```
bodymindarts, line 219: Why have you moved this out of config?

src/config/process.ts
nicolasburtey: running into some weird issue.

```
    TypeError: Cannot read properties of undefined (reading 'tracingServiceName')

      197 |   resource: Resource.default().merge(
      198 |     new Resource({
    > 199 |       [SemanticResourceAttributes.SERVICE_NAME]: tracingConfig.tracingServiceName,
          |                                                                ^
      200 |     }),
      201 |   ),
      202 | })

      at Object.tracingServiceName (src/services/tracing.ts:199:64)
      at Object.<anonymous> (src/domain/payments/payment-flow.ts:7:18)
      at Object.<anonymous> (src/domain/payments/index.ts:6:14)
      at Object.<anonymous> (src/domain/wallets/deposit-fee-calculator.ts:5:19)
      at Object.<anonymous> (src/domain/wallets/index.ts:6:14)
      at Object.<anonymous> (src/config/yaml.ts:14:18)
      at Object.<anonymous> (src/config/process.ts:5:15)
      at Object.<anonymous> (src/config/index.ts:6:31)
      at Object.<anonymous> (src/services/mongodb/index.ts:7:17)
      at Object.require (test/jest-e2e.setup.js:2:34)
```

krtk6160:
I think the commit message needs to be changed to include a '!', not just the PR title because that's what we use to generate changelogs

nicolasburtey:
> The commit message needs to be changed to include a '!', not just the PR title because that's what we use to generate changelogs

done

dev/otel-agent-config.yaml
```diff
@@ -1,16 +1,16 @@
 receivers:
-  jaeger:
+  otlp:
     protocols:
-      thrift_compact: # on port 6831
-      thrift_binary: # on port 6832
+      http:
```
krtk6160, line 4: This is empty, does it work?

nicolasburtey: Yup I think so

nicolasburtey: Worked locally

krtk6160: Okay, just saw it's enabled by default in the otel chart so don't need to explicitly enable it in deployments.

PR #2772, opened 2023-06-20
refactor: change catch type from any to unknown

ntheile:
Attempting to change catch type from any to unknown

tsconfig.ts
```
"useUnknownInCatchVariables": false, // making this true (default setting) 
```
Screen Shot 2023-06-19 at 11 37 57 AM

### DIFF
--- a/src/app/admin/backup.ts
+++ b/src/app/admin/backup.ts
@@ -62,19 +62,13 @@ export const uploadBackup =
             logger.info({ backup }, "Static channel backup to Dropbox successful.")
             addEventToCurrentSpan("Static channel backup to Dropbox successful.")
           } catch (error) {
-            logger.error({ error }, "Static channel backup to Dropbox failed.")
-            if (error instanceof Error) {
-              recordExceptionInCurrentSpan({ error: error, level: ErrorLevel.Warn })
-            } else {
-              recordExceptionInCurrentSpan({
-                error: new Error(
-                  typeof error === "string"
-                    ? error
-                    : "Static channel backup to Dropbox failed.",
-                ),
-                level: ErrorLevel.Warn,
-              })
-            }
+            const fallbackMsg = "Static channel backup to Dropbox failed."
+            logger.error({ error }, fallbackMsg)
+            recordExceptionInCurrentSpan({
+              error: error,
+              level: ErrorLevel.Warn,
+              fallbackMsg,
+            })
           }
         },
       )
@@ -103,19 +97,13 @@ export const uploadBackup =
             logger.info({ backup }, "Static channel backup to GoogleCloud successful.")
             addEventToCurrentSpan("Static channel backup to GoogleCloud successful.")
           } catch (error) {
-            logger.error({ error }, "Static channel backup to GoogleCloud failed.")
-            if (error instanceof Error) {
-              recordExceptionInCurrentSpan({ error: error, level: ErrorLevel.Warn })
-            } else {
-              recordExceptionInCurrentSpan({
-                error: new Error(
-                  typeof error === "string"
-                    ? error
-                    : "Static channel backup to GoogleCloud failed.",
-                ),
-                level: ErrorLevel.Warn,
-              })
-            }
+            const fallbackMsg = "Static channel backup to GoogleCloud failed."
+            logger.error({ error }, fallbackMsg)
+            recordExceptionInCurrentSpan({
+              error: error,
+              level: ErrorLevel.Warn,
+              fallbackMsg,
+            })
           }
         },
       )
@@ -144,19 +132,13 @@ export const uploadBackup =
             logger.info({ backup }, "Static channel backup to Nextcloud successful.")
             addEventToCurrentSpan("Static channel backup to Nextcloud successful.")
           } catch (error) {
-            logger.error({ error }, "Static channel backup to Nextcloud failed.")
-            if (error instanceof Error) {
-              recordExceptionInCurrentSpan({ error: error, level: ErrorLevel.Warn })
-            } else {
-              recordExceptionInCurrentSpan({
-                error: new Error(
-                  typeof error === "string"
-                    ? error
-                    : "Static channel backup to Nextcloud failed.",
-                ),
-                level: ErrorLevel.Warn,
-              })
-            }
+            const fallbackMsg = "Static channel backup to Nextcloud failed."
+            logger.error({ error }, fallbackMsg)
+            recordExceptionInCurrentSpan({
+              error: error,
+              level: ErrorLevel.Warn,
+              fallbackMsg,
+            })
           }
         },
       )
--- a/src/app/admin/backup.ts
+++ b/src/app/admin/backup.ts
@@ -65,11 +65,6 @@ export const uploadBackup =
             logger.error({ error }, "Static channel backup to Dropbox failed.")
             if (error instanceof Error)
               recordExceptionInCurrentSpan({ error: error, level: ErrorLevel.Warn })
-            if (typeof error === "string")
-              recordExceptionInCurrentSpan({
-                error: new Error(error),
-                level: ErrorLevel.Warn,
-              })
           }
         },
       )
--- a/src/app/admin/backup.ts
+++ b/src/app/admin/backup.ts
@@ -63,7 +63,13 @@ export const uploadBackup =
             addEventToCurrentSpan("Static channel backup to Dropbox successful.")
           } catch (error) {
             logger.error({ error }, "Static channel backup to Dropbox failed.")
-            recordExceptionInCurrentSpan({ error: error, level: ErrorLevel.Warn })
+            if (error instanceof Error)
+              recordExceptionInCurrentSpan({ error: error, level: ErrorLevel.Warn })
+            if (typeof error === "string")
+              recordExceptionInCurrentSpan({
+                error: new Error(error),
+                level: ErrorLevel.Warn,
+              })
           }
         },
       )
@@ -93,7 +99,8 @@ export const uploadBackup =
             addEventToCurrentSpan("Static channel backup to GoogleCloud successful.")
           } catch (error) {
             logger.error({ error }, "Static channel backup to GoogleCloud failed.")
-            recordExceptionInCurrentSpan({ error: error, level: ErrorLevel.Warn })
+            if (error instanceof Error)
+              recordExceptionInCurrentSpan({ error: error, level: ErrorLevel.Warn })
           }
         },
       )
@@ -123,7 +130,8 @@ export const uploadBackup =
             addEventToCurrentSpan("Static channel backup to Nextcloud successful.")
           } catch (error) {
             logger.error({ error }, "Static channel backup to Nextcloud failed.")
-            recordExceptionInCurrentSpan({ error: error, level: ErrorLevel.Warn })
+            if (error instanceof Error)
+              recordExceptionInCurrentSpan({ error: error, level: ErrorLevel.Warn })
           }
         },
       )
--- a/src/app/admin/backup.ts
+++ b/src/app/admin/backup.ts
@@ -63,8 +63,18 @@ export const uploadBackup =
             addEventToCurrentSpan("Static channel backup to Dropbox successful.")
           } catch (error) {
             logger.error({ error }, "Static channel backup to Dropbox failed.")
-            if (error instanceof Error)
+            if (error instanceof Error) {
               recordExceptionInCurrentSpan({ error: error, level: ErrorLevel.Warn })
+            } else {
+              recordExceptionInCurrentSpan({
+                error: new Error(
+                  typeof error === "string"
+                    ? error
+                    : "Static channel backup to Dropbox failed.",
+                ),
+                level: ErrorLevel.Warn,
+              })
+            }
           }
         },
       )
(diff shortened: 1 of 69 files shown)


src/graphql/types/object/transaction.ts
```diff
@@ -60,7 +60,8 @@ const Transaction = GT.Object<WalletTransaction>({
         try {
           result = await loaders.txnMetadata.load(source.id)
         } catch (err) {
-          result = err
+          if (err instanceof Error) result = err
+          result = new Error("Unknown error")
```
nicolasburtey, line 64: this is changing the code behavior here?

ntheile: ah yes. good catch. how does this look?

```
        try {
          result = await loaders.txnMetadata.load(source.id)
        } catch (err) {
          if (err instanceof Error) {
            result = err
          } else if (typeof err === "undefined") {
            result = undefined
          } else {
            result = new Error("Unknown error")
          }
        }
```

src/app/admin/backup.ts
```diff
@@ -63,7 +63,13 @@ export const uploadBackup =
...
           } catch (error) {
             logger.error({ error }, "Static channel backup to Dropbox failed.")
-            recordExceptionInCurrentSpan({ error: error, level: ErrorLevel.Warn })
+            if (error instanceof Error)
+              recordExceptionInCurrentSpan({ error: error, level: ErrorLevel.Warn })
+            if (typeof error === "string")
+              recordExceptionInCurrentSpan({
+                error: new Error(error),
+                level: ErrorLevel.Warn,
+              })
```
nicolasburtey, line 72: do we need this branch?

src/domain/shared/safe.ts
```diff
@@ -7,6 +7,12 @@ export const safeBigInt = (num: number): bigint | BigIntConversionError => {
     if (err instanceof RangeError) {
       return new BigIntFloatConversionError(`${num}`)
     }
-    return new UnknownBigIntConversionError(err)
+    if (err instanceof Error) {
+      return new UnknownBigIntConversionError(err.message)
+    }
+    if (typeof err === "string") {
+      return new UnknownBigIntConversionError(err)
+    }
```
nicolasburtey, line 15: do we need this branch?

src/graphql/types/object/transaction.ts
```diff
@@ -60,7 +60,13 @@ const Transaction = GT.Object<WalletTransaction>({
         try {
           result = await loaders.txnMetadata.load(source.id)
         } catch (err) {
-          result = err
+          if (err instanceof Error) {
+            result = err
+          } else if (typeof err === "undefined") {
+            result = undefined
```
nicolasburtey, line 66: do we need this branch?

src/servers/middlewares/auth-router.ts
```diff
@@ -58,7 +58,11 @@ authRouter.post(
         cookies = setCookie.parse(loginResp.cookiesToSendBackToClient)
         kratosUserId = loginResp.kratosUserId
       } catch (err) {
-        recordExceptionInCurrentSpan({ error: err })
+        if (err instanceof Error) {
+          recordExceptionInCurrentSpan({ error: err })
+        } else {
+          recordExceptionInCurrentSpan({ error: { message: "Unknown error" } as Error })
```
nicolasburtey, line 64: this casting sounds awkward to me

src/servers/middlewares/auth-router.ts
```diff
@@ -215,8 +227,13 @@ authRouter.post(
           result: authToken,
         })
       } catch (err) {
-        recordExceptionInCurrentSpan({ error: err })
-        return res.status(500).send({ error: `${err.message}` })
+        if (err instanceof Error) {
+          recordExceptionInCurrentSpan({ error: err })
```
nicolasburtey, line 231:
```suggestion
          recordExceptionInCurrentSpan({ error })
```
maybe lets rename err to error so we don't need this `error: err`?

src/services/bria/index.ts
```diff
@@ -214,8 +233,16 @@ export const NewOnChainService = (): INewOnChainService => {
 
       return response.getId() as PayoutId
     } catch (err) {
-      const errMsg = typeof err === "string" ? err : err.message
-      if (err.code == status.PERMISSION_DENIED && errMsg.includes("DestinationBlocked")) {
+      let errMsg = "Unknown error"
+      if (typeof err === "string") errMsg = err
```
nicolasburtey, line 237: why do bria send a `string` error? is that a typical grpc thing?

src/app/admin/backup.ts
ntheile: you are right, axios should not return a string

src/domain/shared/safe.ts
ntheile: you are right, bigint should not return a string error

src/graphql/types/object/transaction.ts
ntheile: ah yeah probably not going to catch an undefined error

src/servers/middlewares/auth-router.ts
ntheile: any suggestions? i was just trying to be extra cautious because cookies is type `any` and I wanted to record the case that it was not an instanceof Error

src/services/bria/index.ts
ntheile: not sure. any ideas @dolcalmi? This is the simplest way i could figure out how to do this. I could not find a way to get a grpc error type.

src/app/admin/backup.ts
```diff
@@ -63,7 +63,8 @@ export const uploadBackup =
             addEventToCurrentSpan("Static channel backup to Dropbox successful.")
           } catch (error) {
             logger.error({ error }, "Static channel backup to Dropbox failed.")
-            recordExceptionInCurrentSpan({ error: error, level: ErrorLevel.Warn })
+            if (error instanceof Error)
```
bodymindarts, line 66: this is not great you are now potentially silently eating an error. We still need to be alerted of all errors.

src/debug/clean-pending-payments-for-dead-node.ts
```diff
@@ -33,7 +33,8 @@ const listAllPendingPayments = async (): Promise<
 
     return results.filter((tx) => tx.debit > 0).map((tx) => translateToLedgerTx(tx))
   } catch (err) {
-    return new UnknownLedgerError(err)
+    if (err instanceof Error) return new UnknownLedgerError(err.message)
+    return new UnknownLedgerError()
```
bodymindarts, line 37: Why not pass the err in either way? We need to be alerted / find out about the details of the error and not eat it.

bodymindarts (changes requested):
This PR - broadly speaking - is IMO a very bad approach.
It significantly reduces visibility while introducing silent failure modes and adding code.

I'm not principally against having the error in a catch block be `unknown` instead of `any` - but this is clearly not the way to deal with it.
We still must guarantee that we get alerted no matter what the type and never silently eat the error.

nicolasburtey:
> 
> This PR - broadly speaking - is IMO a very bad approach. It significantly reduces visibility while introducing silent failure modes and adding code.
> 
> I'm not principally against having the error in a catch block be `unknown` instead of `any` - but this is clearly not the way to deal with it. We still must guarantee that we get alerted no matter what the type and never silently eat the error.

do you want to make a gist on what you think would be the best way to address the `any` issue in the catch block?

src/debug/clean-pending-payments-for-dead-node.ts
ntheile: If you open this up in vscode you get a type error if you try to pass in err like `return new UnknownLedgerError(err)`. `err` is type `unknown` and `UnknownLedgerError` is expecting a `string | undefined`. One of my original ideas was to check if all errors where type `string` but i think @nicolasburtey had doubts about it. Maybe we could check like this:

```
function isString(value: unknown): value is string {
  return typeof value === "string"
}
```

and use like this:
```
return new UnknownLedgerError(isString(err) ? err : undefined)
```

src/domain/shared/errors.ts
```diff
@@ -9,8 +9,12 @@ export const RankedErrorLevel = [ErrorLevel.Info, ErrorLevel.Warn, ErrorLevel.Cr
...
-    super(message)
+  constructor(message?: string | unknown) {
+    if (typeof message === "string" || typeof message === "undefined") {
+      super(message)
+    } else {
+      super()
+    }
     this.name = this.constructor.name
     this.level = ErrorLevel.Info
   }
```
ntheile, line 20: @nicolasburtey @bodymindarts @dolcalmi ,

I updated the shared `DomainError` to allow messages of type `unknown`. Let me know if this is what you were talking about earlier today. So now in the codebase you can return UnknownXXXError like this:

```
  } catch (err) {
    if (err instanceof Error) return new UnknownLedgerError(err.message)
  }
```

Let me know if this looks good then I can update all over the codebase.

src/app/admin/backup.ts
```diff
@@ -63,7 +63,18 @@ export const uploadBackup =
             addEventToCurrentSpan("Static channel backup to Dropbox successful.")
           } catch (error) {
             logger.error({ error }, "Static channel backup to Dropbox failed.")
-            recordExceptionInCurrentSpan({ error: error, level: ErrorLevel.Warn })
+            if (error instanceof Error) {
```
bodymindarts, line 66: couldn't `recordExceptionInCurrentSpan` be changed to accept `Error | unknown` and pass `"Static channel backup to Dropbox failed."` as a third `fallbackMsg:` arg?

src/app/admin/backup.ts
```diff
@@ -93,7 +104,18 @@ export const uploadBackup =
             addEventToCurrentSpan("Static channel backup to GoogleCloud successful.")
           } catch (error) {
             logger.error({ error }, "Static channel backup to GoogleCloud failed.")
-            recordExceptionInCurrentSpan({ error: error, level: ErrorLevel.Warn })
+            if (error instanceof Error) {
```
bodymindarts, line 107: You have duplicated this logic 3 times in this file - please push it down into `recordExceptionInCurrentSpan`

src/domain/shared/errors.ts
```diff
@@ -9,8 +9,14 @@ export const RankedErrorLevel = [ErrorLevel.Info, ErrorLevel.Warn, ErrorLevel.Cr
...
   level?: ErrorLevel
-  constructor(message?: string) {
-    super(message)
+  constructor(message?: string | unknown | Error) {
+    if (typeof message === "string" || typeof message === "undefined") {
+      super(message)
+    } else if (message instanceof Error) {
+      super(message.message)
+    } else {
+      super()
```
bodymindarts, line 18: we shouldn't eat the error entirely. Pass some kind of stringified error or at least a message saying `the provided error had a type we couldn't handle`

src/servers/graphql-server.ts
```diff
@@ -276,7 +276,8 @@ export const startApolloServer = async ({
           ? reportedError
           : { message: `Error processing GraphQL request ${reportedError.code}` }
       } catch (err) {
-        throw mapError(err)
+        if (err instanceof Error) throw mapError(err)
```
bodymindarts, line 279: can't this be extended to `|| typeof err === "string"` then you can remove the ternary in the line below.

src/servers/middlewares/auth-router.ts
```diff
@@ -57,8 +57,14 @@ authRouter.post(
       try {
         cookies = setCookie.parse(loginResp.cookiesToSendBackToClient)
         kratosUserId = loginResp.kratosUserId
-      } catch (err) {
-        recordExceptionInCurrentSpan({ error: err })
+      } catch (error) {
+        if (error instanceof Error) {
+          recordExceptionInCurrentSpan({ error })
```
bodymindarts, line 62: Same logic again - should be pushed down.

src/domain/shared/errors.ts
```diff
@@ -9,8 +9,14 @@ export const RankedErrorLevel = [ErrorLevel.Info, ErrorLevel.Warn, ErrorLevel.Cr
 export class DomainError extends Error {
   name: string
   level?: ErrorLevel
-  constructor(message?: string) {
```
bodymindarts, line 12: shouldn't the fallback be present here?

src/domain/shared/primitives.ts
```diff
@@ -107,3 +107,27 @@ export const balanceAmountFromNumber = <T extends WalletCurrency>({
     currency,
   }
 }
+
+export const parseErrorMessageFromUnknown = (error: unknown): string => {
```
bodymindarts, line 111: Shouldn't this be applied at the bottom of the Error class hierarchy?

src/graphql/error-map.ts
```diff
@@ -30,10 +30,21 @@ const assertUnreachable = (x: never): never => {
   throw new Error(`This should never compile with ${x}`)
 }
 
-export const mapError = (error: ApplicationError): CustomApolloError => {
+export const mapError = (err: ApplicationError | unknown): CustomApolloError => {
```
bodymindarts, line 33: This should never be `unknown` that would be a significant bug!

src/domain/shared/primitives.ts
ntheile: I sync'd with @vindard and we created an `error-parsers` file

src/graphql/error-map.ts
ntheile: @vindard and i synced and created a new `error-parser` called `parseUnknownDomainErrorFromUnknown` which returns a `DomainError`.  So now this works in mapError with a new fallback type called `UnknownDomainError`

src/services/tracing.ts
```diff
@@ -285,7 +287,17 @@ export const recordExceptionInCurrentSpan = ({
     }
   }
 
-  recordException(span, error, level)
+  if (error instanceof Error) {
+    recordException(span, error, level)
+  } else if (fallbackMsg) {
+    recordException(span, { message: fallbackMsg }, level)
+  } else {
```
bodymindarts, line 294: Shouldn't the serialised object have higher priority than the fallback msg?

bodymindarts (dismissed):

bodymindarts (dismissed):

bodymindarts (approved):

PR #2773, opened 2023-06-20
chore: adding NetInstrumentation

nicolasburtey:
this may help to debug the issue we have on staging. but this may also increase the trace count quite a bit so will have to see if we want to keep that eventually in prod.


dolcalmi:
What issue?

nicolasburtey:
> What issue?

the 502 on staging

bodymindarts (approved):
Just FYI this adds a lot of noise if I remember correctly. I think very early on we had it in but I removed it to save on cost. Lets see if its valuable now.

PR #2774, opened 2023-06-20
ci: add circular dependency check

nicolasburtey:


ntheile:
@nicolasburtey 

I started with 17 circular dependencies and now it's down to 7. I am going to hold off on fixing the rest to avoid a difficult rebase until we finish the `catch unknown` PR https://github.com/GaloyMoney/galoy/pull/2772

src/app/accounts/add-earn.ts
```diff
@@ -18,6 +17,9 @@ import {
 } from "@services/mongoose"
 import { AccountsIpRepository } from "@services/mongoose/accounts-ips"
 
+// TODO ??
```
nicolasburtey, line 20:
```suggestion
```

Makefile
```diff
@@ -145,6 +145,7 @@ check-code: check-implicit
 	yarn eslint-check
 	yarn build
 	yarn check-yaml
+	yarn madge-check
```
nicolasburtey, line 148:
```suggestion
	# yarn madge-check
```

src/app/accounts/add-earn.ts
```diff
@@ -18,6 +17,8 @@ import {
 } from "@services/mongoose"
 import { AccountsIpRepository } from "@services/mongoose/accounts-ips"
 
+import { intraledgerPaymentSendWalletIdForBtcWallet } from "../payments/send-intraledger"
```
bodymindarts, line 20: This does not remove the circular dependency. The issue is payment -> accounts -> payments. 
By changing the import like this you are:
- importing something that we are treating as module-private
- circumventing our tracing setup

src/app/payments/send-lightning.ts
```diff
@@ -724,7 +723,7 @@ const executePaymentViaLn = async ({
     }
 
     if (!rawRoute) {
-      const reimbursed = await Wallets.reimburseFee({
+      const reimbursed = await reimburseFee({
```
bodymindarts, line 726: This breaks tracing.

src/domain/payments/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+  MAX_SATS,
+  paymentAmountFromNumber,
+  WalletCurrency,
+  BtcAmountTooLargeError,
+  UsdAmountTooLargeError,
+} from "@domain/shared"
+
+import { InvalidBtcPaymentAmountError, InvalidUsdPaymentAmountError } from "./errors"
+
+export const checkedToBtcPaymentAmount = (
```
bodymindarts, line 12: `utils` is generally speaking a very bad name as it does not help in communicating structure or domain concepts.
These functions should either be grouped according to their function (eg. `validation`) or - even better - colocated with their related domain concept (`PaymentAmount`).

src/domain/wallets/primitives.ts
```diff
@@ -1,6 +1,20 @@
...
 export const WalletType = {
   Checking: "checking",
 } as const
+
+const UuidRegex =
+  /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i
+
+export const WalletIdRegex = UuidRegex
+
+export const checkedToWalletId = (walletId: string): WalletId | ValidationError => {
```
bodymindarts, line 15: Looks like we need to find out a general pattern on where to locate the `checkedToXxx` methods. Previously. you put them in `utils` now into `primitives`. This is inconsistent. I think best is colocated to where the `Xxx` thing is defined.

src/app/payments/send-lightning.ts
nicolasburtey: it's a use case calling another use case. not sure we intend to trace those calls?
tracing from the use case are meant to be called from the graphql layer?

PR #2783, opened 2023-06-21
tests: new websocket server

nicolasburtey:


nicolasburtey:
api.blink.sv/graphql -> main graphql
ws.blink.sv/graphql -> websocket server

src/servers/ws-server.ts
```diff
@@ -45,26 +45,28 @@ const getContext = async (ctx: Context) => {
   const ipString = isProd
     ? connectionParams?.["x-real-ip"] || connectionParams?.["x-forwarded-for"]
     : connectionParams?.ip
+    ? connectionParams?.ip
+    : ctx.extra
+    ? (ctx.extra as Extra).request?.socket?.remoteAddress
```
nicolasburtey, line 50: casting is generally a bad practice. why is that needed?

src/servers/ws-server.ts
```diff
@@ -45,26 +45,28 @@ const getContext = async (ctx: Context) => {
...
-  //   const tokenPayload = {
-  //     sub: kratosCookieRes.kratosUserId,
-  //   }
-  //   return sessionContext({
-  //     tokenPayload,
-  //     ip: request?.socket?.remoteAddress,
-  //     body: null,
-  //   })
-  // }
+  const cookies = ctx.extra ? (ctx.extra as Extra).request.headers.cookie : undefined
```
nicolasburtey, line 57: casting is generally a bad practice. why is that needed?

src/servers/ws-server.ts
ntheile: yep, I was trying to get the e2e tests to pass and it kept failing with an IP address parsing error, so I committed this to see if it would pass. I cleaned up the code with a fully typed context and nullish coalescing operators.

nicolasburtey (commented):
approved (but can't approved from github because I opened the PR)

ntheile (dismissed):

ntheile (approved):

PR #2784, opened 2023-06-21
feat: update deposit fee structure

dolcalmi:
Update onchain fee receive logic according to ACE 24

### DIFF
--- a/test/unit/domain/wallets/deposit-fee-calculator.spec.ts
+++ b/test/unit/domain/wallets/deposit-fee-calculator.spec.ts
@@ -3,22 +3,54 @@ import { WalletCurrency, ZERO_SATS } from "@domain/shared"
 
 describe("DepositFeeCalculator", () => {
   const calculator = DepositFeeCalculator()
-  const amount = {
-    amount: 100n,
-    currency: WalletCurrency.Btc,
+
+  const onChainFeeConfig = {
+    ratio: 30n as DepositFeeRatioAsBasisPoints,
+    minBankFee: {
+      amount: 3000n,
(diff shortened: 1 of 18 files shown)


test/unit/domain/wallets/deposit-fee-calculator.spec.ts
```diff
@@ -3,23 +3,55 @@ import { WalletCurrency, ZERO_SATS } from "@domain/shared"
...
+      const amount = {
+        amount: 10_000_000n,
+        currency: WalletCurrency.Btc,
+      }
+      const config = { ...onChainFeeConfig, ratio: 200n as DepositFeeRatioAsBasisPoints }
+      const fee = calculator.onChainDepositFee({ ...config, amount })
+      expect(fee).toEqual({ amount: 200000n, currency: WalletCurrency.Btc })
     })
 
     it("applies a depositFeeRatio of 0", () => {
```
vindard, line 30: This test might be flawed. Is it that if the amount is less than the threshold the fee gets applied, but if the amount is above the threshold the fee is zero?

vindard: Ah I just noticed this is intended behaviour below the threshold. Maybe we should add in this same test that if the amount is below threshold that the min fee gets applied.

dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/2784/commits/605a0d719c4d13a872983e092d0bbfe2ca692cb3

vindard (approved):

PR #2786, opened 2023-06-22
feat: add fees information to global query

dolcalmi:
Merge after https://github.com/GaloyMoney/galoy/pull/2784

### DIFF
--- a/src/graphql/main/schema.graphql
+++ b/src/graphql/main/schema.graphql
@@ -283,6 +283,20 @@ type Currency {
   symbol: String!
 }
 
+type DepositFeesInformation {
+  minBankFee: String!
+
+  """
+  below this amount minBankFee will be charged
+  """
+  minBankFeeThreshold: String!
+
+  """
+  ratio to charge as basis points
+  """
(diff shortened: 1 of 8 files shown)


vindard (dismissed):
Small nits on description, otherwise looks good

bodymindarts (approved):

src/graphql/main/schema.graphql
```diff
@@ -283,6 +283,20 @@ type Currency {
...
+
+  """
+  below this amount minBankFee will be charged
+  """
+  minBankFeeThreshold: String!
+
+  """
+  ratio to charge as basis points above minBankFeeThreshold amount
+  """
+  ratio: String!
```
bodymindarts, line 297: why type `String` shouldn't this be a custom scalar?

dolcalmi: because is a bigint :thinking: was setup with that type to do the calcs with payment amounts but let me know if I need to change it to int

nicolasburtey (approved):

PR #2802, opened 2023-06-23
chore: reintroduce-onchain-fee-error

bodymindarts:
Pending fix in bria


src/app/wallets/get-on-chain-fee.ts
```diff
@@ -139,6 +139,13 @@ const getOnChainFee = async <S extends WalletCurrency, R extends WalletCurrency>
     "payOnChainByWalletId.estimatedMinerFee": `${paymentFlow.btcMinerFee}`,
   })
 
+  // avoids lnd balance sniffing attack
+  const balanceCheck = paymentFlow.checkBalanceForSend({
+    amount: balance.amount * 2n,
```
nicolasburtey, line 144: If balance is low, thing 10k sats, this check, therefore getFee might error. But sendAll will actually work. Add some constant value, ie 30k sats, will remove the issue

bodymindarts:
Fixed in bria https://github.com/GaloyMoney/bria/pull/247/

PR #2808, opened 2023-06-26
refactor: remove more circular-dependencies

bodymindarts:


src/graphql/admin/types/object/user.ts
```diff
@@ -1,29 +1,15 @@
...
-import Account from "./account"
-
 const User = GT.Object<IdentityPhone>({
   name: "User",
 
   fields: () => ({
     id: { type: GT.NonNullID },
     phone: { type: GT.NonNull(Phone) },
     language: { type: GT.NonNull(Language) },
-    defaultAccount: {
```
nicolasburtey, line 17: is it not used?

bodymindarts: Admin panel never resolves from user -> account. Only account -> user.
Checked here https://github.com/GaloyMoney/admin-panel/tree/main/graphql

nicolasburtey (approved):

PR #2812, opened 2023-06-26
chore!: upgrade otlp-agent dependencies

nicolasburtey:
This reverts commit 400888f188893d3d2a3bb8a04a8ae2faecd495c5.

- use default variable environement now instead of trying to redefining it (work for the url, not for service name for some reasons)

documentation for the url endpoint: https://opentelemetry.io/docs/concepts/sdk-configuration/otlp-exporter-configuration/

- replacing @opentelemetry/exporter-trace-otlp-proto with @opentelemetry/exporter-trace-otlp-http


krtk6160 (dismissed):

test/unit/app/prices/get-current-price.spec.ts
```diff
@@ -1,25 +1,30 @@
...
 import { CacheKeys } from "@domain/cache"
 import { DisplayCurrency } from "@domain/fiat"
 import { PriceNotAvailableError } from "@domain/price"
 
 import * as PriceServiceImpl from "@services/price"
-import { LocalCacheService } from "@services/cache"
-
-jest.mock("@services/redis", () => ({}))
+import { LocalCacheService } from "@services/cache/local"
+import { getCurrentSatPrice, getCurrentUsdCentPrice } from "@app/prices"
```
dolcalmi, line 7: this is not a good change... I tested locally and the only required change is `import { LocalCacheService } from "@services/cache/local"`

test/unit/app/prices/list-currencies.spec.ts
dolcalmi, line 1: same here, only required change is `import { LocalCacheService } from "@services/cache/local"`

test/unit/services/lnd/onchain-service.spec.ts
```diff
@@ -5,8 +5,6 @@ import {
   extractOutgoingTransactions,
 } from "@services/lnd/onchain-service"
 
-jest.mock("@services/redis", () => ({}))
```
dolcalmi, line 8: did you test this with detectOpenHandles?

src/services/tracing.ts
```diff
@@ -237,21 +231,13 @@ class SpanProcessorWrapper extends SimpleSpanProcessor {
...
-provider.addSpanProcessor(
-  new SpanProcessorWrapper(
-    new JaegerExporter({
-      host: tracingConfig?.jaegerHost,
-      port: tracingConfig?.jaegerPort,
-    }),
-  ),
-)
+
+provider.addSpanProcessor(new SpanProcessorWrapper(new OTLPTraceExporter()))
```
dolcalmi, line 235: so.. are we trusting in some hidden env variable instead of constructor config? also is intentional to remove config dependency?

dolcalmi (dismissed):

test/unit/services/lnd/onchain-service.spec.ts
nicolasburtey: yup, it was connecting to redis.

src/services/tracing.ts
nicolasburtey: I guess it's documented here? https://github.com/GaloyMoney/galoy/pull/2812/files#diff-d33e979799a45c7c51752e9c8d96a3e452015d1a40b1e4b6ec6a98e92c4d8430R99

nicolasburtey: it seems to be the recommended way to set up the opel collector https://opentelemetry.io/docs/concepts/sdk-configuration/otlp-exporter-configuration/

test/unit/app/prices/list-currencies.spec.ts
nicolasburtey: ok I can try to revert some of the change but still only import local cache in those scenarios so that redis get not initialized when only using local cache

src/services/tracing.ts
bodymindarts: I would prefer having explicit code that shows what env vars are being read / injected.
Configuring libraries via env vars that they read out without us explicitly injecting them is really obscure. Isn't there a way to pass the config into the constructor?

bodymindarts: Constructor accepts this config:
```
export interface OTLPExporterConfigBase {
    headers?: Partial<Record<string, unknown>>;
    hostname?: string;
    url?: string;
    concurrencyLimit?: number;
    /** Maximum time the OTLP exporter will wait for each batch export.
     * The default value is 10000ms. */
    timeoutMillis?: number;
}
```

bodymindarts (approved):

PR #2822, opened 2023-06-27
chore(deps): bump bria-api proto

galoybot:


bodymindarts:
Included in https://github.com/GaloyMoney/galoy/pull/2821

PR #2832, opened 2023-06-28
chore: switch from deprecated twilio function

vindard:


src/services/twilio.ts
```diff
@@ -72,7 +72,9 @@ export const TwilioClient = (): IPhoneProviderService => {
 
   const getCarrier = async (phone: PhoneNumber) => {
     try {
-      const result = await client.lookups.phoneNumbers(phone).fetch({ type: ["carrier"] })
+      const result = await client.lookups.v1
```
dolcalmi, line 75: did we get a deprecation warning? I couldn't find it :thinking:

vindard: I did when I removed the twilio mock in some tests

bodymindarts (approved):

PR #2843, opened 2023-06-29
chore!: moving PROXY_CHECK_APIKEY to process.env

nicolasburtey:


src/config/process.ts
```diff
@@ -167,3 +167,5 @@ export const mongodbCredentials = () => {
 export const feedback = {
   mattermostWebhookUrl: process.env.MATTERMOST_WEBHOOK_URL,
 }
+
+export const IP_FETCHER_KEY = process.env.IP_FETCHER_KEY
```
dolcalmi, line 171: please use same notation of other consts

dolcalmi (commented):
only a minor change but requires rebase

PR #2852, opened 2023-06-30
refactor: moving ip to its own collection for better performance on a…

nicolasburtey:
…ccount collection

### DIFF
--- a/src/domain/accounts-ips/index.types.d.ts
+++ b/src/domain/accounts-ips/index.types.d.ts
@@ -1,11 +1,13 @@
-type AccountIPs = {
-  readonly id: AccountId
-  lastIPs: IPType[]
+type AccountIP = {
+  accountId: AccountId
+  ip: IpAddress
+  metadata: IPType
 }
 
-type AccountOptIps = {
-  readonly id: AccountId
-  lastIPs?: IPType[]
+type AccountIPNew = {
+  accountId: AccountId
+  ip: IpAddress
+  metadata?: IPType
 }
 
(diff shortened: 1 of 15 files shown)


src/services/mongoose/accounts-ips.ts
```diff
@@ -40,33 +46,54 @@ export const AccountsIpRepository = (): IAccountsIPsRepository => {
...
     }
   }
 
+  const findLastByAccountId = async (
+    accountId: AccountId,
+  ): Promise<AccountIP | RepositoryError> => {
+    try {
+      const result = await AccountIps.findOne({
+        _accountId: toObjectId<AccountId>(accountId),
+      }).sort({ lastConnection: -1 })
```
dolcalmi, line 74: if we will find only by account  maybe we should add an additional index with lastConnection

src/domain/accounts-ips/index.types.d.ts
```diff
@@ -19,7 +21,15 @@ type IPMetadataValidator = {
   validateForReward(ipMetadata?: IPType): true | ValidationError
 }
 
+type FindByAccountIdAndIpInput = {
```
dolcalmi, line 24: please use Args suffix like in other methods

src/services/mongoose/accounts-ips.ts
nicolasburtey: have been thinking about it. it's currently only used by earn so don't think that's necessary

dolcalmi (approved):

PR #2857, opened 2023-07-01
chore: request change renaming

nicolasburtey:
first PR in preparation for reviewing https://github.com/GaloyMoney/galoy/pull/2713 (aiming at reducing the diff)


bodymindarts (approved):
Looks fine - need to check if we override this in prod. If we do we'll need a PR there otherwise if we're on defaults it can just pass through.

PR #2862, opened 2023-07-03
test: add send lightning e2e tests

vindard:


vindard:
This is now covered in bats testing

PR #2865, opened 2023-07-03
test: explore adding bats e2e setup

vindard:


test/bats/gql/transactions.gql
```diff
@@ -0,0 +1,63 @@
+query transactions($walletIds: [WalletId], $first: Int, $after: String) {
```
nicolasburtey, line 1: we generally use a pattern to have a single `$input` as argument for queries. I think it's good to use that consistently across our different apps

test/bats/helpers.bash
```diff
@@ -0,0 +1,129 @@
...
+# node and then moving on with commands that exercise it for the
+# test.
+#
+# Ensures that BATS' handling of file handles is taken into account;
+# see
+# https://github.com/bats-core/bats-core#printing-to-the-terminal
+# https://github.com/sstephenson/bats/issues/80#issuecomment-174101686
+# for details.
+background() {
+  "$@" > /dev/null 2>&1 &
```
bodymindarts, line 106: You should not change this code - you removed the option to be able to get the output of the background tasks.

bodymindarts (approved):

PR #2870, opened 2023-07-05
feat: websocket server add open telemetry

ntheile:


src/servers/ws-server.ts
```diff
@@ -42,151 +55,189 @@ const getContext = async (
...
+
+            const operationAST = getOperationAST(args.document, args.operationName)
+            if (!operationAST) {
+              // returning `GraphQLError[]` sends an `ErrorMessage` and stops the subscription
+              recordExceptionInCurrentSpan({
+                error: new GraphQLError("Unable to identify operation"),
+                level: ErrorLevel.Warn,
+              })
+              return [new GraphQLError("Unable to identify operation")]
+            }
```
ntheile, line 146: @nicolasburtey ,

What do you recommend as the default error level for these?

nicolasburtey: `warn` sounds good

src/servers/ws-server.ts
```diff
@@ -42,151 +55,189 @@ const getContext = async (
...
+
+            // handle mutation and query requests
```
nicolasburtey, line 148:
```suggestion
            // decline mutation and query requests
```

src/servers/ws-server.ts
```diff
@@ -42,151 +55,189 @@ const getContext = async (
...
+              addAttributesToCurrentSpan({ "ws.authType": "bearer" })
+              return authorizedContexts[ctx.connectionParams?.Authorization]
+            }
+            // anon context
+            addAttributesToCurrentSpan({ "ws.authType": "anon" })
+            const context = await getContext(ctx)
+            return context
+          },
+          onNext: (ctx, msg, args, result) => {
+            baseLogger.debug("Next", { ctx, msg, args, result })
```
nicolasburtey, line 224: this is probably where we should have some span events been register to honeycomb (in addition or replacement of the logger) if we want to have more meaning opentelemetry insights.

nicolasburtey (commented):
few comments, looks good otherwise

nicolasburtey (approved):

PR #2873, opened 2023-07-09
chore: deleted phone into an array

nicolasburtey:


src/domain/users/index.types.d.ts
```diff
@@ -52,6 +52,7 @@ type User = {
   deviceTokens: DeviceToken[]
   phoneMetadata: PhoneMetadata | undefined
   phone?: PhoneNumber | undefined
+  deletedPhone?: PhoneNumber[]
```
bodymindarts, line 55: shouldn't this be plural? `deletedPhones`?

src/migrations/20230709103710-deleted-phone-to-array.ts
```diff
@@ -0,0 +1,28 @@
+module.exports = {
+  async up(db) {
+    console.log("Begin migration for deletedPhone")
+
+    const users = await db.collection("users").find({}).toArray()
+    for (const user of users) {
+      if (user.deletedPhone) {
+        // If deletedPhone exists and it's not an array, convert it to an array
+        if (!Array.isArray(user.deletedPhone)) {
```
bodymindarts, line 9: Does this have to be done pro-actively? Can't it be done lazily as users update their phone numbers?

src/services/mongoose/schema.ts
```diff
@@ -348,7 +348,9 @@ const UserSchema = new Schema(
       unique: true,
       sparse: true,
     },
-    deletedPhone: { type: String },
+    deletedPhone: {
```
bodymindarts, line 351: I would prefer an immutable refactor rather than an update in place. Keep this field and add a new one `deletedPhones`.

nicolasburtey: it's to simplify the state/code. more code/code path to maintain with your approach versus a "clean" upgrade. also simplify the data pipeline down the line.

src/migrations/20230709103710-deleted-phone-to-array.ts
nicolasburtey: doesn't have to, but it make things less complex going forward. https://github.com/GaloyMoney/galoy/pull/2873#discussion_r1257447221

src/domain/users/index.types.d.ts
nicolasburtey: good point I'll update

nicolasburtey: good point I'll update

src/services/mongoose/schema.ts
nicolasburtey: it also aims at making this PR more digestable: https://github.com/GaloyMoney/galoy/pull/2713 (ready to be review/merged btw). if we need to think about immutable refactor on top of it, then it's even more context to keep in your head

bodymindarts (dismissed):

bodymindarts (approved):

PR #2877, opened 2023-07-10
test: expand onchain send bats tests

vindard:
## Description

This PR adds the remaining onchain-send mutations to the new bats test suite and removes them from our prior e2e suite.


test/bats/onchain-payments.bats
```diff
@@ -19,57 +25,276 @@ teardown() {
   [[ "$(balance_for_check)" = 0 ]] || exit 1
 }
 
-@test "onchain payments: setup user" {
+@test "onchain payments: settle trade intraccount" {
```
vindard, line 28: The "trade intra-account" tests may potentially be redundant

PR #2885, opened 2023-07-11
chore: remove language custom resolver

nicolasburtey:


src/graphql/types/object/user.ts
dolcalmi, line 1: this object is used too in admin queries... did you test it in that cases? that is the reason to have a custom resolver

dolcalmi (changes requested):

src/graphql/types/object/user.ts
nicolasburtey: I did not test this. that make sense but that also seems kind of an hacky/hidden behavior. will think about it a bit more.

dolcalmi (approved):

PR #2893, opened 2023-07-12
chore: add revert tx debug script

krtk6160:


src/debug/revert-tx.ts
```diff
@@ -0,0 +1,38 @@
...
+ *	. ./.envrc && yarn ts-node \
+ *		--files \
+ *			-r tsconfig-paths/register \
+ *			-r src/services/tracing.ts \
+ *		src/debug/revert-tx.ts <txId>
+ */
+
+import { isUp } from "@services/lnd/health"
+import { lndsConnect } from "@services/lnd/auth"
+
```
vindard, line 12:
```suggestion
```

src/debug/revert-tx.ts
```diff
@@ -0,0 +1,38 @@
...
+  const journalId = tx._journal.toString()
+
+  await MainBook.void(journalId, "Buggy payment")
+
+  return true
+}
+
+setupMongoConnection(false)
+  .then(async (mongoose) => {
+    await Promise.all(lndsConnect.map((lndParams) => isUp(lndParams)))
```
vindard, line 34:
```suggestion
```

vindard (changes requested):
Tested locally, looks good to me for removing intraledger transactions. I don't think we'd need the lnd-related methods since we're only doing ledger operations here.

Do we plan to use this script for any transaction types other than intraledger ones?

dolcalmi:
ummm this script is too dangerous... we need to have a script for specific cases with logic that guarantees at least basic business rules like `src/debug/void-payment.ts`

dolcalmi:
we can reopen it if it is necessary

PR #2905, opened 2023-07-17
chore: make it possible to select which country can onboard

nicolasburtey:
I just verified the value were passed correctly from `transient_payload`, but I have not tests the e2e flow. I intent to test manually on staging, but let me know if you think a local e2e test is mandatory.


src/app/authentication/login.ts
```diff
@@ -72,7 +91,51 @@ export const loginWithPhoneToken = async ({
...
+
+    if (accountConfig.enablePhoneCheck) {
+      const newPhoneMetadata = await TwilioClient().getCarrier(phone)
+
+      if (newPhoneMetadata instanceof Error) {
+        return new InvalidPhoneMetadataForOnboardingError()
+      }
+
+      const validatedPhoneMetadata =
+        PhoneMetadataValidator(accountConfig).validateForReward(phoneMetadata)
```
vindard, line 126: Is this still `validateForReward` or should it be `validateForOnboarding` too?

src/servers/authorization/kratos-router.ts
```diff
@@ -82,9 +82,12 @@ kratosRouter.post(
           res.status(400).send("invalid phone")
           return
         }
+        console.log({ transient_payload }, "transient_payload callback kratos router")
```
vindard, line 85: `baseLogger` instead?

vindard (commented):
Just a few surface-level comments so far

vindard (changes requested):
This integration test is legitimately failing https://github.com/GaloyMoney/galoy/actions/runs/5587284608/jobs/10221643429

src/domain/accounts-ips/ip-metadata-validator.ts
```diff
@@ -11,7 +11,7 @@ export const IPMetadataValidator = ({
   denyASNs,
   allowASNs,
 }: IPMetadataValidatorArgs): IPMetadataValidator => {
-  const validateForReward = (ipMetadata?: IPType): true | ValidationError => {
+  const validate = (ipMetadata?: IPType): true | ValidationError => {
```
bodymindarts, line 14: I think we are conflating 2 issues here. Validation should only be concerned with the shape and bounds of data - does the input conform to our business rules ie. can it be represented as a domain type.

Wether or not to allow a login given a (valid) IP address is not validation. Is a different flavour of concern.

src/app/authentication/login.ts
```diff
@@ -212,6 +275,48 @@ export const loginDeviceUpgradeWithPhone = async ({
     // a. create kratos account
     // b. and c. migrate account/user collection in mongo via kratos/registration webhook
 
+    // check if account is upgradeable
+    const accountConfig = getAccountCountries()
+
+    if (accountConfig.enableIpCheck) {
```
bodymindarts, line 281: I think these 2 conditionals `enableIpCheck` and `enablePhoneCheck` should not be present in the application layer. It's very procedural ins style.

Better would be to encapsulate the `isAllowedToLogin` logic in a (unit-tested) domain object that gets executed unconditionally.

src/domain/accounts-ips/ip-metadata-validator.ts
nicolasburtey: are you saying the word `validate` is the issue here? because it's been used differently in other context in our app?

src/app/authentication/login.ts
nicolasburtey: I refactored the code https://github.com/GaloyMoney/galoy/pull/2905/commits/1110713ff1c433dfdb0754583109072b12840450 but still kept it at the app level. the thing is this code is mostly dealing with network/services call here, so it's unclear to me how to best handle it if we were to move it into domain.

PR #2911, opened 2023-07-18
test: move subscriber server file

vindard:
## Description

@dolcalmi brought up that `src/debug` is a potentially disposable folder meant for executing scripts in a debug pod, and probably isn't a good dependency folder for tests. This is to propose a new folder where the gql-subscriber server can live


dolcalmi (dismissed):

test/bats/helpers/setup-and-teardown.bash
```diff
@@ -31,7 +31,7 @@ start_server() {
 }
 
 subscribe_to() {
-  background ${REPO_ROOT}/node_modules/.bin/ts-node "${REPO_ROOT}/src/debug/gqlsubscribe.ts" "ws://${GALOY_ENDPOINT}/graphqlws" "$(gql_file $1)" > .e2e-subscriber.log
+  background ${REPO_ROOT}/node_modules/.bin/ts-node "${REPO_ROOT}/test/helpers/servers/gql-subscribe.ts" "ws://${GALOY_ENDPOINT}/graphqlws" "$(gql_file $1)" > .e2e-subscriber.log
```
bodymindarts, line 34: I don't understand the reasoning behind this path.

vindard: Do you mean aside from the explanation in the PR description?

dolcalmi: Reason is in the main description... or do you mean something else?

nicolasburtey (dismissed):

dolcalmi (approved):

PR #2931, opened 2023-07-24
fix: add onchain receive record for dust txs

dolcalmi:
- [ ] Unit tests
- [x] integration tests
- [ ] Add max validation to threshold too


dolcalmi:
closed because requires changes in ledger implementation (usd and fee txs with 0 amounts)

PR #2935, opened 2023-07-24
ci: remove e2e github action

bodymindarts:
bats is taking over the responsibility of all e2e tests. Also it still runs in concourse and so we will see something break there. For now its not worth the effort for debugging its flakyness in github actions as we are moving away from this anyway


vindard (approved):
Fair since we still have coverage in concourse for this

PR #2939, opened 2023-07-24
test: add user-login tests

vindard:


test/bats/user-login.bats
```diff
@@ -0,0 +1,43 @@
...
+  clear_cache
+
+  start_server
+}
+
+teardown_file() {
+  stop_server
+}
+
+@test "user-login: login user" {
```
dolcalmi, line 15: shouldn't this be a helper? (or at least part of it)

dolcalmi (changes requested):

test/bats/user-login.bats
vindard: We already have a helper for it. This is to explicitly test it outside the helper in an isolated way.

Not sure if this approach is repetitive/redundant? I like that we're explicitly asserting this as a test condition.

dolcalmi: I think is not necessary.. if login is not working the other tests will fail but if you want to be explicit you can include tests for no happy paths

bodymindarts: Agree - let's not add this. It just increases the e2e runtime but doesn't assert anything unique.

PR #2940, opened 2023-07-24
feat: poc for apollo-router integation into bats setup

bodymindarts:


dev/apollo-federation/supergraph.graphql
```diff
@@ -0,0 +1,1828 @@
+schema
```
nicolasburtey, line 1: some README or pointer would be useful. I'm not too sure how to interpret this file right now without more context

docker-compose.yml
```diff
@@ -53,6 +55,13 @@ services:
       - lnd-outside-3
       - otel-agent
     restart: on-failure:10
+  apollo-router:
+    image: ghcr.io/apollographql/router:v1.25.0
+    environment:
+      - APOLLO_ROUTER_SUPERGRAPH_PATH=/repo/dev/apollo-federation/supergraph.graphql
```
nicolasburtey, line 61: I think we don't typically hard code stuff in here? instead we use env variable?

dev/apollo-federation/supergraph-config.yaml
```diff
@@ -0,0 +1,6 @@
+federation_version: =2.3.2
+subgraphs:
+  main:
```
nicolasburtey, line 3:
```suggestion
  public:
```

nicolasburtey: we want to converge to `public` instead of `main` I think.

docker-compose.yml
bodymindarts: we do actually hard code paths to config.
See Otel agent for example:
```
  otel-agent:
    image: otel/opentelemetry-collector-contrib:0.79.0
    command: ["--config=/etc/otel-agent-config.yaml"]
    environment:
      - HONEYCOMB_DATASET=${HONEYCOMB_DATASET}
      - HONEYCOMB_API_KEY=${HONEYCOMB_API_KEY}
    volumes:
      - ./dev/otel-agent-config.yaml:/etc/otel-agent-config.yaml
```

dev/apollo-federation/supergraph-config.yaml
bodymindarts: done https://github.com/GaloyMoney/galoy/pull/2940/commits/aeed158af97cbea43267dd6bb6a5ab123a0caaea

dev/apollo-federation/supergraph.graphql
bodymindarts: it gets generated in the `yarn write-sdl` task. Will think about where to document it.

bodymindarts: Added a readme under `dev/apollo-federation`

nicolasburtey (approved):

PR #2941, opened 2023-07-24
test: add transactions bats tests

vindard:


test/integration/app/accounts/get-transactions-for-accounts.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    const txns = await Accounts.getTransactionsForAccountByWalletIds({
+      account: senderAccount,
+      walletIds: [otherWalletDescriptor.id],
+    })
+    if (txns instanceof Error) throw txns
+    expect(txns.error).toBeInstanceOf(InvalidWalletId)
+
+    // Restore system state
+    Transaction.deleteMany({ memo })
```
dolcalmi, line 82: should we use await? in faster machine could create random issues

test/bats/helpers/transactions.bash
```diff
@@ -0,0 +1,16 @@
+count_transactions_by_currency() {
+  transactions_array_query=$1
+  currency=$2
+
+  echo $output \
+    | jq -r "$transactions_array_query" \
+    | jq -r --arg currency "$currency" 'select(.node.settlementCurrency == $currency)' \
```
dolcalmi, line 7: does not work just with ` | length`

test/bats/transactions.bats
```diff
@@ -0,0 +1,93 @@
...
+  wallet_1_currency="$(currency_for_wallet 1)"
+  [[ "$wallet_1_currency" == "USD" ]] || exit 1
+
+  wallet_1_transactions_query='.data.me.defaultAccount.wallets[1].transactions.edges[]'
+
+  wallet_1_usd_count="$(count_transactions_by_currency \
+    $wallet_1_transactions_query \
+    'USD' \
+  )"
+  [[ "$wallet_1_usd_count" -gt "0" ]] || exit 1
```
dolcalmi, line 92: in the previous test schema and some data were validated... dont we want to do that? any specific reason?

dolcalmi (changes requested):

test/bats/transactions.bats
vindard: Yup, the approach I've been using so far is to be very minimal in the bats (e2e) tests to only ensure that all the elements of an api call are connected up properly.

I'm trying to push any validations or business logic behaviour checks into unit tests, service integration tests and "early return" use-case integration tests.

I've tried to capture all that we tested before in the new integration tests I added here. The only thing missing I think is one error validation check that will come in early-return integration tests:
- `it("returns an error if non-owned walletId is included")`

But let me know if you can identify any other checks that I may have missed including here as well.

test/bats/helpers/transactions.bash
vindard: I did it like this for easier readability. Can go either way if we really want to change it.

bodymindarts (approved):

PR #2942, opened 2023-07-24
test: auth e2e test to bats

nicolasburtey:
- first pass on bats test for auth
- removing the min internal for identifier. it has not proven to be useful historically, so safe to remove, specially as it is interfering with our test.


test/bats/auth.bats
```diff
@@ -0,0 +1,230 @@
...
+  exec_graphql 'charlie' 'user-email-registration-validate' "$variables"
+  [[ "$(graphql_output '.data.userEmailRegistrationValidate.me.email.address')" == "$email" ]] || exit 1
+  [[ "$(graphql_output '.data.userEmailRegistrationValidate.me.email.verified')" == "true" ]] || exit 1
+}
+
+@test "auth: log in with email" {
+  email=$(read_value "email")
+
+  # code request
+  curl_request "http://${OATHKEEPER_HOST}:${OATHKEEPER_PORT}/auth/email/code" "{ \"email\": \"$email\" }"
```
bodymindarts, line 99: I believe we are using `GALOY_ENDPOINT` which is pointing to oatheeper in the other tests.

nicolasburtey: updated

bodymindarts (approved):

PR #2943, opened 2023-07-25
feat: email cookie login

ntheile:
Enables email cookie login at this new rest endpoint `/email/login/cookie`

```
const session = await ajax.post(config.galoyAuthEndpoint + "/email/login/cookie", {
  emailLoginId,
  code,
})
```

TODO
- [x] chart and config for new oathkeeper rule https://github.com/GaloyMoney/charts/pull/3902

### DIFF
--- a/src/app/authentication/login.ts
+++ b/src/app/authentication/login.ts
@@ -157,42 +157,8 @@ export const loginWithEmail = async ({
     if (limitOk instanceof Error) return limitOk
   }
 
-  const code = checkedToEmailCode(codeRaw)
-  if (code instanceof Error) return code
-
-  const authServiceEmail = AuthWithEmailPasswordlessService()
-
-  const validateCodeRes = await authServiceEmail.validateCode({
-    code,
-    emailFlowId,
-  })
-  if (validateCodeRes instanceof Error) return validateCodeRes
-
-  const email = validateCodeRes.email
-  const totpRequired = validateCodeRes.totpRequired
-
-  const isEmailVerified = await authServiceEmail.isEmailVerified({ email })
-  if (isEmailVerified instanceof Error) return isEmailVerified
-  if (isEmailVerified === false) return new EmailUnverifiedError()
-
-  await rewardFailedLoginAttemptPerIpLimits(ip)
-
-  const res = await authServiceEmail.loginToken({ email })
-  if (res instanceof Error) throw res
-  return { authToken: res.authToken, totpRequired }
-}
-
-export const loginWithEmailCookie = async ({
-  emailFlowId,
-  code: codeRaw,
-  ip,
-}: {
-  emailFlowId: EmailFlowId
-  code: EmailCode
-  ip: IpAddress
-}): Promise<LoginWithEmailCookieResult | ApplicationError> => {
   {
-    const limitOk = await checkFailedLoginAttemptPerIpLimits(ip)
+    const limitOk = await checkFailedLoginAttemptPerLoginIdentifierLimits(emailFlowId)
     if (limitOk instanceof Error) return limitOk
   }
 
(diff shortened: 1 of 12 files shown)


src/app/authentication/login.ts
```diff
@@ -182,6 +182,45 @@ export const loginWithEmail = async ({
...
+}: {
+  emailFlowId: EmailFlowId
+  code: EmailCode
+  ip: IpAddress
+}): Promise<LoginWithEmailCookieResult | ApplicationError> => {
+  {
+    const limitOk = await checkFailedLoginAttemptPerIpLimits(ip)
+    if (limitOk instanceof Error) return limitOk
+  }
+
```
nicolasburtey, line 198: we need to rate limit by email/identifier also

src/domain/authentication/cookie.ts
```diff
@@ -0,0 +1,11 @@
+export interface ICookieOptions {
```
nicolasburtey, line 1: everything optional is not good practice. is that intentional? or just an oversight?

src/servers/authorization/email.ts
```diff
@@ -206,4 +218,97 @@ authRouter.post(
...
+  wrapAsyncToRunInSpan({
+    namespace: "servers.middlewares.authRouter",
+    fnName: "emailLoginCookie",
+    fn: async (req: express.Request, res: express.Response) => {
+      baseLogger.info("/email/login/cookie")
+
+      const ipString = isProd ? req?.headers["x-real-ip"] : req?.ip
+      const ip = parseIps(ipString)
+
+      if (!ip) {
```
nicolasburtey, line 232: let's add `recordExceptionInCurrentSpan()` for all the error case. that can be useful to debug. (like I start to do in this PR https://github.com/GaloyMoney/galoy/pull/2942/files)

src/servers/authorization/email.ts
```diff
@@ -206,4 +218,97 @@ authRouter.post(
...
+          kratosSessionCookie.name,
+          kratosSessionCookie.value,
+          kratosCookies.formatCookieOptions(kratosSessionCookie),
+        )
+        res.cookie(
+          csrfCookie.name,
+          csrfCookie.value,
+          kratosCookies.formatCookieOptions(csrfCookie),
+        )
+      } catch (error) {
```
nicolasburtey, line 299: try block seems overly broad here

src/services/kratos/auth-email-no-password.ts
```diff
@@ -231,6 +232,35 @@ export const AuthWithEmailPasswordlessService = (): IAuthWithEmailPasswordlessSe
...
+        updateLoginFlowBody: {
+          identifier,
+          method,
+          password,
+          csrf_token: cookieFlow.csrf,
+        },
+      })
+      const cookiesToSendBackToClient: Array<SessionCookie> = result.headers["set-cookie"]
+      // note: this only works when whoami: required_aal = aal1
+      const kratosUserId = result.data.session.identity.id as UserId
```
nicolasburtey, line 257:
```suggestion
      const kratosUserId = result.data.session.identity.id as UserId | undefined
```

src/services/kratos/auth-email-no-password.ts
```diff
@@ -231,6 +232,35 @@ export const AuthWithEmailPasswordlessService = (): IAuthWithEmailPasswordlessSe
...
+        cookie: cookieFlow.cookie,
+        updateLoginFlowBody: {
+          identifier,
+          method,
+          password,
+          csrf_token: cookieFlow.csrf,
+        },
+      })
+      const cookiesToSendBackToClient: Array<SessionCookie> = result.headers["set-cookie"]
+      // note: this only works when whoami: required_aal = aal1
```
nicolasburtey, line 256:
```suggestion
// identity is only defined when identity has not enabled totp
```

comment is/was inacurate.

src/app/authentication/login.ts
ntheile: is there code for this? i only see rate limiters for phone or IP

src/domain/authentication/cookie.ts
ntheile: yeah all the fields for Cookie Options are optional. You can look at how Express implements CookieOptions here https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/express-serve-static-core/index.d.ts#L350-L360.

I did not want to couple the kratos service with express so i just created a common domain model for CookieOptions

src/app/authentication/login.ts
nicolasburtey: there is no longer rate limit for phone. it's now `identifier` to cover either phone or email.

nicolasburtey:
also: we need tests.

ntheile:
> also: we need tests.

not quite sure how to test cookie authentication, i think it requires a browser to work

src/app/authentication/login.ts
ntheile: not sure if it applies to this function since we only have access to the emailFlowId and not the actual email address. also I noticed in the `ratelimits.ts` it was still requiring a PhoneNumber in the functions params instead of LoginIdentifer ...fixed it here https://github.com/GaloyMoney/galoy/pull/2943/commits/028636339183081338f7c5fe1dd1cd37925f717a

src/servers/authorization/email.ts
ntheile: fixed it and made it smaller

src/app/authentication/login.ts
nicolasburtey: oh right, we should have the limit on the request code instead.

nicolasburtey (approved):

PR #2952, opened 2023-07-26
fix: apollo router config for web apps

ntheile:
This enables cors via the apollo router docs here (to make admin/web wallet/pay work): https://www.apollographql.com/docs/router/configuration/cors 

It also adds a rhai script to make sure the cookie response headers are passed back (only request header propagate via yaml, a rhai script is needed for response headers). docs here https://www.apollographql.com/docs/router/configuration/header-propagation/#response-header-propagation

### DIFF
--- a/dev/apollo-federation/router.yaml
+++ b/dev/apollo-federation/router.yaml
@@ -7,8 +7,20 @@ supergraph:
   listen: 0.0.0.0:4004
   introspection: true
   path: /graphql
+cors:
+  match_origins:
+    - "^http://([a-z0-9]+[.])*" # http
+    - "^https://([a-z0-9]+[.])*" # https
+  allow_credentials: true # for cookie auth
 telemetry:
   tracing:
     otlp:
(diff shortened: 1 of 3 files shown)


dev/apollo-federation/router.yaml
```diff
@@ -7,8 +7,20 @@ supergraph:
   listen: 0.0.0.0:4004
   introspection: true
   path: /graphql
+cors:
+  match_origins:
+    - "^http://([a-z0-9]+[.])*" # http
+    - "^https://([a-z0-9]+[.])*" # https
+  allow_credentials: true # for cookie auth
```
nicolasburtey, line 14: suggested value for prod:

cors:
  origins:
    - https://studio.apollographql.com # Keep this so Apollo Studio can still run queries against your router
  match_origins:
    - "^https://([a-z0-9]+[.])*blink[.]sv$"
  allow_credentials: true # for cookie auth

PR #2962, opened 2023-07-28
chore: attempt at removing some overloading of test accounts

nicolasburtey:


bodymindarts (commented):
Is this breaking? How do we migrate test_accounts provisioned in environments

nicolasburtey:
> Is this breaking? How do we migrate test_accounts provisioned in environments

have a bit more work to do. should not be breaking but will see.

test/legacy-integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -387,7 +417,7 @@ describe("UserWallet - Lightning Pay", () => {
     expect(userBTxn.slice.filter(matchTx)[0].settlementVia.type).toBe("intraledger")
   })
 
-  it("sends to another Galoy user a push payment", async () => {
+  it.skip("sends to another Galoy user a push payment", async () => {
```
nicolasburtey, line 420: @dolcalmi @vindard could you have a look at this one? there might be a bug, not too sure. I know there has been reported that username are not always added properly.

now that we have "cleaner" test because we use new users for each tests, we may have uncover a bug. otherwise it may just be that my refactor break the tests but it's unclear. haven't dig a lot into it, I don't know that part of the code too well now.

test/legacy-integration/02-user-wallet/02-send-lightning.spec.ts
```diff
@@ -762,7 +792,7 @@ describe("UserWallet - Lightning Pay", () => {
     expect(txn.memo).toBe(memoFromUser)
   })
 
-  it("filters spam from send to another Galoy user as push payment", async () => {
+  it.skip("filters spam from send to another Galoy user as push payment", async () => {
```
nicolasburtey, line 795: @dolcalmi @vindard could you have a look at this one? there might be a bug, not too sure. I know there has been reported that username are not always added properly.

now that we have "cleaner" test because we use new users for each tests, we may have uncover a bug. otherwise it may just be that my refactor break the tests but it's unclear. haven't dig a lot into it, I don't know that part of the code too well now.

test/e2e/servers/graphql-main-server/no-auth-requests.spec.ts.skip
dolcalmi, line 1: same here

test/e2e/servers/graphql-main-server/galoy-pay-testing.spec.ts.skip
dolcalmi, line 1: same here

test/e2e/helpers/account-creation.ts.skip
dolcalmi, line 1: same here

test/e2e/accounts/device-account.spec.ts
dolcalmi, line 1: was all moved to bats?

test/bats/device-account.bats
```diff
@@ -69,13 +67,11 @@ random_uuid() {
...
   token_name="$DEVICE_NAME"
   phone="$DEVICE_PHONE"
-  code="$DEVICE_CODE"
 
   variables=$(
     jq -n \
     --arg phone "$phone" \
-    --arg code "$code" \
-    '{input: {phone: $phone, code: $code}}'
+    '{input: {phone: $phone, code: "000000"}}'
```
dolcalmi, line 74: not sure about this, you can use a default value for the param but please keep it generic

test/bats/helpers/setup-and-teardown.bash
```diff
@@ -175,11 +173,10 @@ fund_wallet_intraledger() {
 initialize_user_from_onchain() {
   local token_name="$1"
   local phone="$2"
-  local code="$3"
```
dolcalmi, line 178: please keep the param with a default value

src/services/twilio.ts
```diff
@@ -185,6 +185,12 @@ export const isPhoneCodeValid = async ({
   phone: PhoneNumber
   code: PhoneCode
 }) => {
+  if (!isProd) {
+    if (code === "000000") {
```
dolcalmi, line 189: you dont need 2 if to do this validation

src/graphql/admin/schema.graphql
```diff
@@ -230,7 +226,6 @@ scalar Memo
 type Mutation {
   accountUpdateLevel(input: AccountUpdateLevelInput!): AccountDetailPayload!
   accountUpdateStatus(input: AccountUpdateStatusInput!): AccountDetailPayload!
-  accountsAddUsdWallet(input: AccountsAddUsdWalletInput!): WalletDetailsPayload @deprecated(reason: "All accounts have USD wallets by default")
```
dolcalmi, line 233: this requires an update of admin panel, also a couple of weeks ago we had a case with an account without stablesats wallet so maybe is better to check in db prev to this removal

src/config/schema.types.d.ts
```diff
@@ -65,15 +65,8 @@ type YamlSchema = {
     phone: string
   }[]
   test_accounts: {
-    ref: string
     phone: string
     code: string
-    needUsdWallet?: boolean
```
dolcalmi, line 71: probably will require config updates in staging and prod

src/app/bootstrap/index.ts
```diff
@@ -86,6 +86,4 @@ export const bootstrap = async () => {
       return new Error("Expected BTC-currency default wallet")
     }
   }
-
-  return true
```
dolcalmi, line 90: if this is a fix please add return type to the method

src/app/authentication/request-code.ts
```diff
@@ -98,13 +103,18 @@ export const requestPhoneCodeForAuthedUser = async ({
     return new PhoneAlreadyExistsError()
   }
 
-  const testAccounts = getTestAccounts()
-  if (TestAccountsChecker(testAccounts).isPhoneValid(phone)) {
+  // return true if we are in dev or test mode
+  if (!isProd) {
```
dolcalmi, line 107: are we sure this will not affect staging?

dolcalmi (changes requested):

test/legacy-integration/app/users/user-role.spec.ts
```diff
@@ -1,11 +1,12 @@
 import { AccountsRepository } from "@services/mongoose"
 
-import { createUserAndWalletFromUserRef, getUserIdByTestUserRef } from "test/helpers"
+import { createUserAndWalletFromPhone, getUserIdByPhone } from "test/helpers"
 
 describe("Users - role", () => {
   it("persists a role set in test accounts", async () => {
-    await createUserAndWalletFromUserRef("I")
-    const userId = await getUserIdByTestUserRef("I")
+    const phone = "+16505554336" as PhoneNumber
```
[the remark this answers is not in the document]
nicolasburtey: No because it's the phone associated with the editor role from the admin_account array

test/legacy-integration/app/cron/delete-payments.spec.ts
```diff
@@ -12,27 +12,48 @@ import {
...
+  walletIdB = await getDefaultWalletIdByPhone(phone)
+
+  await fundWalletIdFromOnchain({
+    walletId: walletIdB,
+    amountInBitcoin: 0.02,
+    lnd: lnd1,
+  })
+
+  // needed to update the balance for some reasons
+  await checkIsBalanced()
```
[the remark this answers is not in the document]
nicolasburtey: There is harm is using it? 

Otherwise the balance is not updated.

test/helpers/user.ts
```diff
@@ -272,39 +288,15 @@ export const createUserAndWallet = async (
...
+    await addWalletIfNonexistent({
+      currency: WalletCurrency.Usd,
+      accountId: account.id,
+      type: WalletType.Checking,
+    })
   }
 
   if (account instanceof Error) throw account
 
-  if (entry.username) {
```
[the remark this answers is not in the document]
nicolasburtey: We were but the code has been replaced. We can no longer do this now that we don't have a test_account array for tests (will only be used for prod... until this features is entirely deprecated - because even in prod we don't need that anymore now that we have deeper kratos integration)

You can see the setUsername in the send lightning test

test/e2e/servers/ory/oathkeeper.spec.ts.skip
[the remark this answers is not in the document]
nicolasburtey: I want @vindard to delete those files after he verified there is no other tests that needs to be ported to bats or integration

test/e2e/servers/graphql-main-server/with-auth-requests.spec.ts.skip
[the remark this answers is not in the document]
nicolasburtey: I want @vindard to delete those files after he verified there is no other tests that needs to be ported to bats or integration

Btw the tests are already not running anymore, whether it's on GitHub or concourse

test/e2e/servers/graphql-main-server/no-auth-requests.spec.ts.skip
nicolasburtey: I want @vindard to delete those files after he verified there is no other tests that needs to be ported to bats or integration

test/e2e/helpers/account-creation.ts.skip
nicolasburtey: I want @vindard to delete those files after he verified there is no other tests that needs to be ported to bats or integration

test/e2e/servers/graphql-main-server/galoy-pay-testing.spec.ts.skip
nicolasburtey: I want @vindard to delete those files after he verified there is no other tests that needs to be ported to bats or integration

test/e2e/accounts/device-account.spec.ts
nicolasburtey: I think so but @vindard can confirm

src/graphql/admin/schema.graphql
nicolasburtey: Iirc it was no longer used in the admin panel already

src/services/twilio.ts
nicolasburtey: I'm not sure to understand?

src/config/schema.types.d.ts
nicolasburtey: Good point 

But I think we only use it with phone + code. 

I'll have a look. 

Ajv would let us know if that's not the case and not let the pod start

src/app/authentication/request-code.ts
nicolasburtey: Good point

@bodymindarts suggested a `--dev` to enable this behavior to be more explicit

Any thought on how to best handle this otherwise?

I don't think we should introduce the pattern of using command line arguments, it's not super typical for a node js app afaik

Otherwise I could add a enable_account_creation_with_defaut_code env variable or something like that

src/app/bootstrap/index.ts
nicolasburtey: the method was not explicitly typed

test/legacy-integration/app/cron/delete-payments.spec.ts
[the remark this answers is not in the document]
nicolasburtey: this is the proper fix: 984572a37b0ad74c64214951964357b57f18367f

src/graphql/admin/schema.graphql
nicolasburtey: confirmed that adminPanel is not using this method anymore

src/config/schema.types.d.ts
nicolasburtey: https://github.com/GaloyMoney/galoy-deployments/pull/4448

test/legacy-integration/notifications/notification.spec.ts
```diff
@@ -91,7 +91,8 @@ async function toArray<T>(gen: AsyncIterable<T>): Promise<T[]> {
 
 describe("notification", () => {
   describe("sendNotification", () => {
-    it("sends daily balance to active users", async () => {
+    // FIXME
+    it.skip("sends daily balance to active users", async () => {
```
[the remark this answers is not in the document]
nicolasburtey: I dig into this quickly but couldn't find really the reason for why this test was failing. but because:
1/ we don't use this code in production any more
2/ this is a very convoluted test that relies on other tests as an artefact, it's very hard to debug. it's probably something we'll want to refactor with more cleaner/independant integration tests.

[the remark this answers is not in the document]
nicolasburtey: e9f727e0419ffc6d5fe18992623e8281258ba798

.env.ci
```diff
@@ -67,6 +67,7 @@ APPCHECK_ISSUER=${APPCHECK_ISSUER}
 APPCHECK_JWKSURI=${APPCHECK_JWKSURI}
 OTEL_EXPORTER_OTLP_ENDPOINT=${OTEL_EXPORTER_OTLP_ENDPOINT}
 MATTERMOST_WEBHOOK_URL=${MATTERMOST_WEBHOOK_URL}
-KRATOS_PG_CON="postgres://dbuser:secret@kratos-pg:5432/default?sslmode=disable"
+KRATOS_PG_CON={$KRATOS_PG_CON}
```
bodymindarts, line 70:
```suggestion
KRATOS_PG_CON=${KRATOS_PG_CON}
```

bodymindarts (approved):

PR #2963, opened 2023-07-28
fix: bats test on mac

nicolasburtey:


bodymindarts (approved):
confirmed this works on my Mac now.

PR #2971, opened 2023-07-28
test: add remaining galoy-pay tests to bats

vindard:


test/e2e/servers/ory/oathkeeper.spec.ts
```diff
@@ -72,6 +73,14 @@ describe("Oathkeeper graphql endpoints", () => {
 })
 
 describe("idempotencyMiddleware", () => {
+  gql`
```
bodymindarts, line 76: don't understand why this is coming in

vindard: It was deleted from galoy-pay and so there was no definition left for the generated.ts file created from `make codegen` so I needed to add it back here since it's still used here

test/bats/public-ln-receive.bats
```diff
@@ -48,15 +141,28 @@ usd_amount=50
...
+  sleep 3
+  retry 10 1 grep "Data.*lnInvoicePaymentStatus.*PENDING" .e2e-subscriber.log
+
   # Receive payment
   lnd_outside_cli payinvoice -f \
     --pay_req "$payment_request" \
 
   # Check for settled
   retry 15 1 check_ln_payment_settled "$payment_request"
+
```
dolcalmi, line 160: this is using the query and not the subscription... is this on purpose? why not just wait for subscription?

test/bats/public-ln-receive.bats
```diff
@@ -127,3 +233,77 @@ usd_amount=50
   # Check for settled
   retry 15 1 check_ln_payment_settled "$payment_request"
 }
+
+@test "public-ln-receive: fail to create invoice - nonexistent wallet-id" {
```
dolcalmi, line 237: this is not the proper description, this could be `fail to create invoice - invalid wallet id` but you need to create another test with a random UUID to get the `nonexistent wallet-id` error

test/e2e/servers/graphql-main-server/galoy-pay-testing.spec.ts.skip
dolcalmi, line 1: this is similar to the `New` prefix.. why do we want to keep this? are there still pending tests to migrate?

dolcalmi (changes requested):

test/bats/public-ln-receive.bats
vindard: Changed it here to just wait for subscription. In the other tests here I use the query because it's quicker. Subscription only used once here to test the subscription explicitly https://github.com/GaloyMoney/galoy/pull/2971/commits/31dae18c8ce409f8088fc0e330e2c99085833937

test/bats/public-ln-receive.bats
vindard: Good catch, fixed https://github.com/GaloyMoney/galoy/pull/2971/commits/0755cf7bbcc9a8861f7431bb85d187879dbbd559

dolcalmi (approved):

PR #2990, opened 2023-08-01
feat: move cold storage to bria

bodymindarts:


nicolasburtey (commented):
Question

src/services/bria/event-handler.ts
```diff
@@ -85,11 +85,17 @@ export const translate = (rawEvent: RawBriaEvent): BriaEvent | BriaEventError =>
   }
   const payoutInfo = rawAugmentation.getPayoutInfo()
   if (payoutInfo) {
+    const grpcMeta = payoutInfo.getMetadata()
     const info = payoutInfo.toObject()
+    let metadata: undefined | PayoutMetadata = undefined
+    if (grpcMeta) {
+      metadata = grpcMeta.toJavaScript()
```
dolcalmi, line 92: this is a really bad name (not our issue)... are we sure this returns a Json?

src/config/schema.types.d.ts
```diff
@@ -58,6 +58,10 @@ type YamlSchema = {
     queueNames: {
       fast: string
     }
+    coldStorage: {
```
dolcalmi, line 61: this needs to be part of `src/config/schema.ts` too. Also this will introduce a "breaking" change so would be a good PR to test release versions and auto merge in charts

src/services/bria/index.ts
```diff
@@ -264,6 +280,26 @@ export const NewOnChainService = (): INewOnChainService => {
...
   }
 
+  const rebalanceToColdWallet = async (
+    amount: BtcPaymentAmount,
+  ): Promise<PayoutId | OnChainServiceError> => {
+    try {
+      const request = new SubmitPayoutRequest()
+      request.setWalletName(briaConfig.hotWalletName)
+      request.setPayoutQueueName(briaConfig.coldStorage.hotToColdRebalanceQueueName)
+      request.setDestinationWalletName(briaConfig.coldStorage.walletName)
```
dolcalmi, line 290: cold storage address is generated internally in bria?

dolcalmi (changes requested):

src/app/cold-storage/record-income.ts
```diff
@@ -0,0 +1,39 @@
+import { toSats } from "@domain/bitcoin"
+import { LedgerService } from "@services/ledger"
+import { getCurrentPriceAsDisplayPriceRatio } from "@app/prices"
+import { DisplayCurrency } from "@domain/fiat"
+
+export const recordIncome = async ({
```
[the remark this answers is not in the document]
bodymindarts: from the POV of the cold storage its income.

src/services/bria/index.ts
bodymindarts: Yes if you use `setDestinationWalletName` then an address is generated internally. https://github.com/GaloyMoney/bria/blob/897b4faf65183a49524d4d8bb7769c5d15741ca3/proto/api/bria.proto#L342-L345

src/servers/event-handlers/bria.ts
```diff
@@ -44,7 +46,13 @@ export const briaEventHandler = async (event: BriaEvent): Promise<true | DomainE
       })
 
     case BriaPayloadType.PayoutBroadcast:
-      return payoutBroadcastEventHandler({ event: event.payload })
+      if (event.augmentation.payoutInfo === undefined) {
+        return new EventAugmentationMissingError()
```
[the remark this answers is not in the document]
bodymindarts: It is https://github.com/GaloyMoney/galoy/blob/1b79c82492769a84d25d63e258928036cc9b9dd0/src/services/bria/event-handler.ts#L73.

This error should never happen but the types don't reflect that so we re-verify here.

src/app/cold-storage/record-income.ts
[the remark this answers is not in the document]
nicolasburtey: then it's probably an `Asset`, not an `Income`

An income would be in the scenario of a Casa, Unchained or Fidelity that says "hey we're going to custody your assets and charge you 0.1%/year for that service". that 0.1% would be the income

for an internal transfer, there is no income per say, the 2 accounting entries that I see are:
1. transfer of assets
2. record of fees, in the form a mining fees, that reduce the total amount of assets under custody.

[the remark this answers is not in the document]
bodymindarts: Understand so the semantics aren't right here. What would you suggest as a better name? The ledger call that this use case is wrapping is:
```
  const journal = await ledgerService.addColdStorageTxReceive({
```
So perhaps `processReceive` or `processTransfer` or something like that?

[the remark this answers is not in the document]
nicolasburtey: > So perhaps processReceive or processTransfer or something like that?

don't have a strong opinion on that one, but generally I think (ie: I'm not sure if this is industry standard, and is more so I have been using those words) as Receive as something where the other party is outside, versus transfer as an internal thing where both sender/recipient are on the same system it's just na internal transfer.

docker-compose.yml
```diff
@@ -14,6 +14,7 @@ services:
       - bitcoind
       - bitcoind-signer
       - lnd1
+      - lnd2
```
nicolasburtey, line 17: is that necessary? I know it outputs some error because lnd2 is not here, but assuming this is the basis or also the yaml for the "quickstart", it would be good to keep it slim.

maybe I could do a PR to make the yml/env more easily configurable, I think that could help.

galoy.yaml
```diff
@@ -29,13 +29,13 @@ coldStorage:
...
-  walletPattern: specter
-  onChainWallet: specter/coldstorage
-  targetConfirmations: 6
 bria:
   hotWalletName: dev
   queueNames:
     fast: dev
+  coldStorage:
+    walletName: cold
+    hotToColdRebalanceQueueName: dev
```
nicolasburtey, line 38: same name at line 33? is that intentional? do they represent the same queue? if not we should have different name to make it explicit that it's not the same thing as to what is line 33 and line 35

src/app/cold-storage/index.ts
```diff
@@ -1,2 +1,2 @@
-export * from "./get-balances"
 export * from "./rebalance-to-cold-wallet"
+export * from "./record-income"
```
nicolasburtey, line 2: wrong name

src/config/schema.ts
```diff
@@ -182,27 +182,13 @@ export const configSchema = {
...
-        targetConfirmations: { type: "integer" },
       },
-      required: [
-        "minOnChainHotWalletBalance",
-        "minRebalanceSize",
-        "maxHotWalletBalance",
-        "walletPattern",
-        "targetConfirmations",
-      ],
+      required: ["minOnChainHotWalletBalance", "minRebalanceSize", "maxHotWalletBalance"],
```
nicolasburtey, line 186: PR should be written as breaking

galoy.yaml
bodymindarts: 33 is a wallet name and 35 is a queue name. 35 + 38 are indeed the same.

src/services/bria/index.ts
```diff
@@ -104,14 +104,14 @@ export const BriaSubscriber = () => {
 
 const queueNameForSpeed = (speed: PayoutSpeed): string => briaConfig.queueNames[speed]
 
-export const NewOnChainService = (): INewOnChainService => {
+export const OnChainService = (): IOnChainService => {
```
nicolasburtey, line 107: I don't think we have really used the `I` prefix consistently. I think we may have been moving away for that (which is fine - type are in a different domain so they could be spell the same)

src/servers/exporter.ts
```diff
@@ -349,22 +323,23 @@ const getAssetsLiabilitiesDifference = async () => {
...
 
   const lnd = lndBalance instanceof Error ? 0 : lndBalance
   const bria = briaBalance instanceof Error ? 0 : Number(briaBalance.amount)
+  const cold = coldStorage instanceof Error ? 0 : Number(coldStorage.amount)
 
   return (
     lnd + // physical assets
-    bitcoind + // physical assets
+    cold + // physical assets
     bria + // physical assets
```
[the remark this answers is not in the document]
bodymindarts: yes

src/servers/exporter.ts
```diff
@@ -1,5 +1,4 @@
 import express from "express"
-import sumBy from "lodash.sumby"
```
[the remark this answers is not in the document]
bodymindarts: Done

src/config/schema.ts
bodymindarts: Done

src/config/schema.types.d.ts
bodymindarts: done

src/services/bria/event-handler.ts
bodymindarts: Yes I'm sure this works (bats cronjob test runs through this code).

src/servers/exporter.ts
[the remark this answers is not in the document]
bodymindarts: Renamed to make it more clear.

src/services/bria/index.ts
bodymindarts: I haven't really changed any convention here, just used what was previously being used in this module - going through the whole codebase and trying to improve consistency on this particular topic is out of scope of this PR.

src/servers/exporter.ts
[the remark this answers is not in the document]
bodymindarts: removed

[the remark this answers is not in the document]
bodymindarts: It is actually still being used had to add it back.

galoy.yaml
nicolasburtey: I think having different for queue and wallet can help convey the message on what is same and what is different. Using dev for dev doesn't help

nicolasburtey (approved):
Small nit but LGTM otherwise

galoy.yaml
bodymindarts: Yeah good point. These are auto-provisioned in bria as part of 'dev' mode. Can change the names there.

PR #2997, opened 2023-08-02
feat: add ip and phone onboard validation

dolcalmi:
based on https://github.com/GaloyMoney/galoy/pull/2905


src/domain/accounts-ips/ip-metadata-validator.ts
```diff
@@ -6,21 +6,20 @@ import {
...
 export const IPMetadataValidator = ({
-  denyIPCountries,
-  allowIPCountries,
+  denyCountries,
+  allowCountries,
   denyASNs,
   allowASNs,
 }: IPMetadataValidatorArgs): IPMetadataValidator => {
-  const validateForReward = (ipMetadata?: IPType): true | ValidationError => {
+  const validate = (ipMetadata?: IPType): true | ValidationError => {
```
vindard, line 14: There's this comment still about whether `validate` is the right name here https://github.com/GaloyMoney/galoy/pull/2905#discussion_r1267839895

vindard (commented):
Resurfaced 1 older comment, but otherwise looks ok to me

src/domain/accounts-ips/ip-metadata-validator.ts
bodymindarts: IMO this is an `authorisation` type concern. Not a `validation` we could have a `valid` IP that we don't want to  `allow`.

dolcalmi: so what is the suggestion (code change needed)? because the logic to validate (`validate` function) is the same but the config used `IPMetadataValidatorArgs` is different

bodymindarts (approved):

bodymindarts (approved):

PR #2998, opened 2023-08-02
fix: properly handle null walletIds

UncleSamtoshi:
WalletIds is a nullable field, but currently we only check if it is undefined.

### DIFF
--- a/src/graphql/public/types/object/business-account.ts
+++ b/src/graphql/public/types/object/business-account.ts
@@ -130,7 +130,7 @@ const BusinessAccount = GT.Object({
 
         let { walletIds } = args
 
-        if (walletIds === undefined) {
+        if (!walletIds) {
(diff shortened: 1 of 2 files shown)


src/graphql/types/object/business-account.ts
```diff
@@ -131,7 +131,7 @@ const BusinessAccount = GT.Object({
 
         let { walletIds } = args
 
-        if (walletIds === undefined) {
+        if (walletIds === undefined || walletIds === null) {
```
dolcalmi, line 134: is not better just use `!walletIds`?

dolcalmi (changes requested):

src/graphql/types/object/business-account.ts
UncleSamtoshi: Works for me

dolcalmi (approved):

PR #3003, opened 2023-08-03
test: add oathkeeper bats tests

vindard:


test/bats/oathkeeper.bats
```diff
@@ -0,0 +1,120 @@
...
+    run_cmd="run"
+  else
+    run_cmd=""
+  fi
+
+  ${run_cmd} curl -s -I \
+    -X POST \
+    ${AUTH_HEADER:+ -H "$AUTH_HEADER"} \
+    -H "Content-Type: application/json" \
+    "${OATHKEEPER_ENDPOINT}graphql"
```
bodymindarts, line 84: This looks weird. should be `{OATHKEEPER_ENDPOINT}/graphql`

bodymindarts (approved):

test/bats/oathkeeper.bats
vindard: Yea it's because of how we've already configured it in the trailing slash in `galoy.yaml` and are using it in other places. I wasn't sure if changing this would affect things in other places or custom configs so I left it as is

PR #3008, opened 2023-08-04
refactor!: consistent use of env

nicolasburtey:
- removing comingling of yaml and process.env for settings/config
- set a source of trust for process.env with env.ts


src/app/admin/backup.ts
```diff
@@ -22,13 +14,13 @@ export const uploadBackup =
   (logger: Logger) =>
   async ({ backup, pubkey }: { backup: string; pubkey: Pubkey }) => {
     logger.debug({ backup }, "updating scb on dbx")
-    const filename = `${BitcoinNetwork()}_lnd_scb_${pubkey}_${Date.now()}`
+    const filename = `${env.NETWORK}_lnd_scb_${pubkey}_${Date.now()}`
```
bodymindarts, line 17: Regardless of wether or not the value is being injected from the environment or yaml the application / service code should not know about this and continue getting the value from the config module.
There needs to be a translation layer between the config that the app layer consumes and the method of injection. Ideally this refactor should only affect the config module and not config consumers.

Or in other words - the only place where you are allowed to reference `env` directly is in `@config`

bodymindarts (changes requested):

src/app/admin/backup.ts
nicolasburtey: I'm not sure there is a real benefits to that, beyond the boilerplate code?

there is 100 configs that we can/have to inject into the server, and it's already fully validated in env.ts

so the boilerplate just to do that https://github.com/GaloyMoney/galoy/pull/3008/files#diff-f0724f34ef85aed4dc757304299f2d7626984d7e04543f22241755c17a94d1beL90 feels unnecessary (leads to dead code) but also error prone (we had default settings in multiple place with different values)

bodymindarts (approved):

src/config/schema.ts
```diff
@@ -573,42 +523,6 @@ export const configSchema = {
...
-        decisionsApi: { type: "string" },
-      },
-      required: ["urlJkws", "decisionsApi"],
-      additionalProperties: false,
-      default: {
-        urlJkws: `http://${
-          process.env.OATHKEEPER_HOST ?? "oathkeeper"
-        }:4456/.well-known/jwks.json`,
-        decisionsApi: `http://${
-          process.env.OATHKEEPER_HOST ?? "oathkeeper"
```
vindard, line 608: Btw, I was caught out by exactly this in #3020 and #3022 where I couldn't figure out where oathkeeper settings were coming from and hadn't thought to check that an environment variable was being pulled into yaml config

nicolasburtey: yup this PR should clean this all up

nicolasburtey:
fix #2951

krtk6160 (approved):

PR #3010, opened 2023-08-06
test: fix bats test on mac os

nicolasburtey:


test/bats/oathkeeper.bats
```diff
@@ -88,7 +88,7 @@ exec_oathkeeper() {
 
 oathkeeper_jwt() {
   echo $output \
-  | grep -oP 'Authorization: Bearer \K[^ ]+'
+  | echo $output | grep -o 'Authorization: Bearer [^ ]*' | sed 's/Authorization: Bearer //'
```
vindard, line 91: The `echo $output` here is likely redundant?
```suggestion
  | grep -o 'Authorization: Bearer [^ ]*' \
  | sed 's/Authorization: Bearer //'
```

vindard (changes requested):

test/bats/oathkeeper.bats
nicolasburtey: > I'm ok with sticking with sed as well if we prefer that
no preference on my end. code provide by GPT :D

vindard (approved):

PR #3014, opened 2023-08-07
Quickstart

bodymindarts:


docker-compose.override.yml
```diff
@@ -7,7 +9,7 @@ services:
       - "bats-tests:host-gateway"
   otel-agent:
     ports:
-      - "4318:4318" # http receiver
+      - "4318:4318" #! http receiver
```
nicolasburtey, line 12: what does the `!` do?

quickstart/bin/smoke-test.sh
```diff
@@ -0,0 +1,109 @@
...
+  echo
+  echo "Seeding some regtest blocks..."
+  bitcoind_init > /dev/null 2>&1
+  echo "DONE"
+  echo "Getting balance from bria..."
+  bria_cli wallet-balance -w dev > /dev/null 2>&1
+  echo "DONE"
+  echo "Running getinfo on lnd..."
+  lnd_cli getinfo > /dev/null 2>&1
+  echo "DONE"
```
nicolasburtey, line 97: I think 2 things missing:
- clear one channel between lnd1 and lnd-outside-1
- seed one defined internal account with some funds

docker-compose.override.yml
bodymindarts: This is for `ytt` to interpret it as a comment. Unfortunately `ytt` does not accept legal yaml comments starting with `#` only.

quickstart/bin/smoke-test.sh
bodymindarts: Agree - out of scope for now. Want to get this PR merged once e2e integration into ci is done. Then can iterate on making the setup functionally complete.

quickstart/docker-compose.yml
```diff
@@ -0,0 +1,322 @@
...
+    - 4004:4004
+    environment:
+    - APOLLO_ROUTER_SUPERGRAPH_PATH=/repo/dev/apollo-federation/supergraph.graphql
+    - APOLLO_ROUTER_CONFIG_PATH=/repo/dev/apollo-federation/router.yaml
+    volumes:
+    - ${HOST_PROJECT_PATH:-.}/${GALOY_QUICKSTART_PATH:-vendor/galoy-quickstart}/dev:/repo/dev
+  oathkeeper:
+    image: oryd/oathkeeper:v0.40.4-distroless
+    ports:
+    - 4002:4455
```
UncleSamtoshi, line 15: Is it possible to split out the port mapping into a separate docker-compose file similar to how we do it for our docker-compose file for this repo. We need to be able to disable the port mapping so that when the bats tests runs on concourse they only happen within a docker network and dont map to any ports on the concourse worker.

bodymindarts: We could but let me ask - have you observed this being an issue? I think since we restrict our docker host to only run 1 job at a time (via concourse locks) this should no longer be an issue.

bodymindarts: The hiding-of-port-mapping stuff we did was to be able to run multiple jobs that need docker at once (so as not to have port collisions). But with only 1 job running at any time (which is the case now) I think we don't have that problem anymore.

UncleSamtoshi: I didn't know that we switched to one job at a time. I don't think we need it then.

PR #3016, opened 2023-08-07
test: move cron-rebalance test to bats

vindard:
## Description

There is a potential clean-up we can do 1st to fix an edge-case in our accounting for channel opens. I may explore implementing that change 1st before marking this PR as ready-for-review.


src/services/lnd/utils-bos.ts
```diff
@@ -75,13 +85,16 @@ export const rebalancingInternalChannels = async () => {
         destination: selfPubkey,
         ...settings,
       })
+
+      await waitFor({ func: checkNoPendingHtlc, max: 20 })
```
dolcalmi, line 89: why do we need to add this? are we moving tests validations to main code?

test/bats/helpers/ln.bash
```diff
@@ -4,9 +4,38 @@ source $(dirname "$BASH_SOURCE")/_common.bash
...
+    | while read -r channel_point; do
+        funding_txid="${channel_point%%:*}"
+        run_with_lnd "$ln_cli_name" closechannel "$funding_txid"
+      done
+}
+
 lnds_init() {
   # Clean up any existing channels
-  lnd_cli closeallchannels || true
+  close_partner_initiated_channels lnd_cli || true
```
dolcalmi, line 38: where is this method used with lnd2_cli? or why the wrapper?

dolcalmi (changes requested):

src/services/lnd/utils-bos.ts
vindard: This was probably poorly named here, did a refactor https://github.com/GaloyMoney/galoy/pull/3016/commits/9300b6b6df4f9d360035cd2205dea010cc1f459c

It's needed because we have a limitation where we can't correctly account for the escrow of a channel if there are pending htlcs in it (see #3031). This is to protect against an edge-case that seems to consistently happen where there are still pending htlcs in our internal channel from the `rebalance` step when cron moves on to the `updateEscrows` step which causes it to consistently record an escrow that is slightly off (by 430 sats).

test/bats/helpers/ln.bash
vindard: Refactored it here to be more explicit. Ideally this is a function that closes all non-self-initiated channels from any of our internal nodes. It replaces the `closeAllChannels` usage because we can't yet account for channels closed that our internal node has initiated.

https://github.com/GaloyMoney/galoy/pull/3016/commits/fe16e45bf01bd072ae0a441fed31f8efcc679c89

PR #3021, opened 2023-08-07
chore: add account UUID migration

dolcalmi:
Please dont merge until e2e is fixed in CI. this PR needs to be deployed after code update

### DIFF
--- a/src/migrations/20230807205306-add-uuid-to-accounts.ts
+++ b/src/migrations/20230807205306-add-uuid-to-accounts.ts
@@ -4,18 +4,18 @@
 const getUuidByString = require("uuid-by-string")
 
 async function migrateAccounts(db, batchSize = 100) {
-  const cursor = db.collection("accounts").find({ uuid: { $exists: false } })
+  const cursor = db.collection("accounts").find()
 
   let batchCount = 0
   while (await cursor.hasNext()) {
     const batchUpdates = []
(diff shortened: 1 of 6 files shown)


ntheile (dismissed):

ntheile (dismissed):

src/migrations/20230807205306-add-uuid-to-accounts.ts
```diff
@@ -0,0 +1,44 @@
...
+  let batchCount = 0
+  while (await cursor.hasNext()) {
+    const batchUpdates = []
+    for (let i = 0; i < batchSize && (await cursor.hasNext()); i++) {
+      const account = await cursor.next()
+      const uuid = getUuidByString(account._id.toString())
+      batchUpdates.push({
+        updateOne: {
+          filter: { _id: account._id },
+          update: { $set: { uuid } },
```
nicolasburtey, line 18: we call it `id` for the Wallet. I think we should keep the same pattern here ( `_id` for internal mongo object and `id` for the external identifier)

nicolasburtey (changes requested):
I think we should be consistent and use `id` not `uuid`

nicolasburtey:
I've look into all the reference of account.id currently per your input and indeed it seems a big effort, so never mind my comment for now

UncleSamtoshi (approved):

PR #3030, opened 2023-08-08
test: check env for oathkeeper endpoint before config

vindard:
## Description

This refactors the oathkeeper endpoint check to go in order of : env variable -> custom.yaml -> galoy.yaml (default config) (#3020 & #3022)

**This will need to be cleaned up when @nicolasburtey #3008 PR gets merged.** 

_Note: it looks like `OATHKEEPER_PORT` isn't used anywhere_

### DIFF
--- a/test/bats/oathkeeper.bats
+++ b/test/bats/oathkeeper.bats
@@ -34,6 +34,13 @@ fetch_config_value() {
   fi
 }
 
+# TODO: Refactor when we clean up env variable handling
+if [[ -n "$OATHKEEPER_HOST" ]]; then
+  OATHKEEPER_ENDPOINT="${OATHKEEPER_HOST}:4456/decisions/"
+else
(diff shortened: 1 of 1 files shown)


dolcalmi (approved):

test/bats/oathkeeper.bats
```diff
@@ -34,6 +34,13 @@ fetch_config_value() {
   fi
 }
 
+# TODO: Refactor when we clean up env variable handling
+if [[ -n "$OATHKEEPER_HOST" ]]; then
```
bodymindarts, line 38: This should be sue in docker compose for

vindard: Do you mean like this where we simply  ignore the config files?
https://github.com/GaloyMoney/galoy/compare/refactor-oathkeeper-endpoint

PR #3037, opened 2023-08-09
fix: otel vulnerability

dolcalmi:
fix audit issue with otel packages


dolcalmi:
closed in favor of https://github.com/GaloyMoney/galoy/pull/3038 but we need to check if recreating yarn.lock will cause the same effect

PR #3042, opened 2023-08-10
feat: Quickstart

bodymindarts:


dev/otel-agent-config.yaml
```diff
@@ -1,4 +1,8 @@
 receivers:
+  jaeger:
```
nicolasburtey, line 2: which system needs jaeger?

docker-compose.yml
```diff
@@ -388,25 +412,26 @@ services:
       retries: 5
   oathkeeper:
     image: oryd/oathkeeper:v0.40.4-distroless
+    ports: []
+    extra_hosts: []
     command: serve -c /home/ory/oathkeeper.yml --sqa-opt-out
     volumes:
-      - type: bind
```
nicolasburtey, line 393: type bind is by default?

quickstart/bin/bump-mongodb-migrate-image-digest.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
nicolasburtey, line 1: why is something specific for mongodb needed?

dev/otel-agent-config.yaml
bodymindarts: the rust systems (stablesats / bria) are still talking to the jaeger endpoints. An easy 'maintenance' task to move them to otlp. cc @thevaibhav-dixit

docker-compose.yml
bodymindarts: yes the -`<origin>:<mountpoint>` is a simpler way to express the same thing.

quickstart/bin/re-render.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -e
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+GALOY_DEV_DIR=${REPO_ROOT}/quickstart/dev
+GALOY_ROOT_DIR=${REPO_ROOT}/quickstart/galoy
```
nicolasburtey, line 7: this is not very clear to me. why root dir is a sub directory of quickstart?

quickstart/bin/bump-mongodb-migrate-image-digest.sh
bodymindarts: Galoy backend defines and builds an image dedicated to running the migration scripts that we have. This script pins it to the sha that gets input. It's intended to be used from ci as part of the release process.
First the image must get built via https://ci.galoy.io/teams/dev/pipelines/galoy-app/jobs/build-migrate-edge-image/builds/1403 and then the ref will get bumped and committed back as part of the release.

quickstart/dev/appcheck/README.md
```diff
@@ -0,0 +1,140 @@
+## Device Login Flow for App Check
```
nicolasburtey, line 1: I don't think this file is really necessary? maybe it could be moved outside of `/dev/`

quickstart/bin/re-render.sh
bodymindarts: This is just for within the context of `re-render`. It syncs the state of the main repo into `quickstart/galoy`. It's like a snapshot of the latest stable / working / released state of the galoy repo as it was in a situation that we know Quickstart is working.

The idea here is that
```
git clone <galoy-repo>
cd quickstart
docker compose up -d
```

Should remain stable even if there are untested / unreleased commits on the main branch.
`bin/re-render.sh` will be executed as part of the release process once tests have past.
We don't want to reference the files under the 'real' repo root as they have not been vetted yet.

bodymindarts: Its the same technique we use in Galoy-infra where the examples should always point to the latest release of the repo and not the head.

See https://github.com/GaloyMoney/galoy-infra/blob/0fcd4c139a4b1051e7f85ecd25f2b56dcd9721ca/examples/gcp/platform/main.tf#L15

quickstart/dev/appcheck/README.md
bodymindarts: Do you mean deleted from the main repo https://github.com/GaloyMoney/galoy/tree/main/dev/appcheck? Or just not sync it into [quickstart](https://github.com/GaloyMoney/galoy/blob/9bf8a3be989d15897c54a59727732b59eab7f89c/quickstart/vendir/template.yml#L14)?

ntheile:
just tested the quickstart on a fresh project and it works great!

PR #3043, opened 2023-08-10
feat: callback api

nicolasburtey:


docker-compose.yml
```diff
@@ -454,3 +456,35 @@ services:
       - SSL_CERTFILE=/tls.cert
       - SSL_KEYFILE=/tls.key
     command: ["Fulcrum", "/fulcrum.conf"]
+
+  svix:
+    image: svix/svix-server
+    platform: linux/amd64
+    environment:
+      WAIT_FOR: "true" # We want to wait for the default services
```
bodymindarts, line 464: For this to work with Quickstart you have to use `ytt` conform comments (ie. `#!`)

docker-compose.yml
```diff
@@ -454,3 +456,35 @@ services:
...
+      POSTGRES_PASSWORD: postgres
+
+  svix-redis:
+    image: "redis:7-alpine"
+    # Enable persistence
+    command: "--save 60 500 --appendonly yes --appendfsync everysec"
+    volumes:
+      - "svix-redis-data:/data"
+
+volumes:
```
bodymindarts, line 488: This key will probably need to be added to the Quickstart `docker-compose.tmpl.yml`

test/bats/ln-receive.bats
```diff
@@ -69,9 +111,13 @@ usd_amount=50
 
   # Check for subscriber event
   check_for_ln_update "$payment_hash" || exit 1
+
+  exec_graphql "$token_name" 'account-details'
+  account_id="$(graphql_output '.data.me.defaultAccount.id')"
+  cat .e2e-callback.log | grep "$account_id" || exit 1
```
vindard, line 117: `account_id` needs to be brought back into this test. Can be done by caching in 1st test and reading back here

```suggestion
  cat .e2e-callback.log | grep "$(read_value $token_name.account_id) || exit 1
```

test/bats/ln-receive.bats
```diff
@@ -41,9 +39,53 @@ teardown() {
...
+@test "ln-receive: init" {
+  token_name="$ALICE_TOKEN_NAME"
+
+  login_user "$ALICE_TOKEN_NAME" "$ALICE_PHONE" "$CODE"
+
+  # create "application" for svix
+  exec_graphql "$token_name" 'account-details'
+  account_id="$(graphql_output '.data.me.defaultAccount.id')"
+  [[ "$account_id" != "null" ]] || exit 1
+
```
vindard, line 51: Can cache here to read in the next test

```suggestion
cache_value "$token_name.account_id" "$account_id"
```

vindard (changes requested):
Comment on test

test/bats/helpers/ln.bash
```diff
@@ -86,6 +86,10 @@ lnds_init() {
   retry 10 1 mempool_not_empty
   retry 10 1 no_pending_channels
 
+  # FIXME: we may need some check on the graph or something else
+  # NB: I get randomly a "no route" error otherwise
+  sleep 10
```
nicolasburtey, line 91: @vindard I have added a sleep here. AFAIR we had a wait for the network graph to be updated for the e2e test, so we may need the same thing here.

vindard (dismissed):
There's some little things I'd clean up in testing but can tackle them in a follow-on PR. Otherwise looks ok to me

Edit: It seems like the integration tests are failing with callback messages. I won't be able to confirm if this is an actual issue until I can get it working locally (for some reason, bats test doesn't pass for me locally). Dismissed my approval until we can confirm this too.

src/services/callback/index.ts
dolcalmi, line 1: this is a problem, if we dont have an interface then the DDD is difficult to apply because we began to depend on types of inner libraries.

dolcalmi (changes requested):

src/app/payments/send-intraledger.ts
```diff
@@ -376,6 +380,21 @@ const executePaymentViaIntraledger = async <
...
     }
 
+    if (
+      recipientAccount.level === AccountLevel.One ||
+      recipientAccount.level === AccountLevel.Two
+    ) {
+      const callbackService = CallbackService(getCallbackServiceConfig())
+      await callbackService.sendMessage({
+        accountUUID: recipientAccount.uuid,
+        eventType: CallbackEventType.ReceiveIntraledger,
```
dolcalmi, line 390: having this makes me wonder if we should use notification service as a wrapper of svix/callbacks :thinking:

vindard:
> the integration tests logs are flooded with docker-compose that make the integration tests quite unparseable (ie:

Yup I've had this issue as well. I did some preliminary investigations, and so far it looks like something might have changed upstream in `docker compose` and we'll have to find a way to explicitly silence that logging. I wasn't able to replicate the extensive logging locally to try to tackle it, so stopped there for now.

https://github.com/GaloyMoney/galoy/pull/3058

src/services/callback/index.ts
```diff
@@ -0,0 +1,93 @@
...
+        name: `callback for account ${accountUUID}`,
+        uid: accountPath,
+      }
+
+      await svix.application.create(application)
+    } catch (err) {
+      if ((err as SvixError).code === 409) {
+        // we create app on the fly, so we are expect this error and can ignore it
+      } else {
+        console.log(err)
```
dolcalmi, line 56: please remove/use logger (applies to all methods)

src/services/callback/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+
+  // only work for hosted svix
+  const getWebsocketPortal = async (accountUUID: AccountUUID) => {
+    return svix.authentication.appPortalAccess(accountUUID, {})
+  }
+
+  const addEndpoint = async (accountUUID: AccountUUID, url: string) => {
+    return svix.endpoint.create(accountUUID, {
```
dolcalmi, line 79: this need to handle exceptions properly (applies to other methods)

dolcalmi (changes requested):

src/app/payments/send-intraledger.ts
nicolasburtey: You mean notifications should be an endpoint from this callback?

I think that would make sense

src/services/callback/index.ts
nicolasburtey: Ok will do

dolcalmi (approved):

PR #3051, opened 2023-08-13
test: remove success ln-send tests from legacy integration

vindard:


test/bats/ln-send.bats
dolcalmi, line 1: intraledger must check sender and receiver so not sure if we are ready to remove the tests from integration

dolcalmi (changes requested):

test/bats/ln-send.bats
vindard: I added a commit to check the recipient side txn was received 22ca0a9d9228a568a74b7e4d797f32256dfa0e78

I'm not checking contact logic as yet. I've left some tests in legacy with contact logic still, planning to tackle that in another PR

dolcalmi (approved):

PR #3054, opened 2023-08-15
test: setup new integration lnd service tests

vindard:
## Description

Add new lnd-service integration tests and move tests from legacy into here


test/helpers/lightning.ts
```diff
@@ -188,6 +188,65 @@ export const openChannelTesting = async ({
...
+  baseLogger.debug("mining blocks and waiting for channel being opened")
+  await Promise.all([
+    // error: https://github.com/alexbosworth/ln-service/issues/122
+    // once(sub, 'channel_opened'),
+    waitFor(() => lndNewChannel && lndPartnerNewChannel),
+    mineBlockAndSync({ lnds: [lnd, lndPartner] }),
+  ])
+
+  if (lndPartner === lnd1) {
+    expect(lndPartnerNewChannel.is_partner_initiated).toBe(true)
```
dolcalmi, line 239: this is a bad practice... we should not have validation in helper methods

test/integration/services/lnd-service.spec.ts
```diff
@@ -0,0 +1,347 @@
...
+    const paid = await lndService.payInvoiceViaPaymentDetails({
+      decodedInvoice: lnInvoice,
+      btcPaymentAmount,
+      maxFeeAmount: undefined,
+    })
+    if (paid instanceof Error) throw paid
+
+    // Confirm payment exists in lnd
+    const retrievedPayment = await lndService.lookupPayment({ paymentHash })
+    expect(retrievedPayment).not.toBeInstanceOf(Error)
```
dolcalmi, line 329: Please validates against specific types (applies to all the error expects)

test/legacy-integration/02-user-wallet/02-send-lightning.spec.ts.skip
dolcalmi, line 1: are you sure this is completely covered? or why is skipped?

dolcalmi (changes requested):

test/integration/services/lnd-service.spec.ts
vindard: I usually validate against specific error types for positive validation (expect a specific error), but keep it generic when I don't want any error at all coming back from a call

test/legacy-integration/02-user-wallet/02-send-lightning.spec.ts.skip
vindard: It's mostly covered and across the current work will be fully covered in the next day or 2. Not sure it's worth the investment trying to keep these working now given how aggressively we're removing tests from that file and how interconnected the tests are in there.

test/helpers/lightning.ts
vindard: Good catch, I'd just copy-pasted the function just above and removed the accounting lines. Did a bit more direct cleanup of that function here: https://github.com/GaloyMoney/galoy/pull/3054/commits/b5aa657b4cf2b2ba5ee813d8fedac0c68ebbbf51

dolcalmi (approved):

PR #3057, opened 2023-08-15
chore: iteration on quickstart

nicolasburtey:


quickstart/dev/bitcoind_signer_descriptors.json
bodymindarts, line 1: This folder is synced via vendir. Can’t add anything because vendir owns it

quickstart/bin/helpers.sh
```diff
@@ -4,7 +4,7 @@ set -e
 
 COMPOSE_PROJECT_NAME="${COMPOSE_PROJECT_NAME:-quickstart}"
 GALOY_ENDPOINT=${GALOY_ENDPOINT:-localhost:4002}
-CACHE_DIR=tmp/smoke-test-cache
+CACHE_DIR=vendor/galoy-quickstart/tmp/smoke-test-cache
```
bodymindarts, line 7: These paths need to work locally as well as when synced so use `${GALOY_QUICKSTART_PATH:-vendor/galoy-quickstart}` for this path like we do in the docker-compose file. https://github.com/GaloyMoney/galoy/blob/b674f6f9d79f766f11c0e0105027878f38390492/quickstart/docker-compose.yml#L11

nicolasburtey: does it matter for the CACHE_DIR?

bodymindarts: Yes it would matter for all paths where you hard code that

dolcalmi (approved):

PR #3072, opened 2023-08-17
test: add contact check for intraledger settled in bats

vindard:


test/bats/onchain-send.bats
```diff
@@ -134,21 +135,33 @@ teardown() {
   [[ "${settled_status}" = "SUCCESS" ]] || exit 1
 }
 
-@test "onchain-send: settle intraledger" {
+@test "onchain-send: settle intraledger, with no contacts check" {
```
dolcalmi, line 138: with contacts check?

test/bats/onchain-send.bats
```diff
@@ -134,21 +135,33 @@ teardown() {
...
+  recipient_phone="$(random_phone)"
+  login_user \
+    "$recipient_token_name" \
+    "$recipient_phone"  \
+    "$CODE"
+  user_update_username "$recipient_token_name"
+  btc_recipient_wallet_name="$recipient_token_name.btc_wallet_id"
+
+  # Check is not contact before send
+  run is_contact "$alice_token_name" "$recipient_token_name"
```
dolcalmi, line 153: not a blocker but should we rename "alice" prefix ?

dolcalmi (changes requested):

test/bats/onchain-send.bats
vindard: Hm, `alice` is a fixed user but the `recipient` I renamed is a transient one created for the test. Maybe rename would make sense if we made Alice random too, which we probably need right now here

test/bats/onchain-send.bats
vindard: So interestingly I don't think we have contacts for `send-onchain` use-case. I'm not sure if that was deliberate or not so I captured current behaviour here. The assertion check that contacts _isn't_ added after the send.

bodymindarts (approved):

dolcalmi (approved):

PR #3079, opened 2023-08-17
chore: some optimization on latency

nicolasburtey:
callback takes about 700 ms to do both queries, so better to return the result to the user without await message result, specifically as we're not doing anything with its response


vindard (commented):
@dolcalmi would have to confirm but I think in the past we've had issues when we don't await, where the promise gets canceled before completing when the function returns

nicolasburtey:
> @dolcalmi would have to confirm but I think in the past we've had issues when if we don't await, where the promise gets canceled before completing when the function returns

I think the issue we had previously was related to a slow db mutation. so maybe we were thinking issue were related they were not AFAIK. 

the idea of not awaiting for a promise when you don't need to wait for the result is a key way to improve latency and I don't think there are major drawback to the approach

bodymindarts:
One issue is that we may return a success code even though sending the callback errored.

nicolasburtey:
> One issue is that we may return a success code even though sending the callback errored.

what is the correct solution then? sending an error here while the payment still succeed doesn't sound to be a good option either?

bodymindarts (approved):

PR #3112, opened 2023-08-24
fix: skip rebalancingInternalChannels on signet

krtk6160:


nicolasburtey (approved):

dolcalmi:
is this the problem? the problem was with bos not supporting signet, if in this moment is failing is for other reason, probably some bug in the code, this is just hiding it

dolcalmi:
issue fixed now

PR #3117, opened 2023-08-26
test: move some receive-onchain integration tests

vindard:


test/integration/app/wallets/create-on-chain-address.spec.ts
```diff
@@ -8,6 +8,31 @@ import { AccountsRepository } from "@services/mongoose"
 import { createRandomUserAndBtcWallet } from "test/helpers"
 
 describe("onChainAddress", () => {
+  it("can apply requestId as idempotency key when creating new address", async () => {
```
dolcalmi, line 11: should not we use `getLastOnChainAddress` too?

test/bats/onchain-receive.bats
```diff
@@ -27,57 +27,111 @@ teardown() {
   fi
 }
 
-@test "onchain-receive: settle onchain for BTC wallet" {
+@test "onchain-receive: btc wallet, can create new address if current one is unused" {
```
dolcalmi, line 30: if this is true then we have an attack vector

dolcalmi (changes requested):

test/integration/app/wallets/create-on-chain-address.spec.ts
vindard: This one was just to explicitly test the requestId idempotency behavious that we haven't exposed at the graphql level. This would've otherwise been an e2e test. The rest of happy path create/getLast tests are in bats tests.

vindard: This was just to specifically test the idempotency behaviour of that `requestId` param that we haven't exposed at the graphql layer. The rest of happy path tests for create/getLast should be in e2e bats tests

bodymindarts (approved):

PR #3122, opened 2023-08-27
refactor: create different context for public and admin api

nicolasburtey:
cleanup at the graphql layer about the context been passed to the resolver. 

prior to this PR, we always assume a `user` and an `account` where passed to the resolver when the query/mutation was authentication. while this assumption is correct for the public API, it is not for the admin API. 

the "userId" (represented by the `token.sub` in the JWT) should be seen as an auditorId, that doesn't necessarily have an account within the system. this will allow to grant "user" to the admin panel that are for instance issues with `Oauth2 client_credentials` and doesn't have themselves an account in the system.  

note: this PR is not enough yet to be able to do that, one last item remaining is the extraction of "isEditor" out of the account entity and into an authorization layer to make this possible, through casbin for instance.


src/graphql/shared/root/mutation/user-login.ts
```diff
@@ -18,12 +18,16 @@ const UserLoginInput = GT.Input({
   }),
 })
 
-const UserLoginMutation = GT.Field<{
-  input: {
-    phone: PhoneNumber | InputValidationError
-    code: PhoneCode | InputValidationError
+const UserLoginMutation = GT.Field<
+  null,
+  GraphQLPublicContext | GraphQLAdminContext,
```
bodymindarts, line 23: Can't this be parameterised and declared where it gets added to the schema?
Something like:
```
const UserLoginMutation<T:GraphQLPublicContext | GraphQLAdminContext > = GT.Field<
  null,
 T,
  ```
  
  And then when its used
  ```
  UserLoginMutation<GraphQLPublicContext>
  ```
  ?

bodymindarts (approved):
1 Question

src/graphql/shared/root/mutation/user-login.ts
nicolasburtey: yeah but this code is going to be deleted so not worth the improvement

nicolasburtey: ie: replaced by rest endpoint. 

this would remove the complexity of having different context for the same function: auth code is better off being outside of the rest of the app

PR #3129, opened 2023-08-28
test: add notification tests for onchain receive actions

vindard:


test/integration/app/wallets/add-pending-on-chain-transaction.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    await addPendingTransaction({
+      txId: "txId" as OnChainTxHash,
+      vout: 0 as OnChainTxVout,
+      satoshis: btcAmount,
+      address,
+    })
+
+    // Expect sent notification
+    expect(sendNotification.mock.calls.length).toBe(1)
+    expect(sendNotification.mock.calls[0][0].title).toContain("Pending")
```
dolcalmi, line 51: if is not the purpose of this test validate title and body then this should be something similar to `intraledger receive` test, else we need to add title and body validations like in current tests

dolcalmi (changes requested):

test/integration/app/wallets/add-pending-on-chain-transaction.spec.ts
vindard: Changed https://github.com/GaloyMoney/galoy/pull/3129/commits/c869794dff4297c2c02c1093e1b1ce5818b2e6ba

PR #3135, opened 2023-08-30
chore: refactor display amounts conversion

vindard:
## Description

This PR adds a new unit tested `DisplayCurrencyConverter` object and then swaps it into all use-cases that have this conversion implemented in-line.

The prompt for this was to stop explicitly integration testing this conversion for each use-case like we do now, and to instead test it once via a unit test.

It also adds a new switching decision for USD display currencies internally in the converter which will eventually allow removing this from our `Metadata` functions in a follow-on PR.

### DIFF
--- a/src/app/payments/reimburse-fee.ts
+++ b/src/app/payments/reimburse-fee.ts
@@ -1,6 +1,10 @@
-import { displayAmountFromNumber } from "@domain/fiat"
+import { DisplayAmountsConverter, displayAmountFromNumber } from "@domain/fiat"
 import { FeeReimbursement } from "@domain/ledger/fee-reimbursement"
-import { DisplayPriceRatio, WalletPriceRatio } from "@domain/payments"
+import {
+  DisplayPriceRatio,
+  WalletPriceRatio,
+  toDisplayBaseAmount,
+} from "@domain/payments"
 import {
   paymentAmountFromNumber,
   WalletCurrency,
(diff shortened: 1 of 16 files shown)


src/app/payments/reimburse-fee.ts
```diff
@@ -1,6 +1,10 @@
-import { displayAmountFromNumber } from "@domain/fiat"
+import { DisplayAmountsConverter, displayAmountFromNumber } from "@domain/fiat"
 import { FeeReimbursement } from "@domain/ledger/fee-reimbursement"
-import { DisplayPriceRatio, WalletPriceRatio } from "@domain/payments"
+import {
+  DisplayPriceRatio,
```
bodymindarts, line 4: a bit confused as to why some parts of`Display` concept is in `payments` and some is in `fiat`. Not something to fix in this PR - just a remark

src/domain/fiat/primitives.ts
```diff
@@ -0,0 +1,4 @@
+export const DisplayCurrency = {
```
bodymindarts, line 1: This isn't the right list is it? We support a lot more display currencies so I'm confused.

test/unit/domain/fiat/display-amounts-converter.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+      displayFee: {
+        ...expectedResult.displayFee,
+        amountInMinor: 25n,
+        displayInMajor: "0.25",
+      },
+    }
+  }
+
+  describe("non-usd display currency", () => {
+    const currency = "EUR" as DisplayCurrency
```
bodymindarts, line 74: Usually ofr enums we use `DisplayCurrency.EUR` syntax.

bodymindarts (commented):
Its a good direction but confused about a few things.

src/domain/fiat/primitives.ts
vindard: We use a strange hybrid type for this that's a union of 2 values from an "enum" and our usual unique symbol (see [here](https://github.com/GaloyMoney/galoy/blob/90294698bbc8fdc00b8780e296f11f19fb11879f/src/domain/fiat/index.types.d.ts#L6)). It's also just a scalar in the graphql layer vs. `WalletCurrency` which is an enum there.

We're loading these currency types in dynamically from config, so I think we can probably decide now if we'd like to just get rid of the enum part of the type.

test/unit/domain/fiat/display-amounts-converter.spec.ts
vindard: Related to https://github.com/GaloyMoney/galoy/pull/3135#discussion_r1320341780

src/app/payments/reimburse-fee.ts
vindard: I noticed that as well (was confusing at first too), but didn't want to touch it here. I'll probably try cleaning this up in a separate PR if it won't be a big change.

src/domain/fiat/primitives.ts
bodymindarts: Don't see the point in a hybrid. Would load it all dynamically

vindard: Fixed in #3193

bodymindarts (approved):

PR #3147, opened 2023-08-31
feat: add held invoices count to exporter

dolcalmi:
- Add held invoices count to exporter
- Refactor listInvoices in lnd service to get only invoices in the last 24h (default value for btc invoices) -> impact in performance https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/wYkcvsJ3AyP/trace/FRQFckMZxLo?fields[]=c_name&fields[]=c_service.name&span=f60e2dd69019c170 (if it is only used in trigger start up is not a problem but in exporter can impact performance)


bodymindarts:
What if a cron doesn't run and we forget to re-run. Wouldn't we want to catch up a longer time frame like 48h?

vindard (approved):

PR #3158, opened 2023-09-05
feat(old): switch to btc wallet for sub 1 cent usd receipts

vindard:
## Description

For incoming transactions to USD wallets where the sats amount is less than 1 cent, we currently return an error and don't credit the USD wallet or settle the invoice. This PR is to instead settle the sats to BTC wallet instead of returning an error in the following cases:
- [x] ln-initiated intraledger receive to USD wallet
- [x] walletId-initiated intraledger receive to USD wallet
- [x] no-amount invoice receive from external source
- [x] onchain-initiated intraledger receive to USD wallet

### DIFF
--- a/src/app/payments/helpers.ts
+++ b/src/app/payments/helpers.ts
@@ -13,8 +13,8 @@ import {
   SelfPaymentError,
 } from "@domain/errors"
 import {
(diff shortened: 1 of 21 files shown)


vindard:
Take over by #3176

PR #3166, opened 2023-09-05
chore: improving naming from validator to authorizer

nicolasburtey:


src/app/authentication/login.ts
```diff
@@ -362,12 +364,15 @@ export const loginWithDevice = async ({
...
+    const authorizedIPMetadata = IPMetadataAuthorizer(
       ipMetadataValidationSettings,
-    ).validate(ipFetcherInfo)
+    ).authorize(ipFetcherInfo)
 
-    if (validatedIPMetadata instanceof Error) {
-      return new InvalidIPForOnboardingError(validatedIPMetadata.name)
+    if (authorizedIPMetadata instanceof Error) {
+      if (authorizedIPMetadata instanceof MissingIPMetadataError)
+        return new InvalidIpMetadataError(authorizedIPMetadata.name)
```
bodymindarts, line 373: Why are we only passing the name on? Aren't we potentially silently dropping interesting error context?

src/graphql/error.ts
```diff
@@ -365,6 +365,39 @@ export class InvalidPhoneForOnboardingError extends CustomApolloError {
...
+      code: "UNAUTHORIZED_VPN_IP_FOR_REWARD",
+      ...errData,
+    })
+  }
+}
+
+export class UnauthorizedIPForRewardError extends CustomApolloError {
+  constructor(errData: CustomApolloErrorData) {
+    super({
+      message: "This ip is not unauthorized for rewards.",
```
bodymindarts, line 393: double negative? `not unauthorised`?

src/app/authentication/login.ts
nicolasburtey: yeah, this was the previous behavior. just changed it.

bodymindarts (approved):
LGTM

PR #3167, opened 2023-09-05
feat: add enable proxy check setting

dolcalmi:


ntheile (approved):

src/domain/errors.ts
```diff
@@ -128,9 +128,9 @@ export class InvalidIpMetadataError extends ValidationError {
 export class UnauthorizedIPForOnboardingError extends AuthorizationError {}
 
 export class UnauthorizedIPForRewardError extends AuthorizationError {}
-export class UnauthorizedIPMetadataProxyForRewardError extends UnauthorizedIPForRewardError {}
-export class UnauthorizedIPMetadataCountryForRewardError extends UnauthorizedIPForRewardError {}
-export class UnauthorizedIPMetadataASNForRewardError extends UnauthorizedIPForRewardError {}
+export class UnauthorizedIPMetadataProxyError extends UnauthorizedIPForRewardError {}
```
nicolasburtey, line 131:
```suggestion
export class UnauthorizedIPMetadataProxyError extends UnauthorizedIPError {}
```

nicolasburtey: if we are dropping `ForReward` we should also do it for the parent class

ntheile (approved):

PR #3168, opened 2023-09-06
feat: add rebalance to hot wallet

dolcalmi:


src/services/bria/index.ts
```diff
@@ -300,6 +305,47 @@ export const OnChainService = (): IOnChainService => {
...
+    submitPayoutRequest.setSatoshis(Number(amount.amount))
+    submitPayoutRequest.setMetadata(
+      constructMetadata({ galoy: { rebalanceToHotWallet: true } }),
+    )
+
+    const triggerPayoutQueueRequest = new TriggerPayoutQueueRequest()
+    triggerPayoutQueueRequest.setName(briaConfig.coldStorage.coldToHotRebalanceQueueName)
+
+    try {
+      const submitPayoutResponse = await submitPayout(submitPayoutRequest, metadata)
```
bodymindarts, line 326: We should probably check the result here before triggering

src/services/bria/index.ts
```diff
@@ -300,6 +305,47 @@ export const OnChainService = (): IOnChainService => {
...
+
+      await triggerPayoutQueue(triggerPayoutQueueRequest, metadata)
+
+      const payoutRequest = new GetPayoutRequest()
+      payoutRequest.setId(submitPayoutResponse.getId())
+      const getPayoutResponse = await getPayout(payoutRequest, metadata)
+
+      const foundPayout = getPayoutResponse.getPayout()
+      if (!foundPayout) return new PayoutNotFoundError()
+      const batchId = foundPayout.getBatchId()
```
bodymindarts, line 336: We may need to loop here - triggering the queue is not instant. The queue will get instantly scheduled but it still takes some time until it is processed.

An alternative would be to get the batch id from the `PayoutCommitted` event that should result from the queue getting processed.

src/services/bria/index.ts
```diff
@@ -300,6 +305,47 @@ export const OnChainService = (): IOnChainService => {
...
+      constructMetadata({ galoy: { rebalanceToHotWallet: true } }),
+    )
+
+    const triggerPayoutQueueRequest = new TriggerPayoutQueueRequest()
+    triggerPayoutQueueRequest.setName(briaConfig.coldStorage.coldToHotRebalanceQueueName)
+
+    try {
+      const submitPayoutResponse = await submitPayout(submitPayoutRequest, metadata)
+
+      await triggerPayoutQueue(triggerPayoutQueueRequest, metadata)
```
bodymindarts, line 328: Probably should check result (incase of error)

PR #3174, opened 2023-09-06
feat: exposing callback to graphql layer

nicolasburtey:
adding graphql endpoint for callback (add, remove, list)

for now, I'm not exposing the svix dashboard platform, as I don't think it's needed in the short term, and not exposing it will make it easier for us to host svix if we wanted to.


test/bats/callback.bats
```diff
@@ -0,0 +1,79 @@
...
+  if [[ "$(balance_for_check)" != 0 ]]; then
+    fail "Error: balance_for_check failed"
+  fi
+
+}
+
+btc_amount=1000
+usd_amount=50
+
+@test "callback endpoints" {
```
bodymindarts, line 42: Can't quite see - does this assert that the callback gets called?

bodymindarts (approved):

test/bats/callback.bats
nicolasburtey: you mean the 

```
btc_amount=1000
usd_amount=50
```

this has been removed

test/bats/callback.bats
```diff
@@ -0,0 +1,76 @@
...
+}
+
+teardown() {
+  if [[ "$(balance_for_check)" != 0 ]]; then
+    fail "Error: balance_for_check failed"
+  fi
+
+}
+
+@test "setup callback endpoints" {
```
bodymindarts, line 39: We have been using the convention to prefix the test name with the filename `callback: setup callback endpoints`.

bodymindarts (approved):

PR #3186, opened 2023-09-08
fix: trigger restart invoices issue

dolcalmi:
Uses similar strategy to setupPaymentSubscribe to avoid missing invoices when trigger is restarted/deployed

### DIFF
--- a/src/servers/trigger.ts
+++ b/src/servers/trigger.ts
@@ -322,6 +322,20 @@ const listenerExistingHodlInvoices = async ({
   }
 }
 
+export const setupAndProcessInvoiceSubscribe = ({
+  lnd,
+  pubkey,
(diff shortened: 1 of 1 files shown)


dolcalmi:
cron is declining usd invoices so this is not a solution

nicolasburtey:
@dolcalmi I had revamped this PR expecting it would now pass with this PR merged https://github.com/GaloyMoney/galoy/commit/a9f0a26ba4194b7546ea20f43747baf1cfc8bf86, but it doesn't. 

https://github.com/GaloyMoney/galoy/actions/runs/6183203185/job/16785119904?pr=3186

I'll let you dive into it

nicolasburtey (approved):

PR #3187, opened 2023-09-09
chore: extend session dynamically when requests are made

nicolasburtey:


src/app/authentication/extend-sessions.ts
```diff
@@ -80,3 +80,25 @@ export const extendSessions = async (
...
+  expiresAt,
+  monthsToExtend = 12,
+}: {
+  sessionId: string
+  expiresAt: string
+  monthsToExtend?: number
+}): Promise<boolean | void | Error> => {
+  const futureDate = new Date()
+  // related to session.lifespan value in kratos.yaml
+  futureDate.setMonth(futureDate.getMonth() + monthsToExtend)
```
bodymindarts, line 95: How is Month represented?
Does this work when we are crossing a year?
1 (Jan) + 12 = 13 (Jan - 1 year later)?

test/integration/app/authentication/extend.spec.ts
```diff
@@ -0,0 +1,41 @@
+import { maybeExtendSession } from "@app/authentication"
+import { extendSession } from "@services/kratos"
+
+jest.mock("@services/kratos", () => {
+  return {
+    extendSession: jest.fn(),
+  }
+})
+
+it("maybe extend session", async () => {
```
bodymindarts, line 10: Can we test to assert that skipping a 1 year boundary behaved as expected?

src/app/authentication/extend-sessions.ts
bodymindarts: I would not say that the interface `setMonth` can mutate the year is intuitive. But fair enough.

bodymindarts (approved):

PR #3193, opened 2023-09-11
refactor: simplify display currency hybrid

vindard:
## Description

This is to remove the partial `DisplayCurrency` enum and to have it only be a "unique symbol" type with display currencies loaded dynamically.


bodymindarts:
needs rebase @vindard

dolcalmi (approved):
looks good :thinking: although not sure about the purpose of this change

PR #3194, opened 2023-09-11
fix: unkown invalid account id error

dolcalmi:
Fix unknown (CouldNotFindAccountError) error https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/CE42VvXhMqi


nicolasburtey:
@bodymindarts had something to say about it

PR #3207, opened 2023-09-13
feat: add push notification filtering settings

UncleSamtoshi:
- Adds the ability to set and retrieve push notification filters
- Push Notifications will be on by default and before they are sent we will check to make sure notifications are enabled and that the individual notification type and subtype is not disabled
- NOTE: This is implemented using a string for NotificationType instead of an enum. This allows for the notification service to not need any context about the notifications that are passing through it.


src/app/accounts/update-push-notification-settings.ts
```diff
@@ -0,0 +1,31 @@
+import { checkedToPushNotificationSettings } from "@domain/notifications"
```
nicolasburtey, line 1: we have a `app/comm` folder. should probably be there
comm for `communication`. we could use `communication` if `comm` is unclear.

src/graphql/public/schema.graphql
```diff
@@ -79,6 +80,16 @@ type AccountUpdateDisplayCurrencyPayload {
   errors: [Error!]!
 }
 
+input AccountUpdatePushNotificationSettingsInput {
+  enabled: Boolean!
+  settings: [PushNotificationSettingsInput]!
```
nicolasburtey, line 85: I don't quite understand the shape of the input. why is enabled passed at the root?

nicolasburtey (commented):
some comments

src/graphql/public/schema.graphql
UncleSamtoshi: The settings are a combination of "NotificationType" filters that live under `settings` and a general on/off switch for all notifications called `enabled`. Maybe I should change `settings` to be `settingsByNotificationType` or even `filters`.

src/app/accounts/update-push-notification-settings.ts
UncleSamtoshi: `comm` is unclear to me, and I would support moving to `communication`. I could still see this mutation live under the `account` use cases because you are storing account info.

src/domain/primitives/index.types.d.ts
```diff
@@ -11,6 +11,7 @@ type MilliSeconds = number & { readonly brand: unique symbol }
 type Days = number & { readonly brand: unique symbol }
 type JwtToken = string & { readonly brand: unique symbol } // short lived asymmetric token from oathkeeper
 type Memo = string & { readonly brand: unique symbol }
+type PushNotificationType = string & { readonly brand: unique symbol }
```
nicolasburtey, line 14:
```suggestion
type PushNotification = string & { readonly brand: unique symbol }
```

src/graphql/public/root/mutation/account-update-push-notification-settings.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const AccountUpdatePushNotificationSettingsInput = GT.Input({
+  name: "AccountUpdatePushNotificationSettingsInput",
+  fields: () => ({
+    pushNotificationsEnabled: {
+      type: GT.NonNull(GT.Boolean),
+    },
+    disabledPushNotificationTypes: {
+      type: GT.NonNull(GT.List(PushNotificationType)),
+    },
```
nicolasburtey, line 16: I still think this should be segmented into 2 mutations. 

1 mutation = 1 action. 

here both argument are mandatory. so action such as:

"I want to deactivate one notification": it is unclear whether the `pushNotificationsEnabled` should be set to true or false.

src/graphql/public/schema.graphql
```diff
@@ -6,6 +6,7 @@ interface Account {
   id: ID!
   level: AccountLevel!
   limits: AccountLimits!
+  pushNotificationSettings: PushNotificationSettings!
```
nicolasburtey, line 9: I don't think we should make it dedicate to pushNotification. I think for instance we will want to activate notification by email also (in a follow up PR).

this could be something like:

```
notifications: {
   push: []
   email: []
}
```

src/graphql/public/schema.graphql
```diff
@@ -79,6 +80,16 @@ type AccountUpdateDisplayCurrencyPayload {
   errors: [Error!]!
 }
 
+input AccountUpdatePushNotificationSettingsInput {
+  disabledPushNotificationTypes: [PushNotificationType]!
+  pushNotificationsEnabled: Boolean!
+}
+
+type AccountUpdatePushNotificationSettingsPayload {
+  account: ConsumerAccount
```
nicolasburtey, line 89: do we not use `Account` instead for the other similar queries?

src/graphql/public/schema.graphql
```diff
@@ -79,6 +80,16 @@ type AccountUpdateDisplayCurrencyPayload {
   errors: [Error!]!
 }
 
+input AccountUpdatePushNotificationSettingsInput {
+  disabledPushNotificationTypes: [PushNotificationType]!
```
nicolasburtey, line 84: typically we also want to make the inner mandatory (with another !)

the difference is that:
with the outer, the argument must exist, with the inner, the argument must be an array. 

if you just had the inner, it would say that if the argument exist, it must be an array, but it could be undefined also. 

```suggestion
  disabledPushNotificationTypes: [PushNotificationType!]!
```

nicolasburtey (commented):
I still have the same reservation on the ambiguity regarding the usage of the mutations with the current API.
the query looks fine.

src/graphql/public/schema.graphql
UncleSamtoshi: Will make this change.

src/graphql/public/schema.graphql
UncleSamtoshi: At the moment we use `ConsumerAccount` for the other similar queries but maybe this is by mistake?

src/graphql/public/root/mutation/account-update-push-notification-settings.ts
UncleSamtoshi: I think it depends on how you view the `1 action`. Going to the notification settings page and then turning on or off a few notifications and then sending your new notification settings when you leave the page could be seen as one action "updating your setting". Otherwise you could look at each individual toggle as 1 action. I'm open to either one. If you split this into two queries, would you then add two more queries for enabling and disabling email and selecting notifications that are disabled in email?

src/graphql/public/schema.graphql
UncleSamtoshi: If we are planning on doing this soon, It may make sense to do it before we release a version of the app with the notification settings that would cause a breaking change.

src/graphql/public/schema.graphql
```diff
@@ -29,6 +30,24 @@ type AccountDeletePayload {
...
+  channel: NotificationChannel!
+}
+
+input AccountDisableNotificationChannelInput {
+  channel: NotificationChannel!
+}
+
+input AccountEnableNotificationCategoryForChannelInput {
+  category: NotificationCategory!
+  channel: NotificationChannel!
```
nicolasburtey, line 44: I'm wondering if channel should be mandatory. if it was optional, then the mutation would apply to all channels

src/domain/notifications/index.ts
```diff
@@ -8,3 +10,137 @@ export const NotificationType = {
...
+  notificationChannel,
+  notificationCategory,
+}: {
+  notificationSettings: NotificationSettings
+  notificationChannel?: NotificationChannel
+  notificationCategory: NotificationCategory
+}): NotificationSettings => {
+  const notificationChannelsToUpdate: NotificationChannel[] = notificationChannel
+    ? [notificationChannel]
+    : Object.values(NotificationChannel)
```
nicolasburtey, line 106: what is the purpose of this?

src/domain/notifications/index.ts
```diff
@@ -8,3 +10,137 @@ export const NotificationType = {
...
+} as const
+
+export const GaloyNotificationCategories = {
+  Payments: "Payments" as NotificationCategory,
+  Balance: "Balance" as NotificationCategory,
+  AdminPushNotification: "AdminPushNotification" as NotificationCategory,
+} as const
+
+export const checkedToNotificationCategory = (
+  type: string,
```
nicolasburtey, line 25: reading this make me feel like `type` can be a keyword conflicting with js base language. maybe we could use notificationType here?

src/app/accounts/disable-notification-channel.ts
```diff
@@ -0,0 +1,23 @@
...
+  accountId,
+  notificationChannel,
+}: {
+  accountId: AccountId
+  notificationChannel: NotificationChannel
+}): Promise<Account | ApplicationError> => {
+  const account = await AccountsRepository().findById(accountId)
+  if (account instanceof Error) return account
+
+  const newNotificationSettings = setNotificationChannelIsEnabled({
```
nicolasburtey, line 14: name not very consistent with `src/app/accounts/disable-notification-category.ts` where the main action is `domainDisableNotificationCategory` (ie: not starting with set)

src/app/admin/send-admin-push-notification.ts
```diff
@@ -1,4 +1,8 @@
 import { checkedToAccountUuid } from "@domain/accounts"
+import {
+  GaloyNotificationCategories,
```
nicolasburtey, line 3:
```suggestion
  galoyNotificationCategories,
```

src/domain/notifications/index.ts
```diff
@@ -8,3 +10,137 @@ export const NotificationType = {
   OnchainPayment: "onchain_payment",
   LnInvoicePaid: "paid-invoice",
 } as const
+
+export const NotificationChannel = {
+  Push: "push",
+} as const
+
+export const GaloyNotificationCategories = {
+  Payments: "Payments" as NotificationCategory,
```
nicolasburtey, line 19: no Circles?

src/services/notifications/push-notifications.ts
```diff
@@ -112,7 +115,40 @@ export const PushNotificationsService = (): IPushNotificationsService => {
...
+      return {
+        status: SendFilteredPushNotificationStatus.Filtered,
+      }
+    }
+
+    const result = await sendNotification({
+      ...sendNotificationArgs,
+      data: {
+        ...data,
+        NotificationCategory: notificationCategory,
```
nicolasburtey, line 138: not sure what is supposed to be inside { data } versus at the root level (ie: `{ ...sendNotification, ... }`) but reading here would make me think  `NotificationCategory: notificationCategory,` is more a metadata thing and should be one level up in the root

test/legacy-integration/notifications/push-notification.spec.ts
```diff
@@ -0,0 +1,28 @@
+import {
```
nicolasburtey, line 1: I'm wondering how much we should add to the `legacy-integration`. We're aiming at trimming this folder down. @vindard  would have more context

nicolasburtey:
great PR. looks good overall, some nits to look at.

nicolasburtey (approved):
approving, even if this may need change. someelse can reapprove later.

src/domain/notifications/index.ts
UncleSamtoshi: This is the behavior that you requested where it will update all channels if a specific channel in not set.

src/domain/notifications/index.ts
UncleSamtoshi: Will change the name to `notificationCategory` as we already use `notificationType` elsewhere.

src/domain/notifications/index.ts
UncleSamtoshi: These are notification categories that Galoy core knows about. "Circles" will be a NotificationCategory but one that is passed through the graphql api and not one that is known about in our core domain.

src/services/notifications/push-notifications.ts
UncleSamtoshi: That makes sense, I will move it up a level.

src/app/accounts/disable-notification-channel.ts
UncleSamtoshi: I will make the names consistent

src/services/notifications/push-notifications.ts
UncleSamtoshi: Actually I think it may make sense to keep it where it is at. It is preferable that all notifications that have a notification category receive it as a part of their metadata. Since this method already has logic relating to handling notification categories I think it is alright for it to augment the metadata.

nicolasburtey (approved):

PR #3209, opened 2023-09-14
fix: cron usd invoices processing

dolcalmi:


test/unit/domain/wallet-invoices/wallet-invoice-validator.spec.ts
```diff
@@ -17,6 +17,7 @@ describe("WalletInvoiceValidator", () => {
       amount: BigInt(10),
     },
     paid: false,
+    createdAt: new Date(Date.now()),
```
bodymindarts, line 20: Are you sure you need the double Date invocation? Doesn’t now return the correct type? Or new default to now?

dolcalmi: I was having issues with legacy tests (with milliseconds). I will check if it is too much update the tests

nicolasburtey: I'm merging this for now as it fix an important bug in production. we can iterate on the test if needed in a follow up PR

nicolasburtey (approved):

PR #3219, opened 2023-09-15
refactor: collect common bria handler errors into functions

vindard:
### Description

Collected some common error handling into functions so that we don't mistakenly miss handling a particular error (e.g. #3178) that can freeze the queue.

### DIFF
--- a/src/servers/event-handlers/bria.ts
+++ b/src/servers/event-handlers/bria.ts
@@ -68,17 +68,7 @@ export const utxoDetectedEventHandler = async ({
   event: UtxoDetected
 }): Promise<true | DomainError> => {
   const res = await Wallets.addPendingTransaction(event)
-  if (res instanceof AmountLessThanFeeError) {
-    return true
-  }
-  if (res instanceof LessThanDustThresholdError) {
-    return true
-  }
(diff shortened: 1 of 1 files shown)


src/servers/event-handlers/bria.ts
```diff
@@ -160,6 +132,24 @@ export const payoutSettledEventHandler = async ({
   event: PayoutSettled
 }): Promise<true | ApplicationError> => {
   const res = await Wallets.settlePayout(event.id)
+  return handlePayoutHandlerResult(res)
+}
+
+const handleIncomingUtxoHandlerResult = (res: true | ApplicationError) => {
```
bodymindarts, line 138: I think we discussed one time this distinction actually belongs in the use case - not in the top level handler.

bodymindarts (approved):

src/servers/event-handlers/bria.ts
vindard: > I think we discussed one time this distinction actually belongs in the use case - not in the top level handler.

Yea true. I think I'll close this one for now then and tackle that as a separate thing the next time it comes up. Don't want to further normalize this sort of handling here.

vindard:
Closing in favour of handling these errors within the use-cases instead

PR #3222, opened 2023-09-15
chore: optionally pass a deviceToken argument when logging out

### LINKED ISSUE #1388: Notification tokens
It seems like when a user logs out and back in to the mobile app, we store a new token without deleting the old one and that causes the app to notify the user multiple times (for the daily balance, for example)

nicolasburtey:
fix #1388

### DIFF
--- a/src/app/authentication/logout.ts
+++ b/src/app/authentication/logout.ts
@@ -1,8 +1,15 @@
+import { removeDeviceTokens } from "@app/users/remove-device-tokens"
 import { AuthWithPhonePasswordlessService, MissingSessionIdError } from "@services/kratos"
 
-export const logoutToken = async (
-  sessionId: SessionId | undefined,
-): Promise<boolean | ApplicationError> => {
+export const logoutToken = async ({
+  userId,
+  deviceToken,
+  sessionId,
+}: {
+  userId: UserId
+  deviceToken: DeviceToken | undefined
+  sessionId: SessionId | undefined
+}): Promise<boolean | ApplicationError> => {
   if (!sessionId) return new MissingSessionIdError()
 
   const authService = AuthWithPhonePasswordlessService()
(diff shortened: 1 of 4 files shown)


src/app/authentication/logout.ts
```diff
@@ -1,15 +1,27 @@
...
   if (!sessionId) return new MissingSessionIdError()
 
   const authService = AuthWithPhonePasswordlessService()
   const kratosResult = await authService.logoutToken({ sessionId })
   if (kratosResult instanceof Error) {
     return kratosResult
   }
+
+  if (deviceToken) {
+    await removeDeviceTokens({ userId, deviceTokens: [deviceToken] })
```
dolcalmi, line 22: dont we need to return the error?

src/graphql/public/schema.graphql
```diff
@@ -724,7 +724,7 @@ type Mutation {
   userEmailRegistrationValidate(input: UserEmailRegistrationValidateInput!): UserEmailRegistrationValidatePayload!
   userLogin(input: UserLoginInput!): AuthTokenPayload!
   userLoginUpgrade(input: UserLoginUpgradeInput!): UpgradePayload!
-  userLogout: SuccessPayload!
+  userLogout(input: UserLogoutInput!): SuccessPayload!
```
dolcalmi, line 727: if input is required it will be a breaking change

dolcalmi (changes requested):

dolcalmi (approved):

src/graphql/public/schema.graphql
nicolasburtey: > if input is required it will be a breaking change. is this mutation not implemented in mobile/web wallet?

this was not used so we're good

src/app/authentication/logout.ts
nicolasburtey: don't think that's useful or necessary. main action of logout would be to deactivate the session.

PR #3241, opened 2023-09-19
fix: usd onchain incoming pending balance gql query

vindard:
Merge after #3225


vindard:
Having some issues changing the target branch for merge here

PR #3250, opened 2023-09-20
refactor: use relative path for envrc.secrets

krtk6160:


krtk6160:
No longer necessary following #3249

PR #3255, opened 2023-09-21
chore: more debugging for transient_payload

nicolasburtey:


src/servers/event-handlers/kratos.ts
```diff
@@ -31,11 +32,11 @@ kratosCallback.post(
     fn: async (req: express.Request, res: express.Response) => {
       const secret = req.headers.authorization
       const body = req.body
-      addAttributesToCurrentSpan({
```
nicolasburtey, line 34: for some reasons, this span does show up

nicolasburtey: maybe because `addAttributesToCurrentSpan` inside `wrapAsyncToRunInSpan` ? haven't tried to debug it too much

PR #3277, opened 2023-09-28
chore: remove legacy volume types

vindard:


src/services/ledger/volume.ts
```diff
@@ -39,99 +39,100 @@ export const TxnGroups = {
   allTxBaseVolumeSince: Object.values(LedgerTransactionType),
 } as const
 
-const volumeFn =
-  (txnGroup: TxnGroup): GetVolumeSinceFn =>
-  async (args) =>
-    txVolumeSince({ ...args, txnGroup })
+const TxVolumeAmountSinceFactory = () => {
```
dolcalmi, line 42: should we use an interface? not a blocker

dolcalmi (approved):

bodymindarts (approved):

PR #3282, opened 2023-09-30
feat: add new usd-invoice-sats-denominated mutation

vindard:


src/app/wallets/add-invoice-for-wallet.ts
```diff
@@ -208,6 +261,7 @@ const addInvoice = async ({
   const dealer = DealerPriceService()
   const walletInvoiceBuilder = WalletInvoiceBuilder({
     dealerBtcFromUsd: dealer.getSatsFromCentsForFutureBuy,
+    dealerUsdFromBtc: dealer.getCentsFromSatsForFutureBuy,
```
vindard, line 264: **This is a critical addition here**

Is this the right dealer method to do the conversion without introducing new 1-cent attacks?

src/domain/wallet-invoices/wallet-invoice-builder.ts
```diff
@@ -72,22 +70,45 @@ export const WIBWithRecipient = (state: WIBWithRecipientState): WIBWithRecipient
...
+        })
+      }
+
+      const btcAmount = amount as BtcPaymentAmount
+      const usdAmount = await state.dealerUsdFromBtc(btcAmount)
       if (usdAmount instanceof Error) return usdAmount
-      const btcAmount = await state.dealerBtcFromUsd(usdAmount)
-      if (btcAmount instanceof Error) return btcAmount
-
+      if (usdAmount.amount === 0n) {
```
vindard, line 90: We will need to refactor the WalletInvoiceBuilder to be able to switch to the recipient's BTC wallet in this instance, like we did with our payment-flow-builder objects (can be done in a follow-up PR)

bodymindarts (approved):

PR #3288, opened 2023-10-01
chore: converting consent to nextjs

siddhart1o1:


apps/consent-next.js/app/consent/page.tsx
```diff
@@ -0,0 +1,208 @@
...
+  "use server";
+  const consent_challenge = form.get("consent_challenge");
+  const submitValue = form.get("submit");
+  const remember = form.get("remember") === "1";
+  const grantScope = form
+    .getAll("grant_scope")
+    .map((value) => value.toString());
+  // TODO add check from email code.
+
+  if (typeof consent_challenge === "string" && consent_challenge) {
```
nicolasburtey, line 20: I think the if should testing the negative here and return an error if not valid

This would make the code cleaner as it would:
- return an error if condition is not true (we don't do anything here)
- avoid indenting the rest of the function. This makes it easier to reason about

apps/consent-next.js/app/logout/page.tsx
```diff
@@ -0,0 +1,76 @@
...
+        logoutChallenge: logout_challenge,
+      });
+      redirect("https://www.blink.sh/");
+    }
+    const response = await hydraClient.acceptOAuth2LogoutRequest({
+      logoutChallenge: logout_challenge,
+    });
+    redirect(String(response.data.redirect_to));
+  }
+};
```
nicolasburtey, line 31: Should return an error otherwise

PR #3292, opened 2023-10-02
chore: removing legacy login admin panel

nicolasburtey:


core/api/src/app/accounts/mark-account-for-deletion.ts
```diff
@@ -8,11 +8,11 @@ import { addEventToCurrentSpan } from "@/services/tracing"
 export const markAccountForDeletion = async ({
   accountId,
   cancelIfPositiveBalance = false,
-  updatedByUserId,
+  updatedByAuditorId,
```
nicolasburtey, line 11:
```suggestion
  updatedByPriviledgedClientId,
```

bodymindarts (approved):

PR #3294, opened 2023-10-02
feat: basic support for buck2

bodymindarts:


toolchains/simple-pnpm/macros.bzl
```diff
@@ -0,0 +1,103 @@
...
+        "_simple_pnpm_toolchain": attrs.toolchain_dep(
+            default = "toolchains//:simple_pnpm",
+            providers = [SimplePnpmToolchainInfo],
+        ),
+    },
+)
+
+def npm_bin(
+        visibility = ["PUBLIC"],
+        node_modules = ":node_modules",
```
bodymindarts, line 97: @vindard 

I have found a legit reason to have a macro wrapper. In the `attrss ={}` section above you can only specify a `default` target as a dependency that has an aboslute path (like `toolchain//:simple_pnpm`) not a relative one (like `:node_modules` which is in the same CELL).

Still no need to have completely other file just for this IMO.

PR #3301, opened 2023-10-05
chore: update mongodb version

dolcalmi:
merge after charts update

### DIFF
--- a/quickstart/docker-compose.yml
+++ b/quickstart/docker-compose.yml
@@ -135,7 +135,7 @@ services:
     - ALLOW_EMPTY_PASSWORD=yes
     - REDIS_DISABLE_COMMANDS=FLUSHDB,FLUSHALL
   mongodb:
-    image: mongo:${MONGODB_VERSION:-6.0.5}
+    image: mongo:${MONGODB_VERSION:-7.0.2}
     ports:
     - 27017:27017
     environment:
--- a/quickstart/docker-compose.yml
+++ b/quickstart/docker-compose.yml
@@ -135,7 +135,7 @@ services:
     - ALLOW_EMPTY_PASSWORD=yes
     - REDIS_DISABLE_COMMANDS=FLUSHDB,FLUSHALL
   mongodb:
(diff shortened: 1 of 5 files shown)


quickstart/docker-compose.yml
```diff
@@ -135,7 +135,7 @@ services:
     - ALLOW_EMPTY_PASSWORD=yes
     - REDIS_DISABLE_COMMANDS=FLUSHDB,FLUSHALL
   mongodb:
-    image: mongo:${MONGODB_VERSION:-6.0.5}
+    image: mongo:${MONGODB_VERSION:-7.0.2}
```
bodymindarts, line 138: this should not be updated manually. It happens in CI.

quickstart/galoy/docker-compose.yml
```diff
@@ -102,7 +102,7 @@ services:
     environment:
       - MONGODB_ADDRESS=mongodb
   mongodb:
-    image: mongo:${MONGODB_VERSION:-6.0.5}
+    image: mongo:${MONGODB_VERSION:-7.0.2}
```
bodymindarts, line 105: this should not be updated manually. It happens in CI.

nicolasburtey: don't think there is any harm is updating them manually?

but it's another argument to not have the computed quickstart in the same branch but in another branch. it make search/replace harder in practive if we want to enforce the rule of "this should not be updated manually"

quickstart/docker-compose.yml
dolcalmi: Fixed https://github.com/GaloyMoney/galoy/pull/3301/commits/ca426c4de10bd79733d73deb51d87b5bd541469f

quickstart/galoy/docker-compose.yml
dolcalmi: I remove the update.. just in case https://github.com/GaloyMoney/galoy/pull/3301/commits/ca426c4de10bd79733d73deb51d87b5bd541469f

bodymindarts (approved):

PR #3308, opened 2023-10-05
build: simplify tsconfig

bodymindarts:


core/api/tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "sourceMap": true,
     "strict": true,
     "target": "es2020",
-    "noImplicitAny": false,
+    "noImplicitAny": true,
```
nicolasburtey, line 8: it's also the default when `strict: true` so could be removed

bodymindarts: Unfortunately it won't work due to the tests.

nicolasburtey (approved):

Issue #1203, opened 2022-03-28
Separate a proper `WalletTransaction` type in our domain 


vindard:
## Background

We currently return "transactions" to users via our API to be used for different purposes. Most commonly they are used to view a given wallet's transaction history as in our mobile app, but we are starting to see different cases where an end user would want to fetch a transaction or set of transactions, e.g.:
- to see the details of a sent transaction for a given lightning invoice (maybe to retrieve the preimage)
- to see the confirmation status of an incoming or outgoing onchain transaction

### Current state

The transactions we return via our API (for e.g. in our `main` query) are typed as `WalletTransaction` but they are almost one-to-one mappings of `LedgerTransaction` entries in our persistence layer. This becomes tricky to parse meaning as an end-user in cases like:
- a payment is attempted and then fails and the user receives a reimbursement
- a payment is attempted without a fee probe and a fee reimbursement transaction is done
- a payment is attempted multiple times and fails but then eventually succeeds

It also leads to the conflation of different concepts under the same abstraction, e.g.:
- `TxStatus` is used to signal both a transaction's actual completion status (e.g. succeeded/failed payments) and the ledger entry's recorded status (whether an entry is recorded or not) for different entries that make up the transaction

Finally, it can return internal details about how a transaction's life cycle that isn't necessarily relevant to an end-user and in some cases can even be confusing.

---

## Tasks

To solve this issue we can look into separating the `WalletTransaction` type more cleanly from the `LedgerTransaction` type when translating these entities in our domain. We can probably do this in stages, starting with a new `WalletTransactionSummary` type that is return from queries targeted at single transactions (most likely fetched using the transaction hash).

### New `TransactionByHash` query
A good starting point for this could be a new `TransactionByHash` query at the account level that returns a summary for the given transaction hash. We can then implement some sort of aggregation algorithm that can update many different transaction types e.g.:
- standard payments
- non-probed payments (transactions with fee reimbursements)
- failed transactions (2 ledger entries)
- retried successful transactions (failed attempts that eventually succeed)

These conversions should be implemented within our core domain and unit tested. 

### `transactions` selection in `me` query
We can eventually try to move queries/mutations that currently return lists of transaction to return the new `WalletTransactionSummary` types instead of the `LedgerTransaction`-like types they currently do.

### Remove the `WalletTransactionSummary` type
Once the places where `WalletTransaction` is currently used have all been replaced, we can eventually simply rename `WalletTransactionSummary` to `WalletTransaction` across the codebase.

PR #3313, opened 2023-10-06
feat: better expose invoices through gql api

UncleSamtoshi:
- add invoiceByPaymentHash to Wallet gql object
- add walletById to Account gql object
- addTransactionByPaymentHash to Wallet gql object
- addTransactionById to Wallet gql object
- Create IInvoice gql interface
- Remove LnAndWalletInvoice domain type and add LnInvoice to WalletInvoice


core/api/src/graphql/public/types/payload/ln-invoice.ts
```diff
@@ -1,6 +1,6 @@
 import IError from "../../../shared/types/abstract/error"
 
-import LnInvoice from "../object/ln-invoice"
+import LnInvoice from "../../../shared/types/object/ln-invoice"
```
nicolasburtey, line 3: shared object should be in /shared/.

this may mean that `core/api/src/graphql/public/types/payload/ln-invoice.ts` may need to be in `core/api/src/graphql/shared/types/payload/ln-invoice.ts` if it imports an object from shared (maybe, not sure)

nicolasburtey: seems they have been renamed already.

core/api/src/app/wallets/index.ts
```diff
@@ -16,14 +16,35 @@ export * from "./settle-payout-txn"
...
+}: {
+  accountId: AccountId
+  walletId: WalletId
+}): Promise<Wallet | ApplicationError> => {
+  const wallets = WalletsRepository()
+  const wallet = await wallets.findById(walletId)
+
+  if (wallet instanceof Error) return wallet
+
+  if (wallet.accountId !== accountId) {
```
nicolasburtey, line 41: wallet.accountId is probably accountId but should be accountUuid?

I think we may want to do the AccountId/Uuid refactoring (ie: removing the use of `account._id` and solely expose `account.uuid`) before merging this PR?

core/api/src/app/wallets/index.ts
```diff
@@ -16,14 +16,35 @@ export * from "./settle-payout-txn"
...
 
+import { CouldNotFindWalletFromIdError } from "@/domain/errors"
 import { WalletsRepository } from "@/services/mongoose"
 
 export const getWallet = async (walletId: WalletId) => {
   const wallets = WalletsRepository()
   return wallets.findById(walletId)
 }
 
+export const getWalletForAccountById = async ({
```
nicolasburtey, line 29: is it an authorization related function? maybe should be named as such if that the intention of this function

core/api/src/app/wallets/index.ts
nicolasburtey: working on this PR: https://github.com/GaloyMoney/galoy/pull/3316

core/api/src/app/wallets/index.ts
UncleSamtoshi: What would you change the name to? The function will look for a wallet that belongs to an account. If the wallet exists but does not belong to that account, it will return a CouldNotFindWalletError.

nicolasburtey: maybe `checkWalletForAccountById`?

UncleSamtoshi: Not sure if that works here. It's not a check that returns true or false, it either returns a wallet or a CouldNotFindWalletError.

core/api/src/graphql/shared/types/object/usd-wallet.ts
```diff
@@ -123,6 +126,75 @@ const UsdWallet = GT.Object<Wallet>({
         )
       },
     },
+    invoiceByHash: {
+      type: GT.NonNull(LnInvoice),
+      args: {
+        paymentHash: {
+          type: GT.NonNull(PaymentHash),
+        },
+      },
```
nicolasburtey, line 135: description missing (compare to the btc wallet)

core/api/src/graphql/admin/schema.graphql
```diff
@@ -443,7 +460,10 @@ interface Wallet {
   accountId: ID!
   balance: SignedAmount!
   id: ID!
+  invoiceByHash(paymentHash: PaymentHash!): LnInvoice!
   pendingIncomingBalance: SignedAmount!
+  transactionByHash(hash: PaymentHash!): Transaction!
```
nicolasburtey, line 465: should the type be mandatory `Transaction!`, or should it be optional? what happen when the transaction/invoice doesn't exist?

nicolasburtey: current the type we have in the admin panel is not mandatory (not saying it should one or the other, not sure yet, but I think it should be consistent)

core/api/src/graphql/admin/schema.graphql
```diff
@@ -82,8 +82,15 @@ type BTCWallet implements Wallet {
...
   id: ID!
 
+  """
+  The lightning invoice with the matching paymentHash belonging to this wallet.
+  """
+  invoiceByHash(paymentHash: PaymentHash!): LnInvoice!
+
   """An unconfirmed incoming onchain balance."""
   pendingIncomingBalance: SignedAmount!
+  transactionByHash(hash: PaymentHash!): Transaction!
```
nicolasburtey, line 92: could there be multiple transaction by hash?

can hash be only a lightning hash? or could it be something else (like txid?) 

I think some description here would be relevant

there are cases where we can return multiple transaction per hash.

- on lightning, if multiple payment attempt were made (not too sure actually)
- on chain, if address is been reused

core/api/src/graphql/admin/schema.graphql
```diff
@@ -196,6 +203,13 @@ type LightningPayment {
   status: LnPaymentStatus
 }
 
+type LnInvoice {
+  paymentHash: PaymentHash!
```
nicolasburtey, line 207: no status?

core/api/src/graphql/admin/schema.graphql
```diff
@@ -196,6 +203,13 @@ type LightningPayment {
   status: LnPaymentStatus
 }
 
+type LnInvoice {
```
nicolasburtey, line 206: shouldn't we re-use the type `LightningInvoice` ?

core/api/src/graphql/admin/schema.graphql
UncleSamtoshi: I will add status now that I have refactored the types

core/api/src/graphql/admin/schema.graphql
UncleSamtoshi: I've switched this to be `transactionByPaymentHash` so that it only for lightning invoices. This makes it guaranteed to a single transaction. I have also made it nullable.

core/api/dev/apollo-federation/supergraph.graphql
```diff
@@ -545,6 +550,20 @@ input IntraLedgerUsdPaymentSendInput
...
+interface Invoice
+  @join__type(graph: PUBLIC)
+{
+  """The payment hash of the lightning invoice."""
+  paymentHash: PaymentHash!
+
+  """The bolt11 invoice to be paid."""
+  paymentRequest: LnPaymentRequest!
+
+  """The payment secret of the lightning invoice."""
```
nicolasburtey, line 563: There can be confusion between secret and preimage. Is this really the secret here? We are not exposing the image? We should/could after the payment has completed

nicolasburtey (commented):
Question on invoice interface

core/api/dev/apollo-federation/supergraph.graphql
UncleSamtoshi: The [domain type](https://github.com/GaloyMoney/galoy/blob/f052151cdd4ec01b7e418c410be284e89b8df275/core/api/src/domain/bitcoin/lightning/index.types.d.ts#L110) of that field is `PaymentIdetifyingSecret` which I dont think is the preimage, but I could be mistaken. It is confusing because we call the preimage the `secret` on the [WalletInvoice type](https://github.com/GaloyMoney/galoy/blob/f052151cdd4ec01b7e418c410be284e89b8df275/core/api/src/domain/wallet-invoices/index.types.d.ts#L87) . For this pr, I haven't changed the value we are exposing for this field.

nicolasburtey: At least we can clarify with the description

nicolasburtey: I think the preimage is actually not that useful for the invoice/receive side, but it's important on the sender side, because payment request (signed by the recipient) + preimage = proof of payment

core/api/src/graphql/shared/types/object/ln-invoice.ts
```diff
@@ -0,0 +1,45 @@
...
+      resolve: (source) => source.lnInvoice.paymentSecret,
+    },
+    satoshis: {
+      type: SatAmount,
+      resolve: (source) => source.lnInvoice.amount,
+    },
+    paymentStatus: {
+      type: GT.NonNull(InvoicePaymentStatus),
+      resolve: (source) => {
+        const statusCheker = WalletInvoiceStatusChecker(source)
```
vindard, line 37: Typo

core/api/src/domain/bitcoin/lightning/index.types.d.ts
```diff
@@ -110,7 +110,7 @@ type LnInvoice = {
   readonly paymentSecret: PaymentIdentifyingSecret | null
   readonly features: LnInvoiceFeature[]
   readonly expiresAt: Date
-  readonly isExpired: boolean
+  readonly isExpired: boolean // Should we remove this because it can become stale?
```
vindard, line 113: I think this can stay for now since we use it in a live sort of way when we `decodeInvoice` and assign the value against time at function call

vindard (commented):
Small comments

core/api/src/app/wallets/index.ts
```diff
@@ -16,14 +16,35 @@ export * from "./settle-payout-txn"
...
+  accountId,
+  walletId,
+}: {
+  accountId: AccountId
+  walletId: WalletId
+}): Promise<Wallet | ApplicationError> => {
+  const wallets = WalletsRepository()
+  const wallet = await wallets.findById(walletId)
```
vindard, line 37: Same here with new `findByIdAndAccountId` to avoid instantiating `CouldNotFindWalletFromIdError` here

vindard (commented):
Found some potential boundary violations. Maybe @bodymindarts can confirm if these comments make sense?

vindard (commented):
Some comments on the LnInvoice graphql type and resolvers

core/api/src/graphql/shared/types/object/ln-noamount-invoice.ts
```diff
@@ -0,0 +1,39 @@
...
+      type: GT.NonNull(LnPaymentRequest),
+      resolve: (source) => source.lnInvoice.paymentRequest,
+    },
+    paymentHash: {
+      type: GT.NonNull(PaymentHash),
+      resolve: (source) => source.lnInvoice.paymentHash,
+    },
+    paymentSecret: {
+      type: GT.NonNull(LnPaymentSecret),
+      resolve: (source) => source.lnInvoice.paymentSecret,
```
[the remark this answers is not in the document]
UncleSamtoshi: removing this would be a breaking change to the api

core/api/src/graphql/shared/types/object/ln-noamount-invoice.ts
```diff
@@ -0,0 +1,39 @@
...
+  interfaces: () => [IInvoice],
+  isTypeOf: (source) => !source.lnInvoice.amount,
+  fields: () => ({
+    paymentRequest: {
+      type: GT.NonNull(LnPaymentRequest),
+      resolve: (source) => source.lnInvoice.paymentRequest,
+    },
+    paymentHash: {
+      type: GT.NonNull(PaymentHash),
+      resolve: (source) => source.lnInvoice.paymentHash,
```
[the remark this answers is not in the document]
UncleSamtoshi: If lnInvoice is undefined it would have returned a repository error I believe

core/api/src/app/wallets/get-transactions-by-hash.ts
```diff
@@ -3,6 +3,30 @@ import { WalletTransactionHistory } from "@/domain/wallets"
...
+  const ledgerTransactions = await ledger.getTransactionsByHash(paymentHash)
+  if (ledgerTransactions instanceof Error) return ledgerTransactions
+
+  const transactions = WalletTransactionHistory.fromLedger({
+    ledgerTransactions,
+    nonEndUserWalletIds: Object.values(await getNonEndUserWalletIds()),
+    memoSharingConfig,
+  }).transactions
+
+  const transaction = transactions.find(
```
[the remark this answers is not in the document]
UncleSamtoshi: will multiple payment retries result in multiple transactions in the ledger? Is a fee reimbursement tagged with payment hash?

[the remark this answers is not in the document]
vindard: Yes to both

core/api/src/graphql/shared/types/object/ln-noamount-invoice.ts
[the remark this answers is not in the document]
vindard: Ah right, this was here before. Ok nevermind then for now

core/api/src/app/wallets/get-transactions-by-hash.ts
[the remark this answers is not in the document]
vindard: This likely only applies for sent ln payments. To see it in action, checkout this commit and follow the instructions in the commit message: https://github.com/GaloyMoney/galoy/commit/50965f837868b9870fbdeaa77eee9bf1127e94a6

The behaviour is described in more detail here: https://github.com/GaloyMoney/galoy/issues/1203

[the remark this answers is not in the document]
UncleSamtoshi: I'll go with making it a list for right now

core/api/src/services/ledger/index.ts
```diff
@@ -414,7 +459,9 @@ export const LedgerService = (): ILedgerService => {
     fns: {
       updateMetadataByHash,
       getTransactionById,
+      getTransactionForWalletById,
       getTransactionsByHash,
+      getTransactionsForWalletByPaymentHash,
```
vindard, line 464: Small optional nit to be consistent with other methods here
```suggestion
      getTransactionByWalletIdAndId,
      getTransactionsByHash,
      getTransactionsByWalletIdAndPaymentHash,
```

core/api/src/services/mongoose/wallets.ts
```diff
@@ -146,6 +167,7 @@ export const WalletsRepository = (): IWalletsRepository => {
 
   return {
     findById,
+    findForAccountById,
```
vindard, line 170: Similar optional nit here
```suggestion
    findByAccountIdAndId,
```

core/api/src/services/mongoose/wallet-invoices.ts
```diff
@@ -119,23 +141,44 @@ export const WalletInvoicesRepository = (): IWalletInvoicesRepository => {
     persistNew,
     markAsPaid,
     findByPaymentHash,
+    findForWalletByPaymentHash,
```
vindard, line 144: Optional nit here
```suggestion
    findByWalletIdAndPaymentHash,
```

vindard (commented):
Left some small naming nits but otherwise looks good

vindard (approved):

PR #3320, opened 2023-10-10
feat: consent and dashboard app

nicolasburtey:
chore: renaming folder

chore: moving some of the change to the rest layer instead of graphql

chore: some iteration on hydra

feat: adding scope middleware

chore: testing client credentials

chore: adding appId to know whether a request comes from hydra or kratos

chore: update docs

fix: test?


apps/consent/app/graphql/queries/me-query.ts
```diff
@@ -0,0 +1,27 @@
...
+
+gql`
+  query getUserId {
+    me {
+      id
+    }
+  }
+`;
+
+export const getUserId = async (
```
nicolasburtey, line 14: is this still necessary @siddhart1o1 ?

apps/consent/app/login/verification/form.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+interface VerificationFormProps {
+  login_challenge: string;
+  loginId: string;
+  remember: string;
+  loginType: string;
+  value: string;
+}
+
+const VerificationForm: React.FC<VerificationFormProps> = ({
```
nicolasburtey, line 19: `VerificationForm` naming is too broad. it's unclear from the name if this is a Totp or if this is the captcha or something else. we should use more precise naming so that there is more understand of what this is supposed to represented just by looking at the filename or function name.

apps/consent/app/graphql/queries/me-query.ts
siddhart1o1: for 2fa users

apps/consent/app/oidc-cert.ts
```diff
@@ -0,0 +1,89 @@
+// Copyright © 2023 Ory Corp
```
nicolasburtey, line 1: this file should be deleted

apps/consent/app/login/phone/form.tsx
```diff
@@ -0,0 +1,118 @@
+"use client";
```
nicolasburtey, line 1: it's not super clear to me as to why we have a `phone` subfolder but not an `email` subfolder given those are the 2 ways to authenticate.

apps/consent/app/oidc-cert.ts
siddhart1o1: deleted on latest https://github.com/GaloyMoney/galoy/pull/3320

apps/consent/app/graphql/queries/me-query.ts
nicolasburtey: why not adding it to the response?

apps/consent/app/login/phone/form.tsx
siddhart1o1: we can create separate folder for email, but it will require one more redirect.

nicolasburtey: is this the right name then?

nicolasburtey: of the folder?

PR #3323, opened 2023-10-10
fix: phone metadata null validation

dolcalmi:
Fix for https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/zrhkEzdW98r


core/api/src/domain/users/phone-metadata-authorizer.ts
```diff
@@ -14,9 +14,9 @@ export const PhoneMetadataAuthorizer = ({
   const authorize = (
     phoneMetadata: PhoneMetadata | undefined,
```
nicolasburtey, line 15:
```suggestion
    phoneMetadata: PhoneMetadata | undefined | null,
```
is this the actual type we should have?

nicolasburtey (approved):

PR #3325, opened 2023-10-11
feat: consent and dashboard 2

siddhart1o1:


apps/consent/env.ts
```diff
@@ -0,0 +1,17 @@
+import { createEnv } from "@t3-oss/env-nextjs";
+import { z } from "zod";
+
+export const env = createEnv({
+  server: {
+    HYDRA_ADMIN_URL: z.string().default("http://localhost:4445"),
+    AUTH_URL: z.string().default("http://localhost:4002"),
```
nicolasburtey, line 7: - we should probably call it `CORE_AUTH_URL` to differentiate with hydra or kratos auth url
- I think the AUTH_URL should have the `/auth` suffix, ie: `http://localhost:4002/auth` so that we want to change where the auth is pointing to (ie: `auth.domain.com` because we are migrating the auth system to an independant server, then it will be less code to update.)

apps/consent/app/login/phone/form.tsx
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// TODO need to add country call codes
+const LoginForm: React.FC<LoginFormProps> = ({
+  login_challenge,
+  countryCodes,
+}) => {
+  const [phoneNumber, setPhoneNumber] = useState<string>("");
+  //TODO useFormState is not giving type suggestions/errors i.e: not typed
+  const [state, formAction] = useFormState<GetCaptchaChallengeResponse>(
```
nicolasburtey, line 32: if we are working on regtest/locally we don't want to show the captcha.

core/api/dev/ory/oathkeeper_rules.yaml
```diff
@@ -55,6 +55,20 @@
     url: "<(http|https)>://<[a-zA-Z0-9-.:]+>/graphql"
     methods: ["POST", "GET", "OPTIONS"]
   authenticators:
+    - handler: cookie_session
+      config:
+        check_session_url: http://kratos:4433/sessions/whoami
+        preserve_path: true
+        preserve_query: true
+        subject_from: identity.id
+        extra_from: "@this"
```
nicolasburtey, line 64:
```suggestion
```

nicolasburtey: this is probably a rebase artefact. we don't need that anymore

core/api/dev/ory/oathkeeper.yml
```diff
@@ -68,7 +68,7 @@ mutators:
     config:
       jwks_url: file:///home/ory/jwks.json
       issuer_url: "galoy.io"
-      claims: '{"sub": "{{ print .Subject }}" }'
+      claims: '{"sub": "{{ print .Subject }}", card: "{{ print .Ext.card }}" }'
```
nicolasburtey, line 71:
```suggestion
      claims: '{"sub": "{{ print .Subject }}"}'
```

nicolasburtey: this was for a test locally

core/api/src/servers/middlewares/scope.ts
```diff
@@ -0,0 +1,79 @@
...
+) => {
+  const scope = context.scope
+  const appId = context.appId
+
+  // not a delegated token
+  if (appId === undefined || appId === "") {
+    return resolve(parent, args, context, info)
+  }
+
+  if (scope === undefined) {
```
bodymindarts, line 23: this will never hit because of line 19.

core/api/src/servers/middlewares/scope.ts
```diff
@@ -0,0 +1,79 @@
...
+  args: unknown,
+  context: GraphQLPublicContextAuth,
+  info: GraphQLResolveInfo,
+) => {
+  const scope = context.scope
+  const appId = context.appId
+
+  // not a delegated token
+  if (appId === undefined || appId === "") {
+    return resolve(parent, args, context, info)
```
bodymindarts, line 20: does resolve mean that we have been authorised? If so this seem likes a potential issue - I think you mean `&&` here not `||` - right?

core/api/src/servers/middlewares/scope.ts
```diff
@@ -0,0 +1,79 @@
...
+type ValidateWalletIdFn = (
+  resolve: GraphQLFieldResolver<unknown, GraphQLPublicContextAuth>,
+  parent: unknown,
+  args: unknown,
+  context: GraphQLPublicContextAuth,
+  info: GraphQLResolveInfo,
+) => Promise<unknown>
+
+const walletIdMutationFields: { [key: string]: ValidateWalletIdFn } = {}
+for (const key of Object.keys(mutationFields.authed.atWalletLevel)) {
```
bodymindarts, line 70: This seems quite implicit... 
I would generally prefer having authorisation as an app layer concern rather than a server middle ware - but I guess this is okay as a get-started-quickly approach.

core/api/src/servers/graphql-server.ts
```diff
@@ -15,10 +15,10 @@ import { createComplexityPlugin } from "graphql-query-complexity-apollo-plugin"
 
 import jwksRsa from "jwks-rsa"
 
-import authRouter from "./authorization"
 import kratosCallback from "./event-handlers/kratos"
 import healthzHandler from "./middlewares/healthz"
 import { idempotencyMiddleware } from "./middlewares/idempotency"
+import authRouter from "./authorization"
```
bodymindarts, line 21: Noise?

core/api/src/domain/authorization/index.ts
```diff
@@ -0,0 +1,6 @@
+export const ScopesOauth2 = {
+  TransactionsRead: "transactions:read",
+  Offline: "offline",
```
bodymindarts, line 3: Where is Offline used?

core/api/src/servers/middlewares/scope.ts
nicolasburtey: yes right now the authorization is really just a POC and need iteration. the question is: do we want to merge this?

core/api/src/servers/middlewares/scope.ts
nicolasburtey: so, currently, if scope is undefined or appId is undefined, it means this is a kratos authenticated request, not a hydra authenticated request.

if this is a kratos request, it means it's a "primary"/first party user request. then we don't need to look into this middleware because it's solely for delegation request

the way to skip this middleware is to resolve it right away and not continue the function.

core/api/src/servers/middlewares/scope.ts
nicolasburtey: line 19 is a check on appId, not scope?

here the idea is that if this is a hydra/delegation request, I want scope to exist.

core/api/src/servers/graphql-server.ts
nicolasburtey: 🤷

IDE things... probably following a merge. 

there is no way to order the import fully deterministically afaik.

core/api/src/domain/authorization/index.ts
nicolasburtey: it's a Oauth2 convention things. it's basically whether the client should receive an refresh token in addition to the access token.

not sure that will be useful, but I added that as part of my R&D to hydra.

core/api/src/servers/middlewares/scope.ts
bodymindarts: Whoops misread

core/api/src/servers/middlewares/scope.ts
bodymindarts: Perhaps we should make the inclusion of the middleware configurable - then we can keep the feature flag off in prod while we continue to develop.

PR #3333, opened 2023-10-11
chore: record paymentRequest with WalletInvoice

UncleSamtoshi:
- start recording payment requests with each wallet invoice
- this pr will be followed by a migration and then a pr to make paymentRequest mandatory

### DIFF
--- a/core/api/src/domain/wallet-invoices/index.types.d.ts
+++ b/core/api/src/domain/wallet-invoices/index.types.d.ts
@@ -74,7 +74,7 @@ type WIBWithAmountState = WIBWithExpirationState & {
 }
 
 type LnAndWalletInvoice = {
-  walletInvoice: WalletInvoice
+  walletInvoice: WalletInvoice & { paymentRequest: EncodedPaymentRequest }
   lnInvoice: LnInvoice
 }
 
@@ -91,6 +91,7 @@ type WalletInvoice = {
   recipientWalletDescriptor: PartialWalletDescriptor<WalletCurrency>
   paid: boolean
   createdAt: Date
+  paymentRequest?: EncodedPaymentRequest
 }
 
 type WalletAddress<S extends WalletCurrency> = {
@@ -156,7 +157,9 @@ type WalletAddressReceiverArgs<S extends WalletCurrency> = {
   walletAddress: WalletAddress<S>
 }
 
-type WalletInvoicesPersistNewArgs = Omit<WalletInvoice, "createdAt">
+type WalletInvoicesPersistNewArgs = Omit<WalletInvoice, "createdAt"> & {
+  paymentRequest: EncodedPaymentRequest
+}
 
 interface IWalletInvoicesRepository {
   persistNew: (
--- a/core/api/src/domain/wallet-invoices/wallet-invoice-builder.ts
+++ b/core/api/src/domain/wallet-invoices/wallet-invoice-builder.ts
@@ -143,7 +143,7 @@ export const WIBWithAmount = (state: WIBWithAmountState): WIBWithAmount => {
       return new InvalidWalletInvoiceBuilderStateError()
     }
 
-    const walletInvoice: WalletInvoice = {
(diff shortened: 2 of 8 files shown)


core/api/src/services/mongoose/schema.types.d.ts
```diff
@@ -63,6 +63,7 @@ interface WalletInvoiceRecord {
   selfGenerated: boolean
   pubkey: string
   paid: boolean
+  paymentRequest?: string
```
nicolasburtey, line 66:
```suggestion
  paymentRequest?: string // optional because we didn't stored it in the past
```

maybe worth adding a comment on why we're making it optional?

core/api/src/services/mongoose/wallet-invoices.ts
```diff
@@ -135,4 +137,5 @@ const walletInvoiceFromRaw = (result: WalletInvoiceRecord): WalletInvoice => ({
   paid: result.paid as boolean,
   usdAmount: result.cents ? UsdPaymentAmount(BigInt(result.cents)) : undefined,
   createdAt: new Date(result.timestamp.getTime()),
+  paymentRequest: result.paymentRequest as EncodedPaymentRequest,
```
nicolasburtey, line 140: this is a casting of a potentially null/undefined value to string. this is pretty dangerous.
I would do this

```suggestion
  paymentRequest: result.paymentRequest ?? "unknown" as EncodedPaymentRequest,
```

nicolasburtey: actually it's not because `WalletInvoice` marks `paymnetRequest` as optional... I still it make sense to return a string here.

core/api/src/domain/wallet-invoices/wallet-invoice-builder.ts
```diff
@@ -143,7 +143,7 @@ export const WIBWithAmount = (state: WIBWithAmountState): WIBWithAmount => {
       return new InvalidWalletInvoiceBuilderStateError()
     }
 
-    const walletInvoice: WalletInvoice = {
+    const walletInvoice = {
```
nicolasburtey, line 146:
```suggestion
    const walletInvoice: WalletInvoice = {
```

we can keep the type if we don't make paymentRequest optional here https://github.com/GaloyMoney/galoy/pull/3333/files#r1355431350

(note: we can still keep the type optional at the db layer)

nicolasburtey (commented):
few comments on where I think we can improve the type

core/api/src/services/mongoose/wallet-invoices.ts
UncleSamtoshi: It will eventually be non nullable after a migration. But for now I think it makes sense to keep it undefined so we dont accidentally think we have a value when we dont.

core/api/src/domain/wallet-invoices/wallet-invoice-builder.ts
UncleSamtoshi: After a migration, I will make the type non-optional and then we can use wallet invoice.

nicolasburtey (dismissed):

core/api/src/domain/wallet-invoices/index.types.d.ts
```diff
@@ -74,7 +74,7 @@ type WIBWithAmountState = WIBWithExpirationState & {
 }
 
 type LnAndWalletInvoice = {
-  walletInvoice: WalletInvoice
+  walletInvoice: WalletInvoice & { paymentRequest: EncodedPaymentRequest }
```
bodymindarts, line 77: This is temporary right? Seems like we would like `paymentRequest` to permanently be part of `WalletInvoice`?

UncleSamtoshi: Yes it is temporary. Right now we want it to be required when we persist it but optional when we read it.

bodymindarts (approved):

PR #3342, opened 2023-10-13
chore: add payment request to invoice debug script

UncleSamtoshi:
- iterates over all invoices in lnd nodes and adds payment requests to their respective wallet invoices

Steps for test testing:
1. Populate the db with invoices
```
make reset-deps
bats -t test/bats/ln-receive
```
2. Go into mongo db and remove the payment request field from a document in `invoiceusers` collection
3. Run the debug script
```
. ./.envrc && yarn ts-node \                                       
        --files \
                -r tsconfig-paths/register \
                -r src/services/tracing.ts \
        ./src/debug/add-payment-request-all-wallet-invoice.ts
```

### DIFF
--- a/core/api/src/debug/add-payment-request-all-wallet-invoice.ts
+++ b/core/api/src/debug/add-payment-request-all-wallet-invoice.ts
@@ -20,7 +20,7 @@ import {
 import { WalletInvoice } from "@/services/mongoose/schema"
 
 const main = async () => {
-  asyncRunInSpan(
+  return asyncRunInSpan(
     "debug.addPaymentRequestAllWalletInvoices",
     {
       attributes: {
@@ -29,6 +29,15 @@ const main = async () => {
       },
     },
     async () => {
+      const initialNumberOfLnInvoicesMissingPaymentRequest =
+        await WalletInvoice.countDocuments({
+          paymentRequest: { $exists: false },
+        })
+
+      console.log(
+        `Number of wallet invoices missing payment request: ${initialNumberOfLnInvoicesMissingPaymentRequest}`,
+      )
+
       const lndService = LndService()
       if (lndService instanceof Error) return lndService
       const pubkeys = lndService.listActivePubkeys()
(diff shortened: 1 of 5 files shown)


core/api/src/debug/add-payment-request-all-wallet-invoice.ts
```diff
@@ -0,0 +1,145 @@
...
+        })
+
+      console.log(
+        `Number of wallet invoices missing payment request: ${initialNumberOfLnInvoicesMissingPaymentRequest}`,
+      )
+
+      const lndService = LndService()
+      if (lndService instanceof Error) return lndService
+      const pubkeys = lndService.listActivePubkeys()
+      for (const pubkey of pubkeys) {
```
nicolasburtey, line 44: note: we had recycle node in the past (following issue with our node), so there will paymentRequest missing

core/api/src/services/lnd/index.ts
```diff
@@ -884,6 +910,7 @@ export const LndService = (): ILightningService | LightningServiceError => {
       listPendingPayments: listPaymentsFactory(getPendingPayments),
       listFailedPayments,
       listInvoices,
+      listAllInvoices,
```
nicolasburtey, line 913: it's kind of confusing to have a `listInvoices` and `listAllInvoices`

nicolasburtey: if the main reason is because of the yield, maybe `yieldInvoices` would be a better name? we could also still pass `createdAfter` to have the same signature.

core/api/src/debug/add-payment-request-all-wallet-invoice.ts
UncleSamtoshi: Good to know

core/api/src/services/lnd/index.ts
UncleSamtoshi: I will change the name and pass in the createdAfter

nicolasburtey: or alternatively, just modify listInvoices in place and update get-held-invoice  and trigger to be using the `yield` interface. those seems the only 2 place there this is used and using the yield may be a few lines change max

UncleSamtoshi: I changed the interface to use yield

core/api/src/debug/add-payment-request-all-wallet-invoice.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+const addPaymentRequestAllWalletInvoicesForPubkey = wrapAsyncToRunInSpan({
+  fnName: "addPaymentRequestAllWalletInvoicesForPubkey",
+  namespace: "debug",
+  fn: async ({ pubkey }: { pubkey: Pubkey }) => {
+    const lndService = LndService()
+    if (lndService instanceof Error) return lndService
+
+    const invoices = await lndService.listInvoices({ pubkey })
```
nicolasburtey, line 76:
```suggestion
    const invoices = lndService.listInvoices({ pubkey })
```

I guess await is not necessary here?

nicolasburtey (approved):

core/api/src/debug/add-payment-request-all-wallet-invoice.ts
UncleSamtoshi: Good call

nicolasburtey (approved):

PR #3343, opened 2023-10-13
chore: move linting back to the root

nicolasburtey:


bodymindarts:
@vindard is just working on linting FYI.

PR #3356, opened 2023-10-15
feat: adding external id to accounts

nicolasburtey:


bodymindarts:
What use case is this solving for?

PR #3360, opened 2023-10-16
feat: add 'madge-check' to workspace-pnpm toolchain

vindard:
_Merge after #3347_


toolchains/workspace-pnpm/run_in_dir.py
```diff
@@ -31,13 +41,18 @@ def compute_path(arg: str) -> str:
 
     cwd = args.cwd
 
+    dist_path = None
+    if args.package_dir:
```
vindard, line 45: This introduces some overloading here depending on if a `--package-dir` argument is passed in. Do we think this is ok, or should this be a separate toolchain function?

toolchains/workspace-pnpm/macros.bzl
```diff
@@ -581,6 +581,45 @@ def _npm_test_impl(
...
+        ctx,
+        ExternalRunnerTestInfo(
+            type = test_info_type,
+            command = [run_cmd_args],
+        ),
+    ) + [
+        DefaultInfo(default_output = args_file),
+    ]
+
+def _npm_build_and_test_impl(
```
vindard, line 600: I tried rolling this into the `_npm_test_impl` and ran into 2 issues:
- I had to add the `runnable_tsc_build ` attr to 3 rules that didn't need it
- Even after I added it, there was no way I could find to set a default falsy value for that attr to switch on inside `_npm_test_impl`

Also, the conditional branching was visually significant enough inside the function that I think it makes sense to leave these as separate functions in this instance.

PR #3368, opened 2023-10-16
refactor: switch madge circular deps check to ts files

vindard:


core/api/src/app/authentication/login.ts
```diff
@@ -35,7 +35,7 @@ import {
 
 import { upgradeAccountFromDeviceToPhone } from "@/app/accounts"
 import { checkedToEmailCode } from "@/domain/authentication"
-import { isPhoneCodeValid, TwilioClient } from "@/services/twilio"
+import { isPhoneCodeValid, TwilioClient } from "@/services/twilio-service"
```
vindard, line 38: Wasn't sure what to do with this one. It seems that the `import twilio from "twilio"` top-line import in that file caused some dependency check issue.

bodymindarts (approved):

core/api/package.json
```diff
@@ -3,7 +3,8 @@
   "scripts": {
     "tsc-check-noimplicitany": "tsc --noEmit -p tsconfig.no-implicit-any.json",
     "tsc-check": "tsc --noEmit -p tsconfig.d.json && tsc --noEmit",
-    "madge-check": "yarn madge --circular lib/servers/graphql-main-server.js",
+    "madge-check": "yarn madge --circular --extensions ts ./src",
```
bodymindarts, line 6: I think we can also delete `yarn` here right?

bodymindarts: Confirmed - calling `yarn` from within a `package.json` `script` is redundant - by removing it it becomes compatible with `pnpm`.

bodymindarts (approved):

core/api/package.json
vindard: Yes, I actually did do that in the [other PR](https://github.com/GaloyMoney/galoy/pull/3360/files#diff-96fde044b147bac83eceed2332a32601b6400e185e9034973e0110d25c8175baR6) too

PR #3372, opened 2023-10-17
feat: add unit test rule to core/api buck file

vindard:
## Description

This PR adds unit tests to buck with the caveat that unit tests currently have a dependency on env variables and the correct ones must be brought in via a `.env` file.

This requirement can probably be refactored out in future.


core/api/package.json
```diff
@@ -19,9 +19,9 @@
     "exporter": "yarn build && node lib/servers/exporter.js",
     "daily-notif": "yarn build && node lib/servers/daily-balance-notification.js",
     "test": "export JEST_JUNIT_OUTPUT_DIR=\"./artifacts\" && jest --ci --reporters=default --reporters=jest-junit",
-    "test:unit": ". ../../.env && LOGLEVEL=warn jest --config ./test/unit/jest.config.js --bail --verbose $TEST | yarn pino-pretty -c -l",
```
vindard, line 22: Only one step left here so far (to re-add pino-pretty to the buck call)

bodymindarts: Don't see this as a strong requirement - output looks fine.

toolchains/workspace-pnpm/run_in_dir.py
```diff
@@ -3,16 +3,23 @@
...
-def compute_path(arg: str) -> str:
-    if arg.endswith("::abspath"):
-        return os.path.abspath(arg.removesuffix("::abspath"))
+def merge_env_from_file(file_path):
+    # Shell command to source the .env file and then get the environment as JSON using jq
+    if file_path and os.path.exists(file_path):
+        cmd = f'source {file_path} && jq -n env'
     else:
-        return arg
+        cmd = f'jq -n env'
```
bodymindarts, line 16: I would prefer if we could do this in python only...

vindard: Done here: https://github.com/GaloyMoney/galoy/pull/3372/commits/6eac5741fcc11e4ccee8336156a06fae7083fbf0

core/api/BUCK
```diff
@@ -155,6 +156,16 @@ madge_check(
     srcs = [":src"],
 )
 
+test_unit(
+    name = "unit-tests",
+    srcs = [":src"] + [":test_src"] + glob([".env", "galoy.yaml"]),
+    config_file = "test/unit/jest.config.js",
+    env_file = ".env",
```
bodymindarts, line 163: I would prefer to bring this into the file instead of relying on the on disk state like in the [Tiltfile](https://github.com/GaloyMoney/galoy/blob/595cf23073bd9ca50b85c3f5e28e0ab6da254ef9/dev/Tiltfile#L137).

Ideally we would be able to delete all the .env files and rely on environment defined in code. I don't know if that's feasible yet - we can merge as is to not mess with the existing workflow - but I think that should be the goal once legacy workflows are entirely deprecated.

bodymindarts (approved):

PR #3385, opened 2023-10-19
chore: update apollo server

dolcalmi:


core/api/src/servers/graphql-server.ts
```diff
@@ -46,40 +48,34 @@ export const startApolloServer = async ({
...
-    createComplexityPlugin({
-      schema,
-      estimators: [fieldExtensionsEstimator(), simpleEstimator({ defaultComplexity: 1 })],
-      maximumComplexity: 200,
-      onComplete: (complexity) => {
-        // TODO(telemetry): add complexity value to span
-        baseLogger.debug({ complexity }, "queryComplexity")
-      },
-    }),
+    // createComplexityPlugin({
```
nicolasburtey, line 51: this doesn't work with apollo 4?

nicolasburtey: maybe we should just import the plugin to our code, this gist is very small: https://gist.github.com/reconbot/37ccd8b8fcdfb78ff6a16a7055edf103

dolcalmi: I used the code from npm source.. it was a bit different from gist

core/api/src/servers/graphql-server.ts
```diff
@@ -190,3 +188,45 @@ export const startApolloServer = async ({
     })
   })
 }
+
+const ApolloServerPluginGraphQLQueryComplexity = ({
```
dolcalmi, line 192: I had to add it here because adding it in a different file is creating a circular dependency (really not sure why because it only use external packages)

nicolasburtey (approved):

PR #3393, opened 2023-10-19
fix: rate limit in dashboard and consent

siddhart1o1:


apps/dashboard/services/graphql/index.ts
```diff
@@ -1,17 +1,36 @@
...
+  ApolloLink,
+} from "@apollo/client";
+import { registerApolloClient } from "@apollo/experimental-nextjs-app-support/rsc";
+
+const headerSettingLink = new ApolloLink((operation, forward) => {
+  const request = operation.getContext().request;
+  operation.setContext({
+    headers: {
+      "x-real-ip": request?.headers.get("x-real-ip") || "",
+      "x-forwarded-for": request?.headers.get("x-forwarded-for") || "",
```
bodymindarts, line 15: do we really want to default to `""` - isn't it better to leave it undefined? cc @nicolasburtey

nicolasburtey: agree but this may have to do with type?

apps/consent/app/graphql/apollo-config.ts
```diff
@@ -1,15 +1,34 @@
...
+    headers: {
+      ...(authToken ? { authorization: `Bearer ${authToken}` } : undefined),
+    },
   });
 
+  const link = ApolloLink.from([headerSettingLink, httpLink]);
   return new ApolloClient({
     cache: new InMemoryCache(),
-    link: httpLink,
+    link: link,
```
nicolasburtey, line 32:
```suggestion
    link,
```

apps/consent/services/galoy-auth/auth-instance.ts
```diff
@@ -2,8 +2,25 @@ import { env } from "../../env";
...
+axiosInstance.interceptors.request.use(
+  (config) => {
+    const xRealIp = config.headers["x-real-ip"];
+    const xForwardedFor = config.headers["x-forwarded-for"];
+    if (xRealIp) {
+      config.headers["x-real-ip"] = xRealIp;
+    }
+    if (xForwardedFor) {
+      config.headers["x-forwarded-for"] = xForwardedFor;
+    }
```
nicolasburtey, line 18: this feels like a no-op? we' taking a value from `config.headers["x-real-ip"]` and redefining to the same value? same for `config.headers["x-forwarded-for"]`?

apps/dashboard/services/graphql/index.ts
```diff
@@ -1,17 +1,36 @@
...
     return new ApolloClient({
       cache: new InMemoryCache(),
-      link: new HttpLink({
-        headers: {
-          ["Oauth2-Token"]: token,
-        },
-        uri: env.CORE_URL,
-        fetchOptions: { cache: "no-store" },
-      }),
+      link: link,
```
nicolasburtey, line 34:
```suggestion
      link,
```

apps/consent/services/galoy-auth/auth-instance.ts
siddhart1o1: yup this will not work need to add config

nicolasburtey: if you do `axiosInstance.interceptors.request.use(config => config)` this doesn't work?

apps/dashboard/services/graphql/index.ts
siddhart1o1: yes

apps/consent/services/galoy-auth/auth-instance.ts
siddhart1o1: no

bodymindarts (approved):

PR #3396, opened 2023-10-19
chore: genericize prettier config and add to apps/dashboard

vindard:
## Description

This is to help with format-on-save actions in editors like VSCode


bodymindarts (approved):

vindard:
Merging for now since this works for some folks. We can continue debugging this in a follow-up PR.

PR #3400, opened 2023-10-20
fix: add axios option withCredentials to acceptOAuth2LoginRequest

bodymindarts:
This is the equivalent to what we had to do in the web wallet here: https://github.com/GaloyMoney/web-wallet/blob/5bb016a1adbe556ae4d09070daa3c18fb9599ce9/src/components/pages/login-email.tsx#L75 (ie pass options down to axios to forward cross site cookies)

### DIFF
--- a/apps/consent/app/consent/page.tsx
+++ b/apps/consent/app/consent/page.tsx
@@ -38,14 +38,14 @@ const submitForm = async (form: FormData) => {
         error: "access_denied",
         error_description: "The resource owner denied the request",
       },
-    });
+    }, { withCredentials: true });
     redirect(response.data.redirect_to);
   }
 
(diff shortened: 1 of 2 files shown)


nicolasburtey:
it should probably have to be done in the consent/page.ts also, because this is where the redirect to hydra happens

nicolasburtey:
> it should probably have to be done in the consent/page.ts also, because this is where the redirect to hydra happens

even though the final redirect is done as a server side component so I don't think would have much effect

bodymindarts:
Let's see what happened with this rolled out - domain is being set correctly already. This is how we got it working in Kratos. If it doesn't work we'll have to turn off SSR for consent app I guess.

PR #3409, opened 2023-10-23
feat: email verification

siddhart1o1:


apps/dashboard/services/graphql/mutations/totp.ts
```diff
@@ -0,0 +1,87 @@
+// import { gql } from "@apollo/client";
```
nicolasburtey, line 1: we don't want comment files committed in our repo

apps/dashboard/services/graphql/mutations/email.ts
```diff
@@ -0,0 +1,107 @@
...
+      emailRegistrationId
+      errors {
+        message
+        code
+      }
+    }
+  }
+`;
+
+gql`
```
nicolasburtey, line 27:
```suggestion
```

nicolasburtey: you can combine queries/mutation in a single `gql` string

apps/dashboard/app/security/server-actions.ts
```diff
@@ -0,0 +1,121 @@
...
+  const token = session?.accessToken;
+  if (!token && typeof token !== "string") {
+    return {
+      error: true,
+      message: "Invalid Token",
+      data: null,
+    };
+  }
+
+  let data;
```
nicolasburtey, line 37: no implicit any

sandipndev:
I was able to delete the email and then re add it using the code from kratos. Is that what is intended in this PR?

sandipndev (approved):

PR #3410, opened 2023-10-23
chore: fix dashboard bug for invalid token, added dev mode for captcha

siddhart1o1:


apps/consent/app/components/captcha-challenge/index.tsx
```diff
@@ -11,11 +17,22 @@ const CaptchaChallengeComponent: React.FC<{
...
+    channel: string
   }
 }> = ({ id, challenge, formData }) => {
+  const sendMockData = async () => {
+    const mockCaptchaResult = {
+      geetest_challenge: "mockChallenge",
+      geetest_validate: "mockValidate",
+      geetest_seccode: "mockSecCode",
+    }
+    const res = await sendPhoneCode(mockCaptchaResult, formData)
```
nicolasburtey, line 29: maybe I'm not understanding the code correctly, but there is no need to fetch the sendPhoneCode endpoint in dev mode, because no code is been sent. 

the action should be as simple as moving to the next screen instead.

nicolasburtey (changes requested):
I think we need to change the logic on this PR

apps/consent/app/components/captcha-challenge/index.tsx
siddhart1o1: In this, we are just sending mock data. I thought this would mimic production mode better, but yes, we can just redirect.

apps/consent/env.ts
```diff
@@ -8,10 +8,12 @@ export const env = createEnv({
   },
   shared: {
     GRAPHQL_ENDPOINT: z.string().default("http://localhost:4002/graphql"),
+    NODE_ENV: z.string().default("development"),
```
nicolasburtey, line 11: if we have a default to development then the likelihood we stumble upon the issue that it will be set as dev mode in production is quite high. 

I don't think we should have default here.

apps/dashboard/.env
```diff
@@ -1,5 +1,5 @@
 # REPLACE THIS IT IS FOR TESTING
-NEXTAUTH_URL=https://c890-2405-201-301c-5b67-89d6-bd56-6afb-6294.ngrok-free.app
+NEXTAUTH_URL=https://39a0-2405-201-301c-5b67-8cea-1030-d56-e0b3.ngrok-free.app
```
nicolasburtey, line 2: we should not change this for every PR this is creating noise for the reviewer.

it's actually possible to have `localhost` here. look at my PR if that's useful: https://github.com/GaloyMoney/galoy/pull/3407/files

the reason I stumble upon some issues at the beginning was because there were a mix of `127.0.0.1` and `localhost`.

nicolasburtey (approved):

PR #3412, opened 2023-10-23
chore: initial Tiltfile

bodymindarts:
Running:
```
buck2 run //dev:up
```

Will bring up tilt and starts the core / auth stack - enough to be able to develop the consent app and run its integration tests on top of it.

To verify integration tests you can either click on the `consent-test` resource in the tilt panel or run:
```
buck2 test //apps/consent:cypress
```
once the whole tilt file is running.


dev/bin/setup-hydra-client.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
nicolasburtey, line 1: would it make to look into terraform so we have some parity with prod, instead of having a script? or that would not make sense for dev

bodymindarts: IMO too much overhead to bring in another tool for dev.

dev/config/bitcoind/bitcoin.conf
```diff
@@ -0,0 +1,19 @@
+rpcuser=rpcuser
```
vindard, line 1: Is the intention to eventually remove these files from `core/api/dev`?

vindard (approved):
Looks good to me. Local workflows work, and I can see how we might tailor this to be able to run integration/bats test against soon

dev/config/bitcoind/bitcoin.conf
bodymindarts: yes

PR #3414, opened 2023-10-24
feat: add payout cancelled event handler

vindard:
## Description

This adds:
- a `PayoutCancelled` event type to the expected event types Bria produces
- an interim `payoutCancelled` handler to the bria event handler

### DIFF
--- a/core/api/src/servers/event-handlers/bria.ts
+++ b/core/api/src/servers/event-handlers/bria.ts
@@ -5,6 +5,7 @@ import {
   CouldNotFindWalletFromOnChainAddressError,
   LessThanDustThresholdError,
   NoTransactionToUpdateError,
+  NotImplementedError,
 } from "@/domain/errors"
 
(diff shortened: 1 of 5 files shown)


core/api/src/servers/event-handlers/bria.ts
```diff
@@ -131,6 +135,22 @@ export const payoutSubmittedEventHandler = async ({
...
+}: {
+  event: PayoutCancelled
+}): Promise<true | ApplicationError> => {
+  const txns = await LedgerFacade.getTransactionsByPayoutId(event.id)
+  if (txns instanceof Error) return txns
+
+  if (txns.length !== 0)
+    return new NotImplementedError(
+      `Payout cancels not implemented as yet for PayoutId: ${event.id}`,
+    )
```
dolcalmi, line 149: how does this works if payout is initiated in stablesats? (not through api)

dolcalmi (changes requested):

dolcalmi (approved):

PR #3418, opened 2023-10-24
refactor: complete PayoutCancelled handler implementation

vindard:
## Description

This is to complete the implementation to handle payouts originated from the application here.

### DIFF
--- a/core/api/src/servers/event-handlers/bria.ts
+++ b/core/api/src/servers/event-handlers/bria.ts
@@ -21,9 +21,17 @@ import { addAttributesToCurrentSpan } from "@/services/tracing"
 const assertUnreachable = (payloadType: never): Error =>
   new UnknownPayloadTypeReceivedError(payloadType)
 
+const isBriaPayoutEvent = (payload: BriaPayload): payload is BriaPayoutPayload => {
+  return (payload as BriaPayoutPayload).id !== undefined
+}
+
 export const briaEventHandler = async (event: BriaEvent): Promise<true | DomainError> => {
   baseLogger.info(
-    { sequence: event.sequence, type: event.payload.type },
+    {
+      sequence: event.sequence,
+      type: event.payload.type,
+      ...(isBriaPayoutEvent(event.payload) ? { id: event.payload.id } : {}),
+    },
     "bria event handler",
   )
   addAttributesToCurrentSpan({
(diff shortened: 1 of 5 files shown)


bodymindarts (approved):
This is great - now we could even expose this to graphql and end users allowing a longer queue cycle.

core/api/src/servers/event-handlers/bria.ts
```diff
@@ -46,8 +61,17 @@ export const briaEventHandler = async (event: BriaEvent): Promise<true | DomainE
         payoutInfo: event.augmentation.payoutInfo,
       })
 
+    case BriaPayloadType.PayoutCommitted:
+      return true
+
     case BriaPayloadType.PayoutCancelled:
-      return payoutCancelledEventHandler({ event: event.payload })
+      if (event.augmentation.payoutInfo === undefined) {
+        return new EventAugmentationMissingError()
```
dolcalmi, line 69: we cant block event processor because of this... stablesats didnt use externalId, also we should stop using === undefined (payout info can be null)

image

dolcalmi (changes requested):

core/api/src/servers/event-handlers/bria.ts
bodymindarts: It should never be null when passing augment option

bodymindarts: Also don't understand relevance of `stablesats didnt use externalId`? Since rollout on Monday it is actually using `externalId` - but what does that have to do with anything?

vindard: > we cant block event processor because of this... stablesats didnt use externalId, also we should stop using === undefined (payout info can be null)

payoutInfo can't be null in this scenario because we always pass the `augmentation`. In the instances where an externalId isn't explicitly passed to stablesats (before Monday) a random uuid is internally assigned. Can see the difference in the two clusters below.

Traces: https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/qEnUKXWF6pf

bodymindarts: Its not random - its the internal id https://github.com/GaloyMoney/bria/blob/3e25e4a3ebd685967ae422b234a3a6aa895566e8/src/payout/entity.rs#L89-L93

dolcalmi: > Also don't understand relevance of `stablesats didnt use externalId`? Since rollout on Monday it is actually using `externalId` - but what does that have to do with anything?

we need to reprocess txs

PR #3421, opened 2023-10-25
feat: enable quiz completion without reward

UncleSamtoshi:
- add `quizRewardsEnabled` to ConsumerAccount gql object
- enable marking a quiz question complete without payment success


core/api/src/app/payments/add-earn.ts
```diff
@@ -29,7 +29,13 @@ export const addEarn = async ({
 }: {
   quizQuestionId: string
   accountId: string
-}): Promise<QuizQuestion | ApplicationError> => {
+}): Promise<
+  | {
+      quizQuestion: Quiz
+      rewardPaymentError?: ApplicationError
```
vindard, line 35: I wonder if it makes sense to wrap this in some like what we've done for the `PartialResult` type and object?

vindard (commented):
One comment so far on the way we've implemented passing a result + error along

core/api/src/app/payments/add-earn.ts
UncleSamtoshi: I think thats a good idea, I'll wrap it with a Partial result and add a paid field.

vindard (commented):
Looks good to me. Only comment left is, I wonder if it'll be easy to add a bats test for this

core/api/src/graphql/public/types/object/consumer-account.ts
```diff
@@ -190,15 +203,10 @@ const ConsumerAccount = GT.Object<Account, GraphQLPublicContextAuth>({
           paginationArgs,
         })
 
-        if (error instanceof Error) {
+        if (!result || error) {
```
dolcalmi, line 206: if this use PartialResult then should use the new approach

core/api/src/graphql/public/types/object/consumer-account.ts
```diff
@@ -190,15 +203,10 @@ const ConsumerAccount = GT.Object<Account, GraphQLPublicContextAuth>({
...
           throw mapError(error)
         }
 
-        if (!result?.slice) {
```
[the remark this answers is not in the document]
UncleSamtoshi: This just matches the existing behavior, but with different types. As you can see in line 193 if there was any error, even if it was a partial success, we still threw in line 194.

core/api/src/graphql/public/types/object/consumer-account.ts
UncleSamtoshi: I will switch to using `type` on the partial result.

UncleSamtoshi:
> Looks good to me. Only comment left is, I wonder if it'll be easy to add a bats test for this

For testing we could add the unhappy path where the phone carrier is not supported. However I dont think we can get the local setup into a state where phone carrier and ip are valid.

UncleSamtoshi:
One concern I have with this PR is we lose explicitly visibility into whether we actually paid out the reward. Previously the presence of a completed question implied that the reward was paid. However this introduces the possibility that a question is completed but a reward was not paid. Since this is a marketing feature it may not be critical to have this visibility. Additionally we could look through the ledger for a transaction correlating to that id.

In exchange for losing this visibility we now allow people without the ability to earn rewards the ability to still complete the earn section.

nicolasburtey:
> One concern I have with this PR is we lose explicitly visibility into whether we actually paid out the reward. Previously the presence of a completed question implied that the reward was paid. However this introduces the possibility that a question is completed but a reward was not paid. Since this is a marketing feature it may not be critical to have this visibility. Additionally we could look through the ledger for a transaction correlating to that id.
> 
> In exchange for losing this visibility we now allow people without the ability to earn rewards the ability to still complete the earn section.

we should make sure that we can, thought tracing, plot the query where rewards are given or not.

UncleSamtoshi:
> we should make sure that we can, through tracing, plot the query where rewards are given or not.

This should be visible.

vindard:
> However this introduces the possibility that a question is completed but a reward was not paid.

Should we consider adding `rewardPaid` to persistence as well?

PR #3429, opened 2023-10-26
chore: making it possible to use localhost instead of ngrok

nicolasburtey:
it's not needed to use ngrok for local dev. it was an issue with mix between localhost and 127.0.0.1


bodymindarts:
I would like to remove this dev setup for apps.

The tilt setup is preferable IMO - better to move forward with that.

bodymindarts:
Somehow some lint errors got merged to main... https://github.com/GaloyMoney/galoy/pull/3430 I fixed them

apps/consent/dev/get-url.sh
```diff
@@ -1,7 +1,7 @@
 NEXTAUTH_URL="http://localhost:3000/"
 
 code_client=$(hydra create client \
-    --endpoint http://127.0.0.1:4445 \
+    --endpoint http://localhost:4445 \
```
bodymindarts, line 4: We should delete this file

apps/dashboard/.env
```diff
@@ -1,5 +1,5 @@
 # REPLACE THIS IT IS FOR TESTING
-NEXTAUTH_URL=https://c890-2405-201-301c-5b67-89d6-bd56-6afb-6294.ngrok-free.app
+NEXTAUTH_URL=http://localhost:3001
```
bodymindarts, line 2: This needs adapting / integrating with tilt

bodymindarts (approved):

bodymindarts:
I think this PR should be preferred https://github.com/GaloyMoney/galoy/pull/3431

PR #3435, opened 2023-10-27
refactor: change tilt buck target to be runnable from any dir

vindard:
## Description

This would fix the issue where someone would have to be at the top-level dir for `//dev:up` and `//dev:down` to work

### DIFF
--- a/dev/tilt.bzl
+++ b/dev/tilt.bzl
@@ -1,32 +1,10 @@
-def tilt_up_impl(ctx: AnalysisContext) -> list[[DefaultInfo, RunInfo]]:
-    return _invoke_tilt(ctx, "up")
+def tilt_impl(ctx: AnalysisContext) -> list[[DefaultInfo, RunInfo]]:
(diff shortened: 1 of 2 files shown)


dev/tilt.bzl
```diff
@@ -50,15 +18,22 @@ def _invoke_tilt(ctx: AnalysisContext, subcmd: str) -> list[[DefaultInfo, RunInf
         ctx.attrs.tiltfile,
     )
 
+    script = ctx.actions.write("tilt-run.sh", """\
+#!/usr/bin/env bash
+set -euo pipefail
+
+rootpath="$(git rev-parse --show-toplevel)"
```
vindard, line 25: Not sure if git here is an acceptable dependency.

This approach was derived from si after I'd exhausted all ways to get a handle on the root dir from the bzl file otherwise without having to execute from python

vindard: We're bringing in `gitMinimal` so this should be ok

bodymindarts (approved):

PR #3450, opened 2023-10-30
fix: added tracing file in BUCK (consent, dashboard)

siddhart1o1:


apps/consent/BUCK
```diff
@@ -43,11 +43,12 @@ filegroup(
     "public/**",
     "services/**",
     "components/**",
-    "env.ts",
     "tailwind.config.ts",
     "postcss.config.js",
     "next.config.js",
     "tsconfig.json",
+    "*.ts", # env.ts / instrumentation.ts
+    "instrumentation.node.ts"
```
bodymindarts, line 51: if *.ts is working then `instrumentation.node.ts` is redundant - or why does the comment say `instrumentation.ts`?

bodymindarts (approved):

apps/consent/BUCK
bodymindarts: Okay I see there are 2 files.

PR #3463, opened 2023-10-31
chore: buckify update / check super graph + schemas

bodymindarts:


.github/workflows/check-supergraph.yml
dolcalmi, line 1: is this just a test or check-code was moved?

bodymindarts: check code has moved into buck - sdl check was the last thing missing

vindard (approved):

PR #3471, opened 2023-11-01
build: add pnpm workspace context to Dockerfile-migrate

vindard:
## Description

This is to nest the files inside Dockerfile-migrate in the same structure as the monorepo and to bring in pnpm lock and workspace files so that packages can be installed from frozen versions.

Needed to fix: https://github.com/GaloyMoney/galoy/actions/runs/6714989030/job/18249348323?pr=3454


bodymindarts:
close in favour of https://github.com/GaloyMoney/galoy/pull/3472

PR #3483, opened 2023-11-02
feat(dashboard): api keys in dashboard

sandipndev:
Adds the following UI to CRUD api-keys.

| Create  | Created | List |
|---|---|---|
| Description 1 | Description 2 | Description 3 |


apps/dashboard/app/api/api-keys/route.ts
```diff
@@ -0,0 +1,57 @@
+import { createApiKey, revokeApiKey } from "@/services/graphql/mutations/api-keys"
```
siddhart1o1, line 1: We are using server actions; we should be consistent here also, right?

siddhart1o1 (dismissed):

sandipndev:
@bodymindarts @vindard this is ready, waiting for the backend to be deployed.

PR #3488, opened 2023-11-02
feat: add pending tx queries

UncleSamtoshi:
- add pendingTransactions to Account and Wallet GQL objects
- add pendingTransactionsByAddress to Wallet objects

### DIFF
--- a/core/api/dev/apollo-federation/supergraph.graphql
+++ b/core/api/dev/apollo-federation/supergraph.graphql
@@ -32,6 +32,7 @@ interface Account
   level: AccountLevel!
   limits: AccountLimits!
   notificationSettings: NotificationSettings!
+  pendingTransactions(walletIds: [WalletId]): [Transaction!]!
   realtimePrice: RealtimePrice!
   transactions(
     """Returns the items in the list that come after the specified cursor."""
@@ -221,6 +222,11 @@ type BTCWallet implements Wallet
 
   """An unconfirmed incoming onchain balance."""
   pendingIncomingBalance: SignedAmount!
+  pendingTransactions: [Transaction!]!
+  pendingTransactionsByAddress(
+    """Returns the items that include this address."""
+    address: OnChainAddress!
+  ): [Transaction!]!
   transactionById(transactionId: ID!): Transaction!
 
   """A list of BTC transactions associated with this wallet."""
@@ -344,6 +350,7 @@ type ConsumerAccount implements Account
   level: AccountLevel!
   limits: AccountLimits!
   notificationSettings: NotificationSettings!
+  pendingTransactions(walletIds: [WalletId]): [Transaction!]!
 
   """List the quiz questions of the consumer account"""
   quiz: [Quiz!]!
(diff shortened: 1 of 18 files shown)


core/api/dev/apollo-federation/supergraph.graphql
```diff
@@ -1960,6 +1972,25 @@ interface Wallet
...
+  """
+  pendingTransactions: [Transaction!]!
+
+  """
+  Pending OnChain transactions. When transactions
+  are confirmed they will receive a new id and be found in the transactions
+  list. Transactions are ordered anti-chronologically,
+  ie: the newest transaction will be first
+  """
+  pendingTransactionsByAddress(
```
dolcalmi, line 1990: should not this be a filter of `pendingTransactions` ? cc @vindard @bodymindarts @nicolasburtey

core/api/src/app/wallets/get-pending-transactions-by-addresses.ts
```diff
@@ -0,0 +1,25 @@
+import { CouldNotFindError } from "@/domain/errors"
+
+import { WalletOnChainPendingReceiveRepository } from "@/services/mongoose"
+
+export const getPendingTransactionsForWalletsByAddresses = async ({
```
dolcalmi, line 5: I know wallets are passed from source but we must enforce in app layer that wallets belong the user/account (similar to logic in `getPendingOnChainTransactionsForAccountByWalletIds`)

core/api/src/graphql/admin/schema.graphql
```diff
@@ -424,6 +429,11 @@ type UsdWallet implements Wallet {
 
   """An unconfirmed incoming onchain balance."""
   pendingIncomingBalance: SignedAmount!
+  pendingTransactions: [Transaction!]!
+  pendingTransactionsByAddress(
```
dolcalmi, line 433: how is this working if you need to pass wallet ids? did you test it?

core/api/src/graphql/public/types/object/consumer-account.ts
```diff
@@ -206,6 +208,35 @@ const ConsumerAccount = GT.Object<Account, GraphQLPublicContextAuth>({
...
+      resolve: async (source, args) => {
+        let { walletIds } = args
+
+        if (!walletIds) {
+          const wallets = await WalletsRepository().listByAccountId(source.id)
+          if (wallets instanceof Error) {
+            throw mapError(wallets)
+          }
+          walletIds = wallets.map((wallet) => wallet.id)
+        }
```
dolcalmi, line 227: this logic belongs to app layer (I know in some legacy code we use services from graphql layer but we should not introduce more debt on this)

dolcalmi (changes requested):

core/api/src/graphql/admin/schema.graphql
UncleSamtoshi: You dont need to pass wallet ids, this is already on the wallet object, so it uses the id of the wallet it is on.

core/api/src/app/wallets/get-pending-transactions-by-addresses.ts
UncleSamtoshi: I dont necessarily agree with that. The function does what it says which is `getPendingTransactionsForWalletsByAddress`. It's different than the `getPendingOnChainTransactionsForAccountByWalletIds` which says its getting them for a specific account and therefore must verify that the account owns the wallets. I think this is a pretty common pattern across the app layer.

core/api/dev/apollo-federation/supergraph.graphql
UncleSamtoshi: @bodymindarts reviewed this comment over standup and believes we should continue with this pattern

dolcalmi (approved):

PR #3489, opened 2023-11-02
chore(api-keys): auth endpoint boilerplate

bodymindarts:


dev/config/ory/oathkeeper.yml
```diff
@@ -28,8 +28,9 @@ authenticators:
     config:
       check_session_url: http://kratos:4433/sessions/whoami
       preserve_path: true
+      preserve_query: true
       subject_from: identity.id
-      extra_from: identity.traits
```
vindard, line 32: Flagging this in case we were specifically using it for something before

vindard (approved):

PR #3492, opened 2023-11-03
build: add api-keys dockerfile

vindard:


core/api-keys/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+COPY --from=builder /tmp/local-bin/* /usr/local/bin/
+
+USER 1000
+
+ARG COMMITHASH
+ENV COMMITHASH ${COMMITHASH}
+
+CMD [ \
+    "/usr/local/bin/api_keys" \
+    ]
```
vindard, line 32: Is this what we want to stick with as default behaviour?

PR #3498, opened 2023-11-03
feat(api-keys): implement 'list api keys' function for query

vindard:


core/api-keys/src/identity/mod.rs
```diff
@@ -105,4 +105,39 @@ impl Identities {
...
+        &self,
+        subject_id: &str,
+    ) -> Result<Vec<IdentityApiKey>, IdentityError> {
+        let identity = sqlx::query!(
+            r#"SELECT id FROM identities WHERE subject_id = $1"#,
+            subject_id,
+        )
+        .fetch_optional(&self.pool)
+        .await?
+        .ok_or(IdentityError::NoIdentityForSubject)?;
```
vindard, line 119: Should we create the identity here and return an empty array instead?

bodymindarts: not here but in app layer we should return an empty list

core/api-keys/src/identity/mod.rs
```diff
@@ -105,4 +105,39 @@ impl Identities {
...
+
+        let api_keys_records = sqlx::query!(
+            r#"SELECT id, name, created_at, expires_at
```
bodymindarts, line 122: not 2 queries - use a JOIN

core/api-keys/src/identity/mod.rs
```diff
@@ -105,4 +105,39 @@ impl Identities {
...
+               FROM identity_api_keys
+               WHERE identity_id = $1 AND active = true AND expires_at > NOW() AT TIME ZONE 'utc'"#,
```
bodymindarts, line 124: I think we can show `expired` keys - but not active ones. (active == revoked)

core/api-keys/src/identity/mod.rs
vindard: Solved with join

bodymindarts (approved):

PR #3512, opened 2023-11-07
feat: invoice pagination

UncleSamtoshi:
- add paginated invoices to account and wallet objects

DISCLAIMER:
- There is a potential for a bug because we are paginating using the dates of invoices to create an ordering. If two invoices were to have the exact same date and were used as a pagination cursor, one of the invoices would be skipped. It is highly unlikely that two invoices would have the exact same timestamp however it is not impossible. An alternative to this approach would be to sort on timestamp AND id however it would most likely require extra round trips to the database and be a performance hit.

Follow Up Items:

- [ ] Remove pending transactions from transactions pagination
- [ ] Fix transaction backwards pagination
- [ ] Switch transaction pagination types


core/api/src/domain/pagination/index.types.d.ts
```diff
@@ -0,0 +1,19 @@
+type PaginatedResult<T> = import("graphql-relay").Connection<T>
```
bodymindarts, line 1: Pagination is not really its own domain. Also domain code should not have an external dependency - especially not something as heavy as a IO layer framework.

core/api/src/app/wallets/get-invoices-for-wallets.ts
```diff
@@ -1,17 +1,22 @@
...
+    first?: number | null
+    last?: number | null
+    before?: string | null
+    after?: string | null
+  }
+}): Promise<PaginatedQueryResult<WalletInvoice> | ApplicationError> => {
   const walletIds = wallets.map((wallet) => wallet.id)
 
-  const parsedPaginationArgs = parsePaginationArgs({
+  const parsedPaginationArgs = checkedToPaginatedQueryArgs({
```
bodymindarts, line 19: I guess the variable should now be `checkedPaginationArgs`? Or even better `paginationArgs` and rename the above one to `rawPaginationArgs`... the checked type should be the default one (ie simpler name) the _unchecked_ one should have the special name (ie prefix).

core/api/src/domain/ledger/index.types.d.ts
```diff
@@ -2,6 +2,8 @@ type LedgerError = import("./errors").LedgerError
 type FeeDifferenceError = import("./errors").FeeDifferenceError
 type LedgerServiceError = import("./errors").LedgerServiceError
 
+type PaginationArgs = import("graphql-relay").ConnectionArguments
```
bodymindarts, line 5: Domain should not depend on external libraries.

UncleSamtoshi: This was already present before the PR. I think it might be showing up as a diff because I had moved it and then added it back and now there might be some whitespace change. I will have a follow up pr that fixes Transaction pagination by using the new types and removing  pending transactions from it. The follow up pr can not be merged till the mobile app has migrated to using the pendingTransactions endpoint.

bodymindarts (dismissed):

core/api/src/app/wallets/get-invoices-for-wallets.ts
UncleSamtoshi: Will make this change.

bodymindarts (approved):

PR #3529, opened 2023-11-10
refactor: remove optional types from ILedgerTransaction

vindard:


dolcalmi (changes requested):
We should not change Medici related interface.. we duplicate it to add new fields but that change could lead to unintended behavior

PR #3530, opened 2023-11-10
fix: tx pagination

UncleSamtoshi:
- removes pending incoming onchain transactions from transactions query


core/api/src/app/accounts/get-account-transactions-for-contact.ts
```diff
@@ -8,29 +9,50 @@ import { WalletsRepository } from "@/services/mongoose"
...
   contactUsername,
-  paginationArgs,
+  rawPaginationArgs,
 }: {
   account: Account
   contactUsername: Username
-  paginationArgs?: PaginationArgs
-}): Promise<PaginatedArray<WalletTransaction> | ApplicationError> => {
+  rawPaginationArgs: {
+    first?: number | null
```
nicolasburtey, line 17: the type here are `number | undefined | null`. any reason to not use undefined?

```suggestion
    first?: number | undefined
```

UncleSamtoshi: The arguments are technically both nullable and undefinable from the gql layer but it looks like that is actually the case in many places and we still just use undefined. I will switch to that.

core/api/src/app/accounts/get-account-transactions-for-contact.ts
```diff
@@ -8,29 +9,50 @@ import { WalletsRepository } from "@/services/mongoose"
...
+      memoSharingConfig,
+    })
 
-  const confirmedHistory = WalletTransactionHistory.fromLedger({
-    ledgerTransactions: resp.slice,
-    nonEndUserWalletIds: Object.values(await getNonEndUserWalletIds()),
-    memoSharingConfig,
+    return {
+      cursor: edge.cursor,
+      node: transactions[0],
```
dolcalmi, line 53: is it possible that transactions has length 0?

core/api/src/app/wallets/get-transactions-by-addresses.ts
```diff
@@ -1,61 +1,60 @@
...
   addresses: OnChainAddress[]
-  paginationArgs?: PaginationArgs
-}): Promise<PartialResult<PaginatedArray<WalletTransaction>>> => {
-  const walletIds = wallets.map((wallet) => wallet.id)
+  rawPaginationArgs: {
+    first?: number
+    last?: number
+    before?: string
+    after?: string
+  }
```
dolcalmi, line 21: since it is used in different files probably is better to create a type

core/api/src/app/wallets/get-transactions-by-addresses.ts
```diff
@@ -1,61 +1,60 @@
...
+      nonEndUserWalletIds,
+      memoSharingConfig,
+    })
 
-  return PartialResult.ok({
-    slice: transactions,
-    total: transactions.length,
+    return {
+      cursor: edge.cursor,
+      node: transactions[0],
```
dolcalmi, line 55: same question here

core/api/src/app/wallets/get-transactions-for-wallet.ts
```diff
@@ -1,55 +1,53 @@
...
+      nonEndUserWalletIds,
+      memoSharingConfig,
+    })
 
-  return PartialResult.ok({
-    slice: transactions,
-    total: confirmedLedgerTxns.total + pendingHistory.length,
+    return {
+      cursor: edge.cursor,
+      node: transactions[0],
```
dolcalmi, line 48: same here

core/api/src/graphql/public/types/object/consumer-account.ts
```diff
@@ -172,41 +165,19 @@ const ConsumerAccount = GT.Object<Account, GraphQLPublicContextAuth>({
...
-        let { walletIds } = args
-
-        if (!walletIds) {
-          const wallets = await WalletsRepository().listByAccountId(source.id)
-          if (wallets instanceof Error) {
-            throw mapError(wallets)
-          }
-          walletIds = wallets.map((wallet) => wallet.id)
-        }
+        const { walletIds } = args
```
dolcalmi, line 168:
```suggestion
        const { walletIds, ...rawPaginationArgs } = args
```

core/api/src/services/ledger/paginated-ledger.ts
```diff
@@ -9,34 +9,34 @@ import { parseFilterQuery } from "medici/build/helper/parse/parseFilterQuery"
...
-    return new InvalidPaginationArgumentsError()
-  }
+  filters: LedgerQueryFilter
+  paginationArgs: PaginatedQueryArgs
+}): Promise<Error | PaginatedQueryResult<LedgerTransaction<WalletCurrency>>> => {
+  const filterQuery = parseFilterQuery(filters.mediciFilters, MainBook)
+
+  const { first, last, before, after } = paginationArgs
+
+  filterQuery["_id"] = { $exists: true }
```
dolcalmi, line 39: is this needed?

core/api/src/services/ledger/paginated-ledger.ts
```diff
@@ -46,27 +46,59 @@ export const paginatedLedger = async ({
...
 
   return {
-    slice,
-    total,
+    edges: txs.map((tx) => ({
+      cursor: checkedToPaginatedQueryCursor(tx.id),
+      node: tx,
+    })),
+    pageInfo: {
+      startCursor: txs[0]?.id ? checkedToPaginatedQueryCursor(txs[0].id) : undefined,
```
dolcalmi, line 96: what happen if txs lenght is 0?

core/api/src/services/ledger/paginated-ledger.ts
```diff
@@ -46,27 +46,59 @@ export const paginatedLedger = async ({
...
+      endCursor: txs[txs.length - 1]?.id
+        ? checkedToPaginatedQueryCursor(txs[txs.length - 1].id)
```
dolcalmi, line 98: same question here

dolcalmi (changes requested):
looks good, just minor changes/questions

core/api/src/app/wallets/get-transactions-by-addresses.ts
UncleSamtoshi: Will make that change.

core/api/src/services/ledger/paginated-ledger.ts
UncleSamtoshi: Technically the "_id" field is optional in the schema, so I added it in.

core/api/src/services/ledger/paginated-ledger.ts
UncleSamtoshi: Then the end cursor will be undefined, which is the expected behavior.

bats/helpers/onchain.bash
```diff
@@ -45,9 +45,34 @@ get_from_transaction_by_address() {
...
+get_from_pending_transaction_by_address() {
+  property_query=$2
+
+  jq_query='.data.me.defaultAccount.pendingIncomingTransactions[] | select(.initiationVia.address == $address)'
+  echo $output \
+    | jq -r --arg address "$1" "$jq_query" \
+    | jq -r "$property_query"
+}
+
 check_for_broadcast() {
```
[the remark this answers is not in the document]
UncleSamtoshi: I can make that change.

core/api/src/app/accounts/get-account-transactions-for-contact.ts
UncleSamtoshi: Actually I will just change the method to take in a single tx instead of a list.

dolcalmi (approved):

PR #3534, opened 2023-11-12
chore(api-keys): add tracing to graphql schema

sandipndev:
Adds tracing for async_graphql schema


bodymindarts:
I had this added but don’t see much value. It adds lots of spans that aren’t very useful. Very noisy IMO

PR #3547, opened 2023-11-14
test: onchain tests to top level bats

sandipndev:
- Create Random user helper
- E2E test onchain stuff to top level


bats/helpers/_common.bash
```diff
@@ -118,3 +118,11 @@ cache_value() {
 read_value() {
   cat ${CACHE_DIR}/$1
 }
+
+bitcoin_cli() {
```
bodymindarts, line 122: This belongs in helpers/cli.sh

bats/helpers/funding/onchain.bash
```diff
@@ -0,0 +1,84 @@
...
+  exec_graphql "$token_name" 'on-chain-address-create' "$variables"
+  address="$(graphql_output '.data.onChainAddressCreate.address')"
+  [[ "${address}" != "null" ]] || exit 1
+
+  bitcoin_cli sendtoaddress "$address" "$btc_amount_in_btc"
+  bitcoin_cli -generate 2
+  retry 30 1 check_for_onchain_initiated_settled "$token_name" "$address"
+}
+
+check_for_onchain_initiated_settled() {
```
bodymindarts, line 23: I don't think the `check_xx` and `get_from_transaction_by_address` functions belong here under `funding`.

bats/helpers/funding/wallet.bash
```diff
@@ -0,0 +1,11 @@
+balance_for_wallet() {
```
bodymindarts, line 1: not sure about having this under `funding`.
Funding (for me) implies sending / seeding money into a wallet. Getting a balance is a more general function.

What about something like this:
```
helpers
- onchain.bash
- intraledeger.bash
- ln.bash
- wallet.bash
```

Or something like that.

bats/helpers/funding/onchain.bash
sandipndev: FIxed in 9757bd4e56a6b4117e750b1782765b5ef35556e2

bats/helpers/funding/wallet.bash
sandipndev: fixed in 9757bd4e56a6b4117e750b1782765b5ef35556e2

bodymindarts (approved):

PR #3554, opened 2023-11-15
feat(api-keys): add read_only field to api-keys

bodymindarts:


core/api/src/servers/middlewares/scope.ts
```diff
@@ -3,79 +3,77 @@ import { GraphQLResolveInfo, GraphQLFieldResolver } from "graphql"
...
 ) => {
   const scope = context.scope
-  const appId = context.appId
-
-  // not a delegated token
-  if (appId === undefined || appId === "") {
-    return resolve(parent, args, context, info)
-  }
 
+  console.log("SCOPE ", scope)
```
nicolasburtey, line 17:
```suggestion
```

core/api/src/servers/middlewares/scope.ts
```diff
@@ -3,79 +3,77 @@ import { GraphQLResolveInfo, GraphQLFieldResolver } from "graphql"
...
+    return resolve(parent, args, context, info)
   }
 
-  if (scope.find((s) => s === ScopesOauth2.TransactionsRead) !== undefined) {
+  if (scope.find((s) => s === ScopesOauth2.Read) !== undefined) {
     return resolve(parent, args, context, info)
   }
 
-  return mapError(new AuthorizationError("not authorized to read transactions"))
+  return mapError(new AuthorizationError("not authorized to read data"))
```
nicolasburtey, line 27: I'm assuing if we have a read_only, the alternative is a read+write token, so I don't think there is a scenario currently in which we are not authorized to read?

core/api/src/servers/middlewares/scope.ts
```diff
@@ -3,79 +3,77 @@ import { GraphQLResolveInfo, GraphQLFieldResolver } from "graphql"
...
+  if (scope === undefined || scope.length === 0) {
     return resolve(parent, args, context, info)
   }
 
-  if (scope === undefined) {
-    return mapError(new AuthorizationError("appId is defined but scope is undefined"))
-  }
-
-  if (scope.find((s) => s === ScopesOauth2.PaymentsSend) !== undefined) {
+  if (scope.find((s) => s === ScopesOauth2.Write) !== undefined) {
```
nicolasburtey, line 44: will we get an otel span propagate with the scope of the request?

dev/config/ory/oathkeeper_rules.yaml
```diff
@@ -87,7 +87,7 @@
   mutators:
     - handler: id_token
       config: #! TODO: add aud: {"aud": ["https://api/graphql"] }
-        claims: '{"sub": "{{ print .Subject }}", "session_id": "{{ print .Extra.id }}", "expires_at": "{{ print .Extra.expires_at }}", "scope": "{{ print .Extra.scope }}", "client_id": "{{ print .Extra.client_id }}" }'
+        claims: '{"sub": "{{ print .Subject }}", "session_id": "{{ print .Extra.id }}", "expires_at": "{{ print .Extra.expires_at }}", "scope": "{{ print .Extra.scope }}", "client_id": "{{ print .Extra.client_id }}"}'
```
nicolasburtey, line 90: is there a way to know if a request comes from an API token versus kratos or hydra?

core/api/src/servers/middlewares/scope.ts
bodymindarts: Theoretically when logging into dashboard someone may just select 'Offline' scope - though that is unlikely as we are skipping that step in production.

core/api/src/servers/middlewares/scope.ts
bodymindarts: They should be added here but I can't seem to find the information in the query https://github.com/GaloyMoney/galoy/blob/b87bfcd1539d35045551b1457ba478f8841041e2/core/api/src/servers/graphql-main-server.ts#L82

dev/config/ory/oathkeeper_rules.yaml
bodymindarts: I don't believe we currently have a clear differentiation ATM.

vindard (approved):

core/api/src/servers/middlewares/scope.ts
nicolasburtey: I think we can remove the `offline` for now btw, we don't need a refresh token for our current use case. I was using it when developing on hydra to get a deeper understanding of hydra and oauth2

bats/core/api-keys/api-keys.bats
```diff
@@ -67,3 +71,44 @@ new_key_name() {
...
+  key_id=$(echo "$key" | jq -r '.id')
+  cache_value "api-key-id" "$key_id"
+
+  exec_graphql 'api-key-secret' 'api-keys'
+
+  name="$(graphql_output '.data.me.apiKeys[-1].name')"
+  [[ "${name}" = "${key_name}" ]] || exit 1
+}
+
+@test "api-keys: read-only key cannot mutate" {
```
nicolasburtey, line 97: do we have a test checking that readOnly key can read?

bodymindarts: Yes in the test above (that creates the read-only key)

PR #3555, opened 2023-11-15
fix(api): withdrawal volumes handling for failed payments

vindard:
## Description

This fixes an issue that was discovered where a user tried to send $700 twice (and got routing errors) and then on the third attempt was hit was a `WithdrawalLimitExceeded` error since the previous 2 failed attempts were wrongly accounted for when calculating volumes.

_See pattern in [traces](https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/FjAzT7CJvnt)_

### DIFF
--- a/core/api/src/services/ledger/facade/volume.ts
+++ b/core/api/src/services/ledger/facade/volume.ts
@@ -21,6 +21,7 @@ export const TxnGroups = {
   ],
   externalPaymentVolumeSince: [
     LedgerTransactionType.Payment,
+    LedgerTransactionType.LnFeeReimbursement,
     LedgerTransactionType.OnchainPayment,
   ],
   intraledgerTxBaseVolumeSince: [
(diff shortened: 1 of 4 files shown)


core/api/src/services/ledger/facade/volume.ts
```diff
@@ -71,6 +73,34 @@ const TxVolumeAmountSinceFactory = () => {
             $and: [{ timestamp: { $gte: timestamp } }],
           },
         },
+        {
+          $lookup: {
```
vindard, line 77: Not sure if this might have performance implications. Two things we can do:
- add indices for `_journal` (and `_original_journal`?)
- switch to using datetime from `journals` collection instead of `medici_transactions` collection

vindard: Update: `_journal` is already indexed by medici (see [here](https://github.com/flash-oss/medici/blob/master/src/models/transaction.ts#L75)), and we can alternatively change the lookup collection later on if we notice performance issues

dolcalmi (approved):

PR #3557, opened 2023-11-15
chore: prevent WalletInvoice deletion

UncleSamtoshi:
- add `processingCompleted` field to replace deleting canceled invoices

To Do:
- [ ] After this pr has been merged and is live for more than the expiration period of invoices, migrate mongo db to set all non-existent `processingCompleted` fields to `true`
- [ ] After db migration change `processingCompleted` field to be non-optional

### DIFF
--- a/core/api/src/app/wallets/update-pending-invoices.ts
+++ b/core/api/src/app/wallets/update-pending-invoices.ts
@@ -116,10 +116,11 @@ export const declineHeldInvoice = wrapAsyncToRunInSpan({
     })
 
     if (lnInvoiceLookup instanceof InvoiceNotFoundError) {
-      const isDeleted = await walletInvoicesRepo.deleteByPaymentHash(paymentHash)
-      if (isDeleted instanceof Error) {
-        pendingInvoiceLogger.error("impossible to delete WalletInvoice entry")
-        return isDeleted
+      const processingCompletedInvoice =
+        await walletInvoicesRepo.markAsProcessingCompleted(paymentHash)
+      if (processingCompletedInvoice instanceof Error) {
+        pendingInvoiceLogger.error("Unable to mark invoice as processingCompleted")
+        return processingCompletedInvoice
       }
       return false
     }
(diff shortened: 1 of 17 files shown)


core/api/src/app/wallets/update-pending-invoices.ts
```diff
@@ -302,6 +316,7 @@ const updatePendingInvoiceBeforeFinally = async ({
       "invoices.finalRecipient": JSON.stringify(recipientWalletDescriptor),
     })
 
+    // Is this crossing the lock boundary?
```
dolcalmi, line 319: please remove, also not sure... what you mean? this can be done outside the lock but need to be done previous to ledger record (that must be inside lock)

core/api/src/services/mongoose/schema.types.d.ts
```diff
@@ -69,6 +69,7 @@ interface WalletInvoiceRecord {
   currency: string
   timestamp: Date
   selfGenerated: boolean
+  processingCompleted?: boolean
```
dolcalmi, line 72: why do we need to add it as optional? isnt default value working?

dolcalmi (changes requested):
looks good, just few comments

core/api/src/services/mongoose/schema.types.d.ts
UncleSamtoshi: Isn't default value only used on write? I have a database migration that will go back and populate this field and then i will make it non optional.

core/api/src/app/wallets/update-pending-invoices.ts
UncleSamtoshi: I will remove it, but `lnInvoiceLookup` is populated before the lock, so its state could change by the time the lock is acquired.

dolcalmi: ahh I thought you referred to settle... you are right, feel free to move it

dolcalmi (approved):

PR #3558, opened 2023-11-15
fix(api): idempotency bug in registerBroadcastHandler

vindard:
## Description

In PR #3550, by fixing the idempotency issue where an `InvalidLedgerTransactionStateError` is thrown if the handler is re-run and records a reconciliation transaction (see [traces](https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/m1PJPNtR4S2)), we introduced a new bug where multiple reconciliation transaction are created on re-run.

This PR cleans up the fix from #3550 by introducing an explicit `isOnChainFeeReconciliationTxn` check and it fixes the new multiple-reconciliation-txns bug.


core/api/src/services/ledger/facade/onchain-reconcile.ts
```diff
@@ -0,0 +1,78 @@
...
+      account: toLiabilitiesWalletId(bankOwnerWalletId),
+    })
+    /* eslint @typescript-eslint/ban-ts-comment: "off" */
+    // @ts-ignore-next-line no-implicit-any error
+    const txns = results.map((tx) => translateToLedgerTx(tx))
+
+    const reconciliationTxn = txns.find((txn) => isOnChainFeeReconciliationTxn(txn))
+    return reconciliationTxn !== undefined
+  } catch (err) {
+    return new UnknownLedgerError(err)
```
dolcalmi, line 76: needs to use err.message or `parseErrorMessageFromUnknown`

core/api/src/services/ledger/facade/onchain-reconcile.ts
```diff
@@ -0,0 +1,78 @@
...
+): Promise<boolean | LedgerFacadeError> => {
+  try {
+    const bankOwnerWalletId = await getBankOwnerWalletId()
+    const { results } = await MainBook.ledger({
+      payout_id: payoutId,
+      account: toLiabilitiesWalletId(bankOwnerWalletId),
+    })
+    /* eslint @typescript-eslint/ban-ts-comment: "off" */
+    // @ts-ignore-next-line no-implicit-any error
+    const txns = results.map((tx) => translateToLedgerTx(tx))
```
dolcalmi, line 71: why is this returning any? `translateToLedgerTx` is returning proper types

core/api/src/services/ledger/facade/onchain-reconcile.ts
```diff
@@ -0,0 +1,78 @@
...
+  return persistAndReturnEntry({
+    entry,
+    hash: metadata.hash,
+  })
+}
+
+export const isOnChainFeeReconciliationTxn = (
+  txn: LedgerTransaction<WalletCurrency>,
+): boolean =>
+  txn.type === LedgerTransactionType.OnchainPayment && txn.address === undefined
```
dolcalmi, line 58: not sure when we began to use non undefinied/null/empty validation (should be `!txn.address` we already had an issue in prod with this) also.. should not include bankowner wallet id validation?

core/api/src/services/ledger/facade/onchain-reconcile.ts
```diff
@@ -0,0 +1,78 @@
...
+    const { results } = await MainBook.ledger({
+      payout_id: payoutId,
+      account: toLiabilitiesWalletId(bankOwnerWalletId),
+    })
+    /* eslint @typescript-eslint/ban-ts-comment: "off" */
+    // @ts-ignore-next-line no-implicit-any error
+    const txns = results.map((tx) => translateToLedgerTx(tx))
+
+    const reconciliationTxn = txns.find((txn) => isOnChainFeeReconciliationTxn(txn))
+    return reconciliationTxn !== undefined
```
dolcalmi, line 74: same here `!!reconciliationTxn`

core/api/src/app/wallets/register-broadcasted-payout-txn.ts
```diff
@@ -21,6 +21,9 @@ export const registerBroadcastedPayout = async ({
   const { estimatedProtocolFee } = setTxIdResult
   if (estimatedProtocolFee.amount === proportionalFee.amount) return true
 
+  const isRecorded = await LedgerFacade.isOnChainFeeReconciliationRecorded(payoutId)
+  if (isRecorded !== false) return isRecorded
```
dolcalmi, line 25: this is a weird validation.. is it to avoid error validation?

core/api/src/services/ledger/facade/onchain-send.ts
```diff
@@ -168,7 +169,7 @@ export const setOnChainTxIdByPayoutId = async ({
 
   const bankOwnerWalletId = await getBankOwnerWalletId()
   const bankOwnerTxns = txns.filter(
-    (txn) => txn.satsFee && txn.walletId === bankOwnerWalletId,
+    (txn) => !isOnChainFeeReconciliationTxn(txn) && txn.walletId === bankOwnerWalletId,
```
dolcalmi, line 172: can reconciliation be applied to other wallets (different from bankowner)?

dolcalmi (changes requested):

core/api/src/services/ledger/facade/onchain-reconcile.ts
vindard: It looks like we already have this handling in the `DomainError` constructor?
https://github.com/GaloyMoney/galoy/blob/b87bfcd1539d35045551b1457ba478f8841041e2/core/api/src/domain/shared/errors.ts#L15

core/api/src/services/ledger/facade/onchain-reconcile.ts
vindard: In this case I was thinking it should specifically check for the absence of the `address` property (since the `OnChainAddress` can technically also be falsy/empty even though it should never be). Or, what sort of issue has this thrown in prod?

> should not include bankowner wallet id validation
Maybe not necessary since the reconciliation transaction only ever involves `Liabilities:bankOwner` account and `Asset:OnChain` account and I believe we only bring in `Liabilities` accounts in our ledger `get` methods. `address` absence should be enough for this check?

Also for this check, ideally it should be a different tx `type` value instead of also being an `onchain_payment` so we don't even need the check, but that might be a tricky migration to do. And I also considered maybe setting a custom description for these types of transactions but then there was no way to guarantee that someone couldn't set the same description on normal onchain-payment transactions which would break the heuristic. I switched from checking for `satsFee` presence to `address` presence since this is more isolated and might less likely eventually be populated on the reconciliation transactions.

core/api/src/services/ledger/facade/onchain-reconcile.ts
vindard: Ya same question here. I started doing this to state `"is this undefined or a LedgerTransaction"` instead of `"is this a truthy LedgerTransaction"` which implicitly assumes that LedgerTransaction is always truthy.

But yea, I can change if this has been an issue in the past in runtime.

core/api/src/app/wallets/register-broadcasted-payout-txn.ts
vindard: Nope, it's that if there was an error we should return the error here, or if the onchain reconciliation was already recorded (`true`) we should also return here. We would only continue if the `isOnChainReconciliationrRecorded` is `false` in which case the rest of the function continues on to record the reconciliation.

core/api/src/services/ledger/facade/onchain-send.ts
vindard: I don't see why we would do that right now. `recordReceiveOnChainFeeReconciliation` construct entries that only involve the `Liabilities:bankOwner` and `Assets:OnChain` via the `FeeOnlyEntryBuilder`

core/api/src/services/ledger/facade/onchain-reconcile.ts
vindard: Good catch, I'd just copy-pasted the previous implementation to a new file.

core/api/src/services/ledger/facade/onchain-send.ts
vindard: Mm, it shouldn't be since reconciliations only involve `Liabilities:bankOwner` and `Assets:OnChain`. Why do you ask?

core/api/src/app/wallets/register-broadcasted-payout-txn.ts
dolcalmi: yep, I meant that `!isRecorded || isRecorded instanceof Error`

core/api/src/services/ledger/facade/onchain-send.ts
dolcalmi: if that is the case then walletId validation should be in `isOnChainFeeReconciliationTxn`

core/api/src/services/ledger/facade/onchain-reconcile.ts
dolcalmi: here is not an issue because find is not an external service/library

core/api/src/services/ledger/facade/onchain-reconcile.ts
dolcalmi: at least with validations in data from external services we must be more "defensive", we cant guarantee the undefined so it is safer just to use `!txn.address`

FYI: the error was a validation using undefined and a service returning null value (typescript in theory was ok but you know we cant trust TS )

vindard: Oh yea ok that makes sense.

For here though, `address` is set explicitly in our `translateToLedgerTx` function, so there's also less risk of it being something unexpected? We handle the uncertain-falsy-check-from-external-service in the line `tx.payee_addresses && tx.payee_addresses.length > 0`

core/api/src/services/ledger/facade/onchain-send.ts
vindard: Not sure if that would be helpful here though, because we'd still need to do the `txn.walletId === bankOwnerWalletId` check since normal non-reconciliation `onchain_payment` txns can also be from user wallet ids

dolcalmi (approved):

PR #3561, opened 2023-11-16
build: add audit buck build steps

vindard:
## Description

This is to buckify our audit steps and use them in github actions in place of the legacy way of doing audit.

_Example report from fail: https://github.com/GaloyMoney/galoy/actions/runs/6895175112/job/18758564398_


toolchains/workspace-pnpm/run_audit.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python3
+"""
+Runs a program in a directory.
```
bodymindarts, line 3: Probably oversight - not a good comment.

.github/workflows/buck2-audit.yaml
```diff
@@ -0,0 +1,37 @@
+name: Buck2 audit
```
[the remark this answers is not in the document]
vindard: Was thinking that as well. I put it as a separate step for feedback. Consolidated now into `test-unit` (https://github.com/GaloyMoney/galoy/pull/3561/commits/bc4cfa391b3b037f76eda46b1301aaea8ef27716)

toolchains/workspace-pnpm/run_audit.py
vindard: Thanks, yup was from `run_in_dir.py` copy-paste

PR #3562, opened 2023-11-16
ci: remove check-code

bodymindarts:


ci/pipeline.yml
```diff
@@ -6,7 +6,6 @@
 #@   "nodejs_concourse_image",
 #@   "slack_failure_notification",
 #@   "nodejs_check_code",
```
vindard, line 8: Do we still need check-code here too?

vindard (approved):

ci/pipeline.yml
bodymindarts: no good catch

PR #3567, opened 2023-11-17
fix(core): avoid unknown exceptions in setGqlContext

dolcalmi:
Avoid 500 errors for [CouldNotFindAccountFromKratosIdError](https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/6XnXpZ12jLz/trace/f9RTg5P266F?fields[]=s_name&fields[]=s_serviceName&span=1aeaee98a7b2420b)

### DIFF
--- a/core/api/src/servers/graphql-main-server.ts
+++ b/core/api/src/servers/graphql-main-server.ts
@@ -30,14 +30,13 @@ import {
   ACCOUNT_USERNAME,
   SemanticAttributes,
   addAttributesToCurrentSpanAndPropagate,
-  recordExceptionInCurrentSpan,
 } from "@/services/tracing"
 
 import { parseIps } from "@/domain/accounts-ips"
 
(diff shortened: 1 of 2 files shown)


core/api/src/servers/graphql-main-server.ts
```diff
@@ -61,15 +60,6 @@ const setGqlContext = async (
     ip,
   })
 
-  if (gqlContext instanceof Error) {
-    recordExceptionInCurrentSpan({
-      error: gqlContext,
-      fallbackMsg: "error executing sessionPublicContext",
-    })
-    next(gqlContext)
```
nicolasburtey, line 69: the key is to not use this?

dolcalmi: yes, this throw an error and we dont have a global express error handler. the change "hide" the not found error and record it as a critical in the span

nicolasburtey (approved):

PR #3572, opened 2023-11-20
chore(api): clean up api runs for local dev

vindard:
## Description

We were getting a port clash because admin api start was being executed twice (didn't affect functionality, but led to misleading error messages). This fixes that issue and adds dev commands for running public and admin api servers.


dev/Tiltfile
```diff
@@ -177,7 +180,7 @@ local_resource(
             port = 4012,
         ),
     ),
-    deps = _buck2_dep_inputs(api_target),
+    deps = _buck2_dep_inputs(api_target_ci),
```
bodymindarts, line 183: don't see why this is required? The dev target should also know its files.

vindard: Fixed https://github.com/GaloyMoney/galoy/pull/3572/commits/15424efdab70847b6b0b7a298dc0d97c088f40d8

core/api/src/servers/graphql-admin-server.ts
```diff
@@ -112,9 +112,11 @@ export async function startApolloServerForAdminSchema() {
...
-  })
-  .catch((err) => graphqlLogger.error(err, "server error"))
+if (require.main === module) {
+  setupMongoConnection()
+    .then(async () => {
+      activateLndHealthCheck()
+      await startApolloServerForAdminSchema()
+    })
+    .catch((err) => graphqlLogger.error(err, "server error"))
+}
```
nicolasburtey, line 122:
```suggestion
```

nicolasburtey: I think we. can just removed it

the reason I remove the wrapper around `if (require.main === module) {` initially was to make it compatible with `bun`.

core/api/src/servers/graphql-main-server.ts
```diff
@@ -123,16 +123,18 @@ export async function startApolloServerForCoreSchema() {
...
-    ])
-  })
-  .catch((err) => baseLogger.error(err, "server error"))
+      await Promise.race([
+        startApolloServerForCoreSchema(),
+        startApolloServerForAdminSchema(),
+      ])
+    })
+    .catch((err) => baseLogger.error(err, "server error"))
+}
```
nicolasburtey, line 140: we should move that to another file

vindard: Done https://github.com/GaloyMoney/galoy/pull/3572/commits/7b806f1ccd581d03f639c3fe4d1af3a7cd53ef84

PR #3574, opened 2023-11-20
feat: request code can use appcheckjti

nicolasburtey:
this endpoint will serve as an alternative from the captcha endpoint

### DIFF
--- a/core/api/src/servers/authentication/index.ts
+++ b/core/api/src/servers/authentication/index.ts
@@ -78,7 +78,7 @@ authRouter.post("/create/device-account", async (req: Request, res: Response) =>
   const appcheckJti = req.headers["x-appcheck-jti"]
   addAttributesToCurrentSpan({ "appcheck.jti": appcheckJti })
 
-  if (!appcheckJti || typeof appcheckJti !== "string" || appcheckJti === "") {
+  if (!appcheckJti || typeof appcheckJti !== "string") {
     return res.status(401).send({ error: "missing or invalid appcheck jti" })
   }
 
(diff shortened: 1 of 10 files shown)


core/api/src/servers/authentication/index.ts
```diff
@@ -307,6 +295,37 @@ authRouter.post("/phone/code", async (req: Request, res: Response) => {
   })
 })
 
+authRouter.post("/phone/code-appcheck", async (req: Request, res: Response) => {
+  const appcheckJti = req.headers["x-appcheck-jti"]
+  if (!appcheckJti || typeof appcheckJti !== "string" || appcheckJti === "") {
+    return res.status(400).send({ error: "missing or invalid appcheck jti" })
+  }
+
+  const ip = req.originalIp
```
dolcalmi, line 304: should not implement the same code used in setGqlContext?

core/api/src/servers/authentication/index.ts
```diff
@@ -307,6 +295,37 @@ authRouter.post("/phone/code", async (req: Request, res: Response) => {
   })
 })
 
+authRouter.post("/phone/code-appcheck", async (req: Request, res: Response) => {
+  const appcheckJti = req.headers["x-appcheck-jti"]
+  if (!appcheckJti || typeof appcheckJti !== "string" || appcheckJti === "") {
```
dolcalmi, line 300: you can remove `appcheckJti === ""`

core/api/src/servers/authentication/index.ts
```diff
@@ -307,6 +295,37 @@ authRouter.post("/phone/code", async (req: Request, res: Response) => {
...
+
+  const result = await Authentication.requestPhoneCodeWithAppcheckJti({
+    phone,
+    appcheckJti,
+    ip,
+    channel,
+  })
+
+  if (result instanceof Error) {
+    return res.status(400).json({ error: result.name })
```
dolcalmi, line 321: please register error `recordExceptionInCurrentSpan`

dolcalmi (changes requested):

core/api/src/servers/authentication/index.ts
nicolasburtey: `setGqlContext` is for graphql routes. there are already 2 different setGqlContext (one for public and one for admin routes). so I'm not sure it's relevant to use/reuse code from setGqlContext because here it's a different route (`/auth` router) with a different context, but maybe I'm missing something?

dolcalmi (approved):

PR #3579, opened 2023-11-21
chore: adding open telemetry to the auth router

nicolasburtey:


core/api/src/services/tracing.ts
```diff
@@ -205,6 +207,14 @@ registerInstrumentations({
     new MongoDBInstrumentation(),
     new GrpcInstrumentation(),
     new IORedisInstrumentation(),
+    new ExpressInstrumentation({
+      requestHook: (span, request) => {
+        console.log("requestHook", request)
+        console.log("span", span)
+        // span NonRecordingSpan {}
```
dolcalmi, line 214: please remove

core/api/test/bats/device-account.bats
```diff
@@ -22,11 +22,7 @@ jwt="eyJhbGciOiJSUzI1NiIsImtpZCI6IjFiOTdiMjIxLWNhMDgtNGViMi05ZDA5LWE1NzcwZmNjZWI
   username="$(random_uuid)"
   password="$(random_uuid)"
 
-  if [[ "$(uname)" == "Linux" ]]; then
-      basic_token="$(echo -n $username:$password | base64 -w 0)"
-  else
-      basic_token="$(echo -n $username:$password | base64)"
-  fi
+  basic_token="$(echo -n $username:$password | base64 -w 0)"
```
dolcalmi, line 25: do you need to rebase?

dolcalmi (changes requested):

core/api/src/services/tracing.ts
nicolasburtey: I want to keep the log for now. for some reason it's returning `NonRecordingSpan` in dev and I don't know why. I'm not sure if this would behave differently on staging.

dolcalmi (approved):

PR #3580, opened 2023-11-21
build(core): switch Dockerfile-websocket to use pnpm

vindard:
## Description

This PR replaces our last outstanding yarn dependency and then removes the unused `yarn.lock` file


core/api/Dockerfile-websocket
```diff
@@ -30,4 +30,4 @@ USER 1000
 ARG COMMITHASH
 ENV COMMITHASH ${COMMITHASH}
 
-CMD ["lib/servers/ws-server.js"]
+CMD ["dist/servers/ws-server.js"]
```
vindard, line 40: Double-checking, is this dockerfile used? This has been a bug that made the file unusable for the last few weeks

bodymindarts:
Dockerfile is no longer used / removed in https://github.com/GaloyMoney/galoy/pull/3581

PR #3587, opened 2023-11-22
build(core): add dev rules for running trigger & ws servers locally

vindard:


toolchains/workspace-pnpm/macros.bzl
```diff
@@ -1021,13 +1021,15 @@ if [ "$install_node_modules" = "True" ]; then
...
-    args = cmd_args([script, str(ctx.attrs.local_node_modules), ctx.label.package, ctx.attrs.command])
+    package_dir = ctx.attrs.package_dir or ctx.label.package
+    args = cmd_args([script, str(ctx.attrs.local_node_modules), package_dir, ctx.attrs.command])
     args.hidden([ctx.attrs.deps])
     args.hidden([ctx.attrs.srcs])
     return [DefaultInfo(), RunInfo(args = args)]
 
 dev_pnpm_task_binary = rule(impl = pnpm_task_binary_impl, attrs = {
     "command": attrs.string(doc = """pnpm command to run"""),
+    "package_dir": attrs.string(default = "", doc = """package dir to run commands from"""),
```
bodymindarts, line 1032: we should not expose package_dir IMO.

Rather we should define the target in the correct package (ie core/api) and then alias it from where we want to call it (ie. core/api-ws-server)

vindard: Fixed

bodymindarts (approved):

PR #3590, opened 2023-11-22
refactor(core): move 'watch:compile' from make to buck

vindard:
## Description

This change would mean that to watch-compile we would now do:
```bash
$ buck2 run //core/api:watch-compile
```

cc @nicolasburtey @dolcalmi

### DIFF
--- a/core/api/BUCK
+++ b/core/api/BUCK
@@ -199,3 +199,9 @@ dev_pnpm_task_binary(
   srcs = [":src"],
   visibility = ["PUBLIC"],
 )
+
+dev_pnpm_task_binary(
+  name = "watch-compile",
+  command = "watch:compile",
+  srcs = [":src"] + [":test_src"],
+)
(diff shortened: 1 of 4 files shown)


bodymindarts (dismissed):

bodymindarts:
Approved but lets see if it works for @dolcalmi

bats/helpers/subscriber/output.txt
dolcalmi, line 1: what is the purpose of this file?

core/api/output.txt
dolcalmi, line 1: same here

test_script.sh
dolcalmi, line 1: where is this used?

dolcalmi (changes requested):

test_script.sh
vindard: Ah sorry good catch, these got accidentally committed

dolcalmi (approved):

PR #3615, opened 2023-11-28
refactor: target tests more specifically in Tiltfile

vindard:
With auto_init, all other resources are brought up in addition to the targeted tests. With 'enabled_resources' only the tests and their dependency resources are brought up.

This is required to run the 'core' test suite for example, since that suite finishes long before the rest of server resources come up.

This will for example significantly reduce the time for the new `integration-tests` pipeline step from #3612 to execute in CI.

### DIFF
--- a/dev/Tiltfile
+++ b/dev/Tiltfile
@@ -355,11 +355,11 @@ if to_run != []:
 
 to_test = cfg.get("test", [])
 if to_test != []:
-    enabled_resources = [
-        TEST_RESOURCES.get(label)
-        for label in to_test
-        if TEST_RESOURCES.get(label)
-    ]
+    enabled_resources = []
+    for label in to_test:
+        svc = TEST_RESOURCES.get(label)
+        if svc:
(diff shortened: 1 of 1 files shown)


dev/Tiltfile
```diff
@@ -343,6 +351,18 @@ if to_run != []:
...
 
+to_test = cfg.get("test", [])
+if to_test != []:
+    if not is_ci:
+        print("Please re-run tests in 'ci' mode")
+    else:
+        enabled_resources = [
+           TEST_RESOURCES.get(label)
+           for label in to_test
+           if TEST_RESOURCES.get(label)
```
bodymindarts, line 362: is this indentation right?

vindard: Refactored to be more readables (and less pythonic) https://github.com/GaloyMoney/galoy/pull/3615/commits/7ac47dce93177c85f6847364cd35488826a99529

bodymindarts (approved):

PR #3618, opened 2023-11-28
feat: add galoy pay

UncleSamtoshi:


.github/labeler.yml
```diff
@@ -12,6 +12,11 @@ consent:
   - flake.lock
   - pnpm-lock.yaml
 
+pay:
+  - apps/dashboard/*
```
vindard, line 16:
```suggestion
  - apps/pay/*
```

apps/pay/Dockerfile
```diff
@@ -0,0 +1,42 @@
+# Install dependencies only when needed
+FROM node:20-alpine AS deps
```
vindard, line 2: This file can probably be redone similar to ones in `app/dashboard` and `apps/consent`

apps/pay/docker-compose.yml
```diff
@@ -0,0 +1,8 @@
+version: "3.9"
```
vindard, line 1: Not sure if this docker-compose file will be necessary anymore

vindard (changes requested):
I added a commit with some `Tiltfile` additions to have pay be available for the integration github action and for when we bring up the monorepo with `buck2 run //dev:up`

There's some other small changes as well that we might need to make

apps/pay/docker-compose.yml
UncleSamtoshi: Will remove this

vindard (approved):

PR #3626, opened 2023-11-29
chore: return batchInclusionEstimatedAt from sendOnChain

bodymindarts:


dolcalmi:
I think is better to include this after https://github.com/GaloyMoney/galoy/pull/3619

PR #3637, opened 2023-12-01
test: balance for check was missing onchain

sandipndev:
Adds exporter module to Tiltfile as well


bats/core/api/onchain-send.bats
```diff
@@ -4,6 +4,7 @@ load "../../helpers/cli.bash"
 load "../../helpers/user.bash"
 load "../../helpers/onchain.bash"
 load "../../helpers/wallet.bash"
+load "../../helpers/dealer.bash"
```
bodymindarts, line 7: dealer is not the right name for the file where `balance_for_check` comes from.
Perhaps ledger?

bodymindarts (approved):

PR #3641, opened 2023-12-01
feat: move gt3 library

dolcalmi:
- Move https://github.com/GaloyMoney/gt3-server-node-express-bypass 
- Refactor to only include required sdk files
- Update core to use local reference


nicolasburtey (dismissed):

nicolasburtey:
(tests are not passing)

toolchains/workspace-pnpm/macros.bzl
```diff
@@ -176,6 +180,10 @@ build_node_modules = rule(
             default = False,
             doc = "Only install production dependencies"
         ),
+        "local_packages": attrs.string(
+            default = "lib",
```
vindard, line 184: @bodymindarts to accommodate locally vendored npm packages I changed the `build_node_modules.py` code to also copy those vendored files during the turbo install process.

I know we had originally decided that we want to use the top-level `lib` folder for common configs across sub-projects. Is this where we also want to store vendored npm packages, or is there another folder that we'd prefer?

nicolasburtey (approved):

toolchains/workspace-pnpm/macros.bzl
dolcalmi: if `lib` it is a folder for `common configs` then is not a good name, also, `tracing-rs` does not look like a common config but a library

bodymindarts: `lib` is a folder for shared libraries....

Some shared libraries may contain common config (that is imported as a library).

vindard: `si` uses their `lib` folder for all shared libraries so I think you're right, this is the right approach here
https://github.com/systeminit/si/tree/main/lib

vindard:
Toolchain specific changes: https://github.com/GaloyMoney/galoy/compare/6c14818...feat-move-gt3-library

vindard (approved):

PR #3651, opened 2023-12-03
chore(core)!: update totp mutation to start kratos session

siddhart1o1:
TODO
- [x]  handle case if phone is not added.
- [x]  make changes in test

Notes; 
1. Had to use the admin API for delete because we need an elevated session (TOTP verified) token, but the temporary token we are creating is not TOTP elevated.

 2. Users have to log in again after enabling TOTP, as we are creating a temporary token and using that to validate/add TOTP. So, after validation, the temporary token becomes TOTP elevated (TOTP verified), but the original one, i.e., the one the user is using, is not TOTP elevated, so they need to log in again or verify the token via TOTP (TOTP elevate)
 
3. To solve point (2) made token optional in validate mutation.

### DIFF
--- a/core/api/dev/apollo-federation/supergraph.graphql
+++ b/core/api/dev/apollo-federation/supergraph.graphql
@@ -1146,7 +1146,7 @@ type Mutation
   userPhoneRegistrationInitiate(input: UserPhoneRegistrationInitiateInput!): SuccessPayload! @join__field(graph: PUBLIC)
   userPhoneRegistrationValidate(input: UserPhoneRegistrationValidateInput!): UserPhoneRegistrationValidatePayload! @join__field(graph: PUBLIC)
   userTotpDelete(input: UserTotpDeleteInput!): UserTotpDeletePayload! @join__field(graph: PUBLIC)
-  userTotpRegistrationInitiate(input: UserTotpRegistrationInitiateInput!): UserTotpRegistrationInitiatePayload! @join__field(graph: PUBLIC)
+  userTotpRegistrationInitiate: UserTotpRegistrationInitiatePayload! @join__field(graph: PUBLIC)
   userTotpRegistrationValidate(input: UserTotpRegistrationValidateInput!): UserTotpRegistrationValidatePayload! @join__field(graph: PUBLIC)
   userUpdateLanguage(input: UserUpdateLanguageInput!): UserUpdateLanguagePayload! @join__field(graph: PUBLIC)
   userUpdateUsername(input: UserUpdateUsernameInput!): UserUpdateUsernamePayload! @join__field(graph: PUBLIC) @deprecated(reason: "Username will be moved to @Handle in Accounts. Also SetUsername naming should be used instead of UpdateUsername to reflect the idempotency of Handles")
@@ -1992,12 +1992,6 @@ type UserTotpDeletePayload
   me: User
 }
 
-input UserTotpRegistrationInitiateInput
-  @join__type(graph: PUBLIC)
-{
-  authToken: AuthToken!
-}
-
 type UserTotpRegistrationInitiatePayload
   @join__type(graph: PUBLIC)
 {
@@ -2009,7 +2003,6 @@ type UserTotpRegistrationInitiatePayload
 input UserTotpRegistrationValidateInput
   @join__type(graph: PUBLIC)
 {
-  authToken: AuthToken!
   totpCode: TotpCode!
   totpRegistrationId: TotpRegistrationId!
 }
--- a/core/api/dev/apollo-federation/supergraph.graphql
+++ b/core/api/dev/apollo-federation/supergraph.graphql
@@ -1145,7 +1145,7 @@ type Mutation
   userPhoneDelete: UserPhoneDeletePayload! @join__field(graph: PUBLIC)
   userPhoneRegistrationInitiate(input: UserPhoneRegistrationInitiateInput!): SuccessPayload! @join__field(graph: PUBLIC)
   userPhoneRegistrationValidate(input: UserPhoneRegistrationValidateInput!): UserPhoneRegistrationValidatePayload! @join__field(graph: PUBLIC)
-  userTotpDelete(input: UserTotpDeleteInput!): UserTotpDeletePayload! @join__field(graph: PUBLIC)
+  userTotpDelete: UserTotpDeletePayload! @join__field(graph: PUBLIC)
   userTotpRegistrationInitiate: UserTotpRegistrationInitiatePayload! @join__field(graph: PUBLIC)
   userTotpRegistrationValidate(input: UserTotpRegistrationValidateInput!): UserTotpRegistrationValidatePayload! @join__field(graph: PUBLIC)
   userUpdateLanguage(input: UserUpdateLanguageInput!): UserUpdateLanguagePayload! @join__field(graph: PUBLIC)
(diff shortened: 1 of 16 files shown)


core/api/dev/apollo-federation/supergraph.graphql
```diff
@@ -1143,7 +1143,7 @@ type Mutation
   userPhoneRegistrationInitiate(input: UserPhoneRegistrationInitiateInput!): SuccessPayload! @join__field(graph: PUBLIC)
   userPhoneRegistrationValidate(input: UserPhoneRegistrationValidateInput!): UserPhoneRegistrationValidatePayload! @join__field(graph: PUBLIC)
   userTotpDelete(input: UserTotpDeleteInput!): UserTotpDeletePayload! @join__field(graph: PUBLIC)
-  userTotpRegistrationInitiate(input: UserTotpRegistrationInitiateInput!): UserTotpRegistrationInitiatePayload! @join__field(graph: PUBLIC)
+  userTotpRegistrationInitiate: UserTotpRegistrationInitiatePayload! @join__field(graph: PUBLIC)
```
nicolasburtey, line 1146: this is a breaking change.

I think that is ok because we have not deployed this publicly in the mobile app yet.

core/api/src/services/kratos/totp.ts
```diff
@@ -128,34 +128,15 @@ const refreshToken = async (authToken: AuthToken): Promise<void | KratosError> =
...
+export const kratosRemoveTotp = async (userId: UserId) => {
   try {
-    await kratosPublic.updateSettingsFlow({
-      flow,
-      updateSettingsFlowBody: {
-        method: "totp",
-        totp_unlink: true,
-      },
-      xSessionToken: token,
+    const removeTotpResponse = await kratosAdmin.deleteIdentityCredentials({
```
siddhart1o1, line 133: Had to use the admin API here because we need an elevated session (TOTP verified) token, but the temporary token we are creating is not TOTP elevated.

nicolasburtey:
> Users have to log in again after enabling TOTP, as we are creating a temporary token and using that to validate/add TOTP. So, after validation, the temporary token becomes TOTP elevated (TOTP verified), but the original one, i.e., the one the user is using, is not TOTP elevated, so they need to log in again or verify the token via TOTP (TOTP elevate)

that kind of make sense, but I also think that it is not really a good UX for people adding the totp through the mobile app. 

we may need to work with both use case, either by providing 2 endpoints, where one is the use case where we have a kratos auth token, and one where we don't, or authToken should be optional, and we use it if it's given, or create it if it's not provided.

siddhart1o1:
> > Users have to log in again after enabling TOTP, as we are creating a temporary token and using that to validate/add TOTP. So, after validation, the temporary token becomes TOTP elevated (TOTP verified), but the original one, i.e., the one the user is using, is not TOTP elevated, so they need to log in again or verify the token via TOTP (TOTP elevate)
> 
> that kind of make sense, but I also think that it is not really a good UX for people adding the totp through the mobile app.
> 
> we may need to work with both use case, either by providing 2 endpoints, where one is the use case where we have a kratos auth token, and one where we don't, or authToken should be optional, and we use it if it's given, or create it if it's not provided.

Made authToken optional in the validate mutation, as that is the step where the token is elevated by Kratos after validation. Removed authToken from delete and initiate mutation.

core/api/src/graphql/public/root/mutation/user-totp-registration-validate.ts
```diff
@@ -27,9 +26,9 @@ const UserTotpRegistrationValidateMutation = GT.Field<
   GraphQLPublicContextAuth,
   {
     input: {
-      authToken: AuthToken | InputValidationError
       totpCode: TotpCode | InputValidationError
       totpRegistrationId: TotpRegistrationId | InputValidationError
+      authToken: AuthToken | null
```
nicolasburtey, line 31: `AuthToken` can also be an error, we should check for that

core/api/src/app/authentication/totp.ts
```diff
@@ -23,17 +31,70 @@ export const validateTotpRegistration = async ({
...
+  }
+
+  return validateTotpRegistrationWithoutToken({
+    totpCode,
+    totpRegistrationId,
+    userId,
+  })
+}
+
+export const validateTotpRegistrationWithoutToken = async ({
```
nicolasburtey, line 55:
```suggestion
const validateTotpRegistrationWithoutToken = async ({
```

this doesn't need to be exported

core/api/src/app/authentication/totp.ts
```diff
@@ -23,17 +31,70 @@ export const validateTotpRegistration = async ({
...
   if (res instanceof Error) return res
+  if (res.kratosUserId !== userId) return new AuthTokenUserIdMismatchError()
+
+  const me = await UsersRepository().findById(res.kratosUserId)
+  if (me instanceof Error) return me
+
+  return me
+}
 
+export const validateTotpRegistrationWithToken = async ({
```
nicolasburtey, line 82: this doesn't need to be exported

core/api/src/app/authentication/login.ts
```diff
@@ -400,3 +401,36 @@ const checkDeviceLoginAttemptPerAppcheckJtiLimits = async (
...
+  }
+
+  if (kratosResult instanceof Error) {
+    return kratosResult
+  }
+
+  authToken = kratosResult?.authToken
+
+  if (!authToken) {
+    return new IdentifierNotFoundError()
```
nicolasburtey, line 432: doesn't sound like the proper error here

core/api/src/app/authentication/login.ts
```diff
@@ -400,3 +401,36 @@ const checkDeviceLoginAttemptPerAppcheckJtiLimits = async (
     rateLimitConfig: RateLimitConfig.deviceAccountCreate,
     keyToConsume: appcheckJti,
   })
+
+export const getAuthTokenFromUserId = async (
+  userId: UserId,
+): Promise<AuthToken | AuthenticationError> => {
+  const { data } = await kratosAdmin.getIdentity({ id: userId })
+  let authToken = null
+  let kratosResult = null
```
nicolasburtey, line 410: type is missing. 

@vindard would be interesting to know why Typescript doesn't complain

core/api/src/app/authentication/login.ts
```diff
@@ -400,3 +401,36 @@ const checkDeviceLoginAttemptPerAppcheckJtiLimits = async (
     rateLimitConfig: RateLimitConfig.deviceAccountCreate,
     keyToConsume: appcheckJti,
   })
+
+export const getAuthTokenFromUserId = async (
```
nicolasburtey, line 405: I don't think there is an external use case for it?

this method should probably be in the kratos service. @bodymindarts can confirm

nicolasburtey:
the code logics looks good overall. have comments on different improvements we should do or think about

core/api/src/app/authentication/login.ts
```diff
@@ -57,6 +58,7 @@ import { ErrorLevel } from "@/domain/shared"
 import { consumeLimiter } from "@/services/rate-limit"
 import { RateLimitConfig } from "@/domain/rate-limit"
 import { RateLimiterExceededError } from "@/domain/rate-limit/errors"
+import { kratosAdmin } from "@/services/kratos/private"
```
bodymindarts, line 61: The file you are importing is flagged as `private`. That means you are not allowed to depend on it! That's what private means.
Only code inside `@/services/kratos/` should be allowed access.. unfortunately TS doesn't support module private files / functions so we have to enforce it manually.

core/api/src/app/authentication/login.ts
bodymindarts: Seeing that this is using `kratosAdmin` client directly, which should (correctly) be private to the service boundary I would agree that this needs to be refactored. We are leaking boundary / adapter concerns into the application layer. This code belongs on the service.

PR #3657, opened 2023-12-05
test: e2e, add new `public-ln-receive.bats`

Ayush170-Future:
This PR adds the new e2e test `public-ln-receive` in `bats/core/api` and deletes the old one from `core/api`.

I added a new function in the helpers and made some adjustments in the test file to align with the new helper.


Ayush170-Future:
The CI failure seems unrelated. Any suggestions to fix this? Or are these fine for test files?

vindard:
> The CI failure seems unrelated. Any suggestions to fix this? Or are these fine for test files?

That is strange. I'll look into them and let you know on here

vindard:
> The CI failure seems unrelated. Any suggestions to fix this? Or are these fine for test files?

Ok we'll fix the `quickstart` failure in #3661, but the new bats tests themselves are genuinely broken still too. Are you able to execute these tests locally from `bats/core/api/public-ln-receive.bats` and confirm they work?

Ayush170-Future:
Yeah, just give me some time to confirm it. If it doesn't pass, I'll push the fix as soon as possible. Thanks for your quick responses!

Ayush170-Future:
I think I understand the problem here. Will force-push the fix soon!

Ayush170-Future:
- Rebased to the main
- Updated the `ln.bash` with necessary functions.

@vindard I suspect there might be some problem with the query formatting in [`exec_graphql`](https://github.com/GaloyMoney/galoy/blob/main/bats/helpers/_common.bash#L113).

vindard:
> * Rebased to the main
> 
>     * Updated the `ln.bash` with necessary functions.
> 
> 
> @vindard I suspect there might be some problem with the query formatting in [`exec_graphql`](https://github.com/GaloyMoney/galoy/blob/main/bats/helpers/_common.bash#L113).

Ok as far as I can tell you won't even get past the `lnds_init` step in `setup_file` because we deliberately don't have an lnd2 container setup in Tiltfile so far.

I would recommend taking this PR a bit more slowly. The way I would do it is to just start with a single `@test` case and to then only bring in the exact functions you'd need in `ln.bash`. Get that to pass, commit, and then continue to subsequent test cases.

Ayush170-Future:
Thanks, @vindard, for the advice! I'm taking it one test at a time now. Will follow your LN advice as soon as I reach those tests.

I've added the first test in the recent force push, but our old problem still persists. It shows that the GraphQL query is invalid.

vindard:
Taken over by #3765

PR #3669, opened 2023-12-07
ci: build gh-token in flake.nix

krtk6160:


bodymindarts:
Can we add this build definition to a flake in concourse-shared and then reference that from here?

krtk6160:
> Can we add this build definition to a flake in concourse-shared and then reference that from here?

Done in [ae3b3d9](https://github.com/GaloyMoney/galoy/pull/3669/commits/ae3b3d9e10c215c655d37d285c318418bbd414d0).

bodymindarts (approved):

PR #3676, opened 2023-12-07
chore: adding shared component lib

siddhart1o1:


apps/consent/package.json
```diff
@@ -32,13 +32,14 @@
     "edge-csrf": "^1.0.6",
     "graphql": "^16.8.1",
     "libphonenumber-js": "^1.10.51",
-    "next": "14.0.4",
+    "next": "14.0.1",
```
bodymindarts, line 35: why downgrade?

apps/dashboard/BUCK
```diff
@@ -65,6 +65,7 @@ next_build_bin(
 
 dev_deps_srcs = {
     "lib/eslint-config": "//lib/eslint-config:src",
+    "lib/galoy-components": "//lib/galoy-components:src",
```
bodymindarts, line 68: why in dev_deps?

toolchains/workspace-pnpm/macros.bzl
```diff
@@ -494,6 +494,12 @@ _next_build = rule(
             default = "toolchains//:workspace_pnpm",
             providers = [WorkspacePnpmToolchainInfo],
         ),
+        "prod_deps_srcs": attrs.dict(
```
bodymindarts, line 497: Adding this isn't going to have the desired effect. Also need to handle the argument in the implementation cc @vindard

apps/consent/package.json
siddhart1o1: Cypress tests were failing with this version; they were flaky. I will look into this, but I think that would be a topic for another PR.

toolchains/workspace-pnpm/macros.bzl
siddhart1o1: > Adding this isn't going to have the desired effect. Also need to handle the argument in the implementation cc @vindard

https://github.com/GaloyMoney/galoy/blob/3c150ca73d2a8a81277d58d50ef5e2e128272576/toolchains/workspace-pnpm/macros.bzl#L568
I think `prod_deps_srcs`  handles this, but yes, @vindard may have more context.

apps/dashboard/BUCK
siddhart1o1: yes this should be in prod_deps.

bodymindarts (approved):

PR #3678, opened 2023-12-08
chore: update assetlinks.json for google play verification of domain

sandipndev:


dolcalmi (changes requested):
this is not part of galoy pay

PR #3681, opened 2023-12-08
feat(api): add lnurl payment send mutation

UncleSamtoshi:
Open questions:
1. Should the mutation for paying to an lnurl and lnAddress have separate endpoints?
  - Currently they are different endpoints, but they both have the same behavior. I have no strong preference here.
2. How should we handle if the amount is invalid with the given lnurl?
  - Currently we are just returning an error. I don't think it makes sense to return the lnurl params, because if the end user is willing to parse that from our api, they might as well just hit the lnurl directly.
3. Should we allow comments? How should we handle if the lnurl doesn't support comments?
  - Currently we do not support adding comments. This leads to the highest success rate, but I suppose that we might as well give the user the option to use comments and warn them that it might not be supported.
4. @bodymindarts what is the best way to get a `bech32 encoder cli` in the bats env? It is needed for generating an lnurl string to pay to. I found this tool here that we could use but I'm not sure how to add it to our env https://github.com/cmoog/bech32.


dev/Tiltfile
```diff
@@ -236,7 +236,7 @@ core_serve_env = {
   "LND1_TYPE": "offchain,onchain",
   "SVIX_SECRET": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpYXQiOjE2OTE2NzIwMTQsImV4cCI6MjAwNzAzMjAxNCwibmJmIjoxNjkxNjcyMDE0LCJpc3MiOiJzdml4LXNlcnZlciIsInN1YiI6Im9yZ18yM3JiOFlkR3FNVDBxSXpwZ0d3ZFhmSGlyTXUifQ.b9s0aWSisNdUNki4edabBEToLNSwjC9-AiJQr4J3y4E",
   "SVIX_ENDPOINT": "http://localhost:8071",
-  "EXPORTER_PORT": "3002"
+  "EXPORTER_PORT": "3003"
```
dolcalmi, line 239: why change exporter if the addition is galoy pay?

core/api/test/integration/services/lnurl-pay.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  describe("fetchInvoiceFromLnAddressOrLnurl", () => {
+    it("fetches an invoice from an lnurl", async () => {
+      const invoice = await lnurlPayService.fetchInvoiceFromLnAddressOrLnurl({
+        amount: {
+          amount: BigInt(1000),
+          currency: "BTC",
+        } as BtcPaymentAmount,
+        lnAddressOrLnurl:
+          "lnurl1dp68gurn8ghj7urp0yhxymrfde4juumk9uh8wetvdskkkmn0wahz7mrww4excup0w3jhxaqlq07tq", // https://pay.blink.sv/.well-known/lnurlp/test
```
dolcalmi, line 16: please dont implement tests that depends on external resoures

core/api/src/services/lnurl-pay/index.ts
```diff
@@ -0,0 +1,38 @@
...
+          "Lnurl service returned an invoice with an invalid amount",
+        )
+      }
+
+      return invoice.invoice
+    } catch (err) {
+      if (err instanceof Error) {
+        return new ErrorFetchingLnurlInvoice(err.message)
+      }
+      return new ErrorFetchingLnurlInvoice("Unknown error fetching LnUrl invoice")
```
dolcalmi, line 31: Please follow unknown error rule.. this will not trigger an alert

core/api/src/graphql/shared/types/scalar/lnurl.ts
dolcalmi, line 1: if you dont implement validations then just use GT.String

core/api/src/graphql/shared/types/scalar/ln-address.ts
dolcalmi, line 1: same here

core/api/src/domain/bitcoin/lnurl/index.types.d.ts
dolcalmi, line 1: are we sure this is part of bitcoin domain?

core/api/src/domain/bitcoin/lnurl/errors.ts
```diff
@@ -0,0 +1,5 @@
+import { DomainError } from "@/domain/shared"
+
+export class LnurlError extends DomainError {}
+
+export class ErrorFetchingLnurlInvoice extends LnurlError {}
```
dolcalmi, line 5: please implement proper error structure (including level critical for unknown)

core/api/src/domain/bitcoin/lnurl/index.types.d.ts
```diff
@@ -0,0 +1,6 @@
+interface ILnurlPayService {
+  fetchInvoiceFromLnAddressOrLnurl(args: {
+    amount: BtcPaymentAmount
+    lnAddressOrLnurl: string
+  }): Promise<string | ApplicationError>
```
dolcalmi, line 5: application error is not part of domain, please use the base service error like in other implementations

bats/helpers/user.bash
```diff
@@ -95,8 +95,7 @@ user_update_username() {
   )
   exec_graphql "$token_name" 'user-update-username' "$variables"
   num_errors="$(graphql_output '.data.userUpdateUsername.errors | length')"
-  username="$(graphql_output '.data.userUpdateUsername.user.username')"
-  [[ "$num_errors" == "0" || "$username" == "$token_name" ]] || exit 1
```
dolcalmi, line 99: why this change? does not seems to be related to the PR

bats/helpers/ledger.bash
```diff
@@ -2,7 +2,7 @@ CURRENT_FILE=${BASH_SOURCE:-bats/helpers/.}
 source "$(dirname "$CURRENT_FILE")/_common.bash"
 source "$(dirname "$CURRENT_FILE")/cli.bash"
 
-METRICS_ENDPOINT="localhost:3002/metrics"
+METRICS_ENDPOINT="localhost:3003/metrics"
```
dolcalmi, line 5: again this.. why modify exporter endpoint if the new addition in galoy pay?

bats/core/api/lnurl-send.bats
```diff
@@ -0,0 +1,34 @@
...
+  variables=$(
+    jq -n \
+    --arg wallet_id "$(read_value $btc_wallet_name)" \
+    --arg amount 200 \
+    --arg lnurl $lnurl \
+    '{input: {walletId: $wallet_id, amount: $amount, lnurl: $lnurl }}'
+  )
+  exec_graphql 'bob' 'lnurl-payment-send' "$variables"
+  lnurl_payment_send_status="$(graphql_output '.data.lnurlPaymentSend.status')"
+  [[ "${lnurl_payment_send_status}" == "SUCCESS" ]] || exit 1
```
dolcalmi, line 33: how is this working? if lnurl is `invalidlnurl` ?

dolcalmi (changes requested):

bats/helpers/user.bash
UncleSamtoshi: This function was broken and was caching a null value. It is required that we cache the correct username for knowing an ln address we can send to.

bats/helpers/ledger.bash
UncleSamtoshi: See above

dev/Tiltfile
UncleSamtoshi: Exporter was added after galoy-pay and was added incorrectly on the same port as galoy-pay.

core/api/test/integration/services/lnurl-pay.spec.ts
UncleSamtoshi: This service inherently depends on external resources. How would you propose testing it otherwise? I suppose we could try to mock the response of the http requests it will make. Otherwise bringing up the entire galoy pay and galoy api locally seems like it would introduce a lot of overhead and I dont believe the library's validation supports an ln address of the form `name@localhost:3002` where a port is included.

core/api/src/domain/bitcoin/lnurl/errors.ts
```diff
@@ -0,0 +1,7 @@
+import { DomainError, ErrorLevel } from "@/domain/shared"
+
+export class LnurlError extends DomainError {}
```
dolcalmi, line 3: the suffix should be `LnurlServiceError` to match other services

core/api/src/domain/bitcoin/lnurl/errors.ts
```diff
@@ -0,0 +1,7 @@
...
+export class ErrorFetchingLnurlInvoice extends LnurlError {}
+export class UnknownLnurlError extends LnurlError {
```
dolcalmi, line 5: here too

core/api/src/services/lnurl-pay/index.ts
```diff
@@ -0,0 +1,42 @@
...
+      return invoice.invoice
+    } catch (err) {
+      if (err instanceof Error) {
+        return new ErrorFetchingLnurlInvoice(err.message)
+      }
+      return new UnknownLnurlError(err)
+    }
+  }
+
+  return {
```
dolcalmi, line 39: please add the tracing wrapper

dolcalmi (changes requested):
looks good, just minor changes

core/api/src/domain/bitcoin/lnurl/index.types.d.ts
UncleSamtoshi: Where else do you think it should live?

dolcalmi (dismissed):

dolcalmi (approved):

PR #3692, opened 2023-12-11
fix(pay): use req url for ln address

UncleSamtoshi:
- use the url of the req to populate ln address

### DIFF
--- a/apps/pay/app/lnurlp/[username]/callback/route.ts
+++ b/apps/pay/app/lnurlp/[username]/callback/route.ts
@@ -16,6 +16,7 @@ import {
   LnInvoiceCreateOnBehalfOfRecipientMutation,
 } from "../../../../lib/graphql/generated"
 import { client } from "../graphql"
+import { getOriginalRequestInfo } from "../../../../utils/utils"
 
 gql`
   mutation lnInvoiceCreateOnBehalfOfRecipient(
@@ -114,9 +115,11 @@ export async function GET(
     })
   }
 
+  const originalUrlInfo = getOriginalRequestInfo(request)
+
   const metadata = JSON.stringify([
     ["text/plain", `Payment to ${username}`],
-    ["text/identifier", `${username}@${env.PAY_DOMAIN}`],
+    ["text/identifier", `${username}@${originalUrlInfo.hostname}`],
   ])
 
   // lnurl generate invoice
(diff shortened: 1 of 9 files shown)


apps/pay/app/lnurlp/[username]/callback/route.ts
```diff
@@ -8,6 +8,8 @@ import { gql } from "@apollo/client"
 
 import Redis from "ioredis"
 
+import originalUrl from "original-url"
```
Comment (dolcalmi) [inline apps/pay/app/lnurlp/[username]/callback/route.ts:11]: is this required? seems is not maintained (last update 4 years ago)

apps/pay/app/lnurlp/[username]/callback/route.ts
```diff
@@ -114,9 +116,13 @@ export async function GET(
     })
   }
 
+  // eslint-disable-next-line
+  // @ts-ignore
```
Comment (dolcalmi) [inline apps/pay/app/lnurlp/[username]/callback/route.ts:120]: why is this required? please use declare or try to find the types

dolcalmi (changes requested):

apps/pay/app/lnurlp/[username]/callback/route.ts
```diff
@@ -114,9 +116,13 @@ export async function GET(
...
+  const url = originalUrl(request)
```
Comment (UncleSamtoshi) [inline apps/pay/app/lnurlp/[username]/callback/route.ts:121]: We already use this function `originalUrl` elsewhere in the project, however the types do not exactly match up. It is used to get the original url in the case that the request was proxied. A follow up would be to migrate the logic from originalUrl to this repo so that the types can be correct.

apps/pay/app/lnurlp/[username]/callback/route.ts
Comment (UncleSamtoshi) [reply apps/pay/app/lnurlp/[username]/callback/route.ts:11]: I believe it is necessary in the case that the request is proxied. I will try to bring the logic within the app as I think its pretty simple.

apps/pay/app/lnurlp/[username]/callback/route.ts
Comment (UncleSamtoshi) [reply apps/pay/app/lnurlp/[username]/callback/route.ts:121]: I will actually just migrate the logic now.

apps/pay/app/lnurlp/[username]/callback/route.ts
Comment (nicolasburtey) [reply apps/pay/app/lnurlp/[username]/callback/route.ts:11]: > is this required? seems is not maintained (last update 4 years ago)

I think I actually removed this library because of that.

Comment (UncleSamtoshi) [reply apps/pay/app/lnurlp/[username]/callback/route.ts]: It was still present in the repo, but I've now removed it.

apps/pay/app/lnurlp/[username]/callback/route.ts
Comment (UncleSamtoshi) [reply apps/pay/app/lnurlp/[username]/callback/route.ts]: No longer required

PR #3697, opened 2023-12-12
feat: add default wallet field

vindard:
## Description

This is to support being able to get details like an account's default wallet's currency without having to do multiple round trips

### DIFF
--- a/core/api/src/graphql/public/types/object/business-account.ts
+++ b/core/api/src/graphql/public/types/object/business-account.ts
@@ -13,6 +13,8 @@ import Transaction, {
 import RealtimePrice from "./realtime-price"
 import { NotificationSettings } from "./notification-settings"
 
+import PublicWallet from "./public-wallet"
+
 import { connectionArgs } from "@/graphql/connections"
 import { GT } from "@/graphql/index"
 import { mapError } from "@/graphql/error-map"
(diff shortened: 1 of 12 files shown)


core/api/src/graphql/public/types/object/business-account.ts
```diff
@@ -47,6 +49,15 @@ const BusinessAccount = GT.Object({
         domainAccount.defaultWalletId,
     },
 
+    defaultWallet: {
+      type: GT.NonNull(PublicWallet),
+      resolve: (source) =>
+        Wallets.getWalletForAccountById({
```
dolcalmi, line 55: errors here need to be thrown?

core/api/src/graphql/public/types/object/consumer-account.ts
```diff
@@ -83,6 +85,15 @@ const ConsumerAccount = GT.Object<Account, GraphQLPublicContextAuth>({
       resolve: (source) => source.defaultWalletId,
     },
 
+    defaultWallet: {
+      type: GT.NonNull(PublicWallet),
+      resolve: (source) =>
+        Wallets.getWalletForAccountById({
```
dolcalmi, line 91: same here

dolcalmi (changes requested):
looks good, just minor change... also I think is better to have another review because it is an API change

nicolasburtey:
great idea. it would actually remove many lines of code in the mobile repo!

core/api/src/graphql/public/types/object/business-account.ts
vindard: Done https://github.com/GaloyMoney/galoy/pull/3697/commits/6eea1818b188f77c802ca35c8ec22609a2f14645

core/api/src/graphql/public/types/object/consumer-account.ts
vindard: Done https://github.com/GaloyMoney/galoy/pull/3697/commits/6eea1818b188f77c802ca35c8ec22609a2f14645

dolcalmi (approved):

PR #3700, opened 2023-12-12
chore: remove alert on bad code for totp

nicolasburtey:


core/api/src/services/kratos/totp.ts
```diff
@@ -51,6 +50,16 @@ export const kratosValidateTotp = async ({
       xSessionToken: authToken,
     })
   } catch (err) {
+    if (isAxiosError(err)) {
+      if (err.response?.status === 400 && err.response.statusText === "Bad Request") {
+        console.log(err.response.data, "err.response.data")
+        console.log(err.response.status, "err.response.status")
+        console.log(err.response.statusText, "err.response.statusText")
```
dolcalmi, line 57: please remove or use baseLogger

core/api/src/graphql/error-map.ts
```diff
@@ -731,7 +737,7 @@ export const mapError = (error: ApplicationError): CustomGraphQLError => {
         error.message ? ": " + error.message : ""
       })`
       return new UnknownClientError({ message, logger: baseLogger })
-
+      123
```
dolcalmi, line 740: ??

core/api/src/app/authentication/totp.ts
```diff
@@ -90,14 +91,17 @@ const validateTotpRegistrationWithToken = async ({
...
+  const res = await refreshToken(authToken)
+  if (res instanceof Error) return res
+
   const validation = await kratosValidateTotp({ authToken, totpCode, totpRegistrationId })
   if (validation instanceof Error) return validation
 
-  const res = await validateKratosToken(authToken)
-  if (res instanceof Error) return res
-  if (res.kratosUserId !== userId) return new AuthTokenUserIdMismatchError()
+  const res2 = await validateKratosToken(authToken)
```
dolcalmi, line 100: not sure to understand this... first refresh authToken and then validate?

dolcalmi (changes requested):

core/api/src/graphql/error-map.ts
nicolasburtey: if I had a cat I would blame him to walk on the keyboard. but I don't have a cat ^^

core/api/src/app/authentication/totp.ts
nicolasburtey: yes, there is notion of freshness for the login in kratos. certain critical action needs to require to have a recent login, or otherwise kratos will ask you to login again to make sure this is really you. 

the type of actions that requires this are things like changing password or adding TOTP. 

currently the way we have done our login it's not very straightforward to align with what kratos expect because of how we proxy the call to kratos, so we are just here forcing the refresh on the backend.

note: this was already done previously, I just refactor the code to have the refreshToken only if it was just not create prior.

dolcalmi (approved):

PR #3716, opened 2023-12-15
fix(api-keys): connect with api key to WS

bodymindarts:


core/api/src/servers/ws-server.ts
```diff
@@ -69,12 +69,13 @@ const getContext = async (
 
       const ip = parseIps(ipString)
 
+      const apiKey = connectionParams?.["X-API-KEY"] as AuthToken | undefined
       const authz = connectionParams?.Authorization as string | undefined
-      const kratosToken = authz?.slice(7) as AuthToken
+      const bearerToken = authz?.slice(7) as AuthToken
```
nicolasburtey, line 74: maybe for another PR but we should probably not use the same type for different type of token. probably apiKey should be `ApiToken` type

core/api/src/services/oathkeeper/index.ts
```diff
@@ -9,18 +9,24 @@ import {
...
 
-export const sendOathkeeperRequestGraphql = async (
-  token: AuthToken | undefined,
-): Promise<JwtToken | OathkeeperError> => {
+export const sendOathkeeperRequestGraphql = async ({
+  bearerToken,
+  apiKey,
+}: {
+  bearerToken: AuthToken | undefined
+  apiKey: AuthToken | undefined
```
nicolasburtey, line 17: should we return an error if both are set?

bats/helpers/subscriber/src/gql-subscribe.ts
```diff
@@ -15,7 +15,9 @@ const authToken = process.argv[5]
 const variablesString = process.argv[6] || "{}"
 const client = createClient({
   url,
-  connectionParams: authToken ? { Authorization: `Bearer ${authToken}` } : undefined,
+  connectionParams: authToken ? authToken.startsWith("dev_") ? {
+    "X-API-KEY": authToken,
+  } : { Authorization: `Bearer ${authToken}` } : undefined,
```
nicolasburtey, line 20: the double ternary operator make it hard to read and reason about

nicolasburtey (commented):
few comments

nicolasburtey (approved):

PR #3719, opened 2023-12-16
test: move ln tests to top level bats

vindard:
_Takes over from #3593_


.github/workflows/bats.yml
```diff
@@ -21,7 +21,7 @@ jobs:
             //core/api-keys //apps/dashboard //apps/consent //apps/pay
       - name: Run bats tests
         run: |
-          nix develop -c bats --setup-suite-file bats/ci_setup_suite.bash -t bats/core/**
+          . ./.env && nix develop -c bats --setup-suite-file bats/ci_setup_suite.bash -t bats/core/**
```
bodymindarts, line 24: Need to find another way to inject env.

bodymindarts (approved):

PR #3727, opened 2023-12-18
chore: some limits renaming and adjustments

nicolasburtey:


core/api/src/app/authentication/login.ts
```diff
@@ -85,7 +85,6 @@ export const loginWithPhoneToken = async ({
   if (validCode instanceof Error) return validCode
 
   await rewardFailedLoginAttemptPerIpLimits(ip)
-  await rewardFailedLoginAttemptPerLoginIdentifierLimits(phone)
```
nicolasburtey, line 88: we stop rewarding login, such that more than X login in a timeframe will lead to login error

dolcalmi (approved):

PR #3731, opened 2023-12-19
ci: Test fix bats

UncleSamtoshi:


.github/workflows/bats.yml
```diff
@@ -17,8 +17,17 @@ jobs:
         uses: DeterminateSystems/magic-nix-cache-action@v2
       - name: Buck2 build
         run: |
-          nix develop -c buck2 build //core/api //core/api-ws-server \
-            //core/api-keys //apps/dashboard //apps/consent //apps/pay
+          nix develop -c buck2 build //core/api:api //core/api-ws-server:api-ws-server \
+            //core/api-keys:api-keys //apps/dashboard:dashboard //apps/consent:consent //apps/pay:pay
```
nicolasburtey, line 21: isn't this a no-op?

UncleSamtoshi: This was just a test, not moving forward with the pr.

PR #3732, opened 2023-12-19
chore!: refactoring quiz for it to be time based

nicolasburtey:
- removing earn and rewards as part of the naming, converging to quiz (using Quiz because this is what the API layer is using)
- moving the data out of the Account collection into its own collection
-- will make it easier to parse later once and have rules based on time
-- will make it easier to extract to an independent pod if we choose to in the future


core/api/test/unit/domain/users/phone-metadata-authorizer.spec.ts
```diff
@@ -154,14 +154,14 @@ describe("PhoneMetadataAuthorizer - validate", () => {
...
-      getPhoneMetadataRewardsSettings(),
+      getPhoneMetadataQuizzesSettings(),
     ).authorize(undefined)
     expect(validator).toBeInstanceOf(ExpectedPhoneMetadataMissingError)
   })
 
   it("returns error with voip type", () => {
     const validator = PhoneMetadataAuthorizer(
-      getPhoneMetadataRewardsSettings(),
+      getPhoneMetadataQuizzesSettings(),
```
dolcalmi, line 164: quizzes or quiz?

core/api/src/services/mongoose/quiz.ts
```diff
@@ -0,0 +1,38 @@
+import { Quiz } from "./schema"
+
+import { QuizAlreadyPresentError, UnknownRepositoryError } from "@/domain/errors"
+
+interface ExtendedError extends Error {
+  code?: number
+}
+
+export const QuizRepository = (accountId: AccountId) => {
```
dolcalmi, line 9: please move accountId to methods, also this requires an interface

core/api/src/services/mongoose/quiz.ts
```diff
@@ -0,0 +1,38 @@
...
+  const fetchAll = async () => {
+    try {
+      const result = await Quiz.find({ accountId })
+      return result
+    } catch (err) {
+      return new UnknownRepositoryError("quiz issue")
+    }
+  }
+
+  return {
```
dolcalmi, line 34: please add tracing wrapper (at least for add)

core/api/src/services/quiz/index.ts
dolcalmi, line 1: this is not a service, please move to app layer

core/api/src/services/mongoose/schema.ts
```diff
@@ -307,6 +303,26 @@ AccountSchema.index({
 
 export const Account = mongoose.model<AccountRecord>("Account", AccountSchema)
 
+const QuizSchema = new Schema<QuizRecord>({
+  accountId: {
```
dolcalmi, line 307: maybe because it is the first PR but having a key only based on accountId will allow the delete->claim rewards> create cycle

core/api/src/graphql/public/types/object/consumer-account.ts
```diff
@@ -167,7 +169,16 @@ const ConsumerAccount = GT.Object<Account, GraphQLPublicContextAuth>({
     quiz: {
       type: GT.NonNullList(Quiz),
       description: "List the quiz questions of the consumer account",
-      resolve: (source) => source.quiz,
+      resolve: async (source) => {
+        const accountId = source.id
+        const result = await getQuizzesByAccountId(accountId)
```
dolcalmi, line 174: please use a method from app layer

core/api/src/graphql/error.ts
```diff
@@ -381,9 +381,9 @@ export class InvalidPhoneForOnboardingError extends CustomGraphQLError {
 export class UnauthorizedIPMetadataCountryError extends CustomGraphQLError {
   constructor(errData: CustomGraphQLErrorData) {
     super({
-      message: "Country not not authorized for rewards.",
+      message: "Country not not authorized for quizzes.",
       forwardToClient: true,
-      code: "UNAUTHORIZED_COUNTRY_IP_FOR_REWARD",
+      code: "UNAUTHORIZED_COUNTRY_IP_FOR_QUIZZES",
```
dolcalmi, line 386: does this break mobile app error handling?

core/api/src/graphql/error-map.ts
```diff
@@ -286,22 +286,22 @@ export const mapError = (error: ApplicationError): CustomGraphQLError => {
...
-      message = "Reward for quiz question was already claimed."
+    case "QuizAlreadyPresentError":
+      message = "Quiz question was already claimed."
       return new ValidationInternalError({ message, logger: baseLogger })
 
-    case "NotEnoughBalanceForRewardError":
+    case "NotEnoughBalanceForQuizError":
       message =
-        "Rewards wallet temporarily depleted. Please contact support if problem persists."
+        "Quiz wallet temporarily depleted. Please contact support if problem persists."
```
dolcalmi, line 295: probably not a change for this PR but we should not return this info

core/api/src/domain/users/errors.ts
```diff
@@ -7,7 +7,7 @@ export class PhoneCarrierTypeNotAllowedError extends UnauthorizedPhoneError {}
 export class PhoneCountryNotAllowedError extends UnauthorizedPhoneError {}
 
 export class InvalidPhoneForOnboardingError extends UnauthorizedPhoneError {}
-export class InvalidPhoneForRewardError extends UnauthorizedPhoneError {}
+export class InvalidPhoneForQuizError extends UnauthorizedPhoneError {}
```
dolcalmi, line 10: please be consistent Quiz or Quizzes (I prefer quiz but it is up to you)

core/api/src/domain/rate-limit/index.ts
```diff
@@ -85,9 +85,9 @@ export const RateLimitConfig: { [key in RateLimitPrefixKey]: RateLimitConfig } =
     limits: getAppcheckJtiAttemptLimits(),
     error: UserCodeAttemptAppcheckJtiLimiterExceededError,
   },
-  addEarnAttemptPerIp: {
-    key: RateLimitPrefix.addEarnAttemptPerIp,
-    limits: getAddEarnPerIpLimits(),
-    error: UserAddEarnAttemptIpRateLimiterExceededError,
+  addQuizAttemptPerIp: {
```
dolcalmi, line 88: same here, quiz or quizzes?

core/api/src/config/schema.types.d.ts
```diff
@@ -36,7 +36,7 @@ type YamlSchema = {
     android: BuildNumberInput
     ios: BuildNumberInput
   }
-  rewards: {
+  quizzes: {
```
dolcalmi, line 39: this will break custom config in deployments (just to keep in mind before deploy)

core/api/src/config/index.ts
```diff
@@ -13,6 +13,7 @@ export * from "./schema"
 import { ConfigError } from "./error"
 
 import { toDays } from "@/domain/primitives"
+import { QuizzesValue } from "@/app/quiz/config"
```
dolcalmi, line 16: :thinking: shouldn't thits be part of config and not from app layer? it is creating a circular dependency

core/api/src/app/quiz/config.ts
dolcalmi, line 1: move to config or domain, this should not be part of app layer

dolcalmi (changes requested):

core/api/test/unit/domain/users/phone-metadata-authorizer.spec.ts
nicolasburtey: it depend of the context on whether it's singular or plural

core/api/src/services/mongoose/quiz.ts
nicolasburtey: so that `add` would take `{quizId, accountId}` input instead?

core/api/src/config/schema.types.d.ts
nicolasburtey: yup I know

core/api/src/config/index.ts
nicolasburtey: not sure, but I've also been thinking about that. I think this is the config of the quizzes, so we may argue this could be fine? I also try to make the quiz module for self contain.

core/api/src/services/mongoose/schema.ts
nicolasburtey: this is not changing the behavior compared to before? the goal of this PR is to create a time bound earn / section. other way to prevent yields for people re-creating account can be part of a different PR

core/api/src/services/mongoose/quiz.ts
dolcalmi: yes

core/api/src/graphql/error.ts
nicolasburtey: this string doesn't exist on mobile: https://github.com/search?q=repo%3AGaloyMoney%2Fgaloy-mobile%20UNAUTHORIZED_COUNTRY_IP_FOR_REWARD&type=code

nicolasburtey: same for the 2 below

core/api/src/services/mongoose/quiz.ts
nicolasburtey: isn't the tracing already provided by mongodb tracing module? we are not doing this for the other module in this mongoose folder

core/api/src/services/mongoose/quiz.ts
```diff
@@ -0,0 +1,44 @@
...
+      return new UnknownRepositoryError("quiz issue")
+    }
+  }
+
+  const fetchAll = async (accountId: AccountId) => {
+    try {
+      const result = await Quiz.find({ accountId })
+      return result
+    } catch (err) {
+      return new UnknownRepositoryError("quiz issue")
```
dolcalmi, line 36: interesting this is not causing a lint issue, err needs to be passed to UnknownRepositoryError

core/api/src/services/mongoose/quiz.ts
```diff
@@ -0,0 +1,44 @@
...
+      await Quiz.create({ accountId, quizId })
+
+      return true
+    } catch (err) {
+      if (err instanceof Error) {
+        const error = err as ExtendedError
+        if (error?.code === 11000) return new QuizAlreadyPresentError()
+      }
+
+      return new UnknownRepositoryError("quiz issue")
```
dolcalmi, line 27: same here

dolcalmi (changes requested):

dolcalmi (approved):

PR #3743, opened 2023-12-22
feat: introduce time based quizzes

nicolasburtey:


core/api/test/unit/domain/quiz/index.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+      { quizId: "sat" as QuizQuestionId, createdAt: new Date() },
+      { quizId: "whereBitcoinExist" as QuizQuestionId, createdAt: new Date() },
+      { quizId: "whoControlsBitcoin" as QuizQuestionId, createdAt: new Date() },
+      { quizId: "copyBitcoin" as QuizQuestionId, createdAt: new Date() },
+    ]
+    const filledInfo = fillQuizInformation(quizzesCompleted)
+    expect(filledInfo.currentSection).toBe(1)
+
+    console.log(filledInfo.quizzes[4])
+    console.log(filledInfo.quizzes[5])
```
dolcalmi, line 61: remove

core/api/test/unit/domain/quiz/index.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+      completed: true,
+      id: "sat",
+      section: 0,
+      notBefore: undefined,
+    })
+
+    expect(filledInfo.currentSection).toBe(0)
+  })
+
+  it("two elements completed", () => {
```
dolcalmi, line 31: what is the test?

core/api/test/unit/domain/quiz/index.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+    expect(result.find((quiz) => quiz.completed === true)).toBe(undefined)
+  })
+
+  it("default value when not started", () => {
+    const filledInfo = fillQuizInformation([])
+    expect(filledInfo.currentSection).toBe(0)
+    expect(filledInfo.quizzes[0].notBefore).toBe(undefined)
+  })
+
+  it("one element completed", () => {
```
dolcalmi, line 16: same here

core/api/test/integration/services/quiz.spec.ts
```diff
@@ -2,6 +2,7 @@ import crypto from "crypto"
 
 import { QuizAlreadyPresentError } from "@/domain/errors"
 import { QuizRepository } from "@/services/mongoose"
+import { QuizQuestionId } from "@/domain/quiz/index.types"
```
dolcalmi, line 5: this should not be needed

core/api/src/services/mongoose/quiz.ts
```diff
@@ -1,6 +1,9 @@
 import { Quiz } from "./schema"
 
+import { QuizCompleted } from "@/domain/quiz"
+
 import { QuizAlreadyPresentError, UnknownRepositoryError } from "@/domain/errors"
+import { QuizQuestionId } from "@/domain/quiz/index.types"
```
dolcalmi, line 6: should not be required

core/api/src/graphql/public/types/payload/quiz-completed.ts
```diff
@@ -3,6 +3,7 @@ import Quiz from "../object/quiz"
 
 import { GT } from "@/graphql/index"
 
+// deprecated
```
dolcalmi, line 6: if this is deprecated add `deprecationReason` to the query/mutation

core/api/src/graphql/public/types/payload/quiz-claim.ts
```diff
@@ -0,0 +1,18 @@
...
+import { GT } from "@/graphql/index"
+
+const QuizClaimPayload = GT.Object({
+  name: "QuizClaimPayload",
+  fields: () => ({
+    errors: {
+      type: GT.NonNullList(IError),
+    },
+    quizzes: {
+      type: GT.List(Quiz),
```
dolcalmi, line 13: shouldn't be this non null too and return [] if the user has not claims?

core/api/src/graphql/public/types/object/quiz.ts
```diff
@@ -11,6 +13,7 @@ const Quiz = GT.Object({
       description: "The reward in Satoshis for the quiz question",
     },
     completed: { type: GT.NonNull(GT.Boolean) },
+    notBefore: { type: Timestamp },
```
dolcalmi, line 16: is it necessary to return this?

core/api/src/graphql/public/types/object/quiz.ts
```diff
@@ -1,5 +1,7 @@
 import SatAmount from "../../../shared/types/scalar/sat-amount"
 
+import Timestamp from "@/graphql/shared/types/scalar/timestamp"
```
dolcalmi, line 3: use the same import format in all the imports `"@/graphql/`

core/api/src/graphql/public/types/object/consumer-account.ts
```diff
@@ -34,7 +34,7 @@ import DisplayCurrency from "@/graphql/shared/types/scalar/display-currency"
 
 import { listEndpoints } from "@/app/callback"
 import { IInvoiceConnection } from "@/graphql/shared/types/abstract/invoice"
-import { getQuizzesByAccountId } from "@/app/quiz"
+import { listQuizzesByAccountId } from "@/app/quiz"
```
dolcalmi, line 37: is this on purpose? if we use it in this way we dont use the tracing wrapper

core/api/src/domain/quiz/script.ts
dolcalmi, line 1: ???

core/api/src/domain/quiz/sections.ts
```diff
@@ -0,0 +1,212 @@
+export const QuizzesSectionsConfig = [
```
dolcalmi, line 1: instead of having 2 sources of truth we need to add amounts to this `config`

core/api/src/domain/quiz/index.ts
```diff
@@ -0,0 +1,78 @@
...
+  const quizzes = Object.entries(QuizzesValue).map(([id, amount]) => {
+    const quizCompleted = quizzesCompleted.find((quiz) => quiz.quizId === id)
+    const section =
+      QuizzesSections.find((section) => section.quiz.includes(id as QuizQuestionId))
+        ?.order ?? NaN
+
+    const lastQuizCreatedAt =
+      quizzesCompleted[quizzesCompleted.length - 1]?.createdAt ?? new Date()
+
+    const completed = Boolean(quizCompleted)
```
dolcalmi, line 58:
```suggestion
    const completed = !!quizCompleted
```

core/api/src/domain/quiz/index.ts
```diff
@@ -0,0 +1,78 @@
...
+): FillQuizInformationResult => {
+  const currentSection = lastSectionCompleted(quizzesCompleted)
+
+  const quizzes = Object.entries(QuizzesValue).map(([id, amount]) => {
+    const quizCompleted = quizzesCompleted.find((quiz) => quiz.quizId === id)
+    const section =
+      QuizzesSections.find((section) => section.quiz.includes(id as QuizQuestionId))
+        ?.order ?? NaN
+
+    const lastQuizCreatedAt =
```
dolcalmi, line 55: please move to line 62

core/api/src/domain/quiz/index.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export const fillQuizInformation = (
+  quizzesCompleted: QuizCompleted[],
+): FillQuizInformationResult => {
+  const currentSection = lastSectionCompleted(quizzesCompleted)
+
+  const quizzes = Object.entries(QuizzesValue).map(([id, amount]) => {
+    const quizCompleted = quizzesCompleted.find((quiz) => quiz.quizId === id)
+    const section =
```
dolcalmi, line 51: not sure to unerstand the use of NaN... shouldn't we return an error?

core/api/src/domain/quiz/index.ts
```diff
@@ -0,0 +1,78 @@
...
+  )
+
+  // if we completed the last quiz of the current section, we move to the next
+  const section = QuizzesSections[index]?.quiz
+  if (section && section[section.length - 1] === lastQuizCompleted) return index + 1
+
+  return index === -1 ? 0 : index
+}
+
+export const fillQuizInformation = (
```
dolcalmi, line 44: cant this be replaced by flatMap ?

core/api/src/domain/quiz/config.ts
```diff
@@ -1,6 +1,8 @@
+import { QuizQuestionId } from "./index.types"
+
+export const milliSecondsBetweenSections = 60 * 60 * 12 * 1000
```
dolcalmi, line 3: please move to quizzes config (also this is some of the reasons why having a `config` in domain is not good

core/api/src/app/quiz/list.ts
```diff
@@ -0,0 +1,9 @@
+import { fillQuizInformation } from "@/domain/quiz"
+import { QuizRepository } from "@/services/mongoose"
+
+export const listQuizzesByAccountId = async (accountId: AccountId) => {
```
dolcalmi, line 4: please use 1 param rule (object)

core/api/src/app/quiz/claim.ts
```diff
@@ -32,16 +35,26 @@ import {
 import { consumeLimiter } from "@/services/rate-limit"
 import { getFunderWalletId } from "@/services/ledger/caching"
 import { AccountsIpsRepository } from "@/services/mongoose/accounts-ips"
+import { QuizQuestionId } from "@/domain/quiz/index.types"
```
dolcalmi, line 38: this should not be necessary

core/api/src/app/quiz/claim.ts
```diff
@@ -88,7 +101,18 @@ export const completeQuiz = async ({
 
     if (validatedIPMetadata instanceof UnauthorizedIPError) return validatedIPMetadata
 
-    return new UnknownRepositoryError("add quiz error")
+    return new UnknownRepositoryError("claim quiz error")
```
dolcalmi, line 104: didnt see this in the last PR... but `UnknownRepositoryError` should only be used in repositories. do we need a new error here?

core/api/src/app/quiz/claim.ts
```diff
@@ -32,16 +35,26 @@ import {
...
+  completed: boolean
+  notBefore: Date | undefined
+}[]
 
-export const completeQuiz = async ({
+export const claimQuiz = async ({
   quizQuestionId: quizQuestionIdString,
   accountId: accountIdRaw,
   ip,
+  legacy,
```
dolcalmi, line 51: move to a different method.. having logic for this in a refactor is unnecessary

bats/core/api/quiz.bats
```diff
@@ -78,3 +78,84 @@ setup_file() {
   ')
   [[ "$btc_balance_after_retry" == "$btc_balance_after_quiz" ]] || exit 1
 }
+
+@test "quiz: new mutation" {
```
dolcalmi, line 82: please use a proper description

dolcalmi (changes requested):

core/api/src/graphql/public/types/payload/quiz-completed.ts
nicolasburtey: it's a payload, we can only add `deprecationReason` to the mutation, and this has already be added https://github.com/GaloyMoney/galoy/pull/3743/files#diff-dc640d972fe6a4a4fc5e94759ee942cf9d4940224c264530eb012ebacccbff1bR1172

core/api/src/graphql/public/types/object/quiz.ts
nicolasburtey: yup returning a timeframe at which quiz can be claim it's the whole idea behind this PR

core/api/src/graphql/public/types/object/quiz.ts
nicolasburtey: you mean for SatAmount?

core/api/src/graphql/public/types/object/consumer-account.ts
nicolasburtey: not sure to understand the question / point you're making

core/api/src/app/quiz/list.ts
nicolasburtey: is it a new rule? afair, one of our pattern is that we can use non object where there is a single input to the method, we only force the use of object when there are multiple arguments

core/api/src/graphql/public/types/object/consumer-account.ts
dolcalmi: if you dont use the wrapper (`Quiz` import from `@/app`) we will not have the proper tracing.. this technique is used in trigger for example (some cases where we avoid to use it)

core/api/src/app/quiz/list.ts
dolcalmi: the problem with this is tracing... if we dont use objects is difficult to trace and block leaks because param is mapped with a number (which is generic) while if you use an object we will have the name `code.function.params.paramName`

core/api/src/graphql/public/types/payload/quiz-claim.ts
```diff
@@ -0,0 +1,18 @@
+import IError from "../../../shared/types/abstract/error"
```
dolcalmi, line 1: in this cases is better to use `@/graphql` prefix

core/api/src/graphql/public/types/object/quiz.ts
dolcalmi: yes

dolcalmi (changes requested):
additional feedback + not all previous comments (please check `hidden conversations`) addressed/answered

core/api/src/domain/quiz/index.ts
nicolasburtey: it's not supposed to happen; but typescript is complaining without it (it's because find always return `undefined` as a potential type, but we know here we will always return an object

core/api/src/domain/quiz/index.ts
nicolasburtey: I'm not sure, maybe it could be, but what is the issue with the current code?

core/api/src/domain/quiz/script.ts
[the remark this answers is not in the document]
nicolasburtey: it's a script to import the list of list of sections / quizzes from the mobile app (mobile app is currently the source of truth for that)
I think it's useful to keep around if we update the quizzes in the future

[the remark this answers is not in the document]
nicolasburtey: I named in script.ts to highlight it's not really executable code. maybe I can put this script somewhere else but don't think it matters really?

core/api/src/domain/quiz/sections.ts
nicolasburtey: they represents different things:
- this represent which quiz are in which sections
- the other config represent the reward of each quiz

I can merge it but at the end of the day we get the same information, and I feel the config will be more complexed if merged

core/api/src/graphql/public/types/object/quiz.ts
nicolasburtey: yeah but that's almost another refactoring PR :)

core/api/src/domain/quiz/script.ts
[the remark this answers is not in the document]
dolcalmi: yes, it should not be part of domain also include a relative path to galoy mobile (not all devs have the same folder structure) and have a console.log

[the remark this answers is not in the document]
nicolasburtey: I don't want to make a complicated script that download the galoy-mobile, so it has to be a relatively path

But I'll add a comment at least to explain this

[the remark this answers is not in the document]
dolcalmi: maybe we need another PoV... @vindard @bodymindarts do you have comments no this?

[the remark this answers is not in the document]
bodymindarts: My POV:
- if its a dev script it certainly shouldn't live under `src/domain` rather in `<root>/dev`.
- Not happy with mobile being source of truth (circular dependency). Why not use https://github.com/GaloyMoney/earn as the source of truth to bump both repos?
- if its used seldomly just for 1 off sync I guess its _okayish_ to pull from mobile (for pragmatic reasons)... but then I absolutely would want it to be a bash script or use vendir or something that is easily reproducible for all devs and does not rely on implicit folder structure.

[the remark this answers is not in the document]
nicolasburtey: I think you guys may be over thinking this. the last time we updated quiz was in 2020, I don't known when we will update it again, but likely not anytime soon, so I would not spend a lot more time into that script I just made to sync with the mobile repo. you guys should not see that as a core engineering piece, but more a marketing / growth operation, where velocity of execution matters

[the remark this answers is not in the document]
dolcalmi: this is not overthinking... we need to keep the code as clean as possible and having an script in domain is not good and not needed because as you mentioned `the last time we updated quiz was in 2020` so it is not necessary to have it.

if you still want it please move it too dev or debug

dolcalmi (approved):

PR #3758, opened 2023-12-27
fix: check if 'lineTypeIntelligence' property exists

vindard:
## Description

Improves visibility for [this trace](https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/goLn2oFXeQi/trace/y31jmYHA5af?fields[]=s_name&fields[]=s_serviceName&source=query&span=cee07c2f2f2b5477)

### DIFF
--- a/core/api/src/domain/phone-provider/errors.ts
+++ b/core/api/src/domain/phone-provider/errors.ts
@@ -10,6 +10,9 @@ export class RestrictedRegionPhoneProviderError extends PhoneProviderServiceErro
 export class UnsubscribedRecipientPhoneProviderError extends PhoneProviderServiceError {}
 export class RestrictedRecipientPhoneNumberError extends PhoneProviderServiceError {}
 export class InvalidOrApprovedVerificationError extends PhoneProviderServiceError {}
+export class MissingTypePhoneProviderError extends PhoneProviderServiceError {
+  level = ErrorLevel.Critical
+}
 
 export class PhoneCodeInvalidError extends PhoneProviderServiceError {}
 
--- a/core/api/src/services/twilio-service.ts
+++ b/core/api/src/services/twilio-service.ts
@@ -16,6 +16,7 @@ import {
   InvalidOrApprovedVerificationError,
   InvalidPhoneNumberPhoneProviderError,
   InvalidTypePhoneProviderError,
+  MissingTypePhoneProviderError,
   PhoneCodeInvalidError,
   PhoneProviderConnectionError,
   PhoneProviderRateLimitExceededError,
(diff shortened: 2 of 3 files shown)


core/api/src/domain/phone-provider/errors.ts
```diff
@@ -10,6 +10,9 @@ export class RestrictedRegionPhoneProviderError extends PhoneProviderServiceErro
 export class UnsubscribedRecipientPhoneProviderError extends PhoneProviderServiceError {}
 export class RestrictedRecipientPhoneNumberError extends PhoneProviderServiceError {}
 export class InvalidOrApprovedVerificationError extends PhoneProviderServiceError {}
+export class MissingTypePhoneProviderError extends PhoneProviderServiceError {
+  level = ErrorLevel.Critical
```
nicolasburtey, line 14: I'm not sure we should have Critical for this error? not sure in what scenario we would not have this, but this seems to be expected from twilio at times and not much we could about that?

dolcalmi: this should not be critical, we cant control external service response

core/api/src/services/twilio-service.ts
```diff
@@ -57,6 +58,9 @@ export const TwilioClient = (): IPhoneProviderService => {
         fields: "line_type_intelligence",
       })
       // https://www.twilio.com/docs/lookup/v2-api/line-type-intelligence#type-property-values
+      if (!lookup.lineTypeIntelligence) {
```
dolcalmi, line 61: do we really want to stop sending the code if `lineTypeIntelligence`  is not defined?

dolcalmi (changes requested):

core/api/src/domain/phone-provider/errors.ts
vindard: My thinking was to just flag this as critical for now but to identify what's happening more clearly so that we can look into if we should downgrade it to not alert, or if it can serve as a proxy that something is wrong with the general service.

In the last month or so it's only come up 4 time out of thousands of call ([traces](https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/jgWTYE4VR3M))

___

If we think it's better to swallow the error alert now though I can make that change

core/api/src/services/twilio-service.ts
vindard: https://github.com/GaloyMoney/galoy/pull/3758#discussion_r1437226749

dolcalmi (approved):
ok. lets see how this behaves in prod

PR #3761, opened 2023-12-27
fix: time based quizzes

nicolasburtey:
mongodb is not necessarily returning by order of entry, so we need to order the list of quiz to make sure we are comparing date on the last quiz that has been claimed


dolcalmi (approved):
Approved but.. is it not better to sort by db?

nicolasburtey:
Yup but there is still the possibility for someone to claim id not in the expected order (someone using the API, not the mobile app). We may want to robustify this eventually but for now doing ordering in domain is safer than assuming a db order

PR #3768, opened 2024-01-02
feat(dashboard): adding batch payments in dashboard

siddhart1o1:
image
image

csv format needed 
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/siddharth/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/siddharth/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">

</head>

<body link="#0563C1" vlink="#954F72">

username | sats/dollars | memo
-- | -- | --
bob | 25 | test
alice | 68 | test
bob | 15 | test
alice | 39 | test

</body>

</html>


apps/dashboard/components/batch-payments/index.tsx
```diff
@@ -0,0 +1,269 @@
...
+            file={file}
+            setFile={setFile}
+          />
+          <Card
+            sx={{
+              margin: "1em auto",
+              padding: "1em",
+            }}
+          >
+            {`Please upload a CSV file containing the following columns: "username" (the
```
nicolasburtey, line 257: instead of this text, it would be easier to give the user an empty csv with already the right expected column

sandipndev (approved):
LGTM! Would be good is we have an e2e test of this.

siddhart1o1:
> LGTM! Would be good is we have an e2e test of this.



> LGTM! Would be good is we have an e2e test of this.

Yes, will create a PR for that.

apps/dashboard/components/batch-payments/index.tsx
```diff
@@ -0,0 +1,280 @@
...
+            recipient of the bitcoin), "dollars/sats" (this column will indicate the amount;
+            the header of this column will determine which wallet to use. If the header is
+            "dollars," your USD wallet will be utilized, and the amount will be in US dollars. If
+            "sats" is used, your BTC wallet will be employed, and the amount will be in
+            sats), and "memo" (optional)`}
```
nicolasburtey, line 272: I don't think this is intuitive

I might typically want to use my bitcoin wallet for a dollar denominated payment

I think a more intuitive way would be to have an optional walletId to pay from, and if this is not provided, then the defaultWalletId should be used to make the payment

then we can have 2 columns, one called amount, and the other called denomination that can either be USD or BTC

also, we should ensure that the CSV we are using here maps the CSV we have for galoy-cli. I don't think that is the case currently. Maybe the galoy-cli may need some iteration also.

apps/dashboard/app/batch-payments/utils.ts
```diff
@@ -0,0 +1,90 @@
...
+        return new Error("Record with missing username field found.")
+      }
+
+      const amount = walletCurrency === WalletCurrency.Usd ? record.dollars : record.sats
+      if (!amount || Number(amount) <= 0) {
+        return new Error("Record with invalid amount")
+      }
+
+      totalAmount += Number(amount)
+    }
```
nicolasburtey, line 60: we're missing unit tests for the logic in this code

nicolasburtey: we're also missing some end to end tests here, it doens't seem there are any tests for this feature at all

PR #3789, opened 2024-01-05
feat(dashboard): update csv schema in batch payments

siddhart1o1:
image
image

Updated CSV format
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/siddharth/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/siddharth/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">

</head>

<body link="#0563C1" vlink="#954F72">

username | amount | currency | wallet | memo
-- | -- | -- | -- | --
test_user_a | 1.9 | USD | USD | a
test_user_b | 1.142 | USD | BTC | b
test_user_a | 1 | SATS | BTC | c
test_user_b | 1.32 | USD | BTC | d

</body>

</html>


apps/dashboard/app/batch-payments/server-actions.ts
```diff
@@ -159,3 +181,82 @@ export const processPaymentsServerAction = async (
...
+      totalAmountForBTCWallet: number
+      totalAmountForUSDWallet: number
+    }
+    userWalletBalance: {
+      btcWalletBalance: number
+      usdWalletBalance: number
+    }
+  } | null
+}> => {
+  // TODO move this logic to client side by setting up apollo clint
```
nicolasburtey, line 201: what would be the benefits of client side logic?

apps/dashboard/app/batch-payments/server-actions.ts
```diff
@@ -159,3 +181,82 @@ export const processPaymentsServerAction = async (
...
+        },
+      },
+    }
+  }
+
+  return {
+    error: true,
+    message: "Insufficient Balance",
+    responsePayload: null,
+  }
```
nicolasburtey, line 261: to be consistent, we should have the pattern that path with errors are before the success path

apps/dashboard/app/batch-payments/utils.ts
```diff
@@ -2,25 +2,28 @@ import { parse } from "csv-parse/sync"
...
-const CSV_HEADER_SATS = "sats"
-const CSV_HEADER_MEMO = "memo"
+import { WalletCurrency } from "@/services/graphql/generated"
 
-export type CSVRecord = {
-  username: string
-  dollars?: string
-  sats?: string
-  memo?: string
+enum HEADERS {
```
nicolasburtey, line 9: we don't use enum. instead we use `const { ... } as const` in our codebase

apps/dashboard/components/batch-payments/index.tsx
```diff
@@ -64,81 +59,82 @@ export default function BatchPayments() {
...
-        if (processedRecords.error || !processedRecords.responsePayload) {
-          setModalDetails({
-            open: true,
-            heading: "Error",
-            message: processedRecords.message,
-          })
-          setFile(null)
-          return
-        }
+      // VALIDATE ENOUGH BALANCE AND PAYMENT DETAILS
```
nicolasburtey, line 93: lower comment

apps/dashboard/components/batch-payments/sample-table.tsx
```diff
@@ -1,14 +1,13 @@
 import React from "react"
 import Table from "@mui/joy/Table"
-import { Box, Button } from "@mui/joy"
+import { Button } from "@mui/joy"
 import DownloadIcon from "@mui/icons-material/Download"
 
 const SampleCSVTable = () => {
   const createCsvContent = () => {
-    const csvRows = [["username", "dollars", "memo"]]
+    const csvRows = [["username", "amount", "currency", "wallet", "memo"]]
```
nicolasburtey, line 8: what is wallet used for?

siddhart1o1: To indicate the wallet type we are using i.e BTC/USD

apps/dashboard/app/batch-payments/server-actions.ts
siddhart1o1: Not completely sure about moving it to the client side, but the main reason would be to avoid unnecessary API calls, since this server action only checks if the user's balance exceeds the payable amount (doesn't require much compute). We already have the user balance and payable amount available on the client side. The only reason for processing this on the server side currently is the realTime price query. However, we can manage this on the client side by setting up an Apollo provider on client and directly calling the realTimePriceQuery or price subscription there, instead of making a call to the server and then to the Price APIs.
thinking on this not completely sure...

apps/dashboard/components/batch-payments/index.tsx
```diff
@@ -244,35 +240,80 @@ export default function BatchPayments() {
...
+                <li>
+                  <Typography component="div" sx={{ mb: 1 }}>
+                    <strong>Wallet (optional):</strong> The wallet that will be used to
+                    send the payment in USD or BTC. If not provided, the default wallet
+                    will be used. If the currency is SATS and no wallet is provided, BTC
+                    wallet will be used.
+                  </Typography>
+                </li>
+                <li>
+                  <Typography component="div" sx={{ mb: 1 }}>
```
siddhart1o1, line 308: Details about columns.

nicolasburtey: > If the currency is SATS and no wallet is provided, BTC wallet will be used.

I don't think this is the right behavior. this is dangerous.

we should always have currency required. 

if there is a mismatch or error in what the user provides, then we should error and not process any payments

this goes how an API should generally be developed: it should be easy to use, and hard to misuse. and we want to avoid side/surprising effect.

siddhart1o1: > > If the currency is SATS and no wallet is provided, BTC wallet will be used.
> 
> I don't think this is the right behavior. this is dangerous.
> 
> we should always have currency required.
> 
> if there is a mismatch or error in what the user provides, then we should error and not process any payments
> 
> this goes how an API should generally be developed: it should be easy to use, and hard to misuse. and we want to avoid side/surprising effect.

Currency is required. Wallet is optional; if a wallet is not provided, the default wallet will be used. But, SATS only work with a BTC wallet. Therefore, by default, BTC will be used for SATS if a wallet has not already been given.

this might be more clear 
image

nicolasburtey: >  Therefore, by default, BTC will be used for SATS if a wallet has not already been given.

my point is that I think an error here is more appropriate than attempting to use the bitcoin wallet.

siddhart1o1: > > Therefore, by default, BTC will be used for SATS if a wallet has not already been given.
> 
> my point is that I think an error here is more appropriate than attempting to use the bitcoin wallet.

got it made changes.

sandipndev (approved):

PR #3796, opened 2024-01-08
chore: remove legacy endpoint

nicolasburtey:


dolcalmi (approved):
do we still use `quiz-question` query (gql file) in bats?

PR #3806, opened 2024-01-10
test(core): move 'accounts' integration tests

vindard:


core/api/src/debug/create-usd-wallets.ts
dolcalmi, line 1: we can remove this

dolcalmi (changes requested):

dolcalmi (approved):

PR #3821, opened 2024-01-11
feat(notifications): finish migrating gql functionality

UncleSamtoshi:
- switch from account settings to user settings
- add notificationSettings to User
- add userEnableNotificationChannel mutation
- add userDisableNotificationCategory mutation
- add userEnableNotificationCategory mutation


core/notifications/src/app/mod.rs
```diff
@@ -25,6 +25,17 @@ impl NotificationsApp {
         }
     }
 
+    pub async fn notification_settings_for_account(
+        &self,
+        account_id: GaloyAccountId,
+    ) -> Result<AccountNotificationSettings, ApplicationError> {
+        if let Some(settings) = self.settings.find_for_account_id(&account_id).await? {
```
dolcalmi, line 32: probably because I dont like the else blocks :) but... can we replace this with `map_or_else` or something different? (applies to all the if/else blocks in this PR)

UncleSamtoshi: I didn't introduce this pattern, and I'm not sure what is canonical for rust. @bodymindarts any thoughts?

core/notifications/src/graphql/schema.rs
```diff
@@ -13,50 +13,149 @@ pub struct Query;
 #[Object]
 impl Query {
     #[graphql(entity)]
-    async fn consumer_account(&self, id: ID) -> Option<ConsumerAccount> {
```
bodymindarts, line 16: For backwards compatibility won't we need to keep it on consumer_account as well?

Can still return the same
```
app
  .notification_settings_for_user(GaloyUserId::from(self.id.0.clone()))
  .await?;
```

bodymindarts (approved):

core/notifications/src/app/mod.rs
bodymindarts: I agree with @dolcalmi 
```
        let mut user_settings = self
            .settings
            .find_for_user_id(&user_id)
            .await?
            .unwrap_or_else(|| UserNotificationSettings::new(user_id));
```

looks nicer.

core/notifications/src/graphql/schema.rs
UncleSamtoshi: The types are differnent now and I think we will need to keep backwards compatibility by keeping it live on core for a period.

UncleSamtoshi: In addition to the mutations on core.

core/notifications/src/app/mod.rs
UncleSamtoshi: Made this change

core/notifications/src/graphql/schema.rs
bodymindarts: So core will proxy to notifications I for a while I guess.

PR #3835, opened 2024-01-15
test(notifications): add e2e test for grpc server

thevaibhav-dixit:


bats/core/notifications/notifications.bats
```diff
@@ -9,6 +9,9 @@ setup_file() {
   create_user 'alice'
 }
 
+NOTIFICATIONS_GRPC_ENDPOINT="localhost:2478"
+NOTIFICATIONS_PROTO_FILE="./core/notifications/proto/notifications.proto"
```
bodymindarts, line 13: Path should not be relative. Use `REPO_ROOT` from `_common.sh` as an anchor.

bodymindarts (approved):

PR #3846, opened 2024-01-16
test(core): move ledger volume tests

vindard:
## Description

This is an in-place move of volume tests with some refactoring to unroll the tests and make them more readable, and a fix to include `LnFeeReimbursement' to the payments transaction group.

A future iteration would be to also abstract the decision to use `outgoing` vs `incoming` into specific volume functions, instead of it being arbitrarily decided in use-cases.

**Note:**
- the larger amount of new lines is from unrolling the tests to make them more readable and reduce unintuitive de-duplication

- the old `facade.spec.ts` and `service.spec.ts` were somehow the same exact file duplicated


core/api/test/integration/services/ledger-facade.spec.ts
```diff
@@ -650,4 +669,2226 @@ describe("Facade", () => {
...
+            }
+          })
+        }
+      })
+
+      describe("internal send transactions", () => {
+        const internalVolumeAffectingTypes = volumeAffectingTypes.filter(
+          (txType) => txType in internalSendTxFnForType,
+        )
+        if (internalVolumeAffectingTypes.length) {
```
dolcalmi, line 1236: same here

core/api/test/integration/services/ledger-facade.spec.ts
```diff
@@ -650,4 +669,2226 @@ describe("Facade", () => {
...
+            }
+          })
+        }
+      })
+
+      describe("internal receive transactions", () => {
+        const internalVolumeAffectingTypes = volumeAffectingTypes.filter(
+          (txType) => txType in internalReceiveTxFnForType,
+        )
+        if (internalVolumeAffectingTypes.length) {
```
dolcalmi, line 1294: same here

core/api/test/integration/services/ledger-facade.spec.ts
```diff
@@ -650,4 +669,2226 @@ describe("Facade", () => {
...
+                expect(expected).toStrictEqual(actual)
+              })
+            }
+          })
+        }
+
+        const internalNonVolumeAffectingTypes = nonVolumeAffectingTypes.filter(
+          (txType) => txType in internalReceiveTxFnForType,
+        )
+        if (internalNonVolumeAffectingTypes.length) {
```
dolcalmi, line 1324: same here

core/api/test/integration/services/ledger-facade.spec.ts
```diff
@@ -650,4 +669,2226 @@ describe("Facade", () => {
...
+      const {
+        includedTypes: volumeAffectingTypes,
+        excludedTypes: nonVolumeAffectingTypes,
+      } = txTypesForVolumes(intraLedgerSendPaymentTxTypes)
+
+      describe("external out transactions", () => {
+        const externalVolumeAffectingTypes = volumeAffectingTypes.filter(
+          (txType) => txType in externalOutTxFnForType,
+        )
+        if (externalVolumeAffectingTypes.length) {
```
dolcalmi, line 1424: same here

core/api/test/integration/services/ledger-facade.spec.ts
```diff
@@ -650,4 +669,2226 @@ describe("Facade", () => {
...
+                expect(expected).toStrictEqual(actual)
+              })
+            }
+          })
+        }
+
+        const externalNonVolumeAffectingTypes = nonVolumeAffectingTypes.filter(
+          (txType) => txType in externalOutTxFnForType,
+        )
+        if (externalNonVolumeAffectingTypes.length) {
```
dolcalmi, line 1451: same here

core/api/test/integration/services/ledger-facade.spec.ts
```diff
@@ -650,4 +669,2226 @@ describe("Facade", () => {
...
+            }
+          })
+        }
+      })
+
+      describe("external in transactions", () => {
+        const externalVolumeAffectingTypes = volumeAffectingTypes.filter(
+          (txType) => txType in externalInTxFnForType,
+        )
+        if (externalVolumeAffectingTypes.length) {
```
dolcalmi, line 1480: same here

core/api/test/integration/services/ledger-facade.spec.ts
```diff
@@ -650,4 +669,2226 @@ describe("Facade", () => {
...
+                expect(expected).toStrictEqual(actual)
+              })
+            }
+          })
+        }
+
+        const externalNonVolumeAffectingTypes = nonVolumeAffectingTypes.filter(
+          (txType) => txType in externalInTxFnForType,
+        )
+        if (externalNonVolumeAffectingTypes.length) {
```
dolcalmi, line 1507: same here

core/api/test/integration/services/ledger-facade.spec.ts
```diff
@@ -650,4 +669,2226 @@ describe("Facade", () => {
...
+            }
+          })
+        }
+      })
+
+      describe("internal send transactions", () => {
+        const internalVolumeAffectingTypes = volumeAffectingTypes.filter(
+          (txType) => txType in internalSendTxFnForType,
+        )
+        if (internalVolumeAffectingTypes.length) {
```
dolcalmi, line 1536: same here

core/api/test/integration/services/ledger-facade.spec.ts
```diff
@@ -650,4 +669,2226 @@ describe("Facade", () => {
...
+                expect(expected).toStrictEqual(actual)
+              })
+            }
+          })
+        }
+
+        const internalNonVolumeAffectingTypes = nonVolumeAffectingTypes.filter(
+          (txType) => txType in internalSendTxFnForType,
+        )
+        if (internalNonVolumeAffectingTypes.length) {
```
dolcalmi, line 1564: same here

core/api/test/integration/services/ledger-facade.spec.ts
dolcalmi, line 1: this file needs to be splitted

dolcalmi (changes requested):

core/api/test/integration/services/ledger-facade.spec.ts
```diff
@@ -650,4 +669,2226 @@ describe("Facade", () => {
...
+        includedTypes: volumeAffectingTypes,
+        excludedTypes: nonVolumeAffectingTypes,
+      } = txTypesForVolumes(allPaymentTxTypes)
+
+      describe("external out transactions", () => {
+        const externalVolumeAffectingTypes = volumeAffectingTypes.filter(
+          (txType) => txType in externalOutTxFnForType,
+        )
+        if (externalVolumeAffectingTypes.length) {
+          describe("affects volume", () => {
```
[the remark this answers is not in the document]
vindard: Yup this one was tricky. The goal with this particular `volume by tx type` describe block is to take each volume method and then run every transaction type against it to check whether it affects the volume type or not. Example for one volume type, `External payment (withdrawal) txns`, these are the assertions and how they are categorised:

image

This categorization is accomplished by simply defining an array like:
```ts
const externalPaymentTxTypes = [
    "Payment",
    "OnchainPayment",
    "LnFeeReimbursement",
    "LnFailedPayment",
]
```

We have 25 transaction types that can potentially affect each volume, and they are generally sorted into 5 categories. Each of these categories has a slightly different `it` implementation to check volumes against:
1. externalOut
1. externalIn
1. internalOut
1. internalIn
1. noEffect

For these 25 transaction types, we'd need some way to decide whether we will assert that they affect the volume or not.

The methodology I've come up with so far is to:
- group the transaction types into the 5 categories above
- define 2 tests for each category: asset affects volume, or asset does not affect volume
- define an array of the specific transaction types we expect to affect a given volume
- have code to use this array to sort each transaction types into its appropriate category of the 5 categories and then also decide whether to assert if it affects or does not affect the volume

If we were to roll out all these combinations into a distinct test by hand this test suite would become a lot bigger and trickier to reason about. So far this is the best way I've found to condense all the permutations/combinations of tx types that we need to check against each volume, which requires using some if statements in some places to help categorise our assertions and specific implementations of the volume check based on the tx type.

I'm open to other ideas on how we can do this without using the if statements. I'm not sure the right answer is to define every test case individually by hand.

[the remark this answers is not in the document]
vindard: Looking at this again, I think we can actually remove this entire describe. It was brought in when the volume code was a lot messier and the code is at the point now where we can probably test the txVolumeSince function more directly like we've started doing in #

core/api/test/integration/services/ledger-facade.spec.ts
```diff
@@ -650,4 +669,2226 @@ describe("Facade", () => {
...
+          })
+        }
+
+        const internalNonVolumeAffectingTypes = nonVolumeAffectingTypes.filter(
+          (txType) => txType in internalReceiveTxFnForType,
+        )
+        if (internalNonVolumeAffectingTypes.length) {
+          // FIXME: IntraLedgerReceive txns aren't properly accounted for in
+          //        the 'allPaymentVolume' implementation
+          describe.skip("does not affect volume", () => {
```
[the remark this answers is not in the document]
vindard: Is removed in #3870

vindard:
> approved because comments will be resolved after tests deletion

Resolved in #3876

core/api/test/integration/services/ledger-facade.spec.ts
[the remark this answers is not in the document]
vindard: Removed in #3876

PR #3850, opened 2024-01-17
test(pay): add cypress e2e

dolcalmi:
- [x] Add cypress and tests for basic flows
- [ ] Update galoy pay structure (new PR)
- [ ] Update deps (new PR)
- [ ] Add checkout (new PR)


toolchains/workspace-pnpm/macros.bzl
```diff
@@ -549,7 +549,7 @@ def next_build_bin(**kwargs):
         )
     _next_build_bin(
         next = ":{}".format(next_bin),
-        next_build = ":build",
+        next_build = kwargs.pop("next_build", ":build"),
```
bodymindarts, line 552: See line 430 on how to do default args:
```
def prod_tsc_build_bin(
        prod_tsc_build = ":prod_build",
        **kwargs):
    _prod_tsc_build_bin(
        prod_tsc_build = prod_tsc_build,
        **kwargs,
    )
  ```
  
  At leats I think that is what you are trying to do here.

dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/3850/commits/fc1679d739db2318f2305da55c723a414a980dac

bodymindarts (approved):

PR #3856, opened 2024-01-17
chore: make returning map values mandatory

nicolasburtey:


dolcalmi (changes requested):
this does not handle the case where there are no businesses setup. Please update app/graphql layer to do proper validation (repo return an error on a empty list)

nicolasburtey:
> this does not handle the case where there are no businesses setup. Please update app/graphql layer to do proper validation (repo return an error on a empty list)

it's already returning an empty array and not a list. I removed the probably-wrong check for presence of `accounts` because it should always be an array (even if empty) and therefore the condition would always be false

dolcalmi (approved):

PR #3861, opened 2024-01-17
ci: deprecate legacy pipeline

krtk6160:


ci/core/template.lib.yml
```diff
@@ -299,3 +299,84 @@ source:
   password: #@ data.values.docker_registry_password
   repository: #@ data.values.docker_registry + "/galoy-" + component
 #@ end
+
+#@ def proto_src_resource_name(name):
```
bodymindarts, line 303: these are not part of the component template - should be in another file

krtk6160: I thought `template.lib.yml` was for any templates under core - but it's indeed cleaner to have it separate. Done in 509fc64.

bodymindarts (approved):

PR #3868, opened 2024-01-17
refactor(core): limits related volume methods

vindard:
_Merge after #3867_

## Description

This is to codify the application of `incoming` and `outgoing` volumes directly  in the respective volume method

### DIFF
--- a/core/api/src/domain/accounts/index.types.d.ts
+++ b/core/api/src/domain/accounts/index.types.d.ts
@@ -66,17 +66,17 @@ interface IAccountTxVolumeLimitChecker {
 interface IAccountTxVolumeRemaining {
   intraLedger: (args: {
     priceRatio: WalletPriceRatio
-    walletVolumes: TxBaseVolumeAmount<WalletCurrency>[]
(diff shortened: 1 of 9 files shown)


core/api/src/domain/accounts/index.types.d.ts
```diff
@@ -66,17 +66,17 @@ interface IAccountTxVolumeLimitChecker {
 interface IAccountTxVolumeRemaining {
   intraLedger: (args: {
     priceRatio: WalletPriceRatio
-    walletVolumes: TxBaseVolumeAmount<WalletCurrency>[]
+    outWalletVolumes: PaymentAmount<WalletCurrency>[]
```
dolcalmi, line 69: in theory PaymentAmount should be only used for payments...not sure about this type change... cc @bodymindarts

vindard: This is really `WalletAmount`. It's a change we discussed a long time ago but never actually refactored in. We decided there was no need for a distinction among things like `PaymentAmount`, `BalanceAmount` etc.

dolcalmi (approved):

PR #3883, opened 2024-01-19
build: add 'env' variables handling to next_build rule

vindard:
## Description

This will allow for things like
```py
next_build(
    name = "build",
    srcs = [":src"],
    build_env = {
        "NEXT_PUBLIC_CORE_GQL_URL": "http://localhost:4455/graphql",
    }
)
```

### DIFF
--- a/toolchains/workspace-pnpm/build_next_build.py
+++ b/toolchains/workspace-pnpm/build_next_build.py
@@ -19,7 +19,7 @@ if __name__ == "__main__":
         help="Directory of the package",
         )
     parser.add_argument(
-        "--env",
+        "--build-env",
         action="append",
         default=[],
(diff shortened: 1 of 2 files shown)


toolchains/workspace-pnpm/build_next_build.py
```diff
@@ -18,6 +18,12 @@
         "--package-dir",
         help="Directory of the package",
         )
+    parser.add_argument(
+        "--env",
```
bodymindarts, line 22: can we make this `build_env`?

bodymindarts (commented):
If we need build_env - I think this is the way to do it... but we still need a solution that this will work in all our environments without having to re-build.

bodymindarts (approved):

PR #3886, opened 2024-01-22
docs: update API notes on github pages, remove postman collection

openoms:
Updating the API tutorial on github pages (https://galoymoney.github.io/galoy/) with links to dev.galoy.io.

Removing the postman collection files as they are now managed here: https://dev.blink.sv/api/postman

The API reference is also auto-updated weekly and served from https://dev.galoy.io/public-api-reference.html
and
https://dev.galoy.io/admin-api-reference.html
which are linked from the docs.


bodymindarts (dismissed):

openoms:
Removed the yml which is generated when the https://github.com/GaloyMoney/galoy/actions/workflows/generate-graphql-docs.yml action runs (was testing this locally).

The page at https://galoymoney.github.io/galoy will look like this:
image

PR #3887, opened 2024-01-22
refactor(notifications): persist device token in notifications

bodymindarts:


core/api/src/app/users/get-user-language.ts
```diff
@@ -0,0 +1,11 @@
+import { NotificationsService } from "@/services/notifications"
+
+export const getUserLanguage = async (
```
dolcalmi, line 3: I think this is not a proper approach. Language is not only used in notifications and from the domain pov is part of the user not notifications.

if notification sub system needs the language we will need to pass it through notification request or persist it too when the user change the language

dolcalmi (changes requested):

core/api/src/services/notifications/index.ts
```diff
@@ -168,7 +192,12 @@ export const NotificationsService = (): INotificationsService => {
     transaction,
   }: NotificatioSendTransactionArgs): Promise<true | NotificationsServiceError> => {
     try {
-      const hasDeviceTokens = recipient.deviceTokens && recipient.deviceTokens.length > 0
+      const settings = await getUserNotificationSettings(recipient.userId)
```
dolcalmi, line 195: this does not make sense, why do we want to persist settings in an external system if we will request them to then send them again to another method of the external system?

dolcalmi (changes requested):

core/notifications/src/user_notification_settings/entity.rs
```diff
@@ -172,6 +178,40 @@ impl UserNotificationSettings {
...
+    pub fn remove_push_device_token(&mut self, token: PushDeviceToken) {
+        if !self.push_device_tokens().contains(&token) {
+            return;
+        }
+
+        self.events
+            .push(UserNotificationSettingsEvent::PushDeviceTokenRemoved { token })
+    }
+
+    // pub fn remove_push_device_token()
```
thevaibhav-dixit, line 214: redundant comment ?

PR #3891, opened 2024-01-23
refactor(notifications): move device token to notifications

bodymindarts:


core/api/test/integration/app/wallets/send-onchain.spec.ts
```diff
@@ -671,7 +680,7 @@ describe("onChainPay", () => {
       })
 
       // Expect sent notification
-      expect(sendFilteredNotification.mock.calls.length).toBe(1)
+      expect(sendFilteredNotification.mock.calls.length).toBe(2)
```
thevaibhav-dixit, line 683: can you please explain why we needed to change an assert here @UncleSamtoshi

PR #3899, opened 2024-01-25
feat(core): add payment status by hash query/subscription

dolcalmi:


core/api/src/graphql/public/root/query/ln-invoice-payment-status.ts
```diff
@@ -13,20 +13,28 @@ const LnInvoicePaymentStatusQuery = GT.Field({
   },
   resolve: async (_, args) => {
     const { paymentRequest } = args.input
-    if (paymentRequest instanceof Error) throw paymentRequest
+    if (paymentRequest instanceof Error) {
```
vindard, line 16: Same here, I think we usually throw our errors for queries and return for mutations

core/api/src/graphql/public/root/query/ln-invoice-payment-status-by-hash.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const LnInvoicePaymentStatusByHashQuery = GT.Field({
+  type: GT.NonNull(LnInvoicePaymentStatusPayload),
+  args: {
+    input: { type: GT.NonNull(LnInvoicePaymentStatusByHashInput) },
+  },
+  resolve: async (_, args) => {
+    const { paymentHash } = args.input
+    if (paymentHash instanceof Error) {
+      return { errors: [{ message: paymentHash.message }] }
```
vindard, line 17: I believe we usually throw for queries and return only for mutations

core/api/src/graphql/public/queries.ts
```diff
@@ -27,6 +28,7 @@ export const queryFields = {
     realtimePrice: RealtimePriceQuery,
     btcPriceList: BtcPriceListQuery,
     lnInvoicePaymentStatus: LnInvoicePaymentStatusQuery,
```
vindard, line 30: I wonder if we should do something like this just so it's clearer internally?

```suggestion
    lnInvoicePaymentStatus: LnInvoicePaymentStatusByPaymentRequestQuery,
```

core/api/src/graphql/public/root/query/ln-invoice-payment-status.ts
```diff
@@ -1,10 +1,10 @@
 import { Lightning } from "@/app"
+import { WalletInvoiceStatus } from "@/domain/wallet-invoices"
 
 import { GT } from "@/graphql/index"
-import { mapError } from "@/graphql/error-map"
+import { mapAndParseErrorForGqlResponse } from "@/graphql/error-map"
 import LnInvoicePaymentStatusPayload from "@/graphql/public/types/payload/ln-invoice-payment-status"
 import LnInvoicePaymentStatusInput from "@/graphql/public/types/object/ln-invoice-payment-status-input"
```
vindard, line 7: Same comment here for whether we should be more specific internally
```suggestion
import LnInvoicePaymentStatusByRequestInput from "@/graphql/public/types/object/ln-invoice-payment-status-by-request-input"
```

core/api/src/graphql/public/root/subscription/ln-invoice-payment-status.ts
```diff
@@ -16,23 +16,23 @@ const pubsub = PubSubService()
...
 }
 
 type LnInvoicePaymentResolveSource = {
   errors?: IError[]
   status?: string
+  paymentHash?: PaymentHash
+  paymentRequest?: EncodedPaymentRequest
 }
 
 const LnInvoicePaymentStatusSubscription = {
```
vindard, line 30: Same here

```suggestion
const LnInvoicePaymentStatusByRequestSubscription = {
```

vindard (commented):
Few comments about error handling and naming things

core/api/src/graphql/public/root/query/ln-invoice-payment-status.ts
dolcalmi: type payload contains error.. really not sure if was on purpose or because it is used to in subscription... do you remember?

core/api/src/graphql/public/root/query/ln-invoice-payment-status-by-hash.ts
dolcalmi: yes, I copied the structure used in the original status query.. please check https://github.com/GaloyMoney/galoy/pull/3899#discussion_r1484314966

core/api/src/graphql/public/root/subscription/ln-invoice-payment-status.ts
dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/3899/commits/0f67fbcf9456930df5d2d5fc09ebdb2dedd3f9f7

core/api/src/graphql/public/root/query/ln-invoice-payment-status.ts
dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/3899/commits/0f67fbcf9456930df5d2d5fc09ebdb2dedd3f9f7 although name remains the same to avoid breaking changes `name: "LnInvoicePaymentStatusInput"`

core/api/src/graphql/public/queries.ts
dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/3899/commits/0f67fbcf9456930df5d2d5fc09ebdb2dedd3f9f7

vindard (approved):
Looks good to me now

PR #3902, opened 2024-01-25
chore(dashboard): update account id in dashboard

siddhart1o1:


openoms (approved):
Tested locally that the Account ID in Dashboard now matches the `defaultAccount.id` (not just `id`).

Also the copy to clipboard icon works.

PR #3904, opened 2024-01-25
feat(notifications): e2e circle notification use cases

UncleSamtoshi:


core/notifications/src/executor/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        self.create_subscriber_for_galoy_id(event.user_id()).await?;
+        self.update_push_credentials(event.user_id(), &settings.push_device_tokens())
+            .await?;
+
+        let user_id = event.user_id().to_string();
+        let payload = event.into_payload();
+
+        let id = self
```
thevaibhav-dixit, line 49: Should this not be named ```trigger_name```?

thevaibhav-dixit (approved):

core/notifications/src/executor/mod.rs
bodymindarts: that would be more precise yes.

PR #3912, opened 2024-01-29
chore(notifications): fix dead_code warnings

thevaibhav-dixit:


core/notifications/src/executor/novu/subscriber.rs
```diff
@@ -234,6 +234,7 @@ impl Subscribers {
         }
     }
 
+    #[allow(dead_code)]
```
thevaibhav-dixit, line 237: ```[allow(dead_code)]``` is required here. Tried to prefix it as well but then clippy started giving other warnings.

PR #3916, opened 2024-01-30
chore(core): remove swap (via loopd) functionality

vindard:
## Description

Removing as unused component in our system


nicolasburtey:
@vindard there is probably a good chunk of code in our infra/deployment related to this

nicolasburtey (approved):

PR #3923, opened 2024-01-31
test: add bats test for map

nicolasburtey:


bats/core/api/merchant.bats
```diff
@@ -0,0 +1,63 @@
...
+
+  create_user 'merchant'
+  user_update_username 'merchant'
+}
+
+HYDRA_PUBLIC_API="http://localhost:4444"
+HYDRA_ADMIN_API="http://localhost:4445"
+
+
+@test "admin: can retrieve admin token" {
```
vindard, line 17: We already have a test for this so it can probably just be part of the `setup_file` or abstracted as a function

bats/core/api/merchant.bats
```diff
@@ -0,0 +1,63 @@
...
+  -u "$client_id:$client_secret" \
+  -d "grant_type=client_credentials" | jq -r '.access_token'
+  )
+  echo "admin_token: $admin_token"
+  [[ -n "$admin_token" ]] || exit 1
+  cache_value 'admin.token' "$admin_token"
+}
+
+
+@test "merchants: add a merchant with admin api" {
```
vindard, line 38: We usually name the tests after the file it's in

```suggestion
@test "merchant: add a merchant with admin api" {
```

vindard (commented):
Two small nits but otherwise looks good

bats/core/api/merchant.bats
nicolasburtey: yup I just copy and paste it. which option do you think is better? between moving to `setup_file` but where the code would be a test elsewhere (would be more obvious if something break), or having both place where this code run as part of a setup (less code duplicate, not obvious this is a test)?

vindard: I think we can maybe treat it similar to how we have `login_user` in `user.bash`. Would mean not having an explicit test and simply having a function that has assertions within it

vindard (approved):

PR #3925, opened 2024-02-01
refactor(core): remove kratos registration endpoint

vindard:
## Description

Remove registration webhook and move account creation to login use-case


core/api/src/app/authentication/login.ts
```diff
@@ -104,6 +106,12 @@ export const loginWithPhoneToken = async ({
     })
     if (kratosResult instanceof Error) return kratosResult
 
+    const account = await createAccountFromRegistrationPayload({
+      secret: KRATOS_CALLBACK_API_KEY,
```
vindard, line 110: This is the only place we're applying the callback api key now, and it's pretty much redundant here. Should I remove?

dev/config/ory/kratos.yml
```diff
@@ -132,19 +132,6 @@ selfservice:
             #         name: Authorization
             #         value: The-Value-of-My-Key
             #         in: header
```
nicolasburtey, line 134: you could probably remove the 15 lines above also

vindard: https://github.com/GaloyMoney/galoy/pull/3925/commits/ff8786f622994e59d6171cdcf2eac4bdf352432f

core/api/src/app/authentication/login.ts
nicolasburtey: yes `secret` can be removed completely, together with `KRATOS_CALLBACK_API_KEY` (also in charts/deployment)

nicolasburtey: I think you could call `createAccountWithPhoneIdentifier` directly here

nicolasburtey: ie: 

```
  const account = await createAccountWithPhoneIdentifier({
    newAccountInfo: { phone, kratosUserId: userId },
    config: getDefaultAccountsConfig(),
    phoneMetadata,
  })
```

vindard: Resolved https://github.com/GaloyMoney/galoy/pull/3925/commits/ce81cf72bbd549e0434336b0642e9907a0164780

nicolasburtey (approved):

PR #3937, opened 2024-02-03
feat: adding suggest and approve endpoint for the map

nicolasburtey:
new admin panel view:

image


core/api/src/app/merchants/approve-merchant-map.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const approveMerchantById = async (
+  id: MerchantId,
+): Promise<BusinessMapMarker | ApplicationError> => {
+  const merchantsRepo = MerchantsRepository()
+
+  const merchant = await merchantsRepo.findById(id)
+  if (merchant instanceof Error) return merchant
+
+  merchant.validated = true
```
vindard, line 11: Can be a method on the merchants repo similar to how we have things like `markAsPaid` and `markAsProcessingCompleted`

PR #3938, opened 2024-02-03
feat(map): initialising map

siddhart1o1:


.github/labeler.yml
```diff
@@ -22,6 +22,11 @@ admin-panel:
   - flake.lock
   - pnpm-lock.yaml
 
+blink-map:
```
nicolasburtey, line 25: let's just call it map

apps/map/README.md
```diff
@@ -0,0 +1,2 @@
+# Map
+
```
nicolasburtey, line 2: having a command on what to run would be useful

nicolasburtey: for local dev

apps/map/codegen.yml
```diff
@@ -0,0 +1,61 @@
+overwrite: true
+schema:
+  - "http://localhost:4455/graphql"
```
nicolasburtey, line 3: I think it make more sense to use the locally save schema.graphql like we do here https://github.com/GaloyMoney/galoy/blob/main/apps/admin-panel/codegen.yml#L3

apps/map/components/map/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+}
+
+export default function MapComponent({ mapData }: MapComponentProps) {
+  const mapRef = useRef<google.maps.Map>()
+  const [selectedMarker, setSelectedMarker] = useState<
+    BusinessMapMarkersQuery["businessMapMarkers"][number] | null
+  >(null)
+  const [currentLocation, setCurrentLocation] = useState({
+    lat: 13.7942,
+    lng: -88.8965,
```
nicolasburtey, line 20: let's put those in a const

apps/map/components/map/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+  }, [])
+
+  const mapOptions = useMemo<google.maps.MapOptions>(
+    () => ({
+      disableDefaultUI: true,
+      clickableIcons: true,
+      zoomControl: true,
+      maxZoom: 38,
+      minZoom: 3,
+      styles: [...theme, ...unwantedMarkers],
```
nicolasburtey, line 34: what does `unwantedMarkets` means? why is it useful?

apps/map/components/map/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+              strokeWeight: 2,
+              strokeColor: "gold",
+              scale: 0.8,
+            }}
+            key={marker.username}
+            position={{
+              lat: marker.mapInfo.coordinates.latitude,
+              lng: marker.mapInfo.coordinates.longitude,
+            }}
+            title={marker.username as string}
```
nicolasburtey, line 116: casting should not be necessary?

apps/map/components/map/map-theme.ts
```diff
@@ -0,0 +1,80 @@
...
+  {
+    featureType: "road.highway",
+    elementType: "geometry.stroke",
+    stylers: [{ color: "#1f2835" }],
+  },
+  {
+    featureType: "road.highway",
+    elementType: "labels.text.fill",
+    stylers: [{ color: "#f3d19c" }],
+  },
```
nicolasburtey, line 54: why do we have all those elements? if it's copy/paste from some template but is not used, probably best to delete

nicolasburtey:
the "center me" appears to be totally off on my computer (I didn't approve the request when asked to see how that would react... so I would assume it should be a no-op in that case, but it still zoom in in a "random" point in the middle of El Salvador)

apps/map/components/map/index.tsx
siddhart1o1: Remove unnecessary noise/markers from maps. Otherwise, there is too much information on the map.

apps/map/components/map/map-theme.ts
siddhart1o1: Its dark theme for map, it is used.

siddhart1o1:
> the "center me" appears to be totally off on my computer (I didn't approve the request when asked to see how that would react... so I would assume it should be a no-op in that case, but it still zoom in in a "random" point in the middle of El Salvador)

Yes, I know. I was working on it; it's fixed now. Now, the pointer will only be shown if the user allows their location.

apps/map/components/map/index.tsx
nicolasburtey: But where do those markers come from?

nicolasburtey: Assuming it comes from Google maps api, it seems Google maps charge for every single piece of information requested. So if we have to turn all that off, I would think there might be a way to not ask those informations in the first place, which would limit our  billing

siddhart1o1: > Assuming it comes from Google maps api, it seems Google maps charge for every single piece of information requested. So if we have to turn all that off, I would think there might be a way to not ask those informations in the first place, which would limit our billing

Map charges are based on the number of loads; hiding or displaying marker information does not affect it.
https://mapsplatform.google.com/pricing/

nicolasburtey (approved):

PR #3943, opened 2024-02-06
chore(core): move kratos e2e tests and delete e2e dir

vindard:


.env
```diff
@@ -87,7 +87,7 @@ export LOG_LEVEL="info"
 
 export KRATOS_MASTER_USER_PASSWORD="passwordHardtoFindWithNumber123"
 export KRATOS_PG_HOST="localhost"
-export KRATOS_PG_PORT="5433"
+export KRATOS_PG_PORT="5432"
```
dolcalmi, line 90: I dont see changes in docker compose.. are you sure about this?

core/api/.env
```diff
@@ -88,7 +88,7 @@ export LOG_LEVEL="info"
 export KRATOS_MASTER_USER_PASSWORD="passwordHardtoFindWithNumber123"
 export KRATOS_ADMIN_URL="http://localhost:4434"
 export KRATOS_PG_HOST="localhost"
-export KRATOS_PG_PORT="5433"
+export KRATOS_PG_PORT="5432"
```
dolcalmi, line 91: same here

core/api/src/domain/authentication/index.ts
```diff
@@ -35,6 +35,20 @@ export const getSupportedCountries = ({
   return countries
 }
 
+export const getNextPageToken = (link: string): string | undefined => {
```
dolcalmi, line 38: this is specific to service implementation, should not be in the core domain

core/api/src/domain/authentication/index.types.d.ts
```diff
@@ -24,21 +24,39 @@ type IdentityPhone = IdentityBase & {
   phone: PhoneNumber
   email: undefined
   emailVerified: undefined
+
+  username?: undefined
```
dolcalmi, line 28: maybe I am missing something but kratos is not updated when the user sets the username.. so do we have a process syncing this?

.env
vindard: Oh right, this assumes we now target tilt deps and not docker-compose deps anymore. It's probably irrelevant since we're planning to find a way to remove .env and have all the things we would normally need for local dev come through buck either way.

We have some cleaning up of docs, deps and local workflows to do after this PR

core/api/.env
vindard: https://github.com/GaloyMoney/galoy/pull/3943#discussion_r1482094072

core/api/src/domain/authentication/index.ts
vindard: https://github.com/GaloyMoney/galoy/pull/3943/commits/ca20f97e8d59905ea4e665ec9111fbedfc3a6fed

core/api/src/domain/authentication/index.types.d.ts
vindard: This is different. It's added to `IdentityPhone`/`IdentityEmail`/`IdentityPhoneEmail` just to satisfy the `AnyIdentity` interface and it really only has meaning in `IdentityDeviceAccount` where the type is `IdentityUsername` and not our user's account usernames

dolcalmi (approved):

PR #3951, opened 2024-02-07
chore: add dev codegen rules to all `apps` projects

vindard:
## Description

Run all with `buck2 run //dev/codegen`

### DIFF
--- a/apps/admin-panel/BUCK
+++ b/apps/admin-panel/BUCK
@@ -24,6 +24,11 @@ dev_pnpm_task_binary(
     command="cypress:open",
 )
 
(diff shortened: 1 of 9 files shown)


dolcalmi (approved):

bodymindarts:
Shouldn't this be supported via s first-class toolchain so that buck2 can auto detect when it needs re-doing and get all benefits of previous partial compilation steps (like the [rover toolchain](https://github.com/GaloyMoney/galoy/blob/6d648a0c157620085ed06d1dbb976a079d1bc7d8/dev/BUCK#L1))

PR #3960, opened 2024-02-09
chore: remove env file usage from tests

vindard:
## Description

Follow-up task would be to cleanup other places the env files are referenced and then to delete them.

### `TODO`:
- [x] Look for a way to not have duplicated environment variables inlined in `Tiltfile` and `BUCK` files


toolchains/workspace-pnpm/macros.bzl
```diff
@@ -1165,9 +1165,12 @@ _jest_test = rule(
             default = False,
             doc = "Run all tests serially in the current process"
         ),
-        "env_file": attrs.option(
-            attrs.string(),
-            doc = """File name and relative path for env variables required.""",
+        "app_env": attrs.dict(
```
vindard, line 1168: We have `build_env` (for things like next build) and `env` (for test-runner like jest) attributes in other places. I chose to go with `app_env` (for application code) here but open to alternatives

core/api/BUCK
```diff
@@ -180,21 +180,98 @@ dev_pnpm_task_test(
 
 jest_test(
     name = "unit-tests",
-    srcs = [":src"] + [":test_src"] + glob([".env", "galoy.yaml"]),
+    srcs = [":src"] + [":test_src"] + glob(["galoy.yaml"]),
     config_file = "test/unit/jest.config.js",
-    env_file = ".env",
```
bodymindarts, line 185: `env_json = "//dev:api-env"`

vindard: Context: these variables are pretty much dev environment related so nesting at the `dev/` dir level makes sense

toolchains/workspace-pnpm/macros.bzl
```diff
@@ -634,9 +634,9 @@ def _npm_test_impl(
         cmd_args(program_run_info),
     ])
 
-    if hasattr(ctx.attrs, 'env_json_target'):
-        run_cmd_args.add("--env-json-target")
-        run_cmd_args.add(ctx.attrs.env_json_target)
+    if hasattr(ctx.attrs, 'env_json'):
+        run_cmd_args.add("--env-json")
+        run_cmd_args.add(cmd_args(ctx.attrs.env_json))
```
bodymindarts, line 639: you don't need to double wrap in `cmd_args`... in the code I showed you it was required to 'merge' the path (there were 2 arguments).

bodymindarts (approved):

PR #3967, opened 2024-02-09
chore: re-organize tilt file

vindard:
## Description

This is a preparatory step before I do some refactoring to separate out a build step from our Tiltfile

**Note for `BUILD_ARGS`**
`prepare_build_context` is called inside `npm_test_impl` and is not its own rule. It doesn't get build if you call `build` on the `npm_test_impl` targets so this step still gets built in the `test` stage. To get around this, `prepare_build_context` would have to be broken out into its own rule and passed around accordingly.


vindard:
Taken over by other PRs

PR #3974, opened 2024-02-10
chore(notifications): add sqlxmq boilerplate

bodymindarts:


core/notifications/migrations/20210316025847_setup.up.sql
```diff
@@ -0,0 +1,289 @@
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
```
nicolasburtey, line 1: why is there 2021 code migration?

bodymindarts: It’s the way to add sqlxmq. 

https://github.com/Diggsey/sqlxmq/tree/master?tab=readme-ov-file#database-schema

. Not the best thing about the library but it has been working for us reliably.

thevaibhav-dixit (approved):

PR #3975, opened 2024-02-10
fix(api): improve rate limit

nicolasburtey:


bodymindarts (approved):
Can you add a scope ‘fix(api):’

bodymindarts (approved):

PR #3981, opened 2024-02-12
fix(notification): use serde_json in flattened structs

bodymindarts:


core/notifications/src/job/send_push_notification.rs
```diff
@@ -30,3 +30,59 @@ pub async fn execute(
...
+            all_time_circle_size: 1,
+        });
+        let raw = SendPushNotificationData {
+            payload,
+            tracing_data,
+        };
+        let data = serde_json::to_string(&raw).unwrap();
+        let job_data: JobData<SendPushNotificationData> = serde_json::from_str(&data).unwrap();
+        assert!(job_data.data.is_some());
+        assert_eq!(job_data.tracing_data.len(), 3);
```
bodymindarts, line 86: `  assert_eq!(job_data.tracing_data.len(), 3);` asserting 3 because `data` is also present here.

thevaibhav-dixit (approved):

PR #3991, opened 2024-02-13
fix(notifications): remove device tokens when bad request response

thevaibhav-dixit:


core/notifications/src/executor/mod.rs
```diff
@@ -39,9 +39,23 @@ impl Executor {
...
 
-        self.fcm
-            .send(settings.push_device_tokens(), msg, event.deep_link())
-            .await?;
+        let mut should_persist = false;
+        for device_token in settings.push_device_tokens() {
+            let res = self.fcm.send(&device_token, &msg, event.deep_link()).await;
+            if let Err(e) = res {
+                match e {
+                    FcmError::GoogleFcm1Error(google_fcm1::Error::BadRequest(_)) => {
```
bodymindarts, line 47: Are we sure every `BadRequest` is due to the token?

bodymindarts: I feel we should report the error in the span as an event to know what happened

bodymindarts (approved):

core/notifications/src/executor/mod.rs
thevaibhav-dixit: This is what the errors so far indicate. the underlying library ```google_fcm1``` does not have any explicit error for when registration tokens are ```unregistered```. The error enum contains ```BadRequest(Value)```

bodymindarts: For now I've added tracing so we can investigate

PR #3992, opened 2024-02-13
refactor(pay): use app router in pay

siddhart1o1:


apps/pay/app/[username]/layout.tsx
```diff
@@ -0,0 +1,15 @@
...
+export default function UserNameLayout({
+  children,
+  params,
+}: {
+  children: React.ReactNode
+  params: {
+    username: string
+  }
+}) {
+  return <AppLayout username={params.username}>{children}</AppLayout>
```
Comment (nicolasburtey) [inline apps/pay/app/[username]/layout.tsx:14]: are we making proper use of `Layout` here?

Layour are nested, so if we need an app layout, we should just have it at the root and we don't need to reimport an `AppLayout` further down the tree of the router

apps/pay/app/[username]/page.tsx
```diff
@@ -28,11 +31,18 @@ gql`
   }
 `
 
-function ReceivePayment() {
-  const router = useRouter()
-  const { username, memo, display } = router.query
-  const isIOS = /iPad|iPhone|iPod/.test(navigator.userAgent)
-  const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent)
+type ReceivePaymentProps = {
```
Comment (nicolasburtey) [inline apps/pay/app/[username]/page.tsx:34]: I think generally we can just call this `Props`, assuming there is typicall always just one per file, it's not needed to narrow down the variable name. 

```suggestion
type Props = {
```

apps/pay/app/[username]/page.tsx
```diff
@@ -81,7 +91,7 @@ function ReceivePayment() {
 
   return (
     <>
```
Comment (nicolasburtey) [inline apps/pay/app/[username]/page.tsx:93]:
```suggestion
```

Comment (nicolasburtey) [reply apps/pay/app/[username]/page.tsx:93]: those might typically not be needed (I know they were part of the prior code, but it's a good time to trim those placeholder that are creating code indentation that are just making it more difficult to read the code)

apps/pay/app/globals.css
```diff
@@ -1,17 +1,13 @@
-@import url("https://fonts.googleapis.com/css2?family=Source+Sans+Pro:wght@200;300;400;600;700&display=swap");
```
nicolasburtey, line 1: I don't think it's necessary to have the `_` in `_reducer.tsx` same for `_user.module.css` anymore. I would guess this is an artefact of the way the `page` router worked before

(commenting here because I can't comment on a file that has only a file name change)

apps/pay/app/setuppwa/page.tsx
```diff
@@ -1,15 +1,38 @@
...
 
 import CurrencyDropdown from "../../components/Currency/currency-dropdown"
 
 const SetupPwa = () => {
   const router = useRouter()
   const [username, setUsername] = React.useState<string>("")
-  const username_from_local = localStorage.getItem("username")
-  const display_currency_from_local = localStorage.getItem("display")
+
+  let username_from_local: string | null = null
```
nicolasburtey, line 11:
```suggestion
  let usernameFromLocal: string | null = null
```

apps/pay/components/ParsePOSPayment/Receive-Invoice.tsx
```diff
@@ -32,9 +32,11 @@ const USD_MAX_INVOICE_TIME = 5 // minutes
 const PROGRESS_BAR_MAX_WIDTH = 100 // percent
 
 function ReceiveInvoice({ recipientWalletCurrency, walletId, state, dispatch }: Props) {
-  const deviceDetails = window.navigator.userAgent
-  const router = useRouter()
-  const { username, amount, unit, sats, memo } = router.query
+  const deviceDetails = window.navigator?.userAgent
+  const searchParams = useSearchParams()
+  const { username } = useParams()
+  const query = searchParams ? Object.fromEntries(searchParams.entries()) : {}
```
nicolasburtey, line 38: I don't think we're using useSearchParams the way it should be

https://nextjs.org/docs/app/api-reference/functions/use-search-params

we should do more something like 

```
const amount = searchParams.get("amount")
const unit = searchParams.get("unit")
...
```

apps/pay/components/Memo/index.tsx
```diff
@@ -12,35 +13,32 @@ interface Props {
...
+        `${username}?${new URLSearchParams({
+          amount: amount,
+          sats: sats,
+          unit: unit,
+          memo: note,
+          display,
+        }).toString()}`,
       )
     } else {
-      router.push(
```
nicolasburtey, line 37: shouldn't we use the router.push as it's shown here? 

https://nextjs.org/docs/app/api-reference/functions/use-search-params#updating-searchparams

why using the `window.history.pushState` directly?

this sounds very low level to me, and unclear why we would need it?

nicolasburtey: note: there are multiple places in the code where this comment applies

apps/pay/app/globals.css
siddhart1o1: yup, I am going to clean these.

apps/pay/components/Memo/index.tsx
siddhart1o1: App router doesn't support shallow routing: https://nextjs.org/docs/pages/building-your-application/routing/linking-and-navigating#shallow-routing.

So, this is equivalent to that now. I am replacing these by modifying other parts so everything works together perfectly.

Currently trying to make this work together as in page router, then will do cleanup.

apps/pay/app/[username]/page.tsx
Comment (siddhart1o1) [reply apps/pay/app/[username]/page.tsx:90]: > those might typically not be needed (I know they were part of the prior code, but it's a good time to trim those placeholder that are creating code indentation that are just making it more difficult to read the code)

This is needed only single root element can be returned.

apps/pay/components/Memo/index.tsx
nicolasburtey: this page is from the `page` router. isn't this the equivalent for the app router? https://nextjs.org/docs/app/building-your-application/routing/linking-and-navigating#5-soft-navigation

apps/pay/app/[username]/page.tsx
Comment (nicolasburtey) [reply apps/pay/app/[username]/page.tsx:104]: AFAI see from github, there is still only a single element down the tree: `<Container className={styles.payment_container}>`? in that case, this element should be enough

apps/pay/app/[username]/layout.tsx
Comment (nicolasburtey) [reply apps/pay/app/[username]/layout.tsx:14]: Maybe let's call it something else than `AppLayout`. `UsernameLayout` would be more appropriate

apps/pay/components/Memo/index.tsx
siddhart1o1: > this page is from the `page` router. isn't this the equivalent for the `app` router? https://nextjs.org/docs/app/building-your-application/routing/linking-and-navigating#5-soft-navigation

Yes, page router had this option and app router don't.

apps/pay/app/[username]/page.tsx
Comment (siddhart1o1) [reply apps/pay/app/[username]/page.tsx]: Yup, you are right, I missed it.

apps/pay/components/ParsePOSPayment/Receive-Invoice.tsx
siddhart1o1: We are creating an object from search parameters here. I think it is much cleaner and shorter due to destructuring ?
https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/entries

apps/pay/components/Layouts/UsernameLayout.tsx
```diff
@@ -0,0 +1,189 @@
...
+  username: string
+}
+
+const UsernameLayoutContainer = ({ children, username }: Props) => {
+  const router = useRouter()
+  const searchParams = useSearchParams()
+  const memo = searchParams.get("memo")
+  const [openSideBar, setOpenSideBar] = React.useState<boolean>(false)
+  const [copied, setCopied] = React.useState<boolean>(false)
+  const [isClient, setIsClient] = useState(false)
```
nicolasburtey, line 24: I don't really understand what `isClient` is for. could we drop it? seems it starts with false and is immediately set to true 3 lines above, without more logic to that. seems like dead code to me.

apps/pay/app/[username]/layout.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react"
+
+import UsernameLayoutContainer from "@/components/Layouts/UsernameLayout"
```
Comment (nicolasburtey) [inline apps/pay/app/[username]/layout.tsx:3]: I don't really think this is useful to have this file here invoking a component. we could have in place the component it, instead of wrapping it in a call to `component`. I generally see component as something that we should do when it is something that can be invoked multiple time, but here this layout is called only once, and I don't think it will be present elsewhere.

removing the layer of indrection when it feels clear it will never be needed in the future make it easier to parse the code

apps/pay/utils/utils.ts
```diff
@@ -120,3 +123,7 @@ export const getLocaleConfig = (intlConfig?: IntlConfig): LocaleConfig => {
     return prev
   }, defaultConfig)
 }
+
+export const extractSearchParams = (searchParams: ReadonlyURLSearchParams) => {
+  return searchParams ? Object.fromEntries(searchParams.entries()) : {}
+}
```
nicolasburtey, line 129: I still think that using `searchParams.get()` is better than creating a this util function.

```
  const searchParams = useSearchParams()
  const { username } = useParams()
  const query = extractSearchParams(searchParams)
  ```
  
  this code is convoluted, and require more digging than just doing 

```
const searchParams = useSearchParams()
const amount = searchParams.get('amount')
```

additionaly, I think using https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams instead of play javascript when crafting new URL is a way to prevent edge cases (I used to do that before, but there were all case of side issues that was prevent by the use of the proper API, ie: https://github.com/GaloyMoney/galoy-mobile/pull/3070/files)

File: apps/pay/next.config.js [hunk 11]
```diff
@@ -5,6 +5,7 @@ module.exports = {
   experimental: {
     outputFileTracingRoot: require("path").join(__dirname, "../../"),
     instrumentationHook: true,
+    missingSuspenseWithCSRBailout: false,
```
nicolasburtey, line 8: is that really needed?

https://nextjs.org/docs/messages/missing-suspense-with-csr-bailout

`this option will be removed in future version`

nicolasburtey: ie: can't we add a `<Suspense />` wrapper?

apps/pay/config/config.ts
```diff
@@ -36,9 +35,11 @@ export const getClientSideGqlConfig = () => {
 }
 
 export const getClientSidePayDomain = () => {
-  if (env.NEXT_PUBLIC_PAY_DOMAIN || typeof window === "undefined") {
+  if (env.NEXT_PUBLIC_PAY_DOMAIN) {
     return env.NEXT_PUBLIC_PAY_DOMAIN
+  } else if (typeof window !== "undefined") {
+    return new URL(window.location.href).hostname.split(".").slice(-2).join(".")
```
nicolasburtey, line 41:
```suggestion
    return new URL(window.location.href).hostname.split(".").slice(-2).join(".") // Return the last two parts of the hostname (e.g. "blink.sv")
```

feels like this comment is useful becuase it's not clear otherwise what this function aims for

apps/pay/config/config.ts
```diff
@@ -1,29 +1,28 @@
 import { env } from "../env"
+
 export const USD_INVOICE_EXPIRE_INTERVAL = 60 * 5
 export const MAX_INPUT_VALUE_LENGTH = 14
 export const APP_DESCRIPTION = "Blink official lightning network node"
 
 // TODO get rid of this by removing the use of build time env vars in the client
 export const getClientSideGqlConfig = () => {
-  const hostname = new URL(window.location.href).hostname
+  let hostname, coreGqlUrl, coreGqlWebSocketUrl
```
nicolasburtey, line 9: `implicit any`?

apps/pay/components/ParsePOSPayment/index.tsx
```diff
@@ -212,19 +200,14 @@ function ParsePayment({ defaultWalletCurrency, walletId, dispatch, state }: Prop
       amount: amt,
       sats: satsAmt,
       currency: defaultWalletCurrency,
-      unit,
-      memo,
-      display,
+      unit: unit,
+      memo: memo,
+      display: display,
```
nicolasburtey, line 205: this diff should not be needed?

nicolasburtey, line 205:
```suggestion
      unit,
      memo,
      display,
```

apps/pay/components/ParsePOSPayment/index.tsx
```diff
@@ -108,20 +103,17 @@ function ParsePayment({ defaultWalletCurrency, walletId, dispatch, state }: Prop
...
-        { shallow: true },
-      )
+      const newUrl = `${username}?${new URLSearchParams({
+        amount: initialAmount,
+        sats: initialSats,
+        unit: initialUnit,
+        memo: memo ?? "",
+        display: initialDisplay,
+        currency: defaultWalletCurrency,
+      }).toString()}`
```
nicolasburtey, line 113: I think generally we should use the `URL` api for those type of transformation instead of crafting the url manually

apps/pay/components/ParsePOSPayment/index.tsx
```diff
@@ -56,10 +59,18 @@ const defaultCurrencyMetadata: Currency = {
...
+  state,
+  username,
+}: Props) {
   const router = useRouter()
-  const { username, amount, sats, unit, memo, currency } = router.query
-  const { display } = parseDisplayCurrency(router.query)
+  const searchParams = useSearchParams()
+  const { amount, sats, unit, memo } = extractSearchParams(searchParams)
+
+  const display = searchParams?.get("display") ?? localStorage.getItem("display") ?? "USD"
```
nicolasburtey, line 73: this line is confusing to me. why would we use the `localStorage` if display is in the query url, or use `USD` otherwise?

apps/pay/components/Memo/index.tsx
```diff
@@ -1,46 +1,45 @@
...
-        {
-          pathname: `${username}`,
-          query: { memo: note, display },
-        },
-        undefined,
-        { shallow: true },
+      window.history.pushState(
+        {},
+        "",
+        `${username}?${new URLSearchParams({ memo: note, display }).toString()}`,
```
nicolasburtey, line 42: we should converge on either using `memo` or `note`.

apps/pay/components/Memo/index.tsx
```diff
@@ -1,46 +1,45 @@
...
       )
     } else {
-      router.push(
-        {
-          pathname: `${username}`,
-          query: { memo: note, display },
-        },
-        undefined,
-        { shallow: true },
+      window.history.pushState(
```
nicolasburtey, line 39: ok I see this is a new feature of nextjs 14.1: https://nextjs.org/blog/next-14-1#windowhistorypushstate-and-windowhistoryreplacestate

apps/pay/utils/utils.ts
nicolasburtey: and if this needs transformation, this can be used like this:

```
const params = new URLSearchParams(searchParams.toString());
```

like as an example here: https://nextjs.org/blog/next-14-1#windowhistorypushstate-and-windowhistoryreplacestate

apps/pay/components/receive-amount.tsx
```diff
@@ -29,22 +29,28 @@ export default function ReceiveAmount({
...
+  const { amount, currency } = parseQueryAmount(query) // USD or SATs
+  const { display } = parseDisplayCurrency(query)
 
   function toggleCurrency() {
     const newCurrency = currency === "SATS" ? "USD" : "SATS"
     const newAmount =
       newCurrency === "SATS" ? Math.round(usdToSats(amount)) : satsToUsd(amount)
 
     router.push(
-      getUpdatedURL(router.query, { currency: newCurrency, amount: newAmount, display }),
```
nicolasburtey, line 42: `getUpdatedURL` code is quite smelly to me. I think we may want to either refactor it (can't we just use `URL` and `URLSearchParams`?). having some unit tests otherwise would be useful.

apps/pay/components/Currency/currency-dropdown.tsx
```diff
@@ -15,19 +16,20 @@ export default function CurrencyDropdown({
...
   const [isDropDownOpen, setIsDropDownOpen] = React.useState(false)
 
   useEffect(() => {
-    if (router.query?.display && typeof router.query.display === "string") {
-      setSelectedDisplayCurrency(router.query.display)
-    }
+    const newDisplay =
+      display && typeof display === "string"
+        ? display
+        : localStorage.getItem("display") ?? "USD"
```
nicolasburtey, line 30: I think this should come in params because:
- this code is present elsewhere and not specifically related to a component
- this is business logic that we could unit test

apps/pay/app/setuppwa/page.tsx
```diff
@@ -1,51 +1,53 @@
...
 
+  React.useEffect(() => {
+    if (usernameFromLocal) {
+      window.history.pushState(
+        {},
+        "",
+        `${usernameFromLocal}??display=${displayCurrencyFromLocal}`,
+      )
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
nicolasburtey, line 29: when we are using this comment we should explain in comment why we need that. it's typically not needed and a sign that the overall design of the component may be suboptimal when needed

nicolasburtey:
added some more comments. I haven't tried to run it locally yet, will do on the next iteration.

apps/pay/components/Currency/currency-dropdown.tsx
```diff
@@ -1,6 +1,7 @@
+"use client"
```
nicolasburtey, line 1: maybe in a follow up PR, let's also make the filename lower case only, so that it's consistent with the rest of the codebase

apps/pay/app/[username]/layout.tsx
Comment (siddhart1o1) [reply apps/pay/app/[username]/layout.tsx:3]: `UsernameLayoutContainer` is a client component, if we use it in place then the whole child component will become client by default. Currently, I know most of them are client component, but we are going to do refactoring in UI and remove elements from this, then we can use server component there.

https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#supported-pattern-passing-server-components-to-client-components-as-props

apps/pay/utils/utils.ts
siddhart1o1: > I still think that using `searchParams.get()` is better than creating a this util function.
> 
> ```
>   const searchParams = useSearchParams()
>   const { username } = useParams()
>   const query = extractSearchParams(searchParams)
> ```
> 
> this code is convoluted, and require more digging than just doing
> 
> ```
> const searchParams = useSearchParams()
> const amount = searchParams.get('amount')
> ```
> 
> additionaly, I think using https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams instead of play javascript when crafting new URL is a way to prevent edge cases (I used to do that before, but there were all case of side issues that was prevent by the use of the proper API, ie: https://github.com/GaloyMoney/galoy-mobile/pull/3070/files)

Okay, I will make changes, but in some places, we are required to use an object of params. There, we can use this for now, but when we do UI refactoring, we can remove this in those places as well because we are going to remove some UI elements, and I think it would be easier to do refactoring then. Object of params is required because the older `router` used to provide params as an object via `router.query`.

apps/pay/components/Currency/currency-dropdown.tsx
siddhart1o1: Yup, agree.

apps/pay/components/ParsePOSPayment/index.tsx
```diff
@@ -70,36 +81,20 @@ function ParsePayment({ defaultWalletCurrency, walletId, dispatch, state }: Prop
     defaultCurrencyMetadata,
   )
   const [numOfChanges, setNumOfChanges] = React.useState(0)
-
+  const language = typeof navigator !== "undefined" ? navigator?.language : "en"
```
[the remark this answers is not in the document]
siddhart1o1: > is language only used here?

yes for now only here.

apps/pay/components/Layouts/UsernameLayout.tsx
siddhart1o1: this is because of this part: https://github.com/GaloyMoney/galoy/pull/3992/files#diff-f0ad19133e981f4e00ed785d1af692726d2fef1a9fc7a2726721e262955cc9abR103-R111


It is giving a React minified error (external content does not match server-rendered HTML), so here, we want to only show this if the client is ready. This is a quick fix for now.

apps/pay/app/[username]/layout.tsx
Comment (nicolasburtey) [reply apps/pay/app/[username]/layout.tsx:3]: right ok, I forget the nuances of server/client component embedding

apps/pay/components/Layouts/UsernameLayout.tsx
nicolasburtey: add a comment in the code then to explain why we have this. hacky workaround are better documented if they can be confusing (it is confusing to me still, but that's fine I don't know the intricacies of next)

apps/pay/components/ParsePOSPayment/index.tsx
siddhart1o1: this was part of older code but this was wrapped in a funciton. the reason is if you have used the app selected display currency as EUR it would be saved in browser and then if you you go to POS next time on url something like 

`pay.blink.sv/username` (we don't have params here) then the localstroage will be used to set your previously seected currency.

apps/pay/components/receive-amount.tsx
siddhart1o1: This file is not used in app. There are multiple files that are not used in app. We will need a cleanup PR.

apps/pay/components/Currency/currency-dropdown.tsx
siddhart1o1: Unit test may be for another PR ?

apps/pay/components/ParsePOSPayment/index.tsx
nicolasburtey: shouldn't it be then:
```
let display
if (searchParams?.get("display")) {
  display = searchParams?.get("display")
} else {
  if (localStorage.getItem("display"))
    display = localStorage.getItem("display")
else {
    display = "USD"
  }
}
```
}
```

PR #3999, opened 2024-02-14
build: separate build_context rule

vindard:
## Description

This PR separates the `prepare_build_context` function into its own rule so that it can optionally be executed outside of its downstream tasks. This is useful for example in the `Check Code and Unit Test` github action to properly separate build and test run stages.

To compare test performance:
- Sample from before: https://github.com/GaloyMoney/galoy/actions/runs/7857036456/job/21440456836
- Sample from after: https://github.com/GaloyMoney/galoy/actions/runs/7907032197/job/21584589172?pr=3999


vindard:
We decided that `prepare_build_context` isn't actually a build step and is more a "hermetic environment prep" step that can be used before either build or executing things (like tests). Based on that, this PR will be closed in favour of another PR that simply renames the function to reflect this

PR #4008, opened 2024-02-15
test(pay): add 'pay invoice test'

siddhart1o1:


apps/pay/cypress/e2e/point-of-sale.cy.ts
```diff
@@ -54,4 +56,46 @@ describe("Point of Sale", () => {
...
+            const walletId = response.body.data.me.defaultAccount.defaultWalletId
+            cy.log("Wallet ID:", walletId)
+
+            cy.sendInvoicePayment(paymentRequest, walletId, authToken).then(
+              (paymentResponse) => {
+                expect(paymentResponse.body.data.lnInvoicePaymentSend.status).to.equal(
+                  "SUCCESS",
+                )
+              },
+            )
```
nicolasburtey, line 96: aren't we missing a key point here on whether the QR code has been converted to a green checkmark after payment completion?

nicolasburtey (approved):

PR #4035, opened 2024-02-20
chore(notifications): add email port

thevaibhav-dixit:


core/notifications/src/email_executor/smtp/config.rs
```diff
@@ -7,4 +7,5 @@ pub struct SmtpConfig {
     pub password: String,
     pub from_email: String,
     pub relay: String,
+    pub port: u16,
```
bodymindarts, line 10: Why not default it here?

core/notifications/notifications.yml
```diff
@@ -14,5 +14,6 @@ app:
       username: ""
       from_email: ""
       relay: ""
+      port: 587
```
bodymindarts, line 17: This should actually all be commented... its just for documentation

bodymindarts (approved):

PR #4036, opened 2024-02-20
chore(notifications): use serde_default for email_config

thevaibhav-dixit:


core/notifications/src/email_executor/smtp/config.rs
```diff
@@ -2,10 +2,14 @@ use serde::{Deserialize, Serialize};
...
 pub struct SmtpConfig {
+    #[serde(default)]
     pub username: String,
     #[serde(default)]
     pub password: String,
+    #[serde(default)]
     pub from_email: String,
+    #[serde(default)]
     pub relay: String,
+    #[serde(default)]
```
bodymindarts, line 13: This is not a reasonable default or?

thevaibhav-dixit: you mean i should create impl default for port ?

bodymindarts (approved):

PR #4037, opened 2024-02-20
chore(notifications): boilerplate for html engine

thevaibhav-dixit:


core/notifications/src/email_executor/smtp/mod.rs
```diff
@@ -41,6 +41,7 @@ impl SmtpClient {
             .from(Mailbox::new(None, self.from_email.parse()?))
```
thevaibhav-dixit, line 41: Note: we can pass in a name here that should be displayed instead of the complete email.

bodymindarts (approved):

PR #4048, opened 2024-02-22
feat(pay): ui improvements

siddhart1o1:
+ improvements in UI
   + remove SATs and unit notion, now it's completely fiat only.
    image
    image

   + shifted currency toggle to sidebar
   + added a context for invoice
   + some changes in CSS.
   
   In future PRs
   + add BTC currency to pay in sats
   + move memo to sidebar.


apps/pay/app/[username]/layout.tsx
```diff
@@ -1,19 +1,72 @@
...
+export default async function UsernameLayout({ children, params }: Props) {
+  let response: ApolloQueryResult<AccountDefaultWalletsQuery> | { errorMessage: string }
+  try {
+    response = await getClient().query<AccountDefaultWalletsQuery>({
+      query: AccountDefaultWalletsDocument,
+      variables: { username: params.username },
+    })
+  } catch (e) {
+    console.error("error in username-layout.tsx", e)
+    const error = e as ApolloError
```
Comment (nicolasburtey) [inline apps/pay/app/[username]/layout.tsx:36]: if we know this is an ApolloError, then the casting can be done in the catch branch. but generally it's better not to assume a specific type of error and using `if (err instanceof ApolloError)`

apps/pay/app/[username]/layout.tsx
```diff
@@ -1,19 +1,72 @@
...
+  const initialState = {
+    currentAmount: "0",
+    createdInvoice: false,
+    walletCurrency: response?.data?.accountDefaultWallet.walletCurrency,
+    walletId: response?.data?.accountDefaultWallet.id,
+    username: params.username,
+    pinnedToHomeScreenModalVisible: false,
+    memo: "",
+    displayCurrencyMetaData: defaultCurrencyMetadata,
+  }
```
Comment (nicolasburtey) [inline apps/pay/app/[username]/layout.tsx:63]:
```suggestion
  } as const
```

I guess we're not mutating this

apps/pay/app/layout.tsx
```diff
@@ -2,10 +2,10 @@ import type { Metadata } from "next"
 import { Inter_Tight } from "next/font/google"
 
 // eslint-disable-next-line import/no-unassigned-import
-import "./globals.css"
+import "bootstrap/dist/css/bootstrap.css"
 
 // eslint-disable-next-line import/no-unassigned-import
```
nicolasburtey, line 7:
```suggestion
// eslint-disable-next-line import/no-unassigned-import
```

is this still necessary?

apps/pay/app/ssr-client.ts
```diff
@@ -0,0 +1,28 @@
...
+
+      // TODO uncomment this we start using otel
+      //   fetch: (uri, options) => {
+      //     const headersWithTrace = options?.headers || {}
+      //     propagation.inject(context.active(), headersWithTrace)
+      //     return fetch(String(uri), {
+      //       ...options,
+      //       headers: headersWithTrace,
+      //     })
+      //   },
```
nicolasburtey, line 25:
```suggestion
      // TODO uncomment this we start using otel
      //   fetch: (uri, options) => {
      //     const headersWithTrace = options?.headers || {}
      //     propagation.inject(context.active(), headersWithTrace)
      //     return fetch(String(uri), {
      //       ...options,
      //       headers: headersWithTrace,
      //     })
      //   },
```

not sure what is this comment for. maybe this was a TODO? can't we have otel for the node tasks?

apps/pay/components/currency/currency-dropdown.tsx
```diff
@@ -17,26 +16,28 @@ export default function CurrencyDropdown({
   showOnlyFlag?: boolean
 }) {
   const searchParams = useSearchParams()
-  const display = searchParams?.get("display")
+  const display = searchParams?.get("displayCurrency")
```
nicolasburtey, line 19: I would keep display for now. keep the breaking changes / higher risk element for a smaller PR. so we can isolate the effect of it and decide later if that makes sense or not to do it.

apps/pay/components/layouts/username-layout.tsx
```diff
@@ -55,123 +24,16 @@ const UsernameLayoutContainer = ({ children, username }: Props) => {
   }
 
   return (
-    <div className={`${openSideBar && styles.container_bg} ${styles.container}`}>
+    <div className={styles.main_container}>
```
nicolasburtey, line 27:
```suggestion
    <div className={styles.mainContainer}>
```

let's use camelCase name for consistency

apps/pay/components/payment-outcome/receipt.tsx
```diff
@@ -61,15 +68,16 @@ function receipt(props: Props) {
...
+        <div className="link-wrapper">
+          <a
+            className={styles.link}
+            href="https://galoy.io"
+            target="_blank"
+            rel="noreferrer"
+          >
+            Powered by <GaloyIcon />
+          </a>
+        </div>
```
nicolasburtey, line 80: we can remove that. at least as it is currently, it's taking too much real estate for a POS screen (we should not have the ability to scroll on such a screen really, the focus should be on inputing an amount with a keyboard and we should not distract from this action by making the page longer)

nicolasburtey:
Also, this PR is pretty big. If it's possible to break in down in smaller one, that would ease and probably speed up the review process

apps/pay/app/[username]/layout.tsx
Comment (siddhart1o1) [reply apps/pay/app/[username]/layout.tsx:63]: yes

apps/pay/app/layout.tsx
siddhart1o1: yes, these comments work for next line only. We can add comment to ignore in entire file, but not sure if that is better approach.

apps/pay/app/ssr-client.ts
siddhart1o1: Yes, this was a to-do.

apps/pay/components/layouts/username-layout.tsx
siddhart1o1: This is used throughout the project in CSS. Maybe topic for another PR?

apps/pay/components/payment-outcome/receipt.tsx
siddhart1o1: This is part receipt PDF. Not for webpage.

apps/pay/components/currency/currency-dropdown.tsx
siddhart1o1: I added support for display as well, and it was working with URLs with display parameter. However, I am reverting this to make a smaller PR later.

nicolasburtey (approved):

PR #4051, opened 2024-02-22
chore(notifications): refactor messages

UncleSamtoshi:


bodymindarts (dismissed):

core/notifications/src/notification_event/circle_grew.rs
```diff
@@ -1,5 +1,7 @@
 use serde::{Deserialize, Serialize};
 
+use rust_i18n::t;
```
thevaibhav-dixit, line 3: We try to group external deps together. This should be just below serde

UncleSamtoshi: Addressed

bodymindarts (approved):

PR #4065, opened 2024-02-23
chore: increasing limits for invoices creation

nicolasburtey:
@krtk6160 I don't think it's overrided in prod but maybe something to check


bodymindarts (approved):

krtk6160:
> @krtk6160 I don't think it's overrided in prod but maybe something to check

Yes it's not overridden.

PR #4077, opened 2024-02-26
refactor(notifications): move Transaction push notifications to notifications

bodymindarts:


core/api/src/services/notifications/index.ts
```diff
@@ -194,47 +199,34 @@ export const NotificationsService = (): INotificationsService => {
...
+      tx.setType(notificationTypeToProto(type));
+      const settlementAmount = new ProtoMoney();
+      settlementAmount.setMinorUnits(transaction.settlementAmount)
+      settlementAmount.setCurrencyCode(transaction.settlementCurrency);
+      tx.setSettlementAmount(settlementAmount);
       const displayAmountMajor = transaction.settlementDisplayAmount
       const displayCurrency = transaction.settlementDisplayPrice.displayCurrency
-      const displayAmountMinor = roundToBigInt(
-        majorToMinorUnit({ amount: Number(displayAmountMajor), displayCurrency }),
+      const displayAmountMinor =
```
thevaibhav-dixit, line 216: why remove ```roundToBigInt```

bodymindarts: Because we don't need a `bigint` type... we need a `number` type - they are not compatible.

core/api/src/services/notifications/index.ts
```diff
@@ -194,47 +199,34 @@ export const NotificationsService = (): INotificationsService => {
...
+      const displayAmountMinor = 
+        Math.round(majorToMinorUnit({ amount: Number(displayAmountMajor), displayCurrency }))
+      const displayAmount = new ProtoMoney()
+      displayAmount.setMinorUnits(displayAmountMinor)
+      displayAmount.setCurrencyCode(displayCurrency)
+      tx.setDisplayAmount(displayAmount)
+      event.setTransaction(tx)
+      request.setEvent(event)
+
+      const response = await notificationsGrpc.handleNotificationEvent(
```
dolcalmi, line 225: not a blocker but the method name looks more like the name for a callback handler.... maybe something like `sendEvent` is enough

core/api/src/services/notifications/index.ts
```diff
@@ -194,47 +199,34 @@ export const NotificationsService = (): INotificationsService => {
...
-      const displayAmountMinor = roundToBigInt(
-        majorToMinorUnit({ amount: Number(displayAmountMajor), displayCurrency }),
+      const displayAmountMinor = 
+        Math.round(majorToMinorUnit({ amount: Number(displayAmountMajor), displayCurrency }))
+      const displayAmount = new ProtoMoney()
+      displayAmount.setMinorUnits(displayAmountMinor)
+      displayAmount.setCurrencyCode(displayCurrency)
+      tx.setDisplayAmount(displayAmount)
+      event.setTransaction(tx)
+      request.setEvent(event)
```
dolcalmi, line 223: not a blocker but the code looks a bit messy... maybe something like this?

```js
      const type = getPushNotificationEventType(transaction)
      if (!type) return true

      const settlementAmount = new ProtoMoney()
      settlementAmount.setMinorUnits(transaction.settlementAmount)
      settlementAmount.setCurrencyCode(transaction.settlementCurrency)

      const displayAmountMajor = transaction.settlementDisplayAmount
      const displayCurrency = transaction.settlementDisplayPrice.displayCurrency
      const displayAmountMinor = Math.round(majorToMinorUnit({ amount: Number(displayAmountMajor), displayCurrency }))
      const displayAmount = new ProtoMoney()
      displayAmount.setMinorUnits(displayAmountMinor)
      displayAmount.setCurrencyCode(displayCurrency)

      const tx = new TransactionInfo()
      tx.setUserId(recipient.userId)
      tx.setType(notificationTypeToProto(type))
      tx.setSettlementAmount(settlementAmount)
      tx.setDisplayAmount(displayAmount)

      const event = new NotificationEvent()
      event.setTransaction(tx)

      const request = new HandleNotificationEventRequest()
      request.setEvent(event)
```

dolcalmi (changes requested):
Core: looks good, just some comments

core/api/src/services/notifications/index.ts
bodymindarts: well... we've introduced this name already in many places (KYC, Circles). The thing is `send` is not accurate as the question wether or not to send happens in notifications...

core/api/src/services/notifications/index.ts
```diff
@@ -805,3 +797,29 @@ const translateToNotificationType = (
     return NotificationType.OnchainPayment
   }
 }
+
+const notificationTypeToProto = (type: NotificationType): number => {
```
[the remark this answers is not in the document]
bodymindarts: ah yes - I thought we were using it elsewhere but we are not.

dolcalmi (dismissed):
core(api): approved

core/api/src/services/notifications/index.ts
```diff
@@ -194,46 +199,38 @@ export const NotificationsService = (): INotificationsService => {
...
-          currency: displayCurrency,
-        },
-      })
+      const tx = new TransactionInfo()
+      tx.setUserId(recipient.userId)
+      tx.setType(type)
+      tx.setSettlementAmount(settlementAmount)
+      tx.setDisplayAmount(displayAmount)
 
-      const result = await pushNotification.sendFilteredNotification({
```
bodymindarts, line 228: Removing this line causes https://github.com/GaloyMoney/galoy/blob/1852210584ae7583fbe0f65c9ecb16b3ff9a25ba/core/api/test/integration/app/wallets/send-lightning.spec.ts#L899 to fail...

cc @vindard

thevaibhav-dixit (approved):

Issue #4010, opened 2024-02-16
Feature Request: Auto generate the graphQL string including the API key for BTC Pay Server Integration


ThinusXpressed:
**Is your feature request related to a problem? Please describe.**
When creating a link between BTC Pay Server and Blink via graphQL to setup a lightning node; copying and pasting the API
key into the network configuration is a huzzle considering the string needs to be manualy adjust to get the correct format and it's easy to make a mistake or have a typo.

You have to recreate this:

![image](https://github.com/GaloyMoney/galoy/assets/121315637/8e4b2c17-1d82-4b34-901e-ccc91ad19803)

and then add the API key:

![image](https://github.com/GaloyMoney/galoy/assets/121315637/5124a632-08d5-4f4e-9746-6626e96321af)

**Describe the solution you'd like**
My suggestion would be to add a button that auto generate the graphQL string including the API key that can just be copied from the Blink Dashboard and Pasted on the BTC Server Side... This will make it a one step process and eliminate possible mistakes...

![image](https://github.com/GaloyMoney/galoy/assets/121315637/2b628eae-9fae-4936-aee3-6a800ad2e96b)

that can then just be pasted here in the correct string format:

![image](https://github.com/GaloyMoney/galoy/assets/121315637/5f55f5e2-f07e-4e42-a706-5c2ba60777ee)

**Describe alternatives you've considered**
///...

**Additional context**
///...

nicolasburtey:
the `copy graphql string` would be for btcpay server only, right? so it should be more a `copy settings for btcpay server`?

nicolasburtey:
probably need to be on the previous screen, because we don't show the key anymore on the screen you have screenshoted @ThinusXpressed

ThinusXpressed:
> the `copy graphql string` would be for btcpay server only, right? so it should be more a `copy settings for btcpay server`?

Correct!

ThinusXpressed:
> probably need to be on the previous screen, because we don't show the key anymore on the screen you have screenshoted @ThinusXpressed

This would be fine as long as the correctly combined graphQL + API Key is generated and copied to the clipboard.

nicolasburtey:
this has been deployed

PR #4081, opened 2024-02-27
feat(dashboard) separate connection strings for BTCPay wallet

siddhart1o1:
Users can choose between BTC or USD wallet for connection string.
image

updated to 
image

### DIFF
--- a/apps/dashboard/app/api-keys/page.tsx
+++ b/apps/dashboard/app/api-keys/page.tsx
@@ -6,6 +6,8 @@ import { getServerSession } from "next-auth"
 
 import { authOptions } from "../api/auth/[...nextauth]/route"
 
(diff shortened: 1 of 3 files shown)


nicolasburtey:
I think it's a bit convoluted. I don't think showing the string to bring a lot of value. I would rather just have:

BTCPayWallet connection string with BTC Wallet <icon-copy>
BTCPayWallet connection string with USD Wallet <icon-copy>

siddhart1o1:
> I think it's a bit convoluted. I don't think showing the string to bring a lot of value. I would rather just have:
> 
> BTCPayWallet connection string with BTC Wallet BTCPayWallet connection string with USD Wallet

This was requested : https://github.com/GaloyMoney/galoy/issues/4010#issuecomment-1952302725

nicolasburtey (approved):

PR #4095, opened 2024-02-28
chore: fine tuning earn rate limits

nicolasburtey:


core/api/src/config/schema.ts
```diff
@@ -327,7 +329,12 @@ export const configSchema = {
           blockDuration: 86400,
         },
         addQuizPerIp: {
-          points: 125,
+          points: 20,
```
dolcalmi, line 332: probably too low for events but we can update it in prod

dolcalmi (changes requested):
looks good but this will not solve the delete/recreate account issue, also, `buck2 run //dev:update-schemas` is not working properly so probably we will need to fix the buck2 rule first and run it again (with codegen too)

dolcalmi (approved):

PR #4100, opened 2024-02-29
feat(core): add permissions query

dolcalmi:


nicolasburtey:
here we have read and write: https://github.com/GaloyMoney/galoy/blob/main/apps/dashboard/app/api/auth/%5B...nextauth%5D/route.ts#L29

bats/gql/api-keys.gql
```diff
@@ -3,6 +3,7 @@ query apiKeys {
     defaultAccount {
       id
     }
+    scopes
```
bodymindarts, line 6: `scopes` on `me` evokes the idea that this user / account only ever has permission for those scope. But in actual fact it changes depending on the authentication method (which api key for instance).
So I think this needs to be semantically distinguished as its a property of the authentication method.
Ideas are:
`authenticatedScopes`
`currentScopes`
`authentication { scopes }`

Or perhaps something outside of the `me` query

wdyt @vindard

core/api/src/graphql/public/types/object/user.ts
```diff
@@ -126,6 +128,26 @@ const GraphQLUser = GT.Object<User, GraphQLPublicContextAuth>({
       },
     },
 
+    scopes: {
+      type: GT.NonNullList(Scope),
+      description:
+        "Retrieve the list of scopes permitted for the user's token or API key",
+      resolve: async (_source, _args, { domainAccount, scope }) => {
+        const isActive = domainAccount.status === AccountStatus.Active
```
bodymindarts, line 136: This code contains domain logic. Better to extract it into domain helper that we can unit test.

bats/gql/api-keys.gql
vindard: I like the direction of `authentication { scopes }` because then you can eventually add things like `type` (api key, bearer token etc.), `apiKey` (to share some details about the requesting key) etc.

nicolasburtey: I also agree that putting the scope into `me` means we would be merging the concept of `token` and `user`

right now, `me` is really a synonym or `user` so we have:

```
me (user)
  account
    wallet
````

really the token is adding a new layer. this would be away to think about it:
```
token
  user
    account
      wallet
```

the reason token now is a different entities is 2 folds:
- authorization can be different
- there might be budget associated with a token

the budget is probably one of the most important part going forward, and a highly requested features. 

but I don't think we want `budget` to fall under the concept of a `user`, but rather a `token` or `authentication`

Comment (bodymindarts) [reply bats/gql/api-keys.gql]: I think we need another top level query for this information (not nested under `me`).

It is feasible that an api token that is making the call can make / receive payments but _not_ have authorisation to read other user data.

how about:
```
Query {
  authorization {
    scopes
  }
}
```

Or something like that.

core/api/src/graphql/public/types/object/user.ts
dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/4100/commits/11c8b0bc6091b86a6f3a6eccd1dbbf3acdfb7c98

nicolasburtey (approved):

PR #4107, opened 2024-02-29
feat(core): add lnInvoiceCancel mutation

dolcalmi:


core/api/src/app/wallets/cancel-invoice-for-wallet.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    const { pubkey } = walletInvoice
+    const lndService = LndService()
+    if (lndService instanceof Error) return lndService
+
+    const lnInvoiceLookup = await lndService.lookupInvoice({ pubkey, paymentHash })
+    if (lnInvoiceLookup instanceof Error) return lnInvoiceLookup
+
+    const { isSettled, isHeld } = lnInvoiceLookup
+    if (isSettled || isHeld) {
```
dolcalmi, line 49: should we allow to cancel held invoices too? cc @nicolasburtey

nicolasburtey: held invoices are just because it was not processed by our system yet? (because of a trigger restart for instance)

in that case I don't think so. it would expose some of our internal system

core/api/src/graphql/public/root/mutation/ln-invoice-cancel.ts
```diff
@@ -0,0 +1,54 @@
...
+import WalletId from "@/graphql/shared/types/scalar/wallet-id"
+import PaymentHash from "@/graphql/shared/types/scalar/payment-hash"
+import SuccessPayload from "@/graphql/shared/types/payload/success-payload"
+
+const LnInvoiceCancelInput = GT.Input({
+  name: "LnInvoiceCancelInput",
+  fields: () => ({
+    walletId: {
+      type: GT.NonNull(WalletId),
+      description: "Wallet ID for a BTC wallet belonging to the current account.",
```
vindard, line 17: Doesn't this work for stablesats invoices too?

vindard (dismissed):
Looks good, just one comment about gql description

core/api/src/graphql/public/root/mutation/ln-invoice-cancel.ts
dolcalmi: Updated https://github.com/GaloyMoney/galoy/pull/4107/commits/c4221b7a1c0724ea45d5856def3c3101777947e9

vindard (approved):

PR #4113, opened 2024-03-01
fix(core): update notifications to fire and forget

dolcalmi:


bodymindarts:
Merging as previous e2e run had all test passing (even though it has red X).

PR #4115, opened 2024-03-02
refactor: iteration on generating CSV

nicolasburtey:


core/api/src/graphql/public/mutations.ts
```diff
@@ -113,6 +114,8 @@ export const mutationFields = {
 
       callbackEndpointAdd: CallbackEndpointAdd,
       callbackEndpointDelete: CallbackEndpointDelete,
+
+      csvReportGenerate: CSVReportGenerate,
```
dolcalmi, line 118: This would only make sense if we were creating a scheduled csv and another query to receive the csv url... otherwise must not be a mutation

core/api/src/app/accounts/get-csv-for-account.ts
```diff
@@ -7,9 +7,6 @@ export const getCSVForAccount = async (
   const wallets = await WalletsRepository().listByAccountId(accountId)
   if (wallets instanceof Error) return wallets
 
-  const csv = new CsvWalletsExport()
-  for (const wallet of wallets) {
-    await csv.addWallet(wallet.id)
-  }
-  return csv.getBase64()
+  const csvBase64 = await CsvWalletsExporter(wallets.map((w) => w.id)).getBase64()
+  return csvBase64
```
dolcalmi, line 11:
```suggestion
  return CsvWalletsExporter(wallets.map((w) => w.id)).getBase64()
```

bats/core/api/csv.bats
```diff
@@ -0,0 +1,67 @@
...
+  if echo "$content" | base64 --decode &> /dev/null; then
+    decoded_content="$(echo "$content" | base64 --decode)"
+    echo "Decoded content: $decoded_content"
+  else
+    echo "The content is not valid base64 encoded data."
+    exit 1
+  fi
+
+
+}
```
dolcalmi, line 67: please add an enter (\n)

bats/core/api/csv.bats
```diff
@@ -0,0 +1,67 @@
...
+
+  echo "content: $content"
+
+  if echo "$content" | base64 --decode &> /dev/null; then
+    decoded_content="$(echo "$content" | base64 --decode)"
+    echo "Decoded content: $decoded_content"
+  else
+    echo "The content is not valid base64 encoded data."
+    exit 1
+  fi
```
dolcalmi, line 64: not sure about this.. is this a valid assert in bats? @vindard @bodymindarts

bats/core/api/csv.bats
```diff
@@ -0,0 +1,67 @@
...
+  payment_hash="$(echo $invoice | jq -r '.paymentHash')"
+  [[ "${payment_hash}" != "null" ]] || exit 1
+
+  # Receive payment
+  lnd_outside_cli payinvoice -f \
+    --pay_req "$payment_request"
+
+  # Check for settled
+  retry 15 1 check_for_ln_initiated_settled "$token_name" "$payment_hash"
+}
```
dolcalmi, line 48: if this is required for the test please move the content to the proper test... otherwise looks like it belongs to ln-receive use cases

dolcalmi (changes requested):
please run codegen too
```
buck2 clean #required to avoid cache issues in update schemas
buck2 run //dev:update-schemas
./dev/bin/codegen.sh
```

vindard:
> please run codegen too
> 
> ```
> buck2 clean #required to avoid cache issues in update schemas
> buck2 run //dev:update-schemas
> ./dev/bin/codegen.sh
> ```

We have `buck2 run //dev:codegen` as well that does the same as `./dev/bin/codegen.sh`

bats/core/api/csv.bats
nicolasburtey: what would you do? move the `testing csv` within ln-receive use case?

I need to have some transactions to test that the csv return something.

PR #4123, opened 2024-03-04
refactor: improve supergraph rule names

vindard:


dev/BUCK
```diff
@@ -66,13 +66,13 @@ test_suite(
...
 dev_update_file(
-  name = "update-supergraph",
+  name = "update-supergraph-with-existing-schemas",
   generated = ":supergraph",
   out = "config/apollo-federation/supergraph.graphql"
 )
 
 sh_binary(
-  name = "update-schemas",
+  name = "update-schemas-and-supergraph",
```
nicolasburtey, line 75: I think there should be the notion of `update-graph-apps` versus `update-graph-core` to make it more intelligible what part those rules are addressing

nicolasburtey: I think if we have a `update-generated`, maybe it would more clear that there is other buck2 flow

bodymindarts: There should only be 1 command. BUCK figures or the order. 
i agree that we need a similar setup for Gen Code stuff.

vindard:
This was addressed in #4260

PR #4135, opened 2024-03-05
fix(core): register invoice with actual lnd pubkey it was created with

vindard:
## Description

This fixes the issue where invoices were registered in our local persistence with a default pubkey regardless of which lnd the invoice was actually created on.

### DIFF
--- a/core/api/src/app/wallets/add-invoice-for-wallet.ts
+++ b/core/api/src/app/wallets/add-invoice-for-wallet.ts
@@ -1,7 +1,6 @@
 import { validateIsBtcWallet, validateIsUsdWallet } from "./validate"
 
 import { AccountValidator } from "@/domain/accounts"
-import { toSats } from "@/domain/bitcoin"
 import { checkedToWalletId } from "@/domain/wallets"
 import { RateLimitConfig } from "@/domain/rate-limit"
 import { checkedToMinutes } from "@/domain/primitives"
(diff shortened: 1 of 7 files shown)


core/api/src/app/wallets/add-invoice-for-wallet.ts
```diff
@@ -262,8 +261,7 @@ const addInvoice = async ({
   const walletInvoiceBuilder = WalletInvoiceBuilder({
     dealerBtcFromUsd: dealer.getSatsFromCentsForFutureBuy,
     dealerUsdFromBtc: dealer.getCentsFromSatsForFutureBuy,
-    lnRegisterInvoice: (args) =>
-      lndService.registerInvoice({ ...args, sats: toSats(args.btcPaymentAmount.amount) }),
+    lnRegisterInvoice: lndService.registerInvoice,
```
dolcalmi, line 265: Weren't we using sats?

core/api/src/domain/bitcoin/lightning/index.types.d.ts
```diff
@@ -219,7 +215,7 @@ interface ILightningService {
   }): Promise<RawRoute | LightningServiceError>
 
   registerInvoice(
-    args: RegisterInvoiceArgs,
+    args: RegisterInvoiceArgs & { lndsAndPubkeys: LndAndPubkey[] },
```
dolcalmi, line 218: this is not a good interface update, this method is an abstraction that must not require to pass lnds info or "available nodes" (this must be hidden inside service). Probably is better to use pubkey from RegisteredInvoice

dolcalmi (changes requested):

core/api/src/domain/bitcoin/lightning/index.types.d.ts
vindard: Ah true, missed this! I got distracted by trying to find a way to test this from outside the service. Fixed now

core/api/src/app/wallets/add-invoice-for-wallet.ts
vindard: We had a new interface that uses `btcPaymentAmount` instead that we hadn't switched over to, so I did the switchover here as well https://github.com/GaloyMoney/galoy/pull/4135/commits/cfd5055875b53eaa0d85ce65f79ea26bb8b3da8a

dolcalmi (approved):

PR #4149, opened 2024-03-07
feat(pay): sign in with blink

siddhart1o1:
Needs changes in 
+ [x] charts 
+ [x] deployments

### DIFF
--- a/apps/pay/app/[username]/layout.tsx
+++ b/apps/pay/app/[username]/layout.tsx
@@ -2,7 +2,7 @@ import React from "react"
 
 import { ApolloQueryResult } from "@apollo/client"
 
-import { getClient } from "../ssr-client"
+import { apollo } from "../ssr-client"
 
 import { defaultCurrencyMetadata } from "../currency-metadata"
 
@@ -24,10 +24,13 @@ type Props = {
 export default async function UsernameLayout({ children, params }: Props) {
   let response: ApolloQueryResult<AccountDefaultWalletsQuery> | { errorMessage: string }
   try {
-    response = await getClient().query<AccountDefaultWalletsQuery>({
-      query: AccountDefaultWalletsDocument,
-      variables: { username: params.username },
-    })
+    response = await apollo
+      .unAuthenticate()
+      .getClient()
+      .query<AccountDefaultWalletsQuery>({
+        query: AccountDefaultWalletsDocument,
+        variables: { username: params.username },
+      })
   } catch (err) {
     console.error("error in username-layout.tsx", err)
     if (err instanceof Error) {
--- a/apps/pay/app/[username]/layout.tsx
+++ b/apps/pay/app/[username]/layout.tsx
@@ -25,7 +25,7 @@ export default async function UsernameLayout({ children, params }: Props) {
   let response: ApolloQueryResult<AccountDefaultWalletsQuery> | { errorMessage: string }
   try {
     response = await apollo
-      .unauthorized()
+      .unauthenticated()
       .getClient()
       .query<AccountDefaultWalletsQuery>({
         query: AccountDefaultWalletsDocument,
--- a/apps/pay/app/[username]/layout.tsx
+++ b/apps/pay/app/[username]/layout.tsx
@@ -25,7 +25,7 @@ export default async function UsernameLayout({ children, params }: Props) {
   let response: ApolloQueryResult<AccountDefaultWalletsQuery> | { errorMessage: string }
   try {
     response = await apollo
-      .unAuthenticate()
+      .unauthorized()
       .getClient()
       .query<AccountDefaultWalletsQuery>({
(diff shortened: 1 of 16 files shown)


apps/pay/app/[username]/layout.tsx
```diff
@@ -27,10 +27,13 @@ type Props = {
 export default async function UsernameLayout({ children, params }: Props) {
   let response: ApolloQueryResult<AccountDefaultWalletsQuery> | { errorMessage: string }
   try {
-    response = await getClient().query<AccountDefaultWalletsQuery>({
-      query: AccountDefaultWalletsDocument,
-      variables: { username: params.username },
-    })
+    response = await apollo
+      .unAuthenticate()
```
Comment (nicolasburtey) [inline apps/pay/app/[username]/layout.tsx:31]: this sounds like a weird wording. is that supposed to be mean `non authenticated` queries?

Comment (nicolasburtey) [reply apps/pay/app/[username]/layout.tsx:31]: `unauthenticate` can easily be interpreted as the action to log out.

apps/pay/app/api/auth/[...nextauth]/auth.ts
```diff
@@ -0,0 +1,70 @@
...
+const type = "oauth" as const
+export const authOptions: NextAuthOptions = {
+  providers: [
+    {
+      id: "blink",
+      clientId: env.CLIENT_ID,
+      clientSecret: env.CLIENT_SECRET,
+      wellKnown: `${env.HYDRA_PUBLIC}/.well-known/openid-configuration`,
+      authorization: {
+        params: { scope: "read write" },
```
Comment (nicolasburtey) [inline apps/pay/app/api/auth/[...nextauth]/auth.ts:27]: we only need read permission

apps/pay/components/sidebar/index.tsx
```diff
@@ -96,11 +101,27 @@ export function SideBar({ username }: { username: string }) {
...
+            >
+              {signedInUser ? (
+                <>
+                  <p className="text-md font-semibold mb-1">Signed in as</p>
+                  <p className="text-sm mb-0">{signedInUser.username}</p>
+                </>
+              ) : (
+                <p className="text-md font-semibold mb-1">Sign in</p>
+              )}
+            </div>
```
nicolasburtey, line 123: we're also missing a log out / clear cache. not only for the case we are logged in, but also just to remove the username.

I had to change the username doing a demo this week, and I had to find the settings in chrome to clear all cache to be able to do it. we should have the option here.

apps/pay/app/[username]/layout.tsx
```diff
@@ -27,10 +27,13 @@ type Props = {
 export default async function UsernameLayout({ children, params }: Props) {
   let response: ApolloQueryResult<AccountDefaultWalletsQuery> | { errorMessage: string }
   try {
-    response = await getClient().query<AccountDefaultWalletsQuery>({
-      query: AccountDefaultWalletsDocument,
-      variables: { username: params.username },
-    })
+    response = await apollo
+      .unauthorized()
```
Comment (nicolasburtey) [inline apps/pay/app/[username]/layout.tsx:31]:
```suggestion
      .unauthenticated()
```

the correct terminology here is authenticated. 

its important to know the different between authentication and authorization (but sometimes I mess it myself as well)

nicolasburtey:
can we put this feature behind a beta flag until we can test it a bit more?

we can activate a flag in the console manually by doing this for instance: 

```
document.cookie = "beta=true";
```

like this we can iterate on the transactions lists for instance while still not exposing this feature yet. 

I like the fact you are decomposing the PR in small chunk, but then it makes it that people will ask question on what is the benefits of signing in to blink if there is nothing different until more PR are merged. 

the beta flag solves this

siddhart1o1:
> can we put this feature behind a beta flag until we can test it a bit more?
> 
> we can activate a flag in the console manually by doing this for instance:
> 
> ```
> document.cookie = "beta=true";
> ```
> 
> like this we can iterate on the transactions lists for instance while still not exposing this feature yet.
> 
> I like the fact you are decomposing the PR in small chunk, but then it makes it that people will ask question on what is the benefits of signing in to blink if there is nothing different until more PR are merged.
> 
> the beta flag solves this



> can we put this feature behind a beta flag until we can test it a bit more?
> 
> we can activate a flag in the console manually by doing this for instance:
> 
> ```
> document.cookie = "beta=true";
> ```
> 
> like this we can iterate on the transactions lists for instance while still not exposing this feature yet.
> 
> I like the fact you are decomposing the PR in small chunk, but then it makes it that people will ask question on what is the benefits of signing in to blink if there is nothing different until more PR are merged.
> 
> the beta flag solves this

We already have transaction list PR : https://github.com/GaloyMoney/galoy/pull/4149#issuecomment-1987387798 
Which is built on top of this PR, we can merge that PR into this one when you have reviewed them separately.

nicolasburtey (approved):

nicolasburtey (approved):

PR #4152, opened 2024-03-07
fix(core): handle gql validation error in gql layer

vindard:


core/api/src/graphql/public/root/mutation/onchain-payment-send.ts
```diff
@@ -32,7 +32,7 @@ const OnChainPaymentSendMutation = GT.Field<
     input: {
       walletId: WalletId | InputValidationError
       address: OnChainAddress | InputValidationError
-      amount: number
+      amount: number | InputValidationError
```
dolcalmi, line 35: why are not we using SatAmount? https://github.com/GaloyMoney/galoy/blob/main/core/api/src/graphql/shared/types/scalar/sat-amount.ts

dolcalmi (changes requested):

core/api/src/graphql/public/root/mutation/onchain-payment-send.ts
vindard: I think we are?
https://github.com/GaloyMoney/galoy/blob/7352f934d6739983d224665fcc57977e6536af6d/core/api/src/graphql/public/root/mutation/onchain-payment-send.ts#L19

dolcalmi (approved):

PR #4156, opened 2024-03-08
refactor(core): separate payments lock steps into function

vindard:
## Description

This is to create a separate boundary for lock steps and to explicitly pass arguments across into the lock for ln payment related use-cases.

This has the added benefit of exposing meaningful tracing attributes alongside `bundle_completion_state` attribute in same span.

### DIFF
--- a/core/api/src/app/payments/update-pending-payments.ts
+++ b/core/api/src/app/payments/update-pending-payments.ts
@@ -129,26 +129,19 @@ const updatePendingPayment = wrapAsyncToRunInSpan({
     pendingPayment: LedgerTransaction<WalletCurrency>
     logger: Logger
   }): Promise<true | ApplicationError> => {
-    const { paymentHash, pubkey, type: txType, journalId } = pendingPayment
-    addAttributesToCurrentSpan({ walletId, paymentHash, txType })
+    const { paymentHash, pubkey, type: txType } = pendingPayment
+    // If we had PaymentLedgerType => no need for checking the fields
+    if (!paymentHash) {
+      return new InconsistentDataError("paymentHash missing from payment transaction")
+    }
+    if (!pubkey) {
+      return new InconsistentDataError("pubkey missing from payment transaction")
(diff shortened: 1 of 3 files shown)


core/api/src/app/payments/update-pending-payments.ts
```diff
@@ -129,26 +129,32 @@ const updatePendingPayment = wrapAsyncToRunInSpan({
...
+    }
 
+    addAttributesToCurrentSpan({
+      "payment.walletId": walletId,
+      "payment.request.paymentHash": paymentHash,
+      "payment.txType": txType,
+    })
+
+    const persistedLookup = await LnPaymentsRepository().findByPaymentHash(paymentHash)
+    if (persistedLookup instanceof Error) return persistedLookup
```
dolcalmi, line 148: I am not sure about this, LnPaymentsRepository is not the source of truth and previously we had sync issues with that collection.  In previous code is used only after a timeout

dolcalmi (changes requested):
in general looks good, only 1 comment

core/api/src/app/payments/update-pending-payments.ts
dolcalmi: cc @nicolasburtey @bodymindarts

vindard: So far it's purely used for tracing attributes purposes, not critical to the use-case. Used to fetch `isExpired` for the htlc bug edge case and `destination` for context on successful and failed payments. Both can be undefined so the the only issue would be if we think the lookup could somehow wrongly return a `CouldNotFind` error which would early-return the function

vindard: Added handling for case of `CouldNotFind` issue so that it does not stop the use-case https://github.com/GaloyMoney/galoy/pull/4156/commits/7655eead436bc2107c5ef71cec779fad3fc2aaa5

PR #4165, opened 2024-03-09
chore(pay): add change username link

siddhart1o1:
image

### DIFF
--- a/apps/pay/components/parse-pos-payment/parse-payment.module.css
+++ b/apps/pay/components/parse-pos-payment/parse-payment.module.css
@@ -44,11 +44,12 @@
   .digitsGrid > button {
     font-size: 250%;
   }
+  .digitsGrid > button:active {
+    background-color: rgba(233, 233, 236, 0.3); /* Light touch feedback */
+  }
 }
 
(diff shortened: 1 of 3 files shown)


nicolasburtey:
- share / pin to home screen / change username not having the same width is not very pleasing from a UI standpoint
- I don't think it make sense to have Share as a primary button

siddhart1o1:
> * share / pin to home screen / change username not having the same width is not very pleasing from a UI standpoint

updated

siddhart1o1:
> * I don't think it make sense to have Share as a primary button

I will sync with Andre, it was in design.

nicolasburtey:
I think it should be more a "Log out" option, that should be the same button whether it's just someone who has added a username, or someone who had sign in with the Oauth2 flow.

siddhart1o1:
> I think it should be more a "Log out" option, that should be the same button whether it's just someone who has added a username, or someone who had sign in with the Oauth2 flow.

both are different options i think we can change the username of someone we want to pay, but maybe changing the username should not log us out ? But I can update this if you think it would make things more complex for users.

PR #4166, opened 2024-03-09
feat: openAi Assistant

nicolasburtey:


core/api/src/app/support/chat.ts
```diff
@@ -0,0 +1,65 @@
...
+  if (account instanceof RepositoryError) return account
+
+  let threadId = account.threadId
+
+  if (!threadId) {
+    const threadIdOrError = await Assistant().initialize()
+
+    if (threadIdOrError instanceof Error) {
+      return threadIdOrError
+    } else {
```
dolcalmi, line 41: else is not necessary

core/api/src/services/mongoose/schema.types.d.ts
```diff
@@ -90,6 +90,8 @@ interface AccountRecord {
   defaultWalletId: WalletId
   displayCurrency?: string
 
+  threadId?: string
```
dolcalmi, line 93: uhmm not sure about this.. do we really want to include this in account? what is the real purpose? (openai id?)

core/api/src/services/openai/index.ts
```diff
@@ -0,0 +1,114 @@
...
+
+let openai: OpenAI
+
+if (env.OPENAI_API_KEY) {
+  openai = new OpenAI({
+    apiKey: env.OPENAI_API_KEY,
+  })
+}
+
+export const Assistant = () => {
```
dolcalmi, line 15: Please define the interface

core/api/src/services/openai/index.ts
```diff
@@ -0,0 +1,114 @@
...
+  const assistantId = env.OPENAI_ASSISTANT_ID
+  if (!assistantId) {
+    throw new Error("No assistant id found")
```
dolcalmi, line 18: we must not throw

core/api/src/services/openai/index.ts
```diff
@@ -0,0 +1,114 @@
...
+              : item.content[0].text.value,
+          timestamp: item.created_at,
+        }))
+        .reverse()
+
+      return processedResponse
+    } catch (err) {
+      if (err instanceof OpenAIError) {
+        return err
+      }
```
dolcalmi, line 44: why this? if this is from openai library we need to create a proper wrapper (domain error)

core/api/src/services/openai/index.ts
```diff
@@ -0,0 +1,114 @@
...
+      return new UnknownDomainError("openai unknown error")
```
dolcalmi, line 45: please create a specific error for this service

core/api/src/domain/accounts/index.types.d.ts
```diff
@@ -107,6 +107,8 @@ type Account = {
   readonly contacts: AccountContact[]
   kratosUserId: UserId
   displayCurrency: DisplayCurrency
+  threadId: ThreadId | undefined
```
dolcalmi, line 110: is this assistant thread id?

core/api/src/graphql/public/root/mutation/support-chat-message-add.ts
```diff
@@ -0,0 +1,58 @@
...
+  name: "SupportChatMessageAddPayload",
+  fields: () => ({
+    errors: {
+      type: GT.NonNullList(IError),
+    },
+    supportMessage: {
+      type: GT.List(SupportMessage),
+    },
+  }),
+})
```
dolcalmi, line 25: uhmm should we move this to payloads folder?

core/api/src/graphql/public/types/object/support-message.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const SupportMessage = GT.Object({
+  name: "SupportMessage",
+  fields: () => ({
+    id: { type: GT.NonNullID },
+    message: {
+      type: GT.NonNull(GT.String),
+    },
+    role: {
+      // "user" | "assistant"
```
dolcalmi, line 13: is not better an enum?

core/api/src/graphql/public/types/object/support-message.ts
```diff
@@ -0,0 +1,20 @@
+import Timestamp from "@/graphql/shared/types/scalar/timestamp"
+
+import { GT } from "@/graphql/index"
+
+const SupportMessage = GT.Object({
+  name: "SupportMessage",
+  fields: () => ({
+    id: { type: GT.NonNullID },
+    message: {
+      type: GT.NonNull(GT.String),
```
dolcalmi, line 10: do we need to add a limit?

core/api/src/graphql/public/types/object/user.ts
```diff
@@ -124,6 +126,13 @@ const GraphQLUser = GT.Object<User, GraphQLPublicContextAuth>({
       },
     },
 
+    supportChat: {
+      type: GT.NonNullList(SupportMessage),
+      resolve: async (source, args, { domainAccount }) => {
+        return getSupportChatMessages(domainAccount.id)
```
dolcalmi, line 132: dont we want to use the tracing wrapper?

dolcalmi (changes requested):
first pass

core/api/src/services/mongoose/schema.types.d.ts
nicolasburtey: it's the thread of communication you have (in case you decide to use the chatbot) with the open ai assistant

core/api/src/services/openai/index.ts
nicolasburtey: it's a misconfiguration if we ends up here. are you sure throw is not acceptable? I'm fine crashing the app if it's misconfigured / not supposed to work because we miss an env variable

core/api/src/domain/accounts/index.types.d.ts
nicolasburtey: yes

core/api/src/graphql/public/root/mutation/support-chat-message-add.ts
nicolasburtey: I've been thinking about it, I actually think we should get rid of the payload folder, because there is a one to one relation between mutation and payload, and it doesn't make a lot of sense that we have to add/delete/open 2 files when working on a mutation

I'll do this general change in another PR if that's fine with you

core/api/src/graphql/public/types/object/support-message.ts
nicolasburtey: yes, still wip

core/api/src/graphql/public/types/object/support-message.ts
nicolasburtey: we should, but we can in a follow up PR

core/api/src/graphql/public/types/object/user.ts
nicolasburtey: we should, I will add that.

core/api/src/services/openai/index.ts
bodymindarts: Throwing on illegal states at startup is good - don't want it to start running and silently fail when we could have been notified of the issue much earlier via CrashBackoffLoop.

core/api/src/services/mongoose/schema.types.d.ts
bodymindarts: I think the name could be more descriptive. My first interpretation was also an OS thread.

Perhaps its a `chatContextId` or `supportChatId` or something like that?

core/api/.envrc
```diff
@@ -5,3 +5,5 @@ for key in $(jq -r 'keys[]' $INTEGRATION_ENV_FILE); do
   value=$(jq -r --arg k "$key" '.[$k]' $INTEGRATION_ENV_FILE)
   export $key="$value"
 done
+
+source ../../.env.local
```
bodymindarts, line 9: should be wrapped like in the top level `.envrc`

```
if [ -f .env.local ]; then
  source .env.local
fi
```

core/api/src/app/support/chat.ts
```diff
@@ -0,0 +1,65 @@
+import { RepositoryError } from "@/domain/errors"
+import { AccountsRepository, UsersRepository } from "@/services/mongoose"
+import { Assistant } from "@/services/openai"
+
+export const getSupportChatMessages = async (accountId: AccountId) => {
+  const account = await AccountsRepository().findById(accountId)
```
bodymindarts, line 6: don't think account should be the entry point for this module.

Would expect something like:
```
const supportContext = await SupportContextRepository().findForAccountId(accountId)
```

core/api/src/graphql/public/root/mutation/support-chat-message-add.ts
bodymindarts: Strongly agree - locality of information is key! Group by concept (SupportChartMessage) not by role (XxxPayload).

bodymindarts (commented):
Some high level observations.

I think we can improve the semantics. 

Key words I see are:

```
message
thread
conversation
assistant
```

I'm not sure these are grouped correctly in the domain code. Also I'm expecting this module to have its own entity instead of adding the reference to account.

core/api/src/services/openai/index.ts
nicolasburtey: a point here is that it's not going to crash on startup as of now, but also crash when the Assistant() is been called. 

Alternatively, I can make the openai env var mandatory but this requires some synchronization in the rollout with the chart.

core/api/src/domain/accounts/index.types.d.ts
nicolasburtey: do you think I should rename it?

bodymindarts: I don't see why this needs to be on account at all...

rather should be on its own entity `SupportChat.id` or something like that.

core/api/src/app/support/chat.ts
nicolasburtey: so you would expect another mongodb collection that would have the account -> "threadId" (renamed to supportChatId) relatinoship? even if this all there is to keep right now for now?

bodymindarts: In our code base we always run into adding things onto `Account`. It's a common (anti)-pattern to have some kind of god entity that is the base for looking up everything related. But it doesn't have to be that way - we can flip the dependency around.

For example consider wanting to find the user for an account (hypothetical example):
```
const account = AccountsRepository.findById(account)
const user = UserRepository.findById(account.ownerUserId)
```

In this case the dependency graph goes `Account -> User` (via the field `account.ownerUserId`).

But it is easy to avoid the 'bloat' on account by flipping the dependency graph:

```
const user = UserRepository.findByAccountId(accountId)
```

now `User -> Account` (via the field `user.accountId`).

So that is what I am suggesting here... not have `Account` know about `threadId` at all.
Rather have a `Thread` or `SupportChat` or something like that `Entity` that has an `id` (which currently is your `account.threadId`. And to lookup the entity have a `SupportChatRepository.findByAccountId(accountId)` query.

bodymindarts: .. so yes have a new mongodb table (to persist the entity).

nicolasburtey: > Also I'm expecting this module to have its own entity instead of adding the reference to account.

ok I think this is my answer

core/api/src/services/openai/index.ts
dolcalmi: we dont want to crash the entire pod when someone init a support request

bats/core/api/user.bats
```diff
@@ -24,3 +24,37 @@ setup_file() {
   language="$(graphql_output '.data.me.language')"
   [[ "$language" == "$new_language" ]] || exit 1
 }
+
+@test "support: default support message result" {
+  "skip"
```
dolcalmi, line 29: is this working? https://bats-core.readthedocs.io/en/stable/writing-tests.html#skip-easily-skip-tests

bats/core/api/user.bats
```diff
@@ -24,3 +24,37 @@ setup_file() {
...
+
+  exec_graphql 'alice' 'support-messages'
+  language="$(graphql_output '.data.me.supportChat')"
+  
+  # expect an empty array
+  [[ "$language" == "[]" ]] || exit 1
+}
+
+@test "support: ask 2 questions" {
+  "skip"
```
dolcalmi, line 39: same here, I didnt mention before because it was a draft

bats/core/api/user.bats
```diff
@@ -24,3 +24,37 @@ setup_file() {
...
+  local variables=$(
+    jq -n \
+    '{input: {message: "My transaction is stuck"}}'
+  )
+  exec_graphql 'alice' 'support-chat-message-add' "$variables"
+  language="$(graphql_output '.data.supportChatMessageAdd.supportMessage')"
+  length=$(echo "$language" | jq '. | length')
+  [[ $length -eq 4 ]] || exit 1
+
+  echo "$language"
```
dolcalmi, line 59: please remove and if it is the case update the test

core/api/BUCK
dolcalmi, line 1: is this because your IDE config or are we using different identation in buck files? cc @vindard

core/api/src/domain/chat-support/errors.ts
dolcalmi, line 1: isn't the domain support?

core/api/src/domain/chat-support/index.types.d.ts
```diff
@@ -0,0 +1,8 @@
+type SupportChatId = string & { readonly brand: unique symbol }
+
+type Message = {
+  id: string
+  message: string
+  role: "user" | "assistant"
```
dolcalmi, line 6: please add the type DomainSupportRole (similar to other `enums`)

core/api/src/domain/chat-support/index.types.d.ts
```diff
@@ -0,0 +1,8 @@
...
+  timestamp: number
```
dolcalmi, line 7: must be a Date? (if openai api return a number then we need to translate it in service layer)

core/api/src/services/mongoose/support-chat.ts
```diff
@@ -0,0 +1,40 @@
+import { UnknownRepositoryError } from "@/domain/errors"
+import { SupportChat } from "./schema"
+import { ChatSupportNotFoundError } from "@/domain/chat-support/errors"
+
+export const SupportChatRepository = () => {
```
dolcalmi, line 5: please add the interface

core/api/src/services/openai/index.ts
```diff
@@ -0,0 +1,172 @@
...
+
+let openai: OpenAI
+
+if (env.OPENAI_API_KEY) {
+  openai = new OpenAI({
+    apiKey: env.OPENAI_API_KEY,
+  })
+}
+
+interface Assistant {
```
dolcalmi, line 16: this must be at domain layer and implemented in line 34, also please create arg types and follow naming conventions

core/api/test/integration/services/openai.spec.ts
```diff
@@ -0,0 +1,63 @@
+import { Assistant } from "@/services/openai"
+describe("ChatSupport", () => {
+  it("add 2 message to the thread", async () => {
+    // this test needs an OpenAI API key to run
+    // not running this test in CI
+    if (!process.env.OPENAI_API_KEY) {
+      console.log("No OpenAI API key found, skipping test")
+      return
+    }
```
dolcalmi, line 9: this is a bad practice.. if the test is necessary please create the mock, other wise I prefer this test to be removed

dolcalmi (changes requested):

bodymindarts:
I made some commits to clean up (most of) the types and a bit of the structure.

All tests are passing now - but I will not merge pending the open points by @dolcalmi

bats/core/api/user.bats
nicolasburtey: yes locally when the OPENAI key is added.
it's skipped voluntarily on ci/cd because we don't want to pay for openai usage for testing. it's like twilio it's a pay per call

bats/core/api/user.bats
nicolasburtey: it's really to develop locally

core/api/test/integration/services/openai.spec.ts
nicolasburtey: see that more as a local dev tool?

dolcalmi: no, I mean, this is logic that avoid the test.. tests must test something

dolcalmi (approved):

PR #4171, opened 2024-03-11
chore(pay): QR page changes

siddhart1o1:
image

+ updated create new invoice to secondary btn.
+ removed timer-bar.


nicolasburtey:
I think the button would render better if, the text (Back), was centered in the component, and the arrow was on the left end side

What do you think Andrej?

nicolasburtey (approved):

nicolasburtey:
that's already better imo, so approving now. we can iterate more later.

PR #4181, opened 2024-03-12
refactor(core): move intraledger send locked steps into function

vindard:


core/api/src/app/payments/helpers.ts
```diff
@@ -146,28 +146,38 @@ export const getPriceRatioForLimits = wrapAsyncToRunInSpan({
...
+  senderContactEnabled: boolean
+  senderAccountId: AccountId
+  senderUsername: Username | undefined
+
+  recipientContactEnabled: boolean
+  recipientAccountId: AccountId
+  recipientUsername: Username | undefined
 }): Promise<true | ApplicationError> => {
-  if (!(senderAccount.contactEnabled && recipientAccount.contactEnabled)) {
+  if (!(senderContactEnabled && recipientContactEnabled)) {
```
dolcalmi, line 165: if we move enabled fields at this level then does not make sense to do this validation here. The purpose of passing whole objects is to delegate to this method the business logic. (probably helpers is not a good location for this method)

core/api/src/app/payments/send-intraledger.ts
```diff
@@ -236,151 +223,208 @@ const executePaymentViaIntraledger = async <
...
+      }),
+  )
+  if (paymentSendResult instanceof Error) return paymentSendResult
 
-        memoOfPayer: memo || undefined,
-        senderUsername: senderAccount.username,
-        recipientUsername,
-      }))
-    }
+  if (senderAccount.id !== recipientAccount.id) {
```
dolcalmi, line 266: not a blocker but... moving this inside `executePaymentViaIntraledger` is a bit confusing, actually this is not part of payment execution (the same applies to notifications but it is more critical because of the callbacks)

cc @nicolasburtey @bodymindarts

dolcalmi (changes requested):

core/api/src/app/payments/send-intraledger.ts
vindard: I did this more to be consistent with the `executePaymentViaIntraledger` function in `send-lightning` but can move both to be outside the function if that makes more sense

vindard: Kk fixed here in both files https://github.com/GaloyMoney/galoy/pull/4181/commits/a0e1fd49af9d844cd8e1b97e3b826e5337d4e6ad

core/api/src/app/payments/helpers.ts
vindard: Fixed back here. I'd originally changed it because the function was inside the lock and I wanted to isolate the properties that could change (require potential account refetch), but since it's outside the lock now this is no longer relevant

https://github.com/GaloyMoney/galoy/pull/4181/commits/a1ffe7c8a0e522a8134d0ca3255aeb77c5533ceb

core/api/src/app/payments/send-intraledger.ts
vindard: Also I think you're right about this being a distinct interface being clearer if it weren't in helpers. Moved it to `accounts` module here https://github.com/GaloyMoney/galoy/pull/4181/commits/ba94cbbb40be988f046386b1e2de4c24063c2811

core/api/src/app/payments/helpers.ts
vindard: Also I think you're right about this being a distinct interface being clearer if it weren't in helpers. Moved it to `accounts` module here https://github.com/GaloyMoney/galoy/pull/4181/commits/ba94cbbb40be988f046386b1e2de4c24063c2811

core/api/src/domain/payments/payment-flow.ts
```diff
@@ -56,12 +56,7 @@ export const PaymentFlowCommon = <S extends WalletCurrency, R extends WalletCurr
     }
   }
 
-  const recipientDetails = (): {
-    walletDescriptor: WalletDescriptor<R> | undefined
-    recipientPubkey: Pubkey | undefined
-    recipientUsername: Username | undefined
-    recipientUserId: UserId | undefined
-  } => ({
+  const recipientDetails = () => ({
```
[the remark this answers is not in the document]
vindard: Mistakenly removed (I think I mistook them for arg type, not return type at the time)

Re-added here https://github.com/GaloyMoney/galoy/pull/4181/commits/300a027596c3d48e66d2e1f5be6f5c6df43e2c25

dolcalmi (approved):

PR #4195, opened 2024-03-15
chore(pay): changes in sign in flow

siddhart1o1:
- Added a sign-in option on the landing page; this will work as an alternative to typing the username and currency.
- Do not allow sign-in if the user does not have a username.
- If the user signs in from the sidebar and already has a username in the URL, the sign-in username will take precedence.
- Added a logout/change username button to return to the landing page:
    - If not signed in, "Change Username" will be shown.
    - If signed in, "Logout" will be shown.

image

### DIFF
--- a/apps/pay/app/api/auth/[...nextauth]/auth.ts
+++ b/apps/pay/app/api/auth/[...nextauth]/auth.ts
@@ -66,5 +66,15 @@ export const authOptions: NextAuthOptions = {
       session.accessToken = token.accessToken
       return session
     },
+    async signIn({ account }) {
+      if (!account?.access_token) {
+        return false
+      }
+      const res = await fetchUserData({ token: account.access_token })
+      if (res instanceof Error || !res.data.me?.username) {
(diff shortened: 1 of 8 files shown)


apps/pay/app/api/auth/[...nextauth]/auth.ts
```diff
@@ -66,5 +66,15 @@ export const authOptions: NextAuthOptions = {
...
       return session
     },
+    async signIn({ account }) {
+      if (!account?.access_token) {
+        return false
+      }
+      const res = await fetchUserData({ token: account.access_token })
+      if (res instanceof Error || !res.data.me?.username) {
+        return "/error?errorMessage=This account does not have a username. Please update your profile from the mobile app and update your username"
+      }
```
Comment (nicolasburtey) [inline apps/pay/app/api/auth/[...nextauth]/auth.ts:76]: I think there are 2 errors possible here. if the server doesn't respond for whatever reason, it will return an error. but the error message `This account does not have a username` will be wrong

nicolasburtey (approved):
just one comment. can be addressed in a later PR

PR #4197, opened 2024-03-17
test(quickstart): wallet balance is updated when btc is sent to a user's wallet

sandipndev:
This test PR ensures that an account's BTC wallet balance updates correctly after receiving funds via bitcoin-cli, addressing a balance reporting issue in our mobile app's quickstart setup.

### DIFF
--- a/quickstart/bin/quickstart.sh
+++ b/quickstart/bin/quickstart.sh
@@ -51,7 +51,12 @@ main() {
     sleep 1
   done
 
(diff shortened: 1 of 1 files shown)


sandipndev:
Oh I see that the quickstart test is disabled and all existing runs have failed because it ran out of disk space. [Link](https://github.com/GaloyMoney/galoy/actions/workflows/quickstart.yml)
I can try to see if I can reproduce this in mobile repo itself then.

PR #4207, opened 2024-03-18
feat(admin-panel): add marketing notification UI

UncleSamtoshi:


apps/admin-panel/app/notifications/page.tsx
```diff
@@ -0,0 +1,36 @@
...
+import NotificationTestSender from "../../components/notification/test-sender"
+import NotificationFilteredSender from "../../components/notification/filtered-sender"
+
+export default function NotificationsScreen() {
+  const [notification, setNotification] = useState<NotificationContent>({
+    localizedPushContents: [],
+    deepLink: undefined,
+  })
+
+  const notificationContainsEnglishContent = notification.localizedPushContents.some(
```
siddhart1o1, line 17: Notifications can only be triggered if English is included ?

siddhart1o1 (approved):
lgtm.

apps/admin-panel/components/notification/languages.ts
```diff
@@ -0,0 +1,292 @@
...
+  Pashto: "ps",
+  Amharic: "am",
+  Yoruba: "yo",
+  Igbo: "ig",
+  Swahili: "sw",
+  Hausa: "ha",
+  Zulu: "zu",
+}
+
+export const CountryCodes = [
```
siddhart1o1, line 46: Maybe we could also include the name of the country in future PRs? That would make it easier to find.

apps/admin-panel/app/notifications/page.tsx
UncleSamtoshi: yup at the moment

apps/admin-panel/components/notification/languages.ts
UncleSamtoshi: Thats good idea! Will look into adding that in a future pr.

PR #4209, opened 2024-03-19
chore(dashboard): added sats/cents in transaction table.

siddhart1o1:
image

### DIFF
--- a/apps/dashboard/components/transaction-details/index.types.ts
+++ b/apps/dashboard/components/transaction-details/index.types.ts
@@ -7,6 +7,7 @@ export interface RowNode {
   status: string
   settlementFee: number
   settlementCurrency: string
+  settlementAmount: number
 }
 
 export interface TransactionDetailsProps {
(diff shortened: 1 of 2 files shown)


nicolasburtey (changes requested):
- for bitcoin wallet, sats should not be approximates
- I don't think we want to reduce the density of the table by having a double line on the amount

siddhart1o1:
> * for bitcoin wallet, sats should not be approximates
> * I don't think we want to reduce the density of the table by having a double line on the amount

Updated 
According to @pretyflaco's suggestion,
"""
Display Currency that the payment is made in could be "Payment Amount" 
Currency that the settlement is made in could be "Settlement Amount" 
"""

nicolasburtey:
> > * for bitcoin wallet, sats should not be approximates
> > * I don't think we want to reduce the density of the table by having a double line on the amount
> 
> Updated According to @pretyflaco's suggestion, """ Display Currency that the payment is made in could be "Payment Amount" Currency that the settlement is made in could be "Settlement Amount" """

I think we're introducing a new language here (`Payment amount` is not something we have in our stack as of now). not sure this is ideal but for now I think that's ok.

nicolasburtey (approved):

PR #4210, opened 2024-03-19
ci: add mock config for realtime price server

vindard:
## Description

Pulls in the mock functionality introduced in https://github.com/GaloyMoney/price/pull/62


sandipndev (approved):
Works, fund-user task works for me blazingly fast now with this mock provider.

PR #4215, opened 2024-03-20
feat(pay): transaction printout from pos

siddhart1o1:
Receipt:
image

Auto print out on payment success.

### DIFF
--- a/apps/pay/components/payment-outcome/index.tsx
+++ b/apps/pay/components/payment-outcome/index.tsx
@@ -1,6 +1,6 @@
 "use client"
 import { useParams, useSearchParams } from "next/navigation"
-import React, { useRef } from "react"
+import React, { useEffect, useRef } from "react"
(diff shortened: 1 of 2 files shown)


apps/pay/components/payment-outcome/index.tsx
```diff
@@ -33,10 +33,26 @@ gql`
...
+const sendDataToPosCompanion = ({
+  username,
+  amount,
+  paymentHash,
+}: {
+  username: string
+  amount: string
+  paymentHash: string
+}) => {
+  const deepLinkUrl = `blink-pos-companion://print?username=${encodeURIComponent(
```
nicolasburtey, line 50: we should also pass the time here. 

we should also be able to go to past transactions and print a receipt for those transactions. this would assume that time is not always `now()`. 

but approving now, we can iterate on that in follow up PRs

PR #4219, opened 2024-03-21
chore(toolchain): refactor pnpm install from dev_task rules

vindard:
## Description

This fixes the issue in our local dev where multiple concurrent `pnpm_dev_task..` rules would concurrently try to run `pnpm install` inline in their contained scripts. This fixes this by instead introducing workspace pnpm install rule and passing this in as a dependency to these dev_task rules so that the `pnpm install` step is only executed once.


BUCK
```diff
@@ -19,3 +19,8 @@ export_file(
     name = "rustfmt.toml",
     visibility = ["PUBLIC"],
 )
+
+build_workspace_node_modules(
```
bodymindarts, line 23: I don't think this will solve it as it is downloading them into a folder under `buck-out`. I think we need a `dev_` version of this to solve the issue at hand... or?

vindard: I thought so at first when I saw the si implementation, but it works and creates all the node_modules folders locally as well. I believe they did it this way so that buck co-ordinates when this actually gets executed and can co-ordinate its behavior as a dependency and for concurrent jobs

bodymindarts: do you understand why it creates the node_modules locally as well? Can you show me?

bodymindarts (approved):

Issue #3699, opened 2023-12-12
query that should probably be mutations


nicolasburtey:
- ExportCsvSetting
- fet onchain fees

yemmyharry:
@nicolasburtey i would like to contribute to this.

nicolasburtey:
> @nicolasburtey i would like to contribute to this.

sure go ahead @yemmyharry 

the first one is almost done https://github.com/GaloyMoney/galoy/pull/4115 but onchain fees is not

PR #4222, opened 2024-03-22
feat(core): add currencyConversionEstimate query

dolcalmi:
response example:
```
{
  "data": {
    "currencyConversionEstimation": {
      "id": "b4ace506-fa77-5b0e-a851-12e9eb6640c5",
      "timestamp": 1711121477,
      "btcSatAmount": 11371,
      "usdCentAmount": 227,
    }
  }
}
```


dev/config/apollo-federation/supergraph.graphql
```diff
@@ -1587,6 +1604,15 @@ type Query
...
   businessMapMarkers: [MapMarker!]! @join__field(graph: PUBLIC)
+
+  """Returns an estimated conversion rate for the given amount and currency"""
+  currencyConversionEstimate(
+    """Amount in major unit."""
+    amount: Float!
+
+    """Currency of the amount to be converted."""
+    currency: DisplayCurrency = "USD"
+  ): CurrencyConversionEstimate! @join__field(graph: PUBLIC)
```
nicolasburtey, line 1615: the wording make me thing this should be a mutation (ie: it's an action)

nicolasburtey: how is that different from fetching the price directly?

dolcalmi: > how is that different from fetching the price directly?

the purpose is to reduce queries and overhead on price calculation... even for us is difficult to follow and do a proper implementation, for example please check galoy pay implementation https://github.com/GaloyMoney/galoy/blob/main/apps/pay/lib/use-realtime-price.tsx#L85 (or search `** offset`) we did not implement it correctly. 

In the case of ecommerce plugins we need an easy way to convert fiat prices (in major unit) to wallet currency (minor unit) so this query solve 2 things: reduces the amount of requests to create an invoice from 3 (list currencies, realtime, create invoice) to 2 and reduce the friction for developers in terms of proper calculation using base, offset and fractionDigits

dolcalmi: > the wording make me thing this should be a mutation (ie: it's an action)

uhmm for me it is a query because it does not modify anything on the server side... @vindard @bodymindarts wdyt? 

@nicolasburtey I also have doubts about the name, so please let me know if you prefer something else

nicolasburtey: > > the wording make me thing this should be a mutation (ie: it's an action)
> 
> uhmm for me it is a query because it does not modify anything on the server side... @vindard @bodymindarts wdyt?
> 
> @nicolasburtey I also have doubts about the name, so please let me know if you prefer something else

yup I read back https://graphql.org/learn/queries/ and you are right I think, there is no side effect to be expected here. but the cue on using a verb can generally be a good guess (https://github.com/GaloyMoney/galoy/issues/3699)

core/api/src/graphql/public/root/query/currency-conversion-estimate.ts
```diff
@@ -0,0 +1,80 @@
...
+      amount: majorToMinorUnit({
+        amount: amountInMajorUnit,
+        displayCurrency: currency,
+      }),
+      currency,
+    })
+    if (amountInMinor instanceof Error) throw mapError(amountInMinor)
+
+    const amountInSats = satPriceRatio.convertFromDisplayMinorUnit(amountInMinor)
+    const amountInCents = centPriceRatio.convertFromDisplayMinorUnit(amountInMinor)
```
nicolasburtey, line 69: shouldn't this move to `apps`?

core/api/src/graphql/public/schema.graphql
```diff
@@ -338,6 +338,21 @@ type Currency {
   symbol: String!
 }
 
+type CurrencyConversionEstimate {
+  """Amount in satoshis."""
+  btcSatAmount: SatAmount!
+  currency: Currency!
```
nicolasburtey, line 344: should we also return the displayCurrency? that we have as input? for cache purpose for instance

dev/config/apollo-federation/supergraph.graphql
nicolasburtey: how about:
```
walletCurrencyAmount { 
    amount: Float!
    currency: DisplayCurrency
}
```

return
```
type WalletCurrencyAmount
  @join__type(graph: PUBLIC)
{
  btcSatAmount: SatAmount!
  usdCentAmount: CentAmount!

  currency: Currency!
  id: ID!
  timestamp: Timestamp!
}
```

core/api/src/graphql/public/root/query/currency-conversion-estimate.ts
```diff
@@ -0,0 +1,80 @@
...
+  type: GT.NonNull(CurrencyConversionEstimate),
+  description: `Returns an estimated conversion rate for the given amount and currency`,
+  args: {
+    amount: {
+      type: GT.NonNull(GT.Float),
+      description: "Amount in major unit.",
+    },
+    currency: {
+      type: DisplayCurrencyGT,
+      defaultValue: UsdDisplayCurrency,
```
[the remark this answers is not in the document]
dolcalmi: Removed https://github.com/GaloyMoney/galoy/pull/4222/commits/bc722bc1d5c06b0a98c9c32669b084091f774c82#diff-e020ba9cbe96258b7d6050b148cda2072d97e8227adf1c25aae25515bbffff2eR24

core/api/src/graphql/public/schema.graphql
dolcalmi: It is not necessary. Removed https://github.com/GaloyMoney/galoy/pull/4222/commits/bc722bc1d5c06b0a98c9c32669b084091f774c82

dev/config/apollo-federation/supergraph.graphql
dolcalmi: Updated according to ACE meeting https://github.com/GaloyMoney/galoy/pull/4222/commits/bc722bc1d5c06b0a98c9c32669b084091f774c82

nicolasburtey:
LGTM (beyond tests not passing)

nicolasburtey (approved):

PR #4239, opened 2024-03-26
feat(toolchain): include local env in serve_env load

vindard:
## Description

We ran across a situation where someone could want to add secrets to `.env.local` and have them get to the api environment that's brought up in tilt. This is an attempt to address this.


nicolasburtey (approved):

nicolasburtey:
can I merge it? I'll see if this work on my branch

PR #4248, opened 2024-03-28
chore(pay): move boldcard to sidebar

siddhart1o1:
image

### DIFF
--- a/apps/pay/components/parse-pos-payment/index.tsx
+++ b/apps/pay/components/parse-pos-payment/index.tsx
@@ -20,7 +20,6 @@ import Memo from "../memo"
 import DigitButton from "./digit-button"
 import styles from "./parse-payment.module.css"
 import ReceiveInvoice from "./receive-invoice"
-import NFCComponent from "./nfc"
 
 import { satCurrencyMetadata } from "@/app/sats-currency-metadata"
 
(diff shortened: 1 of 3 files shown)


nicolasburtey:
I'm not sure this should be a toggle on or off. I think it should rather be a button to "Activate boltcard" and then if it's activated we could be showing "boltcard activated"

basically the reason we need this is because the browser needs a client led action for it to ask for NFC authorization. a developer can't just access NFC without the browser haivng already approve the access to NFC. and the only way to ask for approval is to respond to an action. 

but once the approval has been made, there is no reason to deactivate it back.

siddhart1o1:
> I'm not sure this should be a toggle on or off. I think it should rather be a button to "Activate boltcard" and then if it's activated we could be showing "boltcard activated"
> 
> basically the reason we need this is because the browser needs a client led action for it to ask for NFC authorization. a developer can't just access NFC without the browser haivng already approve the access to NFC. and the only way to ask for approval is to respond to an action.
> 
> but once the approval has been made, there is no reason to deactivate it back.


that implementation would have disabled the payment acceptance feature if it was toggled off after being enabled and receiving permission, but it has been updated to a button.

nicolasburtey:
> that implementation would have disabled the payment acceptance feature (from boltcards) if it was toggled off after being enabled and receiving permission, but it has been updated to a button.

yeah I understand this, but my point is that I don't see a scenario where someone would want to do this

PR #4259, opened 2024-04-02
chore(core): fix warning in spell check 

thevaibhav-dixit:


core/api/src/domain/quiz/config.ts
```diff
@@ -11,10 +11,10 @@ export const QuizzesValue: Record<QuizQuestionId, Satoshis> = {
   moneyEvolution: 2 as Satoshis,
   whyStonesShellGold: 3 as Satoshis,
   moneyIsImportant: 3 as Satoshis,
-  moneyImportantGovernement: 4 as Satoshis,
+  moneyImportantGovernment: 4 as Satoshis,
```
nicolasburtey, line 14: I don't think this should be changed. this string is also recorded in the db. 
I think we should add an exemption for this word instead

nicolasburtey (changes requested):
it's better to not change `moneyImportantGovernment` otherwise we'll need to do migration. let's add that to the "skip list"

core/api/src/domain/quiz/config.ts
```diff
@@ -11,10 +11,10 @@ export const QuizzesValue: Record<QuizQuestionId, Satoshis> = {
...
   WhatIsFiat: 2 as Satoshis,
   whyCareAboutFiatMoney: 2 as Satoshis,
-  GovernementCanPrintMoney: 2 as Satoshis,
+  GovernmentCanPrintMoney: 2 as Satoshis,
```
thevaibhav-dixit, line 17: So we also leave this as is ? @nicolasburtey

nicolasburtey: yes

nicolasburtey (changes requested):
still need to revert the changes on core/api/src/domain/quiz/config.ts and core/api/src/domain/quiz/sections.ts

typos.toml
```diff
@@ -23,3 +25,5 @@ default.extend-ignore-re = ["\"eyJ[a-zA-Z0-9_\\-\\.]+\""]
 Ons = "Ons"
 Lsat = "Lsat"
 quizs = "quizs"
+Aggrement = "Aggrement"
```
bodymindarts, line 28: I don't think we should allow this typo across the whole code base when it is only present in 1 file. Can we make a narrower exclude?

bodymindarts (approved):

nicolasburtey (approved):

PR #4261, opened 2024-04-03
chore: moving galoy-voucher to monorepo

siddhart1o1:
## Changes and refactoring 
+ Added `Sign in with blink`
+ Remove unfunded links
+ Direct fund voucher with Intraledger


nicolasburtey (approved):

apps/voucher/app/create/page.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const [commissionPercentage, setCommissionPercentage] = useState<string>(
+    storedCommission || "0",
+  )
+  const { currencyToCents, hasLoaded } = useRealtimePrice(currency.id)
+  const [amount, setAmount] = useState<string>("0")
+
+  const [confirmModal, setConfirmModal] = useState<boolean>(false)
+  const [currentPage, setCurrentPage] = useState<string>("AMOUNT")
+
+  const AmountInDollars = Number(
```
nicolasburtey, line 56:
```suggestion
  const amountInDollars = Number(
```

apps/voucher/app/create/page.tsx
```diff
@@ -0,0 +1,126 @@
...
+    typeof window !== "undefined" ? localStorage.getItem("commission") : null
+
+  const [currency, setCurrency] = useState<Currency>(
+    storedCurrency ? JSON.parse(storedCurrency) : DEFAULT_CURRENCY,
+  )
+  const [commissionPercentage, setCommissionPercentage] = useState<string>(
+    storedCommission || "0",
+  )
+  const { currencyToCents, hasLoaded } = useRealtimePrice(currency.id)
+  const [amount, setAmount] = useState<string>("0")
```
nicolasburtey, line 51: string? shouldn't it be a number?

apps/voucher/app/page.tsx
```diff
@@ -0,0 +1,29 @@
+"use client"
+import Link from "next/link"
+
+import Button from "@/components/Button/Button"
+import { useSession } from "next-auth/react"
+// path : /
+//TODO home page or welcome screen needed to be created here
+export default function Home() {
+  const session = useSession()
+  console.log(session)
```
nicolasburtey, line 10:
```suggestion
```

apps/voucher/hooks/useCreateInvoice.ts
```diff
@@ -0,0 +1,67 @@
+// import { useState } from "react"
```
nicolasburtey, line 1: let's delete unused file

apps/voucher/hooks/useRealTimePrice.ts
```diff
@@ -0,0 +1,217 @@
+import React, { useState, useEffect } from "react"
```
nicolasburtey, line 1: maybe with this PR merged we don't need this file anymore

https://github.com/GaloyMoney/galoy/commit/14b56dc3bf8ae84628568b406e2d11c1e2ab65c4

apps/voucher/pages/api/lnurlw/[uniqueHash]/index.ts
```diff
@@ -0,0 +1,50 @@
+import { getWithdrawLinkByUniqueHashQuery } from "@/services/db"
+import { convertCentsToSats } from "@/utils/helpers"
+import { env } from "@/env"
+import { getRealtimePriceQuery } from "@/services/galoy/query/get-real-time-price"
+import { escrowApolloClient } from "@/services/galoy/client/escrow"
+const { NEXT_PUBLIC_LOCAL_URL } = env
+// TODO: Add interface for request and response objects
+
+export default async function handler(req: any, res: any) {
+  if (req.method === "GET") {
```
Comment (nicolasburtey) [inline apps/voucher/pages/api/lnurlw/[uniqueHash]/index.ts:10]: instead of doing this, we should do:

```
if (req.method !== "GET") {
  return // return some error
}
```

like this we don't need to intent this function all the way to the bottom. we also exit early with the error which make the code easier to read

apps/voucher/pages/api/lnurlw/callback/[id]/index.ts
```diff
@@ -0,0 +1,125 @@
...
+import { env } from "@/env"
+import { PaymentSendResult, Status, Wallet } from "@/lib/graphql/generated"
+import { escrowApolloClient } from "@/services/galoy/client/escrow"
+import { fetchUserData } from "@/services/galoy/query/me"
+import { getRealtimePriceQuery } from "@/services/galoy/query/get-real-time-price"
+import { lnInvoicePaymentSend } from "@/services/galoy/mutation/ln-invoice-payment-send"
+const { NEXT_PUBLIC_GALOY_URL } = env
+
+export default async function handler(req: any, res: any) {
+  if (req.method === "GET") {
```
Comment (nicolasburtey) [inline apps/voucher/pages/api/lnurlw/callback/[id]/index.ts:15]: same here

apps/voucher/app/create/page.tsx
siddhart1o1: Yes, but we have some numpad validations that require it to be in a string, but we can do refactoring there to make int number.

apps/voucher/pages/api/lnurlw/[uniqueHash]/index.ts
Comment (siddhart1o1) [reply apps/voucher/pages/api/lnurlw/[uniqueHash]/index.ts]: Shifted to app router, not needed anymore.

apps/voucher/pages/api/lnurlw/callback/[id]/index.ts
Comment (siddhart1o1) [reply apps/voucher/pages/api/lnurlw/callback/[id]/index.ts]: Shifted to app router, not needed anymore.

apps/voucher/components/create/conifrm-modal/index.tsx
```diff
@@ -0,0 +1,200 @@
+import React, { useState, MouseEvent } from "react"
```
nicolasburtey, line 1: type in filename

nicolasburtey: let's also get lowercase filenames

apps/voucher/components/funds-paid.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react"
```
nicolasburtey, line 1: `apps/voucher/components/currency-drop-down.tsx` is empty

apps/voucher/config/wdio.conf.ts
```diff
@@ -0,0 +1,38 @@
+export const config = {
```
nicolasburtey, line 1: `apps/voucher/config/seeds/initial.ts` is empty

apps/voucher/services/galoy/client/escrow/index.ts
```diff
@@ -0,0 +1,23 @@
+import { ApolloClient, HttpLink, InMemoryCache } from "@apollo/client"
+import { env } from "@/env"
+
+export const escrowApolloClient = () => {
+  return new ApolloClient({
+    cache: new InMemoryCache(),
+    link: new HttpLink({
+      headers: {
+        "Authorization": `Bearer ${env.ESCROW_TOKEN}`,
```
nicolasburtey, line 9: I think it should be an API token instead of a kratos token (different header)

apps/voucher/graphql/schema.ts
```diff
@@ -0,0 +1,72 @@
...
+
+enum Status {
+  ACTIVE
+  PENDING 
+  PAID
+}
+
+type WithdrawLink {
+  id: ID!
+  userId: ID!
```
nicolasburtey, line 11: what does the userId represents now?

apps/voucher/graphql/schema.ts
```diff
@@ -0,0 +1,72 @@
...
+
+input CreateWithdrawLinkInput {
+  voucherAmountInCents: Float!
+  walletId: ID!
+  commissionPercentage: Float
+}
+
+input OnChainWithdrawLinkInput {
+  voucherSecret: String!
+  btcWalletAddress: String!
```
nicolasburtey, line 68: it is supposed to be an onchain address?

apps/voucher/graphql/schema.ts
```diff
@@ -0,0 +1,72 @@
...
+
+
+type Query {
+  getWithdrawLink(voucherSecret: String): WithdrawLinkWithSecret
+  getWithdrawLinksByUserId(status: Status, limit: Int, offset: Int): WithdrawLinksByUserIdResult!
+}
+
+type Mutation {
+  createWithdrawLink(input: CreateWithdrawLinkInput!): WithdrawLinkWithSecret!
+  onChainWithdrawLink(input: OnChainWithdrawLinkInput!): onChainWithdrawResult!
```
nicolasburtey, line 57: we need consistency on the pattern of the argument and result.

apps/voucher/graphql/schema.ts
```diff
@@ -0,0 +1,72 @@
...
+  salesAmountInCents: Float!
+  identifierCode: String!
+  voucherSecret: String!
+  commissionPercentage: Float!
+  uniqueHash: String!
+}
+
+type WithdrawLinksByUserIdResult {
+  withdrawLinks: [WithdrawLink!]!
+  totalLinks: Int
```
nicolasburtey, line 37: I don't think there is a need for that unless we do pagination. 
if we do pagination, we should use the same pattern as the backend

apps/voucher/test/e2e/specs/create-voucher-flow.e2e.ts
```diff
@@ -0,0 +1,101 @@
...
+        "body > div.top_page_container > div.NumPad_numpad__yb7Im > div > button:nth-child(1)",
+      )
+      await percentage.click()
+
+      const set_commission = await $(
+        "body > div.top_page_container > div.CreateLink_commission_and_submit_buttons__ydXyN > button",
+      )
+      await set_commission.click()
+
+      // const set_commission = await $("body > div.top_page_container > div:nth-child(4)")
```
nicolasburtey, line 75:
```suggestion
```

apps/voucher/test/e2e/specs/create-voucher-flow.e2e.ts
```diff
@@ -0,0 +1,101 @@
...
+      const total_ammoutn = await $(
+        "body > div.top_page_container > div:nth-child(5)",
+      ).getText()
+      expect(total_ammoutn).toEqual("≈ $4.95")
+
+      await createPage.clickSubmitButton()
+      await confirmModal.clickConfirmButton()
+      const notFundedText = await linkDetailsPage.getNotFundedText()
+
+      expect(notFundedText).toEqual("Not Funded")
```
nicolasburtey, line 95: are we not doing the happy path for the test?

ie: doing the funding an then the redeeming?

apps/voucher/utils/helpers.ts
```diff
@@ -0,0 +1,150 @@
+import { v4 as uuidv4 } from "uuid"
```
nicolasburtey, line 1: `apps/voucher/types/light-bolt11-decoder.d.ts` is another empty file

apps/voucher/graphql/schema.ts
siddhart1o1: User who is creating the voucher

apps/voucher/graphql/schema.ts
siddhart1o1: yes

apps/voucher/app/create/page.tsx
```diff
@@ -0,0 +1,126 @@
...
+    AmountInDollars,
+    commissionPercentage,
+  )
+
+  if (!session?.data?.userData?.me?.defaultAccount.wallets) {
+    return null
+  }
+
+  const { btcWallet, usdWallet } = getWalletDetails({
+    wallets: session?.data?.userData?.me?.defaultAccount?.wallets as Wallet[],
```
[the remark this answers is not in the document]
siddhart1o1: Yes, because we are not using all wallet properties only these.
```
accountId
balance
id
pendingIncomingBalance
walletCurrency
```

apps/voucher/test/e2e/specs/create-voucher-flow.e2e.ts
siddhart1o1: These are old wdio tests. We will replace them with Cypress in another PR.

apps/voucher/hooks/useRealTimePrice.ts
siddhart1o1: Maybe we need ? We are using subscription API here as well.

apps/voucher/graphql/schema.ts
siddhart1o1: Removed pagination for now.

apps/voucher/test/e2e/specs/create-voucher-flow.e2e.ts
siddhart1o1: These are old wdio tests. We will replace them with Cypress in another PR.

apps/voucher/app/api/lnurlw/[uniqueHash]/route.ts
```diff
@@ -0,0 +1,55 @@
+import { getWithdrawLinkByUniqueHashQuery } from "@/services/db"
```
Comment (nicolasburtey) [inline apps/voucher/app/api/lnurlw/[uniqueHash]/route.ts:1]: filename uppercase

apps/voucher/app/create/page.tsx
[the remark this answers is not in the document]
nicolasburtey: I think then you should defined a sub type for it

apps/voucher/graphql/schema.ts
nicolasburtey: does it map to the core UserId or core AccountId?

apps/voucher/hooks/useRealTimePrice.ts
nicolasburtey: if we are using the subcription API to get the price, I think we don't need that anymore with the new query above

siddhart1o1: Not sure if it's ready to use.
```
curl 'http://localhost:4455/graphql' \
  -H 'Accept-Language: en-US,en;q=0.9,ko;q=0.8' \
  -H 'Connection: keep-alive' \
  -H 'DNT: 1' \
  -H 'Origin: http://localhost:3006' \
  -H 'Referer: http://localhost:3006/' \
  -H 'Sec-Fetch-Dest: empty' \
  -H 'Sec-Fetch-Mode: cors' \
  -H 'Sec-Fetch-Site: same-site' \
  -H 'User-Agent: Mozilla/5.0 (iPhone; CPU iPhone OS 16_6 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.6 Mobile/15E148 Safari/604.1' \
  -H 'accept: */*' \
  -H 'content-type: application/json' \
  --data-raw $'{"operationName":"CurrencyConversionEstimation","variables":{"amount":68.85,"currency":"USD"},"query":"query CurrencyConversionEstimation($amount: Float\u0021, $currency: DisplayCurrency\u0021) {\\n  currencyConversionEstimation(amount: $amount, currency: $currency) {\\n    btcSatAmount\\n    id\\n    timestamp\\n    usdCentAmount\\n    __typename\\n  }\\n}"}'
{"data":null,"errors":[{"message":"Unexpected error occurred, please try again or contact support if it persists (code: BigIntFloatConversionError: 6884.999999999999)","locations":[{"line":1,"column":89}],"path":["currencyConversionEstimation"]}]}%                                                            
```

Also, is it the best option for the frontend? Every amount change calls an API, which takes a few milliseconds to update the amount. It does not look as smooth on the frontend, maybe better for backend ?

nicolasburtey: @dolcalmi is there an issue with `CurrencyConversionEstimation`?

nicolasburtey: @siddhart1o1 it's made for frontends. the idea is that you want to have the amount and then you do the conversation.

siddhart1o1: image

Still not sure if this is the right query to use here, as we are displaying the voucher amount `(voucher_amount = input amount - commission)` in USD. For this, we convert the user's currency to USD, for example, EUR is converted to USD. 
The voucher amount is updated with every change in the input amount (numpad click), so we need to calculate the voucher amount on every click, i.e., convert the user amount to USD.
And as this voucher amount is updated with every Numpad click. If we use this new query here, every change/click in numpad will cause an API call. Whereas, the `useRealTimePrice` hook provides base and offset values which we are using to calculate the price on the client and these offeset values are updated by the subscription API when needed.

Anyway, I've updated this to use `CurrencyConversionEstimation` maybe I'm over-optimizing it or missing something.

apps/voucher/graphql/schema.ts
siddhart1o1: core userid: `userData.me.id`

nicolasburtey (approved):

PR #4262, opened 2024-04-03
feat: add `externalId` field to lightning invoices & received transactions

vindard:
## Description

This PR is to allow setting an `externalId` field on lightning-received payments via the invoice creation.

### Implementation notes

`externalId` field was added to both `Invoices` and `Transactions` collections:
- uniqueness is enforced at the `Invoices` collection level, external ids are only added via transactions from invoices so far
- new field was added directly to transactions collection because of complications with querying if it lived in metadata collection

Queries:
- new queries were added for both transactions and invoices at wallet level
- queries support taking a substring and searching by external id pattern

Validation:
- regex used that support case-insensitive alphanumerics, '_' and '-' characters
- max length was (arbitrarily) set to 100 chars so far

### `TODO`
- [x] enforce unique `externalId` for WalletInvoices
- [x] add tests for externalId invoice uniqueness
- [x] add `externalId` field handling transactions in LedgerService (https://github.com/GaloyMoney/galoy/pull/4262/commits/cbdc884504ade901b49581dba6370117b6d074f8)
- [x] pass `externalId` from `WalletInvoice` to `LedgerTransaction` in `updatePendingInvoice` function (https://github.com/GaloyMoney/galoy/pull/4262/commits/d6f62b766553c01df47920510a28da4ad9561f6d)
- [x] translate `externalId` from `LedgerTransaction` to `WalletTransaction`
- [x] check (test) that `externalId` is present in places like callback via `walletTransaction`
- [x] add new `transactionsByExternalId` query
- [x] ~facilitate reading `externalId` via dataloader~ move externalId to transactions collection
- [x] refactor transactionsByExternalId (and make it paginated transaction-connection)
- [x] add e2e tests for `transactionsByExternalId`
- [x] implement `externalId` validation
- [x] (optional) add invoiceByExternalId query
- [x] add a `TxExternalIdSubstring` graphql type
- [x] ~(optional) extend uniqueness check to Ledger collection persist function?~ postpone for now


core/api/src/app/wallets/add-invoice-for-wallet.ts
```diff
@@ -35,6 +37,7 @@ const addInvoiceForSelf = async ({
       recipientWalletDescriptor,
     }: BuildWIBWithAmountFnArgs) =>
       walletInvoiceBuilder
+        .withMetadata(externalId)
```
bodymindarts, line 40: this should be `withExternalId`

core/api/src/domain/wallet-invoices/wallet-invoice-builder.ts
```diff
@@ -13,6 +13,19 @@ import { toSeconds } from "@/domain/primitives"
 export const WalletInvoiceBuilder = (
   config: WalletInvoiceBuilderConfig,
 ): WalletInvoiceBuilder => {
+  const withMetadata = (externalId?: LedgerExternalId) => {
```
bodymindarts, line 16: `withExternalId`

core/api/src/services/ledger/schema.ts
```diff
@@ -172,13 +172,13 @@ const transactionMetadataSchema = new Schema<TransactionMetadataRecord>(
...
+    external_id: String,
   },
   { id: false },
 )
 
-transactionMetadataSchema.index({
-  hash: 1,
-})
+transactionMetadataSchema.index({ hash: 1 })
+transactionMetadataSchema.index({ external_id: 1 })
```
bodymindarts, line 181: Index needs to enforce uniqueness

core/api/src/app/wallets/add-invoice-for-wallet.ts
vindard: Yup I was wondering about whether to "future-prep" this step to take both the external id and a json blob if we decided to add that later. But having external id be its own step makes sense.

bodymindarts: Yeah I had the same thought - we can use 1 fn for both `{externalId, metadataJson}` that's fine. I also don't mind the class name for that state being `WithMetadata` but until we add the Json blob I would still rename the fn.

vindard: Changed in [`force-push`](https://github.com/GaloyMoney/galoy/compare/707494475338d9ad0ebd6ebdeb47d2d16b830b1d..bce2c352786c18b90164c1b3c2519c25c489639e)

core/api/src/domain/wallet-invoices/wallet-invoice-builder.ts
vindard: https://github.com/GaloyMoney/galoy/pull/4262#discussion_r1558054786

core/api/src/services/ledger/schema.ts
vindard: Field lives in `transactions` and `invoices` collections now, uniqueness enforced at `invoices` persistence level, we won't be able to do the same for transactions since we'd have multiple accounting entries for each external id

core/api/src/app/wallets/get-invoices-for-wallets.ts
```diff
@@ -25,3 +26,34 @@ export const getInvoicesForWallets = async ({
     paginationArgs,
   })
 }
+
+export const getInvoicesForWalletsByExternalIdSubstring = async ({
```
bodymindarts, line 30: Is this a fuzzy search for a part of an external id? Is this efficient on DB level?

vindard: It's to potentially support a user who wants to segregate use-cases by id for a single wallet. E.g. `service1-<uuid>`, `service2-<uuid>` etc.

It won't be as efficient as an exact match, but from testing on signet db it takes around 1s.

Now that I think about it though, we probably should explicitly signal the param for a fuzzy search on the API and not overload the exact search  if we decide to keep this in.

vindard: I opted to remove these methods since they're not straightforward to add to the API in a clean way and the exact need isn't known as yet. We can potentially revisit later https://github.com/GaloyMoney/galoy/pull/4262/commits/34dfa15a9964048e8bb4b7ddd80cc11afe5d770f

bodymindarts (approved):

PR #4271, opened 2024-04-07
Auto-set display currency for level 1 users based on phone number prefix

### LINKED ISSUE #4252: Auto-set display currency for level 1 user based on phone number prefix
Currently, display currency defaults to USD for all users. 

It would be better if app auto-set display currency based on phone number prefix

Example: 
New user with +90 (Turkey) creates new account  -> display currency is set to TRY
New user with +49 (Germany) creates new account  -> display currency is set to EUR
etc.

Prakhar-Agarwal-byte:
Fixes #4252 

Matches the `countryCode` in `phoneMetadata` received from Twilio with currency using [country-to-currency](https://www.npmjs.com/package/country-to-currency) package and then updates the default display currency for the new user

### DIFF
--- a/core/api/src/app/accounts/create-account.ts
+++ b/core/api/src/app/accounts/create-account.ts
@@ -9,14 +9,18 @@ import {
   WalletsRepository,
 } from "@/services/mongoose"
 
+import countryToCurrency, { Currencies, Countries } from "country-to-currency"
+
 const initializeCreatedAccount = async ({
   account,
   config,
   phone,
+  phoneMetadata,
 }: {
   account: Account
   config: AccountsConfig
   phone?: PhoneNumber
+  phoneMetadata?: PhoneMetadata
 }): Promise<Account | ApplicationError> => {
   const walletsEnabledConfig = config.initialWallets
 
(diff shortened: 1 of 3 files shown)


core/api/src/app/accounts/create-account.ts
```diff
@@ -51,6 +55,11 @@ const initializeCreatedAccount = async ({
   account.statusHistory = [{ status: config.initialStatus, comment: "Initial Status" }]
   account.level = config.initialLevel
 
+  // Set display currency to match phone numbers country code
+  const countryCode = (phoneMetadata?.countryCode || "US") as Countries
+  const currency = countryToCurrency[countryCode]
+  account.displayCurrency = currency as DisplayCurrency
```
dolcalmi, line 61: we need to validate against our list of currencies (we dont support all the currencies)

dolcalmi (changes requested):

vindard:
@Prakhar-Agarwal-byte thanks for the contribution! I just updated the branch to remove some merge conflicts. Will you be continuing with this PR? There's some feedback from @dolcalmi to address

vindard:
Closing in favour of a different approach we took to tackle this in #4309 and https://github.com/GaloyMoney/price/pull/67

PR #4277, opened 2024-04-09
fix(core): add account deletion limit validation

dolcalmi:


core/api/src/app/accounts/mark-account-for-deletion.ts
```diff
@@ -43,11 +49,20 @@ export const markAccountForDeletion = async ({
...
   const user = await usersRepo.findById(kratosUserId)
   if (user instanceof Error) return user
 
+  const deletedPhones: PhoneNumber[] = user.phone ? [user.phone] : []
+  if (user.deletedPhones) {
+    deletedPhones.push(...user.deletedPhones)
+  }
+  const usersByPhones = await usersRepo.findByDeletedPhones(deletedPhones)
+  if (usersByPhones instanceof Error) return usersByPhones
+  if (usersByPhones.length >= maxDeletions) return new InvalidAccountForDeletionError()
```
nicolasburtey, line 64:
```suggestion
  if (usersByPhones.length > maxDeletions) return new InvalidAccountForDeletionError()
```

I think like this, with `maxDeletions === 2` would be more understandable that it's only possible to delete an account 2 times with the same phone number

dolcalmi: I changed to 2 the default value but if I change validation to use `>` instead of `>=` a user will be able to delete the account 3 times

nicolasburtey (approved):

PR #4289, opened 2024-04-14
feat: using vector store for improving knowledge base

nicolasburtey:


core/api/src/services/openai/assistant.ts
```diff
@@ -4,6 +4,9 @@ import { env } from "@/config/env"
 import { sleep } from "@/utils"
 import { UnknownDomainError } from "@/domain/shared"
 import { UnknownChatAssistantError } from "@/domain/support/errors"
+import assert from "assert"
+import { textToVector } from "./embeddings"
+import { retrieveEmbeddings } from "../pinecone"
```
nicolasburtey, line 9: is this fine?

core/api/src/services/pinecone/index.ts
```diff
@@ -0,0 +1,31 @@
+import { Pinecone } from "@pinecone-database/pinecone"
```
bodymindarts, line 1: I would nest this under the openai service. I think treating the 'assistant' service like a domain service (that wraps multiple sub-components) is valid here as the interactions are more complex than a 1-to-1 mapping between some external service calls and an internal representation.

PR #4291, opened 2024-04-14
feat(notifications): bitcoin price down notification

thevaibhav-dixit:
- [ ] Add Missing Spanish Translation

### DIFF
--- a/core/notifications/locales/es.yml
+++ b/core/notifications/locales/es.yml
@@ -58,8 +58,8 @@ transaction.paid_invoice.title: "%{walletCurrency} Transacción"
 transaction.paid_invoice.body: "+%{formattedCurrencyAmount}"
 transaction.paid_invoice.body_display_currency: "+%{displayCurrencyAmount} | %{formattedCurrencyAmount}"
 
-price_changed.title: "Bitcoin está en movimento!"
-price_changed.body: "Bitcoin ha augmentado %{percent_increase}% en el ultimo dia %{price}!"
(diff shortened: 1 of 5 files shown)


core/notifications/locales/es.yml
```diff
@@ -58,8 +58,8 @@ transaction.paid_invoice.title: "%{walletCurrency} Transacción"
 transaction.paid_invoice.body: "+%{formattedCurrencyAmount}"
 transaction.paid_invoice.body_display_currency: "+%{displayCurrencyAmount} | %{formattedCurrencyAmount}"
 
-price_changed.title: "Bitcoin está en movimento!"
-price_changed.body: "Bitcoin ha augmentado %{percent_increase}% en el ultimo dia %{price}!"
+price_changed.up.title: "Bitcoin está en movimento!"
+price_changed.up.body: "Bitcoin ha augmentado %{change_percent}% en el ultimo dia %{price}!"
```
nicolasburtey, line 62: Why are we not using transifex for notifications?

nicolasburtey (commented):
Question

core/notifications/locales/es.yml
thevaibhav-dixit: We do use transifex. I think we can simply add the Spanish translation in this pr itself. Even if we do not then it will default to English.

thevaibhav-dixit: Decided that it can come in via transifex only.

PR #4294, opened 2024-04-15
chore(pay): previous transaction print

siddhart1o1:
## Added print for previous transactions.

Previously, we were not passing the ID because the subscription API does not provide an ID and createdAt in the response after a successful payment. Therefore, for an instant receipt that comes after payment, use the payment hash and date/time of printing.
However, we get the ID and createdAt in previous transactions; therefore, we are using the ID and createdAt time of transaction.
If the transaction has a transaction ID and a hash, both will be printed; otherwise, only the ID (as this is available for all transactions).
### Changes in POS companion app to take date, time, and ID as input:
POS COMPANION PR: https://github.com/GaloyMoney/pos-print-companion/pull/11

EDIT: for better understanding. 
On the Transaction Outcome page (where we receive a green tick), we are not getting the ID and createdAt because we are using the subscription API there. However, here we are receiving these fields here because we are using the "me" query, so we are passing these fields to the POS companion.


nicolasburtey:
> Previously, we were not passing the ID because the subscription API does not provide an ID and createdAt in the response after a successful payment.

should we add that? @siddhart1o1 @dolcalmi

siddhart1o1:
> > Previously, we were not passing the ID because the subscription API does not provide an ID and createdAt in the response after a successful payment.
> 
> should we add that? @siddhart1o1 @dolcalmi


This PR is ready to review if the subscription API is updated. Changes will be on the transaction outcome page and are not related to this.

dolcalmi:
> > > Previously, we were not passing the ID because the subscription API does not provide an ID and createdAt in the response after a successful payment.
> > 
> > 
> > should we add that? @siddhart1o1 @dolcalmi
> 
> This PR is ready to review if the subscription API is updated. Changes will be on the transaction outcome page and are not related to this.

what does this mean? didnt you test it? also you are using `fetchedData.data.me.defaultAccount.transactions` so seems it is the authed query/subscription

dev/bin/add-test-users-with-usernames.sh
```diff
@@ -41,8 +44,13 @@ echo "Logging in user $user_c"
 auth_token_c=$(login_user "${user_c}")
 echo "Auth token for $user_c: $auth_token_c"
 
+echo "Logging in user $user_d"
+auth_token_d=$(login_user "${user_d}")
+echo "Auth token for $user_d: $auth_token_d"
```
dolcalmi, line 49: why did you add this? cant you use other users?

apps/pay/app/print-companion-service.ts
```diff
@@ -0,0 +1,27 @@
...
+  let deepLinkUrl = `blink-pos-companion://print?username=${encodeURIComponent(
+    username,
+  )}&amount=${encodeURIComponent(amount)}`
+
+  if (paymentHash) deepLinkUrl += `&paymentHash=${encodeURIComponent(paymentHash)}`
+  if (transactionId) deepLinkUrl += `&id=${encodeURIComponent(transactionId)}`
+  if (date) deepLinkUrl += `&date=${encodeURIComponent(date)}`
+  if (time) deepLinkUrl += `&time=${encodeURIComponent(time)}`
+
+  console.log(deepLinkUrl)
```
dolcalmi, line 25: please remove

apps/pay/app/[username]/transaction/page.tsx
```diff
@@ -77,11 +80,35 @@ function TransactionsPage() {
...
+        if ("getInstalledRelatedApps" in navigator) {
+          /* eslint @typescript-eslint/ban-ts-comment: "off" */
+          // @ts-ignore-next-line no-implicit-any error
+          const apps = await navigator.getInstalledRelatedApps()
+          const isAppInstalled = apps.some(
+            (app: { id: string }) => app.id === "com.blink.pos.companion",
+          )
+          setPosCompanionInstalled(isAppInstalled)
+        } else {
+          console.error("getInstalledRelatedApps is not supported in this browser.")
```
Comment (dolcalmi) [inline apps/pay/app/[username]/transaction/page.tsx:102]: what is the purpose on this? dont we want to inform to the user what is happening?

apps/pay/app/[username]/transaction/page.tsx
```diff
@@ -77,11 +80,35 @@ function TransactionsPage() {
...
+        }
+      } catch (error) {
+        console.error("Error checking installed related apps:", error)
```
Comment (dolcalmi) [inline apps/pay/app/[username]/transaction/page.tsx:105]: dont we have  log library in galoy pay? using console.log/error is not good

apps/pay/app/[username]/transaction/page.tsx
```diff
@@ -77,11 +80,35 @@ function TransactionsPage() {
...
     variables: { first: 10 },
   })
 
+  useEffect(() => {
+    const checkRelatedApps = async () => {
+      try {
+        if ("getInstalledRelatedApps" in navigator) {
+          /* eslint @typescript-eslint/ban-ts-comment: "off" */
+          // @ts-ignore-next-line no-implicit-any error
+          const apps = await navigator.getInstalledRelatedApps()
```
Comment (dolcalmi) [inline apps/pay/app/[username]/transaction/page.tsx:96]: cant this be fixed with a declaration?

dolcalmi (changes requested):

dolcalmi:
also.. this PR needs tests

siddhart1o1:
> > > > Previously, we were not passing the ID because the subscription API does not provide an ID and createdAt in the response after a successful payment.
> > > 
> > > 
> > > should we add that? @siddhart1o1 @dolcalmi
> > 
> > 
> > This PR is ready to review if the subscription API is updated. Changes will be on the transaction outcome page and are not related to this.
> 
> what does this mean? didnt you test it? also you are using `fetchedData.data.me.defaultAccount.transactions` so seems it is the authed query/subscription

Not sure if you understood the current state.
On the Transaction Outcome page (where we receive a green tick after payment is received), we are not getting the ID and createdAt because we are using the subscription API there. However, here (previous transactions) we are receiving these fields here because we are using the "me" query, so we are passing these fields to the POS companion.

siddhart1o1:
> also.. this PR needs tests

We currently don't have a way to conduct automatic testing for the Pay with POS companion app. We are doing manual testing for now.

dev/bin/add-test-users-with-usernames.sh
siddhart1o1: This user is funded; this is the same user I am using in the fund-user script. I've added the username for that, helpful while developing.

apps/pay/app/[username]/transaction/page.tsx
Comment (siddhart1o1) [reply apps/pay/app/[username]/transaction/page.tsx:105]: No, I don't think we have a log library in Pay.

apps/pay/app/[username]/transaction/page.tsx
Comment (siddhart1o1) [reply apps/pay/app/[username]/transaction/page.tsx:102]: The purpose is to detect if the POS companion app is installed and show the print button if it is installed.

dolcalmi:
> > also.. this PR needs tests
> 
> We currently don't have a way to conduct automatic testing for the Pay with POS companion app. We are doing manual testing for now.

why not? What is preventing it?

dolcalmi:
> > > > > Previously, we were not passing the ID because the subscription API does not provide an ID and createdAt in the response after a successful payment.
> > > > 
> > > > 
> > > > should we add that? @siddhart1o1 @dolcalmi
> > > 
> > > 
> > > This PR is ready to review if the subscription API is updated. Changes will be on the transaction outcome page and are not related to this.
> > 
> > 
> > what does this mean? didnt you test it? also you are using `fetchedData.data.me.defaultAccount.transactions` so seems it is the authed query/subscription
> 
> Not sure if you understood the current state. On the Transaction Outcome page (where we receive a green tick after payment is received), we are not getting the ID and createdAt because we are using the subscription API there. However, here (previous transactions) we are receiving these fields here because we are using the "me" query, so we are passing these fields to the POS companion.

if you are  using me query then why not use the authed subscription and get the whole transaction?

apps/pay/app/[username]/transaction/page.tsx
Comment (dolcalmi) [reply apps/pay/app/[username]/transaction/page.tsx:105]: please add it, at least a wrapper and if we have otel integrate it please handle error properly

siddhart1o1:
> > > > > > Previously, we were not passing the ID because the subscription API does not provide an ID and createdAt in the response after a successful payment.
> > > > > 
> > > > > 
> > > > > should we add that? @siddhart1o1 @dolcalmi
> > > > 
> > > > 
> > > > This PR is ready to review if the subscription API is updated. Changes will be on the transaction outcome page and are not related to this.
> > > 
> > > 
> > > what does this mean? didnt you test it? also you are using `fetchedData.data.me.defaultAccount.transactions` so seems it is the authed query/subscription
> > 
> > 
> > Not sure if you understood the current state. On the Transaction Outcome page (where we receive a green tick after payment is received), we are not getting the ID and createdAt because we are using the subscription API there. However, here (previous transactions) we are receiving these fields here because we are using the "me" query, so we are passing these fields to the POS companion.
> 
> if you are using me query then why not use the authed subscription and get the whole transaction?



Because it is not necessary to be authenticated to use Blink Pay. Authentication is only necessary if the user wants to see previous transactions.

apps/pay/app/[username]/transaction/page.tsx
Comment (siddhart1o1) [reply apps/pay/app/[username]/transaction/page.tsx]: will look into it.

siddhart1o1:
> > > also.. this PR needs tests
> > 
> > 
> > We currently don't have a way to conduct automatic testing for the Pay with POS companion app. We are doing manual testing for now.
> 
> why not? What is preventing it?

Do we need e2e tests for this @nicolasburtey ?
To test it end-to-end, we will need an Android emulator with the POS companion installed in CI/CD, which will require some work. Even after this, we will need to test on a physical POS machine to see the print-out works. Maybe we can have some unit test.

nicolasburtey:
> > > also.. this PR needs tests
> > 
> > 
> > We currently don't have a way to conduct automatic testing for the Pay with POS companion app. We are doing manual testing for now.
> 
> why not? What is preventing it?

the set up is quite cumbersome. we need a companion app as an APK + the webview from the POS installed as a PWA

then, to really test it, we need a bitcoinize POS with a physical printer

nicolasburtey:
> Do we need e2e tests for this @nicolasburtey ?
> To test it end-to-end, we will need an Android emulator with the POS companion installed in CI/CD, which will require some work. Even after this, we will need to test on a physical POS machine to see the print-out works. Maybe we can have some unit test.

I agree it's ok to not have e2e test here. 
1/ it's not business critical code
2/ it's hard to set up in general as it's related to the integration between 2 apps on android
3/ it's hard to test as it depends on a physical printer

apps/pay/components/payment-outcome/index.tsx
```diff
@@ -16,6 +16,7 @@ import Receipt from "./receipt"
 
 import { extractSearchParams } from "@/utils/utils"
 import { useInvoiceContext } from "@/context/invoice-context"
+import useCheckInstalledApps from "@/hooks/use-check-Installed-apps"
```
nicolasburtey, line 19:
```suggestion
import useCheckInstalledApps from "@/hooks/use-check-installed-apps"
```

nicolasburtey: lowercase filename

nicolasburtey (approved):

PR #4314, opened 2024-04-19
fix: not using cache when price server is down

nicolasburtey:


dolcalmi:
closed in favor of https://github.com/GaloyMoney/galoy/pull/4315

PR #4318, opened 2024-04-19
feat: exposing portal for webhook

nicolasburtey:


core/api/src/app/callback/index.ts
```diff
@@ -54,3 +54,11 @@ export const deleteEndpoint = async ({
   if (success instanceof Error) return success
   return success
 }
+
+export const getPortal = async (accountId: AccountId) => {
+  const callbackService = CallbackService(getCallbackServiceConfig())
+  const res = await callbackService.getPortal(accountId)
+
+  if (res instanceof Error) return res
+  return res
```
dolcalmi, line 63:
```suggestion
  return await callbackService.getPortal(accountId)
```

core/api/src/services/svix/index.ts
```diff
@@ -113,7 +113,7 @@ export const CallbackService = (config: SvixConfig) => {
   }
 
   // only work for hosted svix
-  const getWebsocketPortal = async (accountId: AccountId) => {
+  const getPortal = async (accountId: AccountId) => {
```
dolcalmi, line 116: dont we have an interface for this? we must include return type

core/api/src/app/callback/index.ts
```diff
@@ -54,3 +54,11 @@ export const deleteEndpoint = async ({
   if (success instanceof Error) return success
   return success
 }
+
+export const getPortal = async (accountId: AccountId) => {
```
dolcalmi, line 58: should not be `getPortalUrl`? also please use `{accountId}: {accountId: AccountId}` this creates `code.function.params.accountId` in span attrs

core/api/src/services/svix/index.ts
```diff
@@ -113,7 +113,7 @@ export const CallbackService = (config: SvixConfig) => {
...
     const accountCallbackId = getAccountCallbackId(accountId)
```
dolcalmi, line 117: please include line 117-119 inside try block

dolcalmi (changes requested):

core/api/src/app/callback/index.ts
nicolasburtey: oh I didn't know that.

core/api/src/app/callback/index.ts
nicolasburtey: we may not need/want the `await`?

core/api/src/services/svix/index.ts
nicolasburtey: `getAccountCallbackId(accountId)` is just a string operation. it won't fail

core/api/src/services/svix/index.ts
```diff
@@ -31,15 +31,36 @@ function prefixObjectKeys(
   }, {})
 }
 
-export const CallbackService = (config: SvixConfig) => {
+interface CallbackService {
```
dolcalmi, line 34: please move to https://github.com/GaloyMoney/galoy/blob/main/core/api/src/domain/callback/index.types.d.ts

core/api/src/services/svix/index.ts
```diff
@@ -31,15 +31,36 @@ function prefixObjectKeys(
...
+  sendMessage: (args: {
+    eventType: string
+    accountId: AccountId
+    walletId: WalletId
+    payload: Record<string, JSONValue>
+  }) => Promise<ApplicationError | true>
+  getPortalUrl: (accountId: AccountId) => Promise<ApplicationError | string>
```
dolcalmi, line 41: ApplicationError is from app layer, please create CallbackServiceError `export class CallbackServiceError extends CallbackError {}` here https://github.com/GaloyMoney/galoy/blob/main/core/api/src/domain/callback/errors.ts

dolcalmi (changes requested):

core/api/src/services/svix/index.ts
dolcalmi: also please rename to `ICallbackService`

dolcalmi (approved):

PR #4338, opened 2024-04-23
chore(voucher): updated withdraw and redeem page to public

siddhart1o1:


apps/voucher/env.ts
```diff
@@ -1,7 +1,5 @@
 import { createEnv } from "@t3-oss/env-nextjs"
 import { z } from "zod"
-import dotenv from "dotenv"
-dotenv.config({ path: "../.env" })
```
siddhart1o1, line 4: Not needed anymore.

nicolasburtey (approved):

Issue #4324, opened 2024-04-21
Show some simple graph on dashboard


nicolasburtey:
Currently the dashboard only show current balance

It would be useful (asked my some users) to show an history of the balance over the days/weeks/months

Maybe with a usd and btc and total graph 

I think for now everything could be calculated on the browser. This might work for most client

Of course if the user were to have more than 100k or so tx, maybe this would not work, but for now this approach would work for most users

nicolasburtey:
> Should we stick to the one-month (max) graph due to too much calculation, as we don't have balance at time of transaction?

I think we should show balance over time, and calculate it dynamically.
we should start by doing some unit testing here to test the behavior

I think it's better to get granular option from the user. I don't think month only is good. I think someone should be able to choose the day they want. 

> Also, for the total amount graph, we might need to convert the balance to a common currency, so calculation will be required there as well + also BTC movement calculation?

I'm not too sure. I would suggest starting with dollar, and when we have something we like we can iterate on having more currencies

siddhart1o1:
>I think it's better to get granular option from the user. I don't think month only is good. I think someone should be able to choose the day they want.or

Yes, but the maximum duration should be one month or maybe six months. One year might involve too much calculation at runtime.

nicolasburtey:
> Yes, but the maximum duration should be one month or maybe six months ? One year might involve too much calculation at runtime.

this has less to do with how many months, but more how many transactions there is. I think we should not have a limit here to get started, but rather see how this behave and then adjust as needed.

> If we want a total amount graph, then we will need data points equal to what we get from the btcPriceList query to match the timestamp and get the dollar price.

good point. maybe to get started, we can show the balance of the bitcoin wallet in bitcoin, and the stablesats balance in dollar, and see how this feels, and then can iterate from it.

PR #4339, opened 2024-04-23
feat(dashboard): balance history graph

siddhart1o1:
Fixed : https://github.com/GaloyMoney/galoy/issues/4324

image


apps/dashboard/app/page.tsx
```diff
@@ -4,9 +4,130 @@ import { authOptions } from "./api/auth/[...nextauth]/route"
...
+    transactions = response?.edges || []
+    pageInfo = response?.pageInfo
+    allTransactions.push(...transactions)
+
+    lastTransactionDate =
+      transactions.length > 0 ? transactions[transactions.length - 1].node.createdAt : 0
+  }
+
+  return allTransactions
+}
```
nicolasburtey, line 46: let's create a separate "library" or folder to make those calculation and add a batteries of unit test on some mock data to verify those are corrects

nicolasburtey:
- the x axis show the same value (same date) for all items

what date are we picking up? can this be changed by the user?

siddhart1o1:
> * the x axis show the same value (same date) for all items

Same date data.

> what date are we picking up?

We are currently displaying all-time data, but with a maximum limit of 500 transactions. We are not bound by time at the moment, but by transaction count.

>  can this be changed by the user?

No, not for now.

apps/dashboard/app/page.tsx
```diff
@@ -1,18 +1,53 @@
...
+            width: "100%",
+            margin: "0 auto",
+          }}
+        >
+          <PriceContainer walletDetails={walletDetails}></PriceContainer>
+          <TransactionChart
+            currentBtcBalance={btcWallet?.balance || 0}
+            currentUsdBalance={usdWallet?.balance || 0}
+            transactions={response}
+          ></TransactionChart>
```
nicolasburtey, line 49:
```suggestion
          <TransactionChart
            currentBtcBalance={btcWallet?.balance || 0}
            currentUsdBalance={usdWallet?.balance || 0}
            transactions={response}
          />
```

apps/dashboard/lib/get-all-transactions.ts
```diff
@@ -0,0 +1,70 @@
...
+export async function fetchAllTransactionsTillDate({
+  token,
+  days,
+}: {
+  token: string
+  days?: number
+}) {
+  const allTransactions = []
+
+  let daysAgo
```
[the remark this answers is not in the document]
nicolasburtey: let's 1/ fix the linting. when we see this is erroring properly, we can then 2/ fix the underlying `no-implicit-any`

apps/dashboard/lib/get-all-transactions.ts
```diff
@@ -0,0 +1,70 @@
...
+  let pageInfo = response?.pageInfo
+
+  allTransactions.push(...transactions)
+
+  let lastTransactionDate =
+    transactions.length > 0 ? transactions[transactions.length - 1].node.createdAt : 0
+
+  while (pageInfo?.hasNextPage && (!days || (daysAgo && lastTransactionDate > daysAgo))) {
+    const nextCursor = pageInfo.endCursor
+    response = await fetchPaginatedTransactions(token, "next", nextCursor)
```
nicolasburtey, line 31: we are using object instead of positional argument in our codebase

nicolasburtey: the only except is if there is a single argument

apps/dashboard/lib/get-all-transactions.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export async function fetchAllTransactionsByCount({
+  token,
+  fetchCount,
+}: {
+  token: string
+  fetchCount: number
+}) {
+  const allTransactions = []
+  let response = await fetchFirstTransactions(token)
```
nicolasburtey, line 51: passing the token around like this feels like an anti pattern

apps/dashboard/lib/get-balance-for-transactions.ts
```diff
@@ -0,0 +1,78 @@
...
+  const btcTransactions: ProcessedTransaction[] = []
+  const usdTransactions: ProcessedTransaction[] = []
+
+  // for now commented this as we are not restricting transaction by date.
+  // const daysAgo = Math.floor(Date.now() / 1000) - days * 24 * 60 * 60
+
+  for (const { node } of transactions) {
+    // if (node.createdAt < daysAgo) {
+    //   break
+    // }
```
nicolasburtey, line 40:
```suggestion
```

apps/dashboard/lib/get-balance-for-transactions.ts
```diff
@@ -0,0 +1,78 @@
...
+  const maxBalance = {
+    usd: Number((currentUsdBalance / 100).toFixed(2)),
+    btc: currentBtcBalance,
+  }
+
+  const btcTransactions: ProcessedTransaction[] = []
+  const usdTransactions: ProcessedTransaction[] = []
+
+  // for now commented this as we are not restricting transaction by date.
+  // const daysAgo = Math.floor(Date.now() / 1000) - days * 24 * 60 * 60
```
nicolasburtey, line 35:
```suggestion
```

nicolasburtey:
few comments skimming through the PR

apps/dashboard/lib/get-all-transactions.ts
siddhart1o1: Yes, I am aware. This is an old function that we are using in other places as well. I will make a PR to convert all these to take an object.

nicolasburtey: we don't necessarily need to rush into adding new features like the balance history graph

I think it's good to focus on code quality for the time being, which means we can start with this refactoring and rebase this PR when the refactoring PR is merged to main.

apps/dashboard/lib/get-all-transactions.ts
nicolasburtey: similarly this comment: https://github.com/GaloyMoney/blink/pull/4339#discussion_r1585295774 let's fix our "design weakness" before adding new features

apps/dashboard/app/page.tsx
```diff
@@ -1,18 +1,53 @@
...
+  const btcWallet = session.userData.data.me?.defaultAccount.wallets.find(
+    (wallet) => wallet.walletCurrency === "BTC",
+  )
+  const usdWallet = session.userData.data.me?.defaultAccount.wallets.find(
+    (wallet) => wallet.walletCurrency === "USD",
+  )
+
+  const response = await fetchAllTransactionsByCount({
+    token,
+    fetchCount: 5, // Maximum fetch rotation count if set to 5; the API will be called a maximum of 5 times, which gives 500 transactions.
```
nicolasburtey, line 29: magic value should be documented with a `const`

apps/dashboard/app/page.tsx
```diff
@@ -1,18 +1,53 @@
...
       <ContentContainer>
-        <PriceContainer walletDetails={walletDetails}></PriceContainer>
+        <Box
+          sx={{
+            maxWidth: "90em",
+            width: "100%",
+            margin: "0 auto",
+          }}
+        >
+          <PriceContainer walletDetails={walletDetails}></PriceContainer>
```
nicolasburtey, line 44: `PriceContainer` is a misleading name. Pricing would typically imply something like bitcoin historical price, or current bitcoin market price, there is none of that here. 

It should be something like `WalletBalanceContainer`. similarly for this "naming debt", maybe let's do a PR and get this balance history rebased from it.

apps/dashboard/components/chart/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+import CircularProgress from "@mui/joy/CircularProgress"
+
+import useProcessedTransactionsForChart from "@/hook/use-processed-transactions-for-chart"
+import { TransactionEdge } from "@/services/graphql/generated"
+
+const CustomTooltip = ({
+  active,
+  payload,
+  walletCurrency,
+}: TooltipProps<number, string> & { walletCurrency: "USD" | "Sats" }) => {
```
nicolasburtey, line 23: There is a lack of consistency in the capitalization. It should probably either be `USD | BTC` or `CENTS | SATS`

```suggestion
}: TooltipProps<number, string> & { walletCurrency: "USD" | "Sats" }) => {
```

apps/dashboard/lib/get-all-transactions.ts
```diff
@@ -0,0 +1,37 @@
+import { TransactionEdge } from "@/services/graphql/generated"
+import {
+  fetchFirstTransactions,
+  fetchPaginatedTransactions,
+} from "@/services/graphql/queries/get-transactions"
+
+export async function fetchAllTransactionsByCount({
```
nicolasburtey, line 7:
```suggestion
export async function fetchTransactionsByCount({
```

naming is misleading. you won't fetch all if you are asking for a count number

apps/dashboard/app/page.tsx
nicolasburtey: it;s not very intuitive that each count is 100 transactions. might be better to have something like `txCount = 500`. this would make the comment unecessary

apps/dashboard/lib/get-balance-for-transactions.ts
```diff
@@ -0,0 +1,69 @@
...
+      btcTransactions.push({
+        balance: btcBalance,
+        date: formatMonth(node.createdAt),
+        dateTime: formatDateTime(node.createdAt),
+      })
+
+      if (node.direction === "RECEIVE") {
+        btcBalance -= Math.abs(node.settlementAmount)
+      } else if (node.direction === "SEND") {
+        btcBalance += Math.abs(node.settlementAmount) + node.settlementFee
```
nicolasburtey, line 63: @vindard @bodymindarts the idea of exposing a "normal balance" would be quite useful to avoid those pieces of code in the frontend

apps/dashboard/app/page.tsx
siddhart1o1: > it;s not very intuitive that each count is 100 transactions. might be better to have something like `txCount = 500`. this would make the comment unecessary

For the current implementation, we don't have to do any extra processing since the maximum number of transactions per request is 100. Therefore, we just run it a maximum of 5 times, which gives us 500, with no slicing or other processing required. But, i agree that this might be confusing for others, so I've updated it to take the transaction count, and now we are taking `500` as input for 500 transactions.

apps/dashboard/lib/get-balance-for-transactions.ts
```diff
@@ -0,0 +1,73 @@
+import {
+  TransactionEdge,
+  TxDirection,
+  WalletCurrency,
+} from "@/services/graphql/generated"
+import { ProcessedTransaction } from "./index.types"
+import { formatDateTime, formatMonth } from "@/app/utils"
+
+export const getBalanceForTransactions = ({
```
nicolasburtey, line 9: the name is not great. IIUC, it should be more `getMinMaxBalanceAndPerWalletTransaction`

nicolasburtey:
just a comment but LG otherwise

nicolasburtey (approved):

nicolasburtey (approved):

PR #4341, opened 2024-04-23
chore(notifications): make push msg optional

UncleSamtoshi:
- Make push optional for notifications

### DIFF
--- a/core/notifications/src/notification_event/mod.rs
+++ b/core/notifications/src/notification_event/mod.rs
@@ -48,12 +48,17 @@ pub trait NotificationEvent: std::fmt::Debug + Send + Sync {
     fn deep_link(&self) -> Option<DeepLink> {
         None
     }
-    fn to_localized_push_msg(&self, locale: &GaloyLocale) -> LocalizedPushMessage;
+    fn should_send_push(&self) -> bool {
+        false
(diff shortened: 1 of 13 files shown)


core/notifications/src/notification_event/mod.rs
```diff
@@ -58,8 +61,11 @@ pub trait NotificationEvent: std::fmt::Debug + Send + Sync {
     fn should_be_added_to_history(&self) -> bool {
         false
     }
-    fn to_localized_persistent_message(&self, _locale: GaloyLocale) -> LocalizedStatefulMessage {
-        unimplemented!()
+    fn to_localized_persistent_message(
+        &self,
+        _locale: GaloyLocale,
+    ) -> Option<LocalizedStatefulMessage> {
+        None
```
bodymindarts, line 68: This is the wrong way. The interface should encapsulate the contract of the should method is true we  always return the message. Anything else is an implementation bug that is ok to crash (since it should never happen). 

So if you refactor remove the other options instead of adding this one.

UncleSamtoshi: Makes sense, I will make the change

bodymindarts (approved):

PR #4346, opened 2024-04-23
chore: remove unused oauth2 client

nicolasburtey:


quickstart/dev/Tiltfile
```diff
@@ -447,7 +447,6 @@ local_resource(
     allow_parallel = True,
     auto_init = run_apps,
     resource_deps = [
```
bodymindarts, line 449: Please don’t update anything under QuickStart. It gets auto bumped In concourse

nicolasburtey: still a no-op? the same diff also applied on the "principal"/source

dolcalmi (approved):

PR #4349, opened 2024-04-24
refactor(core): serve language from notifications subgraph

thevaibhav-dixit:


nicolasburtey (changes requested):
we should probably talk about this during daily dev. I agree with @dolcalmi that the idea of moving the language to the notification pod is not the best solution. And the downtime we had last week was a manifestation of the issue. 

the logic was that there is only one use case for the language, but I think this is not correct:

there were 2 use case:
- sending notification
- serving the language for the mobile users

now there are more use cases, like with the integration of GPT

this make the notification part of the runtime to serve most requests from the mobile app. this make it that the odds of something going wrong increase. as opposed to have the notification pods as a non mandatory part of the runtime for frequent non critical requests

bodymindarts:
> we should probably talk about this during daily dev. I agree with @dolcalmi that the idea of moving the language to the notification pod is not the best solution. And the downtime we had last week was a manifestation of the issue.
> 
> the logic was that there is only one use case for the language, but I think this is not correct:
> 
> there were 2 use case:
> 
> * sending notification
> * serving the language for the mobile users (it felt like this use case was omitted)
> 
> now there are more use cases, like with the integration of GPT
> 
> this make the notification part of the runtime to serve most requests from the mobile app. this make it that the odds of something going wrong increase. as opposed to have the notification pods as a non mandatory part of the runtime for frequent non critical requests

Agree that we should revisit this discussion - probably at ACE (don't think daily will have enough time as the topic is quite deep).
This PR is intended to improve resiliency in the short term pending the outcome of that discussion. But we can also hold off on merging until we have talked about it.

PR #4352, opened 2024-04-24
feat: exposing sessions to the api

nicolasburtey:
This PR is exposing the kratos/hydra sessions related to a user. 
The idea is show those sessions then into the dashboard.

A follow up PR will enable the mutation to disabled other sessions.

This PR also aim at understanding how to enable long term read only sessions, something Andrew has been asking for

### DIFF
--- a/core/api/package.json
+++ b/core/api/package.json
@@ -161,7 +161,6 @@
     "@typescript-eslint/parser": "^7.7.0",
     "@xascode/tscpaths": "0.1.4",
     "axios-mock-adapter": "^1.22.0",
-    "child_process": "^1.0.2",
     "eslint": "^8.57.0",
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-import": "^2.29.1",
--- a/core/api/test/integration/app/user/consent-list.spec.ts
+++ b/core/api/test/integration/app/user/consent-list.spec.ts
@@ -1,9 +1,9 @@
-import { exec } from "child_process"
 import puppeteer from "puppeteer"
 
 import { Admin } from "@/app"
 import { consentList } from "@/services/hydra"
 import { sleep } from "@/utils"
+import knex from "knex"
 
 let userId: UserId
 const email = "test@galoy.io" as EmailAddress
@@ -15,26 +15,31 @@ beforeAll(async () => {
   userId = account.kratosUserId
 })
 
+const db = knex({
+  client: "pg",
+  connection: "postgres://dbuser:secret@localhost:5432/default?sslmode=disable",
+})
+
 const getOTP = (email: string): Promise<string> => {
-  return new Promise((resolve, reject) => {
-    const query = `docker exec -i galoy-dev-kratos-pg-1 psql -U dbuser -d default -t -c "SELECT body FROM courier_messages WHERE recipient='${email}' ORDER BY created_at DESC LIMIT 1;"`
-    exec(query, (error, stdout, stderr) => {
-      if (error) {
-        reject(`error: ${error.message}`)
-        return
+  return db<{ body: string }>("courier_messages") // Specifying the expected row structure
+    .select("body")
+    .where("recipient", email)
+    .orderBy("created_at", "desc")
+    .limit(1)
+    .then((rows) => {
+      if (rows.length === 0) {
+        throw new Error("OTP not found in the message")
       }
-      if (stderr) {
-        reject(`stderr: ${stderr}`)
-        return
-      }
-      const otpMatch = stdout.match(/(\d{6})/)
+      const otpMatch = rows[0].body.match(/(\d{6})/)
       if (otpMatch && otpMatch[1]) {
-        resolve(otpMatch[1])
+        return otpMatch[1]
       } else {
-        reject("OTP not found in the message")
+        throw new Error("OTP not found in the message")
       }
     })
-  })
+    .catch((error) => {
+      throw new Error(`Error retrieving OTP: ${error.message}`)
+    })
 }
 
 async function performOAuthLogin() {
@@ -60,17 +65,17 @@ async function performOAuthLogin() {
   )
   await page.type('[data-testid="email_id_input"]', email)
   screenshots && (await page.screenshot({ path: "screenshot2.png" }))
-  await sleep(250)
+  await sleep(1000)
 
   await page.click("#accept")
 
-  await sleep(500)
+  await sleep(1000)
   screenshots && (await page.screenshot({ path: "screenshot3.png" }))
 
   const otp = await getOTP(email)
 
   await page.waitForSelector("#code")
-  await page.type("#code", otp, { delay: 100 })
+  await page.type("#code", otp)
 
   screenshots && (await page.screenshot({ path: "screenshot4.png" }))
   await sleep(1500)
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1408,9 +1408,6 @@ importers:
       axios-mock-adapter:
         specifier: ^1.22.0
         version: 1.22.0(axios@1.6.8)
-      child_process:
-        specifier: ^1.0.2
-        version: 1.0.2
       eslint:
         specifier: ^8.57.0
         version: 8.57.0
@@ -2969,7 +2966,7 @@ packages:
       '@babel/traverse': 7.24.1
       '@babel/types': 7.24.0
       convert-source-map: 2.0.0
-      debug: 4.3.4(supports-color@8.1.1)
+      debug: 4.3.4(supports-color@5.5.0)
       gensync: 1.0.0-beta.2
       json5: 2.2.3
       semver: 6.3.1
@@ -3227,7 +3224,7 @@ packages:
       '@babel/core': 7.24.4
       '@babel/helper-compilation-targets': 7.23.6
       '@babel/helper-plugin-utils': 7.24.0
-      debug: 4.3.4(supports-color@8.1.1)
+      debug: 4.3.4(supports-color@5.5.0)
       lodash.debounce: 4.0.8
       resolve: 1.22.8
     transitivePeerDependencies:
@@ -6979,7 +6976,7 @@ packages:
       '@babel/helper-split-export-declaration': 7.22.6
       '@babel/parser': 7.24.4
       '@babel/types': 7.24.0
-      debug: 4.3.4(supports-color@8.1.1)
+      debug: 4.3.4(supports-color@5.5.0)
       globals: 11.12.0
     transitivePeerDependencies:
       - supports-color
@@ -7734,7 +7731,7 @@ packages:
     engines: {node: ^12.22.0 || ^14.17.0 || >=16.0.0}
     dependencies:
       ajv: 6.12.6
-      debug: 4.3.4(supports-color@8.1.1)
+      debug: 4.3.4(supports-color@5.5.0)
       espree: 9.6.1
       globals: 13.24.0
       ignore: 5.3.1
@@ -10238,7 +10235,7 @@ packages:
     engines: {node: '>=10.10.0'}
     dependencies:
       '@humanwhocodes/object-schema': 2.0.2
-      debug: 4.3.4(supports-color@8.1.1)
+      debug: 4.3.4(supports-color@5.5.0)
       minimatch: 3.1.2
     transitivePeerDependencies:
       - supports-color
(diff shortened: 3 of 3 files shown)


apps/dashboard/cypress/e2e/consent-integration.cy.ts
```diff
@@ -32,22 +32,6 @@ describe("Consent integration Test", () => {
...
-      cy.contains("label", "write").should("exist")
-      cy.contains("label", "write").should("be.visible")
-      cy.contains("label", "write").should("not.be.disabled")
-      cy.contains("label", "write").click()
-
-      cy.get("[data-testid=submit_consent_btn]").should("exist")
-      cy.get("[data-testid=submit_consent_btn]").should("be.visible")
-      cy.get("[data-testid=submit_consent_btn]").should("not.be.disabled")
-      cy.get("[data-testid=submit_consent_btn]").click()
-
```
dolcalmi, line 50: why this was removed? not related to the PR

apps/voucher/cypress/e2e/consent-integration.cy.ts
```diff
@@ -32,22 +32,6 @@ describe("Consent integration Test", () => {
...
-      cy.contains("label", "write").should("exist")
-      cy.contains("label", "write").should("be.visible")
-      cy.contains("label", "write").should("not.be.disabled")
-      cy.contains("label", "write").click()
-
-      cy.get("[data-testid=submit_consent_btn]").should("exist")
-      cy.get("[data-testid=submit_consent_btn]").should("be.visible")
-      cy.get("[data-testid=submit_consent_btn]").should("not.be.disabled")
-      cy.get("[data-testid=submit_consent_btn]").click()
-
```
dolcalmi, line 50: same here

bats/core/api/user.bats
```diff
@@ -5,6 +5,8 @@ setup_file() {
 }
 
 @test "account: updates language" {
+  "skip"
```
dolcalmi, line 8: why?

core/api/package.json
```diff
@@ -161,6 +161,7 @@
     "@typescript-eslint/parser": "^7.7.0",
     "@xascode/tscpaths": "0.1.4",
     "axios-mock-adapter": "^1.22.0",
+    "child_process": "^1.0.2",
```
dolcalmi, line 164: last update 8 years ago https://www.npmjs.com/package/child_process

core/api/package.json
```diff
@@ -182,6 +183,7 @@
     "pino-pretty": "^10.3.1",
     "prettier": "^3.2.5",
     "protoc-gen-js": "^3.21.2",
+    "puppeteer": "^22.7.1",
```
dolcalmi, line 186: not sure about this.. integrations tests are already flacky and now introducing a package like this that runs a headless chromium can be problematic.. is not another way to test it?

bats/gql/list-sessions.gql
```diff
@@ -0,0 +1,17 @@
...
+      id
+      expiresAt
+      issuedAt
+    }
+    delegations {
+      app
+      handledAt
+      remember
+      scope
+    }
```
dolcalmi, line 15: probably something to discuss at ACE or another meeting... why not just have a sessions query?

core/api/src/app/users/list-sessions.ts
```diff
@@ -0,0 +1,10 @@
+import { consentList } from "@/services/hydra"
+import { listSessions as listSessionsService } from "@/services/kratos"
+
+export const listMobileSessions = async (userId: UserId) => {
```
dolcalmi, line 4: please include return type and change param to `{ userId }: {userId: UserId }` in this way params is passed correctly for tracing otherwise will be code.function.params.0

core/api/src/app/users/list-sessions.ts
```diff
@@ -0,0 +1,10 @@
...
+  return listSessionsService(userId)
+}
+
+export const listDeleguations = async (userId: UserId) => {
```
dolcalmi, line 8: same here

core/api/src/graphql/public/types/object/user.ts
```diff
@@ -75,6 +77,22 @@ const GraphQLUser = GT.Object<User, GraphQLPublicContextAuth>({
       },
     },
 
+    mobileSessions: {
+      type: GT.NonNullList(MobileSession),
+      description: "List of mobile sessions",
+      resolve: async (source, args, { user }) => {
+        return Users.listMobileSessions(user.id)
```
dolcalmi, line 84: we must guarantee that this method does not return an error otherwise will block me query. Please add validation or guarantee at app layer that we always return a mobile session

core/api/src/graphql/public/types/object/user.ts
```diff
@@ -75,6 +77,22 @@ const GraphQLUser = GT.Object<User, GraphQLPublicContextAuth>({
...
+      },
+    },
+
+    delegations: {
+      type: GT.NonNullList(Delegation),
+      description: "List of Oauth2 delegations",
+      resolve: async (source, args, { user }) => {
+        return Users.listDeleguations(user.id)
```
dolcalmi, line 92: same here + fix typo

core/api/src/services/hydra/index.ts
```diff
@@ -0,0 +1,38 @@
...
+  // @ts-ignore-next-line no-implicit-any error
+  let res: AxiosResponse<any, any>
+
+  try {
+    res = await axios.get(
+      `${hydraUrl}/admin/oauth2/auth/sessions/consent?subject=${userId}`,
+    )
+  } catch (err) {
+    // TODO
+    console.error(err)
```
dolcalmi, line 16: remove and handle errors correclty

core/api/src/services/hydra/index.ts
```diff
@@ -0,0 +1,38 @@
+import axios, { AxiosResponse } from "axios"
+
+const hydraUrl = process.env.HYDRA_ADMIN_URL || "http://localhost:4445"
+
+export const consentList = async (userId: UserId): Promise<Delegation[]> => {
+  /* eslint @typescript-eslint/ban-ts-comment: "off" */
+  // @ts-ignore-next-line no-implicit-any error
+  let res: AxiosResponse<any, any>
```
dolcalmi, line 8: please remove the ignore and create the types

core/api/src/services/hydra/index.ts
dolcalmi, line 1: please create the Interface at domain level like in the other services

core/api/src/services/hydra/index.ts
```diff
@@ -0,0 +1,38 @@
...
+    console.error(err)
+    return []
+  }
+
+  let sessions: Delegation[]
+
+  try {
+    /* eslint @typescript-eslint/ban-ts-comment: "off" */
+    // @ts-ignore-next-line no-implicit-any error
+    sessions = res.data.map((request) => ({
```
dolcalmi, line 25: same here, remove ignore (probably fixed after add types)

core/api/src/services/hydra/index.ts
```diff
@@ -0,0 +1,38 @@
...
+      scope: request.grant_scope,
+      handledAt: new Date(request.handled_at),
+      remember: request.remember,
+      app: request.consent_request.client.client_name,
+    }))
+  } catch (err) {
+    // TODO
+    console.error(err)
```
dolcalmi, line 33: remove + handle errors

core/api/src/services/kratos/private.ts
```diff
@@ -30,16 +25,12 @@ export const getKratosPostgres = () =>
...
-  if (!session.expires_at) throw new MissingExpiredAtKratosError()
-
+export const toDomainSession = (session: KratosSession): MobileSession => {
   return {
     id: session.id as SessionId,
-    identity: toDomainIdentity(session.identity),
-    expiresAt: new Date(session?.expires_at),
+    identity: toDomainIdentity(session.identity as Identity),
+    expiresAt: new Date(session?.expires_at as string),
+    issuedAt: new Date(session?.issued_at as string),
```
dolcalmi, line 33: I checked the Session interface and expires_at and issued_are are already strings... cast is not necessary also can hide a crash because of conversion (Date)

core/api/test/integration/app/user/consent-list.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+beforeAll(async () => {
+  const account = await Admin.getAccountByUserEmail(email)
+  if (account instanceof Error) throw account
+
+  userId = account.kratosUserId
+})
+
+const getOTP = (email: string): Promise<string> => {
+  return new Promise((resolve, reject) => {
+    const query = `docker exec -i galoy-dev-kratos-pg-1 psql -U dbuser -d default -t -c "SELECT body FROM courier_messages WHERE recipient='${email}' ORDER BY created_at DESC LIMIT 1;"`
```
dolcalmi, line 20: not sure about this.. is not better to use knex? we already have the package.. wdyt @vindard

core/api/test/integration/app/user/consent-list.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+  const screenshots = false
+
+  const browser = await puppeteer.launch()
+  // const browser = await puppeteer.launch({ headless: true })
+  const page = await browser.newPage()
+
+  // Navigate the page to a URL
+  await page.goto("http://localhost:3001/api/auth/signin")
+
+  screenshots && (await page.screenshot({ path: "screenshot0.png" }))
```
dolcalmi, line 50: is this only help for debug? what is the purpose of take screenshots?

core/api/test/integration/app/user/consent-list.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+  it("get an empty consent list", async () => {
+    const res = await consentList(userId)
+    expect(res).toEqual([])
+  })
+
+  it("get consent list when the user had perform oauth2 login", async () => {
+    await performOAuthLogin()
+
+    const res = await consentList(userId)
+    expect(res).toEqual([
```
dolcalmi, line 93: if we are only testing this then this test is not needed :thinking:  or it is better to use a mock (which probably does not make sense for what service is doing)

dolcalmi (changes requested):

apps/voucher/cypress/e2e/consent-integration.cy.ts
nicolasburtey: It is, we're skipping consent in the tests

https://github.com/GaloyMoney/blink/pull/4352/files#diff-837b4f391cfd25241cfd9947e05f126a3078d0b3109844600ec949dcc3dd1ebdR25

core/api/package.json
nicolasburtey: doesn't matter? it's on dev only

core/api/package.json
nicolasburtey: I understand your point. 

I tried with axios and basically spent a day on this, and I could not make it work, so I gave up.

the issue with a OAuth2 flow is that there is a lot of context carried around different screen, whether it's via parameters over the URL or via Cookie. it's quite hard to replicate this, specially for someone not having the full context of OAuth2 in its head. 

using puppeteer make it a lot more straightforward.

I suggest we could try it and revert/change if this introduce flackiness to the test suite?

bats/gql/list-sessions.gql
nicolasburtey: you mean, as opposed to segmenting between delegation and sessions? we also have API keys btw.

I think the reason to not have them combined is because they have different properties.

for instance, delegated token can currently have smaller scope than a mobile token, which currently is always full access. delegated token are also attached to a particular application.

core/api/src/app/users/list-sessions.ts
nicolasburtey: adding that to ACE... so we can understand where it's necessary to have an object as opposed to an simple argument when there is a single scalar passed to the functions

core/api/src/services/kratos/private.ts
nicolasburtey: it can be undefined otherwise

image

before I had this but it's more ugly

  if (!session.expires_at) throw new MissingExpiredAtKratosError()

core/api/test/integration/app/user/consent-list.spec.ts
nicolasburtey: that's fair. I copied it from the dashboard.

core/api/test/integration/app/user/consent-list.spec.ts
nicolasburtey: yup indeed to help debugging going over the different screens

core/api/test/integration/app/user/consent-list.spec.ts
nicolasburtey: To test/develop the API, I had to get a client authenticated through an OAuth2 flow to be able to see the result. 

Open to other suggestion, but that feels like a mock would not achieve the tests here, because it's really testing the integration with hydra.

core/api/src/app/users/list-sessions.ts
```diff
@@ -0,0 +1,10 @@
+import { consentList } from "@/services/hydra"
+import { listSessions as listSessionsService } from "@/services/kratos"
+
+export const listMobileSessions = async ({ userId }: { userId: UserId }) => {
+  return listSessionsService(userId)
+}
+
+export const listDeleguations = async ({ userId }: { userId: UserId }) => {
```
bodymindarts, line 8: typo in `Deleguations`

PR #4354, opened 2024-04-24
chore(core): fallback for language query

UncleSamtoshi:
- use empty language when notifictions service returns an error
- add timeout to notification settings grpc call

### DIFF
--- a/core/api/src/domain/users/index.ts
+++ b/core/api/src/domain/users/index.ts
@@ -39,6 +39,8 @@ export const checkedToEmailAddress = (
   return emailAddress as EmailAddress
 }
 
+export const DefaultLanguage = "" as UserLanguageOrEmpty
+
 export const checkedToLanguage = (
   language: string,
 ): UserLanguageOrEmpty | ValidationError => {
--- a/core/api/src/services/notifications/index.ts
+++ b/core/api/src/services/notifications/index.ts
@@ -29,11 +29,10 @@ import * as notificationsGrpc from "./grpc-client"
 
 import { handleCommonNotificationErrors } from "./errors"
 
-import { getCallbackServiceConfig } from "@/config"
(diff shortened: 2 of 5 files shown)


core/api/src/domain/users/index.ts
```diff
@@ -39,6 +39,8 @@ export const checkedToEmailAddress = (
   return emailAddress as EmailAddress
 }
 
+export const DefaultLanguage = "" as UserLanguageOrEmpty
```
bodymindarts, line 42: With the default being set the type `UserLanguageOrEmpty` becomes irrelevant. Perhaps you could find something semantically more meaningful. Or if it's too complicated to untangle we can leave it for another PR.

Looks good otherwise.

UncleSamtoshi: Not sure I fully understand. The `DefaultLanguage` is used as a fallback in the case of the error, but even in the case there is not an error the notifications service could still return an empty string implying that the mobile application should use the OS system default. It's definitely a little bit of a mess that is a relic of how the mobile application was handling the language.

UncleSamtoshi: I think it may make sense to save that refactoring for a separate pr.

core/api/src/services/notifications/index.ts
```diff
@@ -65,6 +64,9 @@ export const NotificationsService = (): INotificationsService => {
       const response = await notificationsGrpc.getNotificationSettings(
         request,
         notificationsGrpc.notificationsMetadata,
+        {
```
dolcalmi, line 67: :thinking: shouldn't we use something similar to https://github.com/GaloyMoney/galoy/blob/main/core/api/src/services/lnd/index.ts#L801 ?

UncleSamtoshi: That would work too, but I don't see a problem with relying on grpc.

dolcalmi (approved):

PR #4357, opened 2024-04-24
chore(notifications): refactor marketing notification

UncleSamtoshi:
- update deep link options
- update MarkingNotificationEvent to allow for different channels
- use updated marketingNotificationTrigger mutation in admin-panel


core/notifications/proto/notifications.proto
```diff
@@ -237,19 +237,49 @@ message PriceChanged {
...
   optional DeepLink deep_link = 3;
+  bool should_send_push = 4;
+  bool should_add_to_history = 5;
+  bool should_add_to_bulletin = 6;
 }
 
-message LocalizedPushContent {
+message LocalizedContent {
   string title = 2;
   string body = 3;
```
thevaibhav-dixit, line 249: why 2, 3 instead of 1,2 .. . This is a breaking change so might as well fix this also

UncleSamtoshi: Yeah thats a good point, I will fix it

core/notifications/src/graphql/convert.rs
```diff
@@ -10,10 +10,11 @@ impl From<history::StatefulNotification> for types::StatefulNotification {
         Self {
             deep_link: notification.deep_link().map(|d| d.to_link_string()),
             id: ID(notification.id.to_string()),
-            title: notification.message.title,
-            body: notification.message.body,
+            title: notification.message.title.clone(),
+            body: notification.message.body.clone(),
             created_at: types::Timestamp::from(created_at),
             acknowledged_at: acknowledeg_at.map(types::Timestamp::from),
+            add_to_bulletin: notification.add_to_bulletin(),
```
bodymindarts, line 17: move these above line 13 - then you can de-structure without needing to clone. Best to avoid clone where possible.
Before you start moving attributes you can call fns.

core/notifications/src/user_notification_settings/repo.rs
```diff
@@ -63,11 +63,11 @@ impl UserNotificationSettingsRepo {
 
     pub async fn find_for_user_ids(
         &self,
-        user_ids: impl IntoIterator<Item = &GaloyUserId>,
+        user_ids: &[GaloyUserId],
```
bodymindarts, line 66: Why was this changed? The original was broader in what it accepts.

PR #4359, opened 2024-04-25
chore: update cargo lock

dolcalmi:
reduce start up times for rust projects in local env (`buck2 run //dev:up`)


bodymindarts:
Its better to update each dev explicitly to keep the root and third-party in sync.

PR #4367, opened 2024-04-25
ci: separate build steps in integration tests

vindard:
## Description

Separate build and run steps for integration tests to help with debugging and reliability

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -22,6 +22,7 @@ jobs:
         id: prepare_args
         run: |
           ARGS=""
+          BUILD_ARGS=""
 
           cat <<EOF > labels.json
           ${{ toJSON(github.event.pull_request.labels.*.name) }}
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-test.yml
```diff
@@ -35,17 +36,65 @@ jobs:
...
             case "$LABEL" in
-              dashboard|consent|pay|core|admin-panel|map|voucher)
+              core|dashboard|consent|pay|admin-panel|map|voucher)
+                BUILD_ARGS+=" //core/api:prod_build"
+                BUILD_ARGS+=" //core/notifications:notifications"
+
+                break
+                ;;
+            esac
+          done
```
dolcalmi, line 49: if we are not using #LABEL why do we need a for

.github/workflows/integration-test.yml
```diff
@@ -35,17 +36,65 @@ jobs:
...
+
+          # Handle root 'node_modules' dependency
+          for LABEL in "${LABELS[@]}"; do
+            case "$LABEL" in
+              dashboard|consent|pay|admin-panel|map|voucher)
+                BUILD_ARGS+=" "//:node_modules""
```
dolcalmi, line 55: same here

.github/workflows/integration-test.yml
```diff
@@ -35,17 +36,65 @@ jobs:
...
+                ;;
+            esac
+          done
+
+          # Handle 'consent' dependencies
+          for LABEL in "${LABELS[@]}"; do
+            case "$LABEL" in
+              dashboard|voucher)
+                BUILD_ARGS+=" //apps/consent:consent"
+                BUILD_ARGS+=" //core/api-keys:api-keys"
```
dolcalmi, line 67: and here

dolcalmi (changes requested):

.github/workflows/integration-test.yml
vindard: We only execute line Ln 55 if it's one of the `app` labels right now

dolcalmi (approved):

PR #4375, opened 2024-04-26
ci: add importer to direct jest non-error output to stdout

vindard:
## Description

`jest` has an issue where non-error reporting also goes to stderr (see this issue https://github.com/jestjs/jest/issues/5064). This seems to be a long-running issue and one of the solution that's emerged is to use the `jest-standard-reporter` package as a reporter in jest config to properly direct output.

Our trigger for this is to be able to see test reporting as tests are executed in buck instead of having to wait to the end of the test run when stderr is collected and dumped to the terminal.

This segregation has additional potential benefits if we ever end up using any ci tooling that depends on error messages going to stderr and regular output going to stdout.


vindard:
Closing because STDOUT also doesn't get printed to console in realtime

PR #4384, opened 2024-04-29
feat(notifications): expose unacknowledgedStatefulNotificationsCount field

thevaibhav-dixit:


core/notifications/src/app/mod.rs
```diff
@@ -307,6 +307,18 @@ impl NotificationsApp {
         Ok(notification)
     }
 
+    #[instrument(name = "app.unacknowledged_notifications_count", skip(self), err)]
```
thevaibhav-dixit, line 310: note: fix the tracing name

core/notifications/src/history/repo.rs
```diff
@@ -59,11 +59,36 @@ impl PersistentNotifications {
         notification: &mut StatefulNotification,
     ) -> Result<(), NotificationHistoryError> {
         let mut tx = self.pool.begin().await?;
+        sqlx::query!(
+            r#"UPDATE stateful_notifications
+            SET acknowledged = true
```
bodymindarts, line 64: how do you know this is acknowledged = true? Need to inject the setting from the entity.

core/notifications/src/history/mod.rs
```diff
@@ -105,4 +105,11 @@ impl NotificationHistory {
     ) -> Result<(Vec<StatefulNotification>, bool), NotificationHistoryError> {
         self.repo.list_for_user(user_id, first, after).await
     }
+
+    pub async fn unacknowledged_notifications_count(
+        &self,
+        user_id: GaloyUserId,
+    ) -> Result<i64, NotificationHistoryError> {
```
bodymindarts, line 112: why `i64`? can this every be negative?

core/notifications/src/history/mod.rs
```diff
@@ -105,4 +105,11 @@ impl NotificationHistory {
     ) -> Result<(Vec<StatefulNotification>, bool), NotificationHistoryError> {
         self.repo.list_for_user(user_id, first, after).await
     }
+
+    pub async fn unacknowledged_notifications_count(
```
bodymindarts, line 109: this name is not idiomatic ` unacknowledged_notifications_count` looks like you have cargo culted from GQL layer.

Should probably have something like `count_unacknowledged_notifications`

core/notifications/src/history/repo.rs
```diff
@@ -59,11 +59,37 @@ impl PersistentNotifications {
...
         tx.commit().await?;
         Ok(())
     }
 
+    pub async fn unacknowledged_notifications_count(
+        &self,
+        user_id: GaloyUserId,
+    ) -> Result<i64, NotificationHistoryError> {
+        let count = sqlx::query_scalar!(
+            r#"SELECT COUNT(*)
```
bodymindarts, line 82: SELECT COUNT(*) AS "count!"

core/notifications/src/history/repo.rs
```diff
@@ -74,19 +74,20 @@ impl PersistentNotifications {
...
-            r#"SELECT COUNT(*)
+            r#"SELECT COUNT(*) as "count!"
             FROM stateful_notifications
             WHERE galoy_user_id = $1 AND acknowledged = FALSE"#,
             user_id.as_ref(),
         )
         .fetch_one(self.read_pool.inner())
         .await?
-        .expect("count can not be None");
+        .try_into()
```
bodymindarts, line 89: an just use `as u64`

thevaibhav-dixit: yeah thought about that as well but I think I want to go for expect... do you have a strong reason not to use it ?

PR #4387, opened 2024-04-29
feat: reissue chatbotId when errored

nicolasburtey:


core/api/src/app/support/chat.ts
```diff
@@ -1,8 +1,25 @@
 import { RepositoryError, CouldNotFindError } from "@/domain/errors"
+import { ChatAssistantNotFoundError } from "@/domain/support/errors"
 import { AccountsRepository, UsersRepository } from "@/services/mongoose"
 import { SupportChatRepository } from "@/services/mongoose/support-chat"
 import { Assistant } from "@/services/openai"
 
+const getMessages = async ({
+  supportChatId,
+  accountId,
+}: {
```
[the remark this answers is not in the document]
nicolasburtey: only used in test

core/api/src/app/support/chat.ts
```diff
@@ -1,20 +1,63 @@
...
+    level,
+    countryCode,
+    language,
+  })
+  if (supportChatId instanceof Error) return supportChatId
+
+  const res = await SupportChatRepository().create({ supportChatId, accountId })
+  if (res instanceof RepositoryError) return res
+
+  return supportChatId
```
dolcalmi, line 45:
```suggestion
  return SupportChatRepository().create({ supportChatId, accountId })
```

core/api/src/app/support/chat.ts
```diff
@@ -1,20 +1,63 @@
...
+
 export const getSupportChatMessages = async (accountId: AccountId) => {
   const supportChatId = await SupportChatRepository().findNewestByAccountId(accountId)
 
   if (supportChatId instanceof CouldNotFindError) {
     return []
-  }
-
-  if (supportChatId instanceof Error) {
+  } else if (supportChatId instanceof Error) {
```
dolcalmi, line 53: not sure to understand this change... else block is not necessary

core/api/src/app/support/chat.ts
```diff
@@ -1,20 +1,63 @@
...
+  const messages = Assistant().getMessages(supportChatId)
+  if (messages instanceof ChatAssistantNotFoundError) {
+    await initializeSupportChat({ accountId })
+    return []
+  }
+
+  return messages
+}
+
+export const initializeSupportChat = async ({ accountId }: { accountId: AccountId }) => {
```
dolcalmi, line 24: please add return type

bats/core/api/user.bats
```diff
@@ -25,30 +25,35 @@ setup_file() {
...
   [[ $length -eq 6 ]] || exit 1
+
+  exec_graphql 'alice' 'support-chat-reset' "$variables"
+  success="$(graphql_output '.data.supportChatReset.success')"
+  [[ "$success" == "true" ]] || exit 1
+
+  exec_graphql 'alice' 'support-chat'
+  messages="$(graphql_output '.data.me.supportChat')"
+  length=$(echo "$messages" | jq '. | length')
+  [[ $length -eq 2 ]] || exit 1
```
dolcalmi, line 58: should not be 0 ?

dolcalmi (dismissed):

core/api/src/app/support/chat.ts
nicolasburtey: it is, there is a different between the prior block which is checking `CouldNotFindError` and this block checking for the more general `Error`

bats/core/api/user.bats
nicolasburtey: no because initialized chat create the first 2 messages, where the user is sharing with the bot his country code / language / account level.

core/api/src/app/support/chat.ts
dolcalmi: again not undestanding.. not a blocker but the else is unnecessary because you are using return

dolcalmi (approved):

bodymindarts (approved):

PR #4407, opened 2024-05-01
refactor(dashboard): moving token to apollo client

siddhart1o1:
Before, we were passing the client from the core API (Blink API) function invocation to the Apollo client instance. Now, we are using `getServerSession` to directly get the token when calling the Apollo client instance.


nicolasburtey (approved):
I just skimmed through it and that sound a better approach

If the other apps had the same approach we should also refactor them

PR #4418, opened 2024-05-03
fix(core): update admin queries balanceSnapshotSec

dolcalmi:
Reduce balanceSnapshotSec for admin balance queries

### DIFF
--- a/core/api/src/services/ledger/admin-legacy.ts
+++ b/core/api/src/services/ledger/admin-legacy.ts
@@ -1,4 +1,4 @@
-import { MainBook } from "./books"
+import { MainBookAdmin } from "./books"
 import { getBankOwnerWalletId } from "./caching"
 import {
   assetsMainAccount,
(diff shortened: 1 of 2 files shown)


vindard (approved):

core/api/src/services/ledger/admin-legacy.ts
```diff
@@ -54,7 +56,7 @@ export const updateLndEscrow = async (amount: Satoshis) => {
     return { ...escrowData, updated: false }
   }
 
-  const entry = MainBook.entry("escrow")
+  const entry = MainBookAdmin.entry("escrow")
```
vindard, line 59: I wonder if we should keep `MainBook` for writes and use `MainBookAdmin` just for queries 🤔

dolcalmi: initially I had `MainBookReadOnly` but even balance needs write permissions to save snapshot.. I will merge it to validate and if it is the case I will update it in the next PR

Issue #4393, opened 2024-04-30
 Inconsistent Interaction for Dashboard Profile Menu Options


amitamrutiya2210:
**Problem**

When clicking on the profile avatar in the Blink dashboard, three menu options are visible:

-  Account ID
-  Logout
-  Dark Mode

Currently, clicking on the **Logout** tile successfully logs out the user. However, clicking on **Account ID** or **Dark Mode** does not trigger any action. To copy the Account ID or enable Dark Mode, users need to click on a small icon within these tiles.

**Proposed Improvement**

To enhance user experience and maintain consistency across the dashboard, it would be beneficial to make all menu options consistent in their interaction method. Either:

-  Make the entire tile clickable, triggering the intended action upon click.
-  Make only icon clickable, triggering the intended action upon click.

**Steps to Reproduce**

-  Navigate to the Blink dashboard and log in with your account.
-  Locate the profile icon in the top right corner.
-  Click on the profile icon to reveal the three menu options.

If this issue is deemed valid by the project mentor, I will fix this issue.

![image](https://github.com/GaloyMoney/blink/assets/91112485/b2f3aca7-3d7a-4fff-a528-039cf46de3e8)

nicolasburtey:
probably a issue worth taclking

amitamrutiya2210:
> Make the entire tile clickable, triggering the intended action upon click.

I think this solution is better. What do you think, @nicolasburtey?

PR #4422, opened 2024-05-05
fix(dashboard): inconsistent interaction for dashboard profile menu options

amitamrutiya2210:
Fix Issue #4393 

Make the entire `MenuTile` clickable while maintaining the existing functionality. Additionally, I have added text `Light Mode` so that users can see light mode text when the dashboard is in dark mode, and dark mode text when the dashboard is in light mode.


siddhart1o1:
Thanks for the PR.

I tried it locally, and when I used it for the first time and clicked on the theme toggle, it stayed in light mode, but the text changed.

image

image

siddhart1o1:
Also, in dark mode, it shows "light mode" and in light mode, it shows "dark."
image

image

amitamrutiya2210:
> Also, in dark mode, it shows "light mode" and in light mode, it shows "dark." 

I have explicitly replaced this text with an icon. 

> I tried it locally, and when I used it for the first time and clicked on the theme toggle, it stayed in light mode, but the text changed.

It's weird. I am not able to reproduce this same error.

siddhart1o1:
> It's weird. I am not able to reproduce this same error.

I am using Brave; maybe try in a private window of Brave so you don't have cookies or other data stored.

amitamrutiya2210:
@siddhart1o1 I tried in brave private window still not able to reproduce this error 😟

siddhart1o1:
> Thanks for the PR.
> 
> I tried it locally, and when I used it for the first time and clicked on the theme toggle, it stayed in light mode, but the text changed.
> 
> image image

Here is how you can fix this issue: the mode is not just dark and light, but it can also be system, so we need to handle that case as well.

Something like 
```
export default function ColorSchemeToggle() {
  const { mode, setMode, systemMode } = useColorScheme()
  const currentMode = mode === "system" ? systemMode : mode   // if mode is system use `systemMode`

  return (
    <MenuItem
      onClick={() => {
        if (currentMode === "light") {
          setMode("dark")
        } else {
          setMode("light")
        }
      }}
      sx={{
        display: "flex",
        justifyContent: "space-between",
        fontWeight: "500",
      }}
    >
      {currentMode === "light" ? "Light Mode" : "Dark Mode"}.        // Also if currentMode is light show `Light Mode` text same for dark i think this is better for understanding.
      {currentMode === "light" ? <LightModeIcon /> : <DarkModeRoundedIcon />}
    </MenuItem>
  )
}

```

Issue #4396, opened 2024-04-30
Unexpected API Call Triggered by Cancel Button on Login Screen


amitamrutiya2210:
**Description:**

On the login screen, clicking the "Cancel" button causes the "Next" button to display a loading state momentarily before returning to the previous screen. However, this behavior triggers an API call for login with empty details, which is unexpected and unnecessary.

**Expected Behavior**

When the user clicks the "Cancel" button, they should be immediately redirected back to the previous screen without triggering any API calls.

**Steps to Reproduce**

1. Navigate to the login screen.
2. Press the "Cancel" button.
3. Observe that the "Next" button briefly shows a loading state.

**Screenshot**
![Screencast from 30-04-24 11 24 46 PM IST](https://github.com/GaloyMoney/blink/assets/91112485/60cc04a0-29fc-4b45-9b0d-643f77cdd8d9)

This issue results in an unnecessary API call upon clicking the "Cancel" button, which should be addressed to improve the user experience.If this issue seems valid, then I am happy to contribute.

nicolasburtey:
also a good issue to tackle

PR #4423, opened 2024-05-05
feat(dashboard): add different functionality for cancel button on login.

amitamrutiya2210:
Fix issue #4396 

Give a different type to the cancel button and add functionality to go back when the cancel button is clicked. Previously, the cancel button would submit the form.


siddhart1o1 (commented):
I don't think this is an issue. There is a reason when you click on "Cancel," the form is submitted; i.e., when we click "Cancel," this section is triggered:
https://github.com/GaloyMoney/blink/blob/d10a8884d09e35c6958c77db561077bd9aad7f09/apps/consent/app/login/email-login-server-action.ts#L37

amitamrutiya2210:
Thanks @siddhart1o1 for the review. I have a question: Is there a specific reason why we make an API call when the cancel button is pressed? I believe if the user presses the cancel button, we shouldn't need to make an API call.

siddhart1o1:
> I have a question: Is there a specific reason why we make an API call when the cancel button is pressed? I believe if the user presses the cancel button, we shouldn't need to make an API call.


Because it's part of the form, we are using server actions to handle form submission, where we are also handling the deny/cancel case.
https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations

PR #4436, opened 2024-05-07
ci: set open-pull-requests-limit in dependabot.yml

bodymindarts:


.github/dependabot.yml
```diff
@@ -12,7 +12,9 @@ updates:
   directory: "/" # Location of package manifests
   schedule:
     interval: "daily"
+  open-pull-requests-limit: 2
 - package-ecosystem: "cargo"
   directory: "/"
   schedule:
     interval: "daily"
+  open-pull-requests-limit: 2
```
dolcalmi, line 20: please remove 1

dolcalmi (changes requested):

.github/dependabot.yml
nicolasburtey: ??

nicolasburtey: why removing 1?

PR #4453, opened 2024-05-10
feat(notifications): query and counts for non bulletins

sandipndev:
This PR updates notification backend for the following use cases:

1. Notification History screen requires querying only non bulletins.
2. The count displayed on top of bell icon indicating number of unacknowledged notifications must show the count for only unacknowledged non bulletins.


thevaibhav-dixit (approved):
LGTM

PR #4483, opened 2024-05-20
fix: assistant function call when multple tool are needed in the same…

nicolasburtey:
… query

### DIFF
--- a/core/api/src/services/openai/assistant.ts
+++ b/core/api/src/services/openai/assistant.ts
@@ -11,6 +11,7 @@ import { sleep } from "@/utils"
 import { UnknownDomainError } from "@/domain/shared"
 import {
   ChatAssistantNotFoundError,
+  TimeoutAssistantError,
   UnknownChatAssistantError,
 } from "@/domain/support/errors"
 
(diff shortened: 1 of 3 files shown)


core/api/src/services/openai/assistant.ts
```diff
@@ -134,30 +135,36 @@ export const Assistant = (): ChatAssistant => {
...
 
-    const name = action.submit_tool_outputs.tool_calls[0].function.name
-    assert(name === "queryBlinkKnowledgeBase")
+    const outputs: string[] = []
 
-    const args = action.submit_tool_outputs.tool_calls[0].function.arguments
-    const query = JSON.parse(args).query_str
+    for (const toolCall of action.submit_tool_outputs.tool_calls) {
+      const name = toolCall.function.name
+      assert(name === "queryBlinkKnowledgeBase")
```
dolcalmi, line 146: not a blocker but this is not a standard way to throw an exception. if we want to use it please catch properly `AssertionError` in the catch block

core/api/src/services/openai/assistant.ts
```diff
@@ -222,12 +234,12 @@ export const Assistant = (): ChatAssistant => {
...
 
       if (responseThread.content[0]?.type !== "text") {
-        return new UnknownChatAssistantError("last message is not text")
+        return new UnknownChatAssistantError("Last message is not text")
       }
 
       return true
     } else {
-      return new UnknownChatAssistantError("issue running the assistant")
+      return new UnknownChatAssistantError("Issue running the assistant")
```
dolcalmi, line 242: else block is not necessary

dolcalmi (approved):

PR #4571, opened 2024-08-16
fix(core): update onchain address validation

dolcalmi:
Update onchain address validation to avoid an unknown error https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/9YBTiwAtusr 

ps: an alternative could be to use something like https://github.com/ruigomeseu/bitcoin-address-validation


core/api/test/unit/domain/bitcoin/onchain/checked-to-on-chain-address.spec.ts
```diff
@@ -2,19 +2,36 @@ import { BtcNetwork } from "@/domain/bitcoin"
...
+    network: "regtest",
+    address: "bcrt1p0xlxvlhemja6c4dqv22uapctqupfhlxm9h8z3k2e72q4k9hcz7vqc8gma6",
+  },
+]
+
+const invalidAddresses = [
+  { network: "mainnet", address: "bc1q5a25l8l5r6kgestvadghsd88pjcj923mwed7ku4" },
+  {
+    network: "mainnet",
+    address: "bc1qrp33g0q5c5txsp9arysrx4k6zdkfs4nce4xj0gdcxswfvpysxf3qccfmv3",
```
vindard, line 34: Why was this previously valid? (curiosity question)

vindard (approved):
Looks ok to me, just left one question

core/api/test/unit/domain/bitcoin/onchain/checked-to-on-chain-address.spec.ts
dolcalmi: previously it was valid because we only had regex validation and not decode validation... you can check here https://cointools.org/valid-address-checker/?address=bc1qrp33g0q5c5txsp9arysrx4k6zdkfs4nce4xj0gdcxswfvpysxf3qccfmv3

PR #4576, opened 2024-08-28
chore: add memo to lnurl and lnaddress send mutations

vindard:
## Description

Add memo field when sending via lnurl or lnaddress


vindard:
Closing because for now we'd like to prioritize the lnurl/ln-address memo from the invoice description as the final transaction memo

PR #4609, opened 2024-10-08
fix(deps): PG version sync

k3yss:
This PR syncs the postgresql versions for the docker-compose files as per the current infrastructure/

### DIFF
--- a/quickstart/docker-compose.yml
+++ b/quickstart/docker-compose.yml
@@ -92,7 +92,7 @@ services:
     depends_on:
     - hydra-pg
   hydra-pg:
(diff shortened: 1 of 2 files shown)


quickstart/docker-compose.yml
dolcalmi, line 1: quickstart is updated automatically, should not be update it in this PR

dolcalmi (changes requested):

k3yss:
@dolcalmi Can you re-review?

dolcalmi (approved):

PR #4614, opened 2024-10-17
perf(core): update username index

dolcalmi:
Avg duration is more than 1 second https://ui.honeycomb.io/galoy/datasets/galoy-bbw/result/71kKbg8xRg6?hideCompare

- Add to username index only records with a valid value
- Remove regex query (improves performance)
- Force the use of username index (I tested it directly and for some reason is not using the index without the hint)


openoms (approved):
LGTM with all tests passing

PR #4627, opened 2024-10-20
test(apps): fix dashboard/voucher/consent tests

dolcalmi:
**Github Actions**
- **Integration tests**:  reduces execution time ~10 mins with the benefit of restart failed jobs for specific apps (caused by infra issues). [before](https://github.com/GaloyMoney/blink/actions/runs/11488091878?pr=4627) -> [after](https://github.com/GaloyMoney/blink/actions/runs/11488091869?pr=4627)
- **E2E tests**: add additional builds to avoid bats setup issues because of timeouts (caused by builds)

**Tests**
- Reverts next version because of a known issue with URL helper method that cause tests crashes
- Fix cookie name issues caused for hydra update (now is a long token that cause errors with tests)
- Fix csrf issue with consent login test
- Fix dashboard issue that caused tests to fail because of an unintended refresh
- Refactor api key creation component in dashboard
- Refactor voucher and dashboard auth for tests (remove NEXT_AUTH_SESSION_TOKEN hack and now use cy.session)
- Fix mui deprecations
- Fix Tilt resources deps


openoms (approved):
LGTM with the tests passing.

Will need to test the API key creation in the Dashboard and the Voucher flow on staging.

PR #4646, opened 2024-11-05
perf(core): update mongodb default connection settings

dolcalmi:


core/api/src/services/ledger/admin-legacy.ts
```diff
@@ -29,9 +29,7 @@ const getWalletBalance = async (account: string, query = {}) => {
 
   const balancePromise = (async () => {
     try {
-      const { balance } = await MainBookAdmin.balance(params, {
-        readPreference: "secondaryPreferred",
```
dolcalmi, line 33: now it is not required because readPreference is setup at server level

core/api/src/services/ledger/index.ts
```diff
@@ -312,12 +312,9 @@ export const LedgerService = (): ILedgerService => {
   ): Promise<Satoshis | LedgerError> => {
     const liabilitiesWalletId = toLiabilitiesWalletId(walletId)
     try {
-      const { balance } = await MainBook.balance(
-        {
-          account: liabilitiesWalletId,
-        },
-        { readPreference: "primaryPreferred" },
```
dolcalmi, line 319: now it is not required because readPreference default value is primaryPreferred

core/api/src/services/mongodb/index.ts
```diff
@@ -17,6 +17,36 @@ import { WalletCurrency } from "@/domain/shared"
...
+    w: "majority", // Wait for majority acknowledgment
+    j: true, // Wait for journal commit
+    wtimeout: 10000, // Write timeout
+  },
+
+  retryReads: true,
+  maxStalenessSeconds: 90,
+  readPreference: "primaryPreferred", // Prefer primary but allow secondary reads
+  readConcern: { level: "majority" }, // Read from majority-committed data
+} as const
```
openoms, line 48: Compared the defaults there is compression introduced and more connections with bigger pool sizes, these settings expected to increase the performance.
Can you link the baseline you are comparing to?

Then in writeConcern and readConcern these settings should improve the consistency, hopefully not with too much tradeoff in performance.

openoms (approved):

PR #4686, opened 2024-12-13
chore: remove redundant words

majorteach:


dolcalmi:
thanks for you contribution, already fixed in a different pr

PR #4750, opened 2025-02-20
feat(core): get account default wallet by username method updated to …

tiankii:
**Sending Sats via SMS & Onboarding**

imagen
Reference: [Click here](https://github.com/blinkbitcoin/blink-tianki/issues/4)

- This is an extension of the username query needed to find by phone number, required  on the (**sending Sats via SMS & Onboarding**) task.


core/api/src/domain/users/index.ts
```diff
@@ -18,7 +18,7 @@ export * from "./phone-metadata-authorizer"
 
 // TODO: we could be using https://gitlab.com/catamphetamine/libphonenumber-js#readme
 // for a more precise "regex"
-const PhoneNumberRegex = /^\+\d{7,14}$/i // FIXME {7,14} to be refined
+export const PhoneNumberRegex = /^\+\d{7,14}$/i // FIXME {7,14} to be refined
```
dolcalmi, line 21: please dont export... use checkedToPhoneNumber

core/api/src/graphql/public/root/query/account-default-wallet.ts
```diff
@@ -22,11 +21,7 @@ const AccountDefaultWalletQuery = GT.Field({
       throw username
     }
 
-    const account = await AccountsRepository().findByUsername(username)
-    if (account instanceof Error) {
-      throw mapError(account)
-    }
-
+    const account = await Accounts.getAccountByUsernameOrPhone(username)
```
dolcalmi, line 24: you need to move the wallet selection too to app layer. including `throw mapError(new CouldNotFindWalletFromUsernameAndCurrencyError(username))` but please dont throw just return `new CouldNotFindWalletFromUsernameAndCurrencyError(username)`

dolcalmi (changes requested):

core/api/src/app/accounts/get-account-for-username-or-phone.ts
```diff
@@ -0,0 +1,22 @@
+import { UsernameRegex } from "@/domain/accounts"
+import { mapError } from "@/graphql/error-map"
+import { AccountsRepository, UsersRepository } from "@/services/mongoose"
+
+export const getAccountByUsernameOrPhone = async (
```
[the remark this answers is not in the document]
tiankii: About this do you mean just a renaming change "getAccountByUsernameOrPhone" function to "getDefaultWalletByUsernameOrPhone"?

core/api/src/graphql/public/root/query/account-default-wallet.ts
tiankii: It is not clear to me what I should do, since I based on the other validations of the same file.

I also have doubts about the CouldNotFindWalletFromUsernameAndCurrencyError class, since that was already there, and we did not make any changes.

core/api/src/app/accounts/get-account-for-username-or-phone.ts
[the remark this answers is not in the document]
dolcalmi: yes.. you need to query the wallets that logic must not be at graphql layer

core/api/src/graphql/public/root/query/account-default-wallet.ts
dolcalmi: you need to move the wallet selection logic to app layer and in app layer you must not use throw

core/api/src/graphql/public/root/query/default-wallet-by-phonenumber.ts
dolcalmi, line 1: why did you do this?

dolcalmi (changes requested):
checks are not passing and you didn't implement the requested tests

core/api/src/app/accounts/get-default-wallet-for-username-phone.ts
```diff
@@ -0,0 +1,28 @@
...
+): Promise<Wallet | RepositoryError> => {
+  const checkedUsername = checkedToUsername(value)
+  if (!(checkedUsername instanceof Error)) {
+    const account = await AccountsRepository().findByUsername(checkedUsername)
+    if (account instanceof Error) return account
+    return getWalletFromAccount(account, walletCurrency, value)
+  }
+
+  const checkedPhoneNumber = checkedToPhoneNumber(value)
+  if (checkedPhoneNumber instanceof Error) return checkedPhoneNumber
```
dolcalmi, line 19: although this is the usual pattern, in this case please return `CouldNotFindWalletFromUsernameAndCurrencyError(value)`
```suggestion
  if (checkedPhoneNumber instanceof Error) return new CouldNotFindWalletFromUsernameAndCurrencyError(value)
```

core/api/src/graphql/shared/types/scalar/username.ts
```diff
@@ -23,6 +24,12 @@ function validUsernameValue(value: string) {
   if (value.match(UsernameRegex)) {
     return value.toLowerCase()
   }
+
+  const checkedPhoneNumber = checkedToPhoneNumber(value)
+  if (!(checkedPhoneNumber instanceof Error)) {
+    return value
```
dolcalmi, line 30: if TS allow it you need to return `checkedPhoneNumber` instead of `value` (probably not for this PR but for the next one we will need to handle phone numbers without + so we will need to add them in checkedToPhoneNumber)

core/api/src/app/accounts/get-wallet-from-account.ts
```diff
@@ -0,0 +1,23 @@
+import { CouldNotFindWalletFromUsernameAndCurrencyError } from "@/domain/errors"
+import { WalletsRepository } from "@/services/mongoose"
+
+export const getWalletFromAccount = async (
+  account: Account,
+  walletCurrency?: WalletCurrency,
+  value?: Username | PhoneNumber,
```
dolcalmi, line 7: this param is not necessary

core/api/src/app/accounts/get-wallet-from-account.ts
```diff
@@ -0,0 +1,23 @@
...
+): Promise<Wallet | RepositoryError> => {
```
dolcalmi, line 8: Application layer methods should not return `RepositoryErrors` please use `ApplicationError`

core/api/src/app/accounts/get-wallet-from-account.ts
```diff
@@ -0,0 +1,23 @@
...
+  const wallets = await WalletsRepository().listByAccountId(account.id)
+  if (wallets instanceof Error) return wallets
+
+  if (!walletCurrency) {
+    return (
+      wallets.find((wallet) => wallet.id === account.defaultWalletId) ??
+      new CouldNotFindWalletFromUsernameAndCurrencyError(value)
```
dolcalmi, line 15: if you split this query you need to use proper errors... `CouldNotFindDefaultWalletForAccount` (which should be critical)

core/api/src/app/accounts/get-wallet-from-account.ts
```diff
@@ -0,0 +1,23 @@
...
+    )
+  }
+
+  return (
+    wallets.find((wallet) => wallet.currency === walletCurrency) ??
+    new CouldNotFindWalletFromUsernameAndCurrencyError(value)
```
dolcalmi, line 21: same here `CouldNotFindWalletForCurrency` (which should be critical too)

core/api/src/app/accounts/get-default-wallet-for-username-phone.ts
```diff
@@ -0,0 +1,28 @@
+import { getWalletFromAccount } from "./get-wallet-from-account"
+
+import { checkedToUsername } from "@/domain/accounts"
+import { checkedToPhoneNumber } from "@/domain/users"
+import { AccountsRepository, UsersRepository } from "@/services/mongoose"
+
+export const getDefaultWalletByUsernameOrPhone = async (
+  value: Username | PhoneNumber,
```
dolcalmi, line 8: value is too generic.. in this case please use `usernameOrPhone`

bats/core/api/public-ln-receive.bats
```diff
@@ -110,6 +110,56 @@ usd_amount=50
   [[ "$error_msg" == "Account does not exist for username idontexist" ]] || exit 1
 }
 
+@test "public-ln-receive: account details - can fetch default wallet with phone number" {
```
dolcalmi, line 113: you need to test default, btc and usd... also this name is a bit confusing because it says default but you are testing BTC wallet

dolcalmi (changes requested):

core/api/src/app/accounts/get-default-wallet-for-username-phone.ts
```diff
@@ -0,0 +1,30 @@
...
+  const checkedUsername = checkedToUsername(usernameOrPhone)
+  if (!(checkedUsername instanceof Error)) {
+    const account = await AccountsRepository().findByUsername(checkedUsername)
+    if (account instanceof Error) return account
+    return getWalletFromAccount(account, walletCurrency)
+  }
+
+  const checkedPhoneNumber = checkedToPhoneNumber(usernameOrPhone)
+  if (checkedPhoneNumber instanceof Error)
+    return new CouldNotFindWalletFromUsernameAndCurrencyError(checkedPhoneNumber)
```
dolcalmi, line 21:
```suggestion
    return new CouldNotFindWalletFromUsernameAndCurrencyError(usernameOrPhone)
```
please pass the original value, if we pass checkedPhoneNumber it will be another error

dolcalmi (changes requested):

dolcalmi (approved):

PR #4763, opened 2025-03-02
chore(core): update invoice default expiration

dolcalmi:


openoms (approved):

openoms:
Will need to update the FAQ with the final values: 
https://faq.blink.sv/using-blink/when-do-lightning-invoices-expire

PR #4787, opened 2025-03-28
fix(core): persist external id for invoice settled intraledger

dolcalmi:


openoms (approved):
LGTM with thorough tests

PR #4835, opened 2025-04-30
fix(pay): make server side LNURL request to avoid CORS issue

### LINKED ISSUE #4825: pay.blink.sv PoS CORS error when using BTCPayServer issued bolt cards
Continuing here with the Blink POS boltcard issue Kemal was helping to test.
We have confirmed using multiple boltcards connected to different btcpayservers that the CORS problem only affects the Blink POS.

NFC capable PoS -es used to test with 1 sat only:
LNbits TPOS:
https://lnbits.huszonegy.world/tpos/Kag37N4AKqBMtoyBMHqHZf
BTCPay:
https://tips.diynodes.com
Blink POS:
https://pay.blink.sv/lightningatm

Could make get my Android browser terminal on the desktop like this:

Image

What I found that both LNbits and BTCpayserver are submitting the invoice to the boltcard lnurlw link in one step:

Image

Image

the payload has the lnurl link from the bolt card

the Blink POS tries to query the bolt card link first which fails with the CORS error:

Image

the payload there is the parameters from the bolt card lnurl:

Image

I wonder is this an issue with js-lnurl library we are using in https://github.com/GaloyMoney/blink/blob/main/apps/pay/components/parse-pos-payment/nfc.tsx or is the implementation which needs to be changed?

all three tools are web based
what I see is that the lnbits and btcpay implementation of web pos does not make this same kind of request which is CORS blocked with the Blink PoS.

openoms:
# Fix CORS error when using BTCPayServer-issued bolt cards

## Problem

The Blink POS system experiences CORS errors when trying to use BTCPayServer-issued bolt cards. This happens because the system attempts to make direct browser-to-LNURL-server requests, which fail due to CORS restrictions.

Other POS systems (LNbits TPOS, BTCPay) don't have this issue because they handle LNURL requests server-side.

## Solution

Implemented a server-side proxy approach to avoid CORS restrictions:

1. Created a server-side proxy endpoint for LNURL requests
2. Modified the NFC component to send requests through the proxy instead of directly using js-lnurl
3. Fixed a typo in the UI ("Bold card" → "Bolt card")

## How It Works

- When a bolt card is tapped, the NFC component sends both the LNURL and payment request to our server-side proxy
- The proxy handles all communication with the LNURL server, avoiding CORS issues
- The proxy returns the result to the client

This approach mirrors how LNbits and BTCPay handle bolt cards, which is why they don't experience CORS issues.

## Testing

Verify that the solution works with BTCPayServer-issued bolt cards by confirming:
- No CORS errors appear in the browser console
- Payments process successfully

Fixes #4825

### DIFF
--- a/apps/pay/app/api/lnurl-proxy/route.ts
+++ b/apps/pay/app/api/lnurl-proxy/route.ts
@@ -1,0 +1,51 @@
+import { NextRequest, NextResponse } from "next/server"
+import { getParams } from "js-lnurl"
+
+export async function POST(request: NextRequest) {
+  try {
+    const { lnurl, paymentRequest } = await request.json()
+
+    if (!lnurl) {
+      return NextResponse.json(
+        { error: "Missing lnurl parameter" },
+        { status: 400 }
+      )
+    }
+
+    // If we have both lnurl and paymentRequest, we're doing a direct payment
+    if (lnurl && paymentRequest) {
+      const lnurlParams = await getParams(lnurl)
+
+      if (!("tag" in lnurlParams && lnurlParams.tag === "withdrawRequest")) {
+        return NextResponse.json(
+          { error: "Not a properly configured lnurl withdraw tag" },
+          { status: 400 }
+        )
+      }
+
+      const { callback, k1 } = lnurlParams
+
+      const urlObject = new URL(callback)
+      const searchParams = urlObject.searchParams
+      searchParams.set("k1", k1)
+      searchParams.set("pr", paymentRequest)
+
+      const url = urlObject.toString()
+
+      const result = await fetch(url)
+      const data = await result.json()
+
+      return NextResponse.json(data, { status: result.ok ? 200 : 400 })
+    }
+
+    // If we only have lnurl, just get the params
+    const params = await getParams(lnurl)
+    return NextResponse.json(params)
+  } catch (error) {
+    console.error("Error processing LNURL request:", error)
+    return NextResponse.json(
+      { error: "Failed to process LNURL request" },
+      { status: 500 }
+    )
+  }
+}
--- a/apps/pay/app/api/lnurl-proxy/route.ts
+++ b/apps/pay/app/api/lnurl-proxy/route.ts
@@ -21,7 +21,7 @@ export async function POST(request: NextRequest) {
 
     const lnurlParams = await getParams(lnurl)
 
-    if (lnurlParams.tag !== "withdrawRequest") {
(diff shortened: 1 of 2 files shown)


apps/pay/app/api/lnurl-proxy/route.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    if (!lnurl) {
+      return NextResponse.json(
+        { error: "Missing lnurl parameter" },
+        { status: 400 }
+      )
+    }
+
+    // If we have both lnurl and paymentRequest, we're doing a direct payment
+    if (lnurl && paymentRequest) {
```
dolcalmi, line 16: if we need both then it is cleaner to add paymentRequest validation and remove this if `if (lnurl && paymentRequest)`
```js
    if (!paymentRequest) {
      return NextResponse.json(
        { error: "Missing paymentRequest parameter" },
        { status: 400 }
      )
    }
```

apps/pay/app/api/lnurl-proxy/route.ts
```diff
@@ -0,0 +1,51 @@
...
+    const { lnurl, paymentRequest } = await request.json()
+
+    if (!lnurl) {
+      return NextResponse.json(
+        { error: "Missing lnurl parameter" },
+        { status: 400 }
+      )
+    }
+
+    // If we have both lnurl and paymentRequest, we're doing a direct payment
```
dolcalmi, line 15: comment is not necessary

apps/pay/app/api/lnurl-proxy/route.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (lnurl && paymentRequest) {
+      const lnurlParams = await getParams(lnurl)
+
+      if (!("tag" in lnurlParams && lnurlParams.tag === "withdrawRequest")) {
```
dolcalmi, line 19: are you getting TS error without `"tag" in lnurlParams` in theory it should be enough with `lnurlParams.tag !== "withdrawRequest"`

apps/pay/components/parse-pos-payment/nfc.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { useState, useEffect } from "react"
-import { getParams } from "js-lnurl"
+// Import removed as we'll use our proxy instead
+// import { getParams } from "js-lnurl"
```
dolcalmi, line 3: please remove

dolcalmi (changes requested):

